PR #2, opened 2019-06-26
[KED-824] Adopt a new JSON format

tolomea:
## Notice

- [ ] I acknowledge and agree that, by checking this box and clicking “Submit Pull Request”:

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

## Motivation and Context
Why was this PR created?

## How has this been tested?
What testing strategies have you used?

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added new entries to the `RELEASE.md` file
- [x] Added tests to cover my changes
- [x] Assigned myself to the PR
- [x] Added `Type` label to the PR

PR #8, opened 2019-07-03
[KED-824][KED-788] Use new JSON API in app, and calculate transitive links on the fly

richardwestenra:
# Description

**This PR contains breaking changes for the API data format.** It must be merged after #2. When merged into master, we should bump the version number to v2.

## KED-824
We are moving to a new JSON data/API structure (See #2), which will allow us to add a great deal more metadata about each node, edge and tag in future. Instead of a list of task nodes, we're passing a list of snapshots, each of which will contain lists of the nodes (both task and data), edges and tags, associated with that snapshot. This is much closer to the data structure used in rendering the chart, so the data formatting required when receiving the initial data is less intense: Now instead of generating the nodes and edges from incomplete data then normalising it, all that needs to be done is to normalise the data. A lot of the calculations that were previously being handled in client-side JavaScript are now able to be handled when the data is generated, in Python.

## KED-788
I have also changed the way transitive links are generated, so that they are calculated on the fly when a chart is rendered, rather than when the initial data is formatted. Previously, they were only calculated for the node(task)/data views, and only worked to jump a single hidden node. This meant that when you manually disabled a node, the nodes on either side would be cut off from each other.

Now, when you hide a node that connects two other nodes, a new connection is formed, and you can do this for as many hidden nodes in a chain as you want, as long as there is a complete path running from the ultimate parent to the final child.

# Development notes

- There are a lot of files modified in this PR, but that's mainly because I split data.mock into data.mock and state.mock (because the mock data got really big with the API change).
- A lot of tests got updated in this PR, and I added/fixed some existing ones too.

# QA notes

- Transitive links (between the parents and children of disabled nodes) work differently now.
- I've had to update the random data generator and the way parameters work, so those should be checked too.
- I've changed the min-height of the node list on small screens to 400px, as 150px was too damn small.

## Checklist

## Notice

### DIFF
--- a/src/selectors/edges.js
+++ b/src/selectors/edges.js
@@ -88,10 +88,8 @@ export const addNewEdge = (source, target, { edgeIDs, sources, targets }) => {
  */
 export const findTransitiveEdges = (
   activeSnapshotEdges,
-  edgeSources,
-  edgeTargets,
-  nodeDisabled,
-  transitiveEdges
+  transitiveEdges,
+  { edgeSources, edgeTargets, nodeDisabled }
 ) => {
   /**
    * Recursively walk through the graph, stepping over disabled nodes,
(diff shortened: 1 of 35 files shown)


src/selectors/edges.js
```diff
@@ -66,23 +66,123 @@ export const getEdgeDisabled = createSelector(
 );
 
 /**
- * Format edges and return them as an array
+ * Create a new transitive edge fromn the first and last edge in the path
```
ottis: Fromn -> from

PR #10, opened 2019-07-08
[KED-786] Fix zoom offset

richardwestenra:
## Description
Previously, when nav sidebar was open, zooming the chart would kind of skew the viewport to the left. This change fixes this, and also helpfully removes an extra animation element, so performance will hopefully be improved, if anything.

## Development notes
I've also improved React ref usage on the Flowchart component:
- Improve naming (remove underscores etc, as these are not a widely-recognised naming pattern).
- Use React.createRef instead of callback refs.

## QA notes
Check zoom when nav is both open and closed. Check different screen sizes, and hover on tooltips.

## Checklist

## Legal notice

### DIFF
--- a/src/components/flowchart/index.js
+++ b/src/components/flowchart/index.js
@@ -135,13 +135,8 @@ export class FlowChart extends Component {
    * Render chart to the DOM with D3
    */
   drawChart() {
-    const { chartSize, layout, textLabels } = this.props;
+    const { layout, textLabels } = this.props;
     const { nodes, edges } = layout;
-
-    // Update SVG dimensions
-    this.el.svg
-      .attr('width', chartSize.outerWidth)
-      .attr('height', chartSize.outerHeight);
 
     // Create selections
     this.el.edges = this.el.edgeGroup
@@ -304,6 +299,7 @@ export class FlowChart extends Component {
    * Render React elements
    */
   render() {
+    const { outerWidth, outerHeight } = this.props.chartSize;
     const {
       tooltipVisible,
       tooltipIsRight,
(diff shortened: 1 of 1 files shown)


src/components/flowchart/index.js
```diff
@@ -328,18 +329,13 @@ export class FlowChart extends Component {
...
+          <g ref={this.wrapperRef}>
+            <g className="pipeline-flowchart__edges" ref={this.edgesRef} />
+            <g
+              id="nodes"
+              className="pipeline-flowchart__nodes"
+              ref={this.nodesRef}
+            />
           </g>
         </svg>
         <div
```
ottis: Have you considered using D3 just for it's helper methods (e.g. output paths, curve data etc) so the rendering is consistent i.e. via react?

ottis:
Have you explored other ways to get the nav offset width? Could the current implementation (below) become brittle over time if we were to change the side bar width in the design (have to change in multiple places)

```
getNavOffset(width) {
    const navWidth = width > 480 ? 300 : 0; // could these numbers be in some shared config?
    return this.props.visibleNav ? navWidth : 0;
  }
```

src/components/flowchart/index.js
richardwestenra: Yeah it's something I'm increasingly considering, especially as D3-generated DOM nodes are proving so hard to test with Jest/Enzyme.

I originally generated the chart with D3 over a year ago because I wanted to have more control over the animations. I'm slowly moving towards using React for more of the rendering, but refactoring it would basically only help with adding extra tests at the moment, so it's not a high priority for me.

richardwestenra:
> Have you explored other ways to get the nav offset width? Could the current implementation (below) become brittle over time if we were to change the side bar width in the design (have to change in multiple places)

Yeah it's not ideal, but it's using both JS variables and Sass variables:
```sass
$nav-width: 300px;
$breakpoint-small: 480px;
```
Without an easy way to share variables between Sass and JS (and I think it'd be overkill to implement a complicated tooling solution for this use-cas), I'm inclined to leave as is. But perhaps it could be documented better?

src/components/flowchart/index.js
ottis: Ok cool, let's keep it in the back of our heads, definitely agree with the testing aspect - might be a bigger endeavour down the line 😄

ottis:
@richardwestenra ok cool let's go with the documentation, maybe there's a good way of doing this - will investigate

PR #17, opened 2019-07-16
Redo Flask caching

tolomea:
## Notice

## Motivation and Context
Flask was treating all of the JS build artifacts as "static" and applying default cache forever headers.
I've changed it so only the static directory gets this treatment and all other file fetches are sent with max_age=0

## How has this been tested?
What testing strategies have you used?

## Checklist

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -33,20 +33,21 @@ from collections import defaultdict
 from pathlib import Path
 
 import click
-from flask import Flask, jsonify
+from flask import Flask, jsonify, send_from_directory
 from kedro.cli import get_project_context
 
 app = Flask(  # pylint: disable=invalid-name
-    __name__,
-    static_folder=str(Path(__file__).parent.absolute() / "html"),
-    static_url_path="",
+    __name__, static_folder=str(Path(__file__).parent.absolute() / "html" / "static")
 )
 
 
 @app.route("/")
-def root():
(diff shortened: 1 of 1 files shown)


package/kedro_viz/server.py
```diff
@@ -122,6 +114,15 @@ def pretty_name(name):
     return jsonify({"snapshots": [{"nodes": nodes, "edges": edges, "tags": tags}]})
 
 
+@app.route("/")
+@app.route("/<path:subpath>")
```
921kiyo: `@app.route` is moved because we do not want to affect `@app.route("/logs/nodes.json")` and `@app.route("/api/nodes.json")`? If that's the case, when we wanted to add another `@app.route`, it might be a bit confusing whether future developers should add it before or after `@app.route("/<path:subpath>")`.

tolomea: do you have a concrete suggestion?
personally I feel this largely comes down to responsibility to have a general understanding of the tools you are using and to test things

921kiyo: Okay, I think they will notice if they put a new decorator below as it won't get executed.

921kiyo: Probably nice to leave a comment in the docstring like `only the static directory gets this treatment and all other file fetches are sent with max_age=0`, so people understand why `send_from_directory` was needed in the future.

PR #20, opened 2019-07-18
[KED-812] Toggle linked-node active state on click

richardwestenra:
## Description

- When clicking on a node, keep the linked-node path open even after mouseleave, until you click  somewhere else.
- Add highlighted/focus states for nodes, in addition to active states.
- Clean up the flowchart SCSS, make styling more consistent, and fix some bugs.

## Development notes

- Add highlighted state as well as active state.
- Fix active state bug in linked-node calculations, which meant that the active state from linked-nodes was being overridden in the main flowchart js.
- Use a full list of colour variables
- Separate the SCSS into several different modules

## QA Notes

Lots of small changes in the node hover/click interactions.

## Notice

## Checklist


richardwestenra:
Python E2E tests are currently failing but it looks like a pip dependency issue

src/components/flowchart/flowchart.test.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import $ from 'cheerio';
 import FlowChart, { mapStateToProps, mapDispatchToProps } from './index';
 import { mockState, setup } from '../../utils/state.mock';
-import { getActiveSnapshotNodes } from '../../selectors/nodes';
+import { getActiveSnapshotNodes } from '../../selectors/index';
```
ottis: Could this be changed to:

```
import { getActiveSnapshotNodes } from '../../selectors
```

src/components/flowchart/index.js
```diff
@@ -35,8 +40,11 @@ export class FlowChart extends Component {
     this.nodesRef = React.createRef();
 
     this.handleWindowResize = this.handleWindowResize.bind(this);
+    this.handleNodeClick = this.handleNodeClick.bind(this);
+    this.handleChartClick = this.handleChartClick.bind(this);
     this.handleNodeMouseOver = this.handleNodeMouseOver.bind(this);
     this.handleNodeMouseOut = this.handleNodeMouseOut.bind(this);
+    this.handleNodeKeyDown = this.handleNodeKeyDown.bind(this);
   }
```
ottis: Have you considered using something like https://babeljs.io/docs/en/babel-plugin-proposal-class-properties to use arrow functions instead of having to bind context in the constructor? Would change

```
this.handleNodeKeyDown = this.handleNodeKeyDown.bind(this)
```

to

```
class FlowChart extends Component {
  constructor() {
    // no need for bind
  }

  handleNodeKeyDown = () => {
   // woohoo bound context
  }
}
```

src/components/flowchart/index.js
```diff
@@ -139,9 +147,12 @@ export class FlowChart extends Component {
    * Render chart to the DOM with D3
    */
   drawChart() {
-    const { layout, textLabels } = this.props;
+    const { focusedNode, layout, linkedNodes, textLabels } = this.props;
     const { nodes, edges } = layout;
 
+    // Only apply callback if focusedNode is truthy
+    const ifFocused = callback => (focusedNode ? callback : false);
```
ottis: Would "isFocused" be a better name?

src/components/flowchart/index.js
```diff
@@ -241,30 +265,62 @@ export class FlowChart extends Component {
...
+    }
+  }
+
+  /**
+   * Handle keydown event when a node is focused
+   * @param {Object} node Datum for a single node
+   */
+  handleNodeKeyDown(node) {
+    const ENTER = 13;
+    const ESCAPE = 27;
```
ottis: Do you use key codes elsewhere in the app? Could be worth creating a helper map

src/components/flowchart/styles/_node-icon.scss
```diff
@@ -0,0 +1,80 @@
+@import './variables';
+
+.node--icon {
+  circle,
+  rect {
```
ottis: Do we have any control to add class names to these tags?

src/components/flowchart/styles/_node-text.scss
```diff
@@ -0,0 +1,86 @@
+@import './variables';
+
+.node--text {
+  text {
+    font-weight: bold;
+    pointer-events: none;
+    transition: all ease 0.1s;
+    font-weight: 600;
```
ottis: Two font-weights here

src/components/flowchart/styles/_tooltip.scss
```diff
@@ -0,0 +1,63 @@
+@import './variables';
+
+.pipeline-flowchart__tooltip {
+  position: absolute;
+  top: -60px;
+  left: -20px;
+  padding: 12px 20px;
```
ottis: Where do these numbers come from?

src/components/flowchart/styles/_tooltip.scss
```diff
@@ -0,0 +1,63 @@
...
+  color: white;
+  opacity: 0;
+  visibility: hidden;
+  transition: opacity ease 0.1s, visibility ease 0.1s;
+  pointer-events: none;
+  z-index: 9;
```
ottis: Might be worth having z-index in some variables file e.g. $z-lowest, $z-highest - might help z-index issues in the future, what do you think?

src/components/flowchart/styles/flowchart.scss
```diff
@@ -0,0 +1,20 @@
...
+
+.pipeline-flowchart {
+  height: 100%;
+}
+
+.pipeline-flowchart__graph {
+  display: block;
+}
+
+.node {
```
ottis: Could node be too general a name? Might be worth introducing css modules or following your .pipeline-flowchart__node as a class name

src/reducers/reducers.test.js
```diff
@@ -2,7 +2,7 @@ import mockData from '../utils/data.mock';
 import { mockState } from '../utils/state.mock';
 import reducer from './index';
 import * as action from '../actions';
-import { getActiveSnapshotNodes } from '../selectors/nodes';
+import { getActiveSnapshotNodes } from '../selectors/index';
```
ottis: Could this be import { getActiveSnapshotNodes } from '../selectors/'

src/components/flowchart/index.js
richardwestenra: Not too fussed either way, but it works out of the box in the current version of react-scripts, so, sure why not 🤷‍♂

src/components/flowchart/index.js
richardwestenra: I'd usually use `isFocused` for a Boolean flag, whereas this is a function that executes a callback only if focusedNode is truthy. So it makes more sense to me to call it `ifFocused`, because it will only execute the function if focused (e.g. `.classed('node--highlight', ifFocused(node => linkedNodes[node.id]))`)

src/components/flowchart/index.js
richardwestenra: There's actually a helper in Kedro UI, and I considered using it, but felt that because this is such a short and simple bit of code, this is better because it's more explicit and less abstact, and replacing it with a helper would only make it more difficult to understand.

src/components/flowchart/styles/_node-icon.scss
richardwestenra: We do, and I started on it, then changed my mind and left it as is, because it either ended up making things a bit more complicated on the JS end, or made the CSS rules more specific.

I'm not ruling out changing it to use classes instead at some point, but I didn't want to bite off any more than I already am in this PR, because it's big enough already.

src/components/flowchart/styles/_node-text.scss
richardwestenra: Good catch, thank you

src/components/flowchart/styles/_tooltip.scss
richardwestenra: From here. It's just the styling/positioning for the tooltip?

src/components/flowchart/styles/_tooltip.scss
richardwestenra: I don't mind too much either way - it's not something I've done much in the past. I'm torn about whether this is beneficial or detrimental abstraction, or premature/appropriate optimisation?

I'll do some research for appropriate z-index ordering systems. I suspect that a system with names that reference the things themselves, but which can be reordered sensibly (so no numbers or 'lowest'/'highest') might be wise.

src/components/flowchart/styles/flowchart.scss
richardwestenra: 'node' and 'edge' are names I've used across the app for a specific type of element, so I don't think it's confusing at all - it is after all, a graph with edges and nodes, so I think it's clear that they refer to those. The only time I could imagine it being a problem is if the names were to conflict with those used on other internal apps (:shipit:) where it might be imported. I'd rather not use a longer BEM name with the parent included, because there are so many modifiers used on this class (see below), and it's used in so many places, so I'm trying to keep it nice and concise.

Example of the modifiers used:
```
.classed('node--data', node => node.type === 'data')
.classed('node--task', node => node.type === 'task')
.classed('node--icon', !textLabels)
.classed('node--text', textLabels)
.classed('node--active', node => node.active)
.classed('node--highlight', ifFocused(node => linkedNodes[node.id]))
.classed('node--faded', ifFocused(node => !linkedNodes[node.id]))
```

src/components/flowchart/index.js
ottis: Ah ok makes sense 👍

src/components/flowchart/index.js
ottis: Cool cool, suppose if we use key events again somewhere we can look at using the helper or making this more general purpose 👍

src/components/flowchart/styles/_node-icon.scss
ottis: Cool cool something to consider in the future then 🙂

src/components/flowchart/styles/_tooltip.scss
ottis: Just seems like magic number positioning, what's it based off?

magic

src/components/flowchart/styles/_tooltip.scss
ottis: Definitely should be a consideration if you find yourself using z-indexes elsewhere, the notorious z-index: 9999 is not far behind without putting some structure around it

src/components/flowchart/styles/flowchart.scss
ottis: I meant more interacting when used inside other applications, some style pollution, maybe css modules would be a way to mitigate this going forward?

ottis:
@richardwestenra do you have an update on those failing builds? i.e. why they are failing and if they are related to this branch

src/components/flowchart/styles/_tooltip.scss
richardwestenra: image

Padding and array offset for the tooltip. It's just styling - if it's magic numbers then so is all CSS? 🤷‍♂

src/components/flowchart/index.js
ottis: Nice, cleans it up loads, allows you to focus on important aspects of the constructor instead of the cruft 👍

richardwestenra:
@ottis The builds are failing due to some sort of Python/PyPI issue - the Jest tests are passing, it's just the Python E2E tests that are failing, and they're likely not affected by this PR, but I'll wait til they're fixed before merging.

@921kiyo is working on it, but he says it's a bit more low-priority than another thing he's got on, so it's on hold for now. Hopefully fixed in the next day or so?

src/components/flowchart/styles/flowchart.scss
richardwestenra: It's a possibility, and worth considering, but I reckon I'll cross that bridge when I come to it? Seems low risk for now?

src/components/flowchart/styles/_tooltip.scss
richardwestenra: 👍 Definitely. That being said, this is the only place in the application where z-index is being used, so I'll leave it for now?

richardwestenra:
@tolomea Can you take a look at [this innocuous commit](https://github.com/quantumblacklabs/kedro-viz/pull/20/commits/e157e56498f3074f50af8fd6c44f6ca55be50497) when you get a sec, and chuck up a ✅ if you're happy?

tolomea:
I'm ok with the circle changes

PR #42, opened 2019-09-07
[KED-990] Remove snapshot history

richardwestenra:
**BREAKING CHANGE - do not merge until v3 release**

## Description

As phase one of the coming redesign, we're removing the snapshot history panel, and instead that info will be displayed elsewhere.

### Major changes:
- Delete snapshot history, tabs and description components
- Remove changeActiveSnapshot action/reducer
- Change API data format: Remove the top level snapshots array prop, so now there is only a single pipeline datum
- Change the mock data format, so that instead of an array of two snapshots, now there are two named data objects ('lorem' and 'animals')
- Refactor the data formatter/normaliser, to break it up into shorter functions and simplify it
- Remove/rename mentions of snapshots

## Notice

## Checklist


CONTRIBUTING.md
```diff
@@ -63,12 +63,13 @@ First, clone this repo. Then install dependencies (`npm i`). Now you're ready to
 
 If you want to use a particular dataset, you'll first need to place it in `/public/logs/nodes.json`. Otherwise, you can serve randomly-generated data.
 
-Kedro-Viz uses an environment variable to configure the data source. You can set it when starting up the dev server, e.g. `DATA=random npm start` will serve random procedurally-generated data (refreshed on each page-load). When random or mock data is enabled, certain other features like snapshot history are also enabled by default to help with browser testing. This is usually the most useful setting for local development.
+Kedro-Viz uses an environment variable to configure the data source. You can set it when starting up the dev server, e.g. `DATA=random npm start` will serve random procedurally-generated data (refreshed on each page-load). This is usually the most useful setting for local development.
```
ottis: Would it be beneficial to have a script like "start:dev": "DATA=random npm start", I know i forget this when running and end up just using npm start - then staring at a blank pipeline lol

src/components/app/index.js
```diff
@@ -22,29 +22,25 @@ class App extends React.Component {
...
    * @param {Object} prevData Previous data prop
    * @param {Object} newData New data prop
    */
   dataWasUpdated(prevData, newData) {
-    // Check just the schema IDs of incoming data updates
-    const dataID = ({ snapshots }) =>
-      Array.isArray(snapshots) && snapshots.map(d => d.schema_id).join('');
-
-    return dataID(prevData) !== dataID(newData);
+    return prevData.schema_id !== newData.schema_id;
```
ottis: Do you plan this method to be used in the future? i.e. I can see it used once in componentDidUpdate - do you think it benefits from separating out the check into it's own function?

src/components/app/load-data.js
```diff
@@ -48,16 +38,18 @@ export const getInitialState = (
...
     case 'random':
-      return formatSnapshots(getRandomHistory());
-    case 'mock':
-      return formatSnapshots(mockData);
+      return formatData(getRandomPipeline());
+    case 'lorem':
+      return formatData(mockData.lorem);
+    case 'animals':
+      return formatData(mockData.animals);
     case 'json':
```
ottis: This feels odd having the random / lorem / animals in the loadData function i.e. mixing test / mock related items with loading real data. Could the config be changed whereby the env var is a path to the data? e.g.

```
DATA_DIR='./mock-data/animals.json' && npm start
```

As there are 2 different mental models with this e.g. real json is config based while the others are key based ("animals", "lorem")

src/components/wrapper/index.js
```diff
@@ -57,22 +55,20 @@ export class Wrapper extends Component {
...
               {
-                'pipeline-sidebar__hide-menu--offset': !showHistory,
                 'pipeline-sidebar__hide-menu--visible': visibleNav
               }
             )}
             onClick={this.toggleNav.bind(this)}>
             <Icon type="close" title="Close" theme={theme} />
           </button>
-          <SidebarTabs />
+          <ChartUI />
```
ottis: Could the naming of this be better? i.e. ChartUI isn't very semantically named (don't really know what i'm getting when I use it etc)

src/utils/format-data.js
```diff
@@ -5,169 +5,109 @@
...
+ * @param {string} type - 'data' or 'task'
+ * @param {Array} tags - List of associated tags
+ */
+const addNode = state => node => {
+  const { id } = node;
+  if (state.nodeName[id]) {
+    return;
+  }
+  state.nodes.push(id);
+  state.nodeID[id] = id;
```
ottis: Silly question - we have 

```
if (state.nodeName[id]) {
```

That checks if it exists, then we have state.nodeID[id] = id; - are the nodeName and nodeID arrays the same? Why is it passing an ID to state.nodeName[id] rather than something called "name"

src/utils/format-data.js
```diff
@@ -5,169 +5,109 @@
...
+  state.nodeName[id] = node.name;
+  state.nodeFullName[id] = node.full_name;
+  state.nodeType[id] = node.type;
+  state.nodeIsParam[id] = Boolean(node.is_parameters);
```
ottis: Why is "is_parameters" not a boolean from the outset?

src/utils/format-data.js
ottis: Is this for convenience for looking it up based on the ID?

src/utils/format-data.js
richardwestenra: I _think_ it is - I suspect this is just to force it into this format if not

src/utils/format-data.js
richardwestenra: Ah yeah `state.nodeID` exists because previously I needed a way to get the original node ID, because the one used for lookups was getting the snapshot ID added to it. I might be able to remove this but I'll need to confirm.

nodeName and nodeID are not necessarily the same though.

richardwestenra: I've removed it, as it doesn't appear to do anything anymore 🤦‍♂

src/components/wrapper/index.js
richardwestenra: What would you recommend? `<SidebarContent />`?

src/components/app/load-data.js
richardwestenra: I think the issue is that there are 3 data loading methods: Direct import from a file, passing it as a prop, and async file request. This function/prop determines which of these to use and where to source it from. There are legitimate reasons why they can't all use the same source input method, as the 3 methods are used in different circumstances (i.e. testing, usage as a React component, and usage as a standalone app). The only reason I've added `lorem` and `animals` here is to allow one to run the test dataset in the browser, to make debugging tests easier.

src/components/app/index.js
richardwestenra: I don't really mind either way - I had it as its own function previously because it was verbose check. Now it's simple. Could collapse it to make it more succinct, but this is easier to  understand, so I don't feel strongly either way.

CONTRIBUTING.md
richardwestenra: Could do 🤷‍♂ though I'd need to add it several times. for each of the different options. I don't feel strongly either way.

src/components/wrapper/index.js
richardwestenra: I think something like SidebarFilters would be a good name for it. I'm about to completely redesign this section, so I'll address the naming in a subsequent PR.

CONTRIBUTING.md
richardwestenra: I'm going to leave this as-is for now, but I'll think about whether this could be done better. Right now I think the current setup (plus the URL query string method) work ok for changing the data input

richardwestenra:
@ottis Thanks a lot for your feedback! Really useful comments. I've addressed and resolved the two that I think need to be fixed in this PR. There's a few others that I think are really good feedback, but are more generally feedback about KedroViz architecture rather than this PR specifically, so I'll make a note of them and address them in future PRs. If that's okay with you, can you please chuck a ✅  on this PR so that merging is unblocked?

PR #50, opened 2019-09-19
Detect and use the OS's default theme if set

richardwestenra:
## Description

Use [window.matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) and [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) to detect the OS's default theme (light/dark) if set, and use it by default

See https://hankchizljaw.com/wrote/create-a-user-controlled-dark-or-light-mode/

## Notice

## Checklist


richardwestenra:
After discussing it with @yetudada, we've agreed that users would probably prefer to just use the dark theme by default, rather than have it be based on relatively unknown OS settings. We can revisit this if it comes up in the future, as it's an interesting use case.

PR #53, opened 2019-09-23
[KED-1051] [KED-1052] [KED-1053] API improvements

tolomea:
## Notice

## Motivation and Context

Viz API updates
KED-1051 Make parameters a separate type from data
KED-1052 Change task full names to be the underlying function name
KED-1053 Replace node id's with hashes

## How has this been tested?
Updated unit tests

## Checklist


package/kedro_viz/server.py
```diff
@@ -70,35 +75,35 @@ def pretty_name(name):
     all_tags = set()
 
     for node in sorted(pipeline.nodes, key=lambda n: n.name):
-        task_id = "task/" + node.name.replace(" ", "")
+        task_id = _hash(str(node))
```
richardwestenra: If there is a node and a dataset with the same name, would `_hash` return the same hash ID for both of them? Is it worth adding the type to the hash seed, just to be sure that they'd get different IDs?

package/kedro_viz/server.py
```diff
@@ -70,35 +75,35 @@ def pretty_name(name):
...
+            edges.append({"source": _hash(namespace), "target": task_id})
             namespace_tags[namespace].update(node.tags)
         for data_set in node.outputs:
             namespace = data_set.split("@")[0]
-            edges.append({"source": task_id, "target": "data/" + namespace})
+            edges.append({"source": task_id, "target": _hash(namespace)})
             namespace_tags[namespace].update(node.tags)
 
     for namespace, tags in sorted(namespace_tags.items()):
+        is_param = bool("param" in namespace.lower())
```
921kiyo: Shouldn't this be something like `bool("params:" in namespace.lower()) or "parameters" == namespace.lower()` (or we will display parameter icon whenever the dataset name included "param")?

package/tests/test_server.py
```diff
@@ -39,93 +39,82 @@
...
     "tags": [{"name": "Bob", "id": "bob"}],
 }
 
 
 def create_pipeline():
-    def func(a, b):  # pylint: disable=unused-argument
+    def func1(a, b):  # pylint: disable=unused-argument
+        return a
+
+    def func2(a, b):  # pylint: disable=unused-argument
```
921kiyo: Can't we pass `func1` to both nodes, as `func2` is identical to `func1` (or would that cause a problem with hashing?)?

tolomea: I wanted to see them separately in the expected output

package/kedro_viz/server.py
tolomea: ideally the catalog should remember if it was added with add_feeddict and we should decide based on that
in the absence of that it's always been like this (it used to be done in the JS) and yes it could incorrectly pick up something that's not parameters

richardwestenra:
Let me know when this is ready, and I'll commit my front-end changes to this branch (?)

package/kedro_viz/server.py
```diff
@@ -70,35 +75,35 @@ def pretty_name(name):
...
     for node in sorted(pipeline.nodes, key=lambda n: n.name):
-        task_id = "task/" + node.name.replace(" ", "")
+        task_id = _hash(str(node))
         nodes.append(
             {
                 "type": "task",
                 "id": task_id,
                 "name": getattr(node, "short_name", node.name),
-                "full_name": str(node),
+                "full_name": getattr(node, "_func_name", str(node)),
```
DmitriiDeriabinQB: If 2 nodes are using the same underlying function, do we expect them to have the same `full_name`?

tolomea: They will if names have not been manually specified, that is fine because they will still have different inputs and outputs in the diagram

tolomea:
@richardwestenra good to go

richardwestenra:
@tolomea @921kiyo @DmitriiDeriabinQB I've added my front-end changes. It's now ready to go. Let me know if you want to review my commit.

Only issue now is that tests appear to be failing at the Python installation stage, likely due to a dependency origin being down or something. I'll keep re-running tests until they work?

src/components/flowchart/icon.js
```diff
@@ -15,6 +16,10 @@ const paths = {
...
+  ],
+  // sliders icon
+  parameters: [
+    'M11.2 18a3 3 0 0 0 0 2H3v-2h8.2zm9.8 0v2h-4.2a3 3 0 0 0 0-2H21zM5.2 11a3 3 0 0 0 0 2H3v-2h2.2zM21 11v2H10.8a3 3 0 0 0 0-2H21zm-9-6c0 .4 0 .7.2 1H3V4h9.2a3 3 0 0 0-.2 1zm9-1v2h-3.2a3 3 0 0 0 0-2H21zM14 17a2 2 0 1 1 0 4 2 2 0 0 1 0-4zm-6-7a2 2 0 1 1 0 4 2 2 0 0 1 0-4zm7-7a2 2 0 1 1 0 4 2 2 0 0 1 0-4z'
```
lorenabalan: Sure.

richardwestenra: ah lol yeah I see what you mean about the series of unintelligible letters and numbers! Yeah this is just an [SVG path definition attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d). It tells the browser the shape of the icon to draw. You don't need to be able to read it - it's basically just image data. :)

PR #72, opened 2019-10-29
FEAT Added optional pipeline argument

### LINKED ISSUE #69: Add Ability to pass pipeline into get_data_from_kedro
## Description
I would like to have the ability to create kedro-viz data and save it to a json file outside of running the kedro-viz server.  The main motivation for this is to keep a simple live version of every pipeline my team is maintaining right at our fingertips for anyone to view.

## Context

## Possible Implementation

https://github.com/quantumblacklabs/kedro-viz/blob/develop/package/kedro_viz/server.py#L110

``` diff
- def get_data_from_kedro():
+ def get_data_from_kedro(pipeline=None):
    """ Get pipeline data from Kedro and format it appropriately """

    def pretty_name(name):
        name = name.replace("-", " ").replace("_", " ")
        parts = [n[0].upper() + n[1:] for n in name.split()]
        return " ".join(parts)

-     pipeline = get_project_context("create_pipeline")()
+    if pipeline == None:
+       pipeline = get_project_context("create_pipeline")()
```

I looked at making this change but do not understand how `kedro-viz` gets packaged.  I tried to inspect kedro.server, but do not even see `get_data_from_kedro`

## Possible Alternatives

Copy the `get_data_from_kedro` function into my own codebase and make the change myself.  I have done this, but do not want to maintain a "fork" of this function so that we can easily follow any data structure changes.

## Checklist

- [x] Make `get_data_from_kedro` importable
- [x] Allow pipeline data to be passed into `get_data_from_kedro`

WaylonWalker:
## Notice

## Motivation and Context

I want to be able to generate pipeline data from within python for visualizing multiple pipelines in a web app. Resolves #69

## How has this been tested?

I manually ran this on one of my pipelines.

## Checklist

- [ ] Added `Type` label to the PR

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,6 +3,9 @@ Add release notes for the upcoming release here.
 
 # Release 2.1.1:
 
+## Major features and improvements
+- Added optional pipeline argument to `get_data_from_kedro` (#70)
+
 ## Bug fixes and other changes
 - Don't ignore gh-pages branch in CircleCI (#33)
 - Document the React props and data format (#34)
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,10 +1,10 @@
 # Upcoming release:
 Add release notes for the upcoming release here.
 
-# Release 2.1.1:
-
 ## Major features and improvements
 - Added optional pipeline argument to `get_data_from_kedro` (#72)
+
+# Release 2.1.1:
 
 ## Bug fixes and other changes
 - Don't ignore gh-pages branch in CircleCI (#33)
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,7 +4,7 @@ Add release notes for the upcoming release here.
 # Release 2.1.1:
 
 ## Major features and improvements
-- Added optional pipeline argument to `get_data_from_kedro` (#70)
+- Added optional pipeline argument to `get_data_from_kedro` (#72)
 
 ## Bug fixes and other changes
 - Don't ignore gh-pages branch in CircleCI (#33)
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,10 @@
 Add release notes for the upcoming release here.
 
 ## Major features and improvements
-- Added optional pipeline argument to `get_data_from_kedro` (#72)
+
+- renamed `get_data_from_kedro` to `format_pipeline_data` (#72)
+- Added optional pipeline, and catalog arguments to `format_pipeline_data` (#72)
+
 
 # Release 3.0.0:
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,7 +4,7 @@ Add release notes for the upcoming release here.
 ## Major features and improvements
 
 - renamed `get_data_from_kedro` to `format_pipeline_data` (#72)
-- Added optional pipeline, and catalog arguments to `format_pipeline_data` (#72)
+- Added pipeline, and catalog arguments to `format_pipeline_data` (#72)
 
 
 # Release 3.0.0:
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -109,7 +109,7 @@ def run_viz(port=None, line=None) -> None:
 
 def get_data_from_kedro(pipeline=None):
     """
-    Get pipeline data from Kedro and format it appropriately 
+    Get pipeline data from Kedro and format it appropriately
 
     Args:
         pipeline: (optional) Kedro pipeline object, if None pipeline is loaded from context.
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -107,7 +107,7 @@ def run_viz(port=None, line=None) -> None:
     display(HTML(wrapper))
 
 
-def get_data_from_kedro(pipeline=None):
+def get_data_from_kedro(pipeline: Pipeline):
     """
     Get pipeline data from Kedro and format it appropriately
 
@@ -119,9 +119,6 @@ def get_data_from_kedro(pipeline=None):
         name = name.replace("-", " ").replace("_", " ")
         parts = [n[0].upper() + n[1:] for n in name.split()]
         return " ".join(parts)
-
-    if pipeline is None:
-        pipeline = get_project_context("create_pipeline")()
 
     nodes = []
     edges = []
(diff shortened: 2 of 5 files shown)


RELEASE.md
```diff
@@ -3,6 +3,9 @@ Add release notes for the upcoming release here.
 
 # Release 2.1.1:
 
+## Major features and improvements
+- Added optional pipeline argument to `get_data_from_kedro` (#72)
```
richardwestenra: Not major, but do you mind moving this up a bit? It should go under `# Upcoming release`. 2.1.1 was a previous release.

WaylonWalker:
@tolomea I really like those suggestions.  Perhaps my fix was a bit short sighted in trying to get something that just works.  As a user of kedro a `to_dict` method on the pipeline object itself would feel a bit more natural.  I know this moves the ownership to a different repo.  I am open to whichever method you prefer.

tolomea:
the format of the data is tuned to be most convenient to the viz JS and changes to that format generally have to happen in tandem to both the JS and PY sides of viz so I think it's best where it is

WaylonWalker:
I have made the `pipeline` argument optional.

package/kedro_viz/server.py
```diff
@@ -107,16 +107,19 @@ def run_viz(port=None, line=None) -> None:
...
 
 
-def get_data_from_kedro():
-    """ Get pipeline data from Kedro and format it appropriately """
+def get_data_from_kedro(pipeline):
+    """
+    Get pipeline data from Kedro and format it appropriately
+
+    Args:
+        pipeline: (optional) Kedro pipeline object, if None pipeline is loaded from context.
```
tolomea: this comment is out of date

package/kedro_viz/server.py
```diff
@@ -107,16 +107,19 @@ def run_viz(port=None, line=None) -> None:
     display(HTML(wrapper))
 
 
-def get_data_from_kedro():
-    """ Get pipeline data from Kedro and format it appropriately """
+def get_data_from_kedro(pipeline):
```
tolomea: could you please add catalog as well for future proofing

WaylonWalker: Does the name `get_data_from_kedro` still make sense since the objects are now passed in?  I made the catalog argument optional for now  since it is not being utilized.  Let me know your thoughts.

tolomea: if it's optional now then code that uses it in the future will have to be conditional on it being provided or it will still be a compatibility break

tolomea: the name doesn't really make sense, format_pipeline_data might be better

WaylonWalker: Those changes are now in

package/kedro_viz/server.py
```diff
@@ -210,7 +214,9 @@ def _call_viz(host=None, port=None, browser=None, load_file=None, save_file=None
                 click.echo("Invalid file, top level key '{}' not found.".format(key))
                 sys.exit(1)
     else:
-        data = get_data_from_kedro()
+        pipeline = get_project_context("create_pipeline")()
+        catalog = None  # currently unused variable for future proofing
```
tolomea: `get_project_context("create_catalog")()`

RELEASE.md
```diff
@@ -1,6 +1,12 @@
 # Upcoming release:
 Add release notes for the upcoming release here.
 
+## Major features and improvements
+
+- renamed `get_data_from_kedro` to `format_pipeline_data` (#72)
+- Added optional pipeline, and catalog arguments to `format_pipeline_data` (#72)
```
tolomea: they are not optional

package/kedro_viz/server.py
tolomea: this is what it will be in future, best to do it right now so that anyone who copies this sees the future proof way

RELEASE.md
```diff
@@ -1,6 +1,12 @@
 # Upcoming release:
 Add release notes for the upcoming release here.
 
+## Major features and improvements
+
+- renamed `get_data_from_kedro` to `format_pipeline_data` (#72)
+- Added pipeline, and catalog arguments to `format_pipeline_data` (#72)
```
richardwestenra: @tolomea I think Waylon has already added release notes? It's not super critical anyway - provided the PR description is decent I usually add it later on during the release process

WaylonWalker:
> and fix the test failures

@tolomea It looks like most of them came in after adding `catalog = get_project_context("create_catalog")()`

I tried to run locally and got this error.

``` 
File ".../kedro-viz/package/kedro_viz/server.py", line 218, in _call_viz
    catalog = get_project_context("create_catalog")()
TypeError: <lambda>() missing 1 required positional argument: 'config'
```


Any suggestions?  I am not very familiar with this part of kedro.

921kiyo:
> > and fix the test failures
> 
> @tolomea It looks like most of them came in after adding `catalog = get_project_context("create_catalog")()`
> 
> I tried to run locally and got this error.
> 
> ```
> File ".../kedro-viz/package/kedro_viz/server.py", line 218, in _call_viz
>     catalog = get_project_context("create_catalog")()
> TypeError: <lambda>() missing 1 required positional argument: 'config'
> ```
> 
> Any suggestions? I am not very familiar with this part of kedro.

`get_project_context("create_catalog")()` expects `config` arg because [this is how we used to create catalog object](https://github.com/quantumblacklabs/kedro/blob/95ba3de5556a0891dbfca82016c1b12760bb7def/kedro/template/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/run.py#L136) when we open sourced Kedro. 

As https://github.com/quantumblacklabs/kedro/blob/develop/kedro/cli/cli.py#L528-L536 said, 
`get_project_context("create_catalog")` and `get_project_context("create_pipeline")` were duplicated, and replaced by `get_project_context("context")`. 

I would suggest to use `context = get_project_context("context")` so that you can get both pipeline and catalog object by `context.pipeline` and `context.catalog`. @tolomea Do you agree with this change?

tolomea:
@921kiyo oops, that was a mistake, we should have made it `*a` but that boat has kinda sailed now, @WaylonWalker this code will need to be changed to pass a None for that arg

tolomea:
@921kiyo using `get_project_context("context")` in a plugin is not a good idea, the thing you get back has changed with every release

package/kedro_viz/server.py
```diff
@@ -215,7 +215,7 @@ def _call_viz(host=None, port=None, browser=None, load_file=None, save_file=None
                 sys.exit(1)
     else:
         pipeline = get_project_context("create_pipeline")()
-        catalog = get_project_context("create_catalog")()
+        catalog = None
```
tolomea: sorry I should have been clearer, I meant `get_project_context("create_catalog")(None)`

WaylonWalker: This causes test failures, any suggestions?

tolomea:
to fix the tests make this change in `test_server.py`
```
def get_project_context(key):
    return {"create_pipeline": create_pipeline, "create_catalog": lambda x: None}[key]
```

WaylonWalker:
Thanks @tolomea, that seemed to solve the issue.

PR #80, opened 2019-11-20
[KED-1218] Add App props to specify theme and hide theme/label icon buttons

richardwestenra:
## Description

JavaScript component consumers of the application have requested the ability to override the theme, and to hide the theme button. This PR adds extra props to facilitate this and also allow users to hide future features like the upcoming export button.

This PR also removes leftover traces of the `created_ts` and `message` data properties, which are no longer used.

Finally, it also turns on text labels by default

## Notice

## Checklist

### DIFF
--- a/src/components/app/index.js
+++ b/src/components/app/index.js
@@ -53,14 +53,21 @@ App.propTypes = {
       nodes: PropTypes.array.isRequired,
       tags: PropTypes.array
     })
-  ])
+  ]),
+  theme: PropTypes.oneOf(['dark', 'light']),
+  visible: PropTypes.shape({
+    labelBtn: PropTypes.bool,
+    themeBtn: PropTypes.bool
+  })
 };
 
 App.defaultProps = {
   /**
    * String (e.g. 'json') or pipeline data
    */
-  data: null
+  data: null,
+  theme: null,
+  visible: {}
 };
 
 export default App;
--- a/src/components/app/index.js
+++ b/src/components/app/index.js
@@ -45,7 +45,7 @@ class App extends React.Component {
 
 App.propTypes = {
   data: PropTypes.oneOfType([
-    PropTypes.string,
+    PropTypes.oneOf(['random', 'lorem', 'animals', 'demo', 'json']),
     PropTypes.shape({
       schema_id: PropTypes.string,
       edges: PropTypes.array.isRequired,
@@ -62,10 +62,25 @@ App.propTypes = {
 
 App.defaultProps = {
   /**
-   * String (e.g. 'json') or pipeline data
+   * Determines what pipeline data will be displayed on the chart.
+   * You can supply one of the following strings:
+     - 'random': Use randomly-generated data
+     - 'lorem': Use data from the 'lorem-ipsum' test dataset
+     - 'animals': Use data from the 'animals' test dataset
+     - 'demo': Use data from the 'demo' test dataset
+     - 'json': Load data from a local json file (in /public/api/nodes.json)
+   * Alternatively, you can supply an object containing lists of edges, nodes, tags.
+   * See /src/utils/data for examples of the expected data format.
    */
   data: null,
+  /**
+   * Specify the theme: Either 'light' or 'dark'.
+   * If set, this will override the localStorage value.
+   */
   theme: null,
+  /**
+   * Show/hide the icon buttons with { labelBtn:false } and/or { themeBtn:false }
+   */
   visible: {}
 };
 
--- a/src/components/app/index.js
+++ b/src/components/app/index.js
@@ -47,9 +47,8 @@ App.propTypes = {
   data: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.shape({
-      created_ts: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+      schema_id: PropTypes.string,
       edges: PropTypes.array.isRequired,
-      message: PropTypes.string,
       nodes: PropTypes.array.isRequired,
       tags: PropTypes.array
     })
--- a/src/components/app/load-data.test.js
+++ b/src/components/app/load-data.test.js
@@ -31,16 +31,15 @@ describe('load-data', () => {
       });
     });
 
-    it('returns mutated visible state if prop is set', () => {
-      const visible = { themeBtn: false };
-      expect(getInitialState(loremData, { visible })).toMatchObject({
-        visible: { labelBtn: true, themeBtn: false }
-      });
-    });
-
     it('uses prop values instead of defaults if provided', () => {
-      expect(getInitialState(loremData, { theme: 'light' })).toMatchObject({
-        theme: 'light'
+      expect(
+        getInitialState(loremData, {
+          theme: 'light',
+          visible: { themeBtn: false }
+        })
+      ).toMatchObject({
+        theme: 'light',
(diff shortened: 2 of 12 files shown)


src/components/app/index.js
```diff
@@ -45,22 +45,43 @@ class App extends React.Component {
...
       nodes: PropTypes.array.isRequired,
       tags: PropTypes.array
     })
-  ])
+  ]),
+  theme: PropTypes.oneOf(['dark', 'light']),
+  visible: PropTypes.shape({
+    labelBtn: PropTypes.bool,
+    themeBtn: PropTypes.bool
+  })
```
ottis: Would providing a config object be more flexible than baking the names into this prop types e.g.

```
config: {
   visibility: {
       labelBtn: false
   }
}
```

With this you could create a HOC to show / hide it e.g.

```
function withVisibility(WrappedComponent, config, componentName) {
  // ...and returns another component...
  return class extends React.Component {
    render() {
      if (typeof config.visibility[componentName] === 'boolean') {
          if (!config.visibility[componentName]) return null
      }
      return <WrappedComponent {...this.props} />;
    }
  };
}
```

You then use it like

``` 
const LabelButton = ({ onToggle, textLabels }) => (
  <button
    aria-live="polite"
    className={classnames('pipeline-toggle-labels pipeline-icon-button', {
      'pipeline-toggle-theme--show-labels': textLabels
    })}
    onClick={() => onToggle(!textLabels)}>
    <span>{textLabels ? 'Hide' : 'Show'} text labels</span>
    <LabelIcon className="pipeline-icon" />
  </button>
);

export withVisibility(LabelButton, config, 'labelBtn')
```

The reason this might something to explore is that if hiding parts of the application is needed more and more it becomes trivial to enable this functionality for any component, rather than having lots of guarded checks dotted around the application (like in src/components/icon-toolbar/index.js)

src/components/app/load-data.test.js
```diff
@@ -0,0 +1,74 @@
...
+    const loremData = loadData('lorem');
+
+    it('returns an object', () => {
+      expect(getInitialState(loremData)).toEqual(expect.any(Object));
+    });
+
+    it('does not require the second argument', () => {
+      expect(getInitialState(loremData)).toEqual(
+        getInitialState(loremData, {})
+      );
```
ottis: Not sure this is testing something of use i.e. you're just testing javascript initial parameter values works rather than some inner workings of the getInitialState method

src/components/app/load-data.test.js
```diff
@@ -0,0 +1,74 @@
...
+      expect(getInitialState(loremData, { visible })).toMatchObject({
+        visible: { labelBtn: true, themeBtn: false }
+      });
+    });
+
+    it('uses prop values instead of defaults if provided', () => {
+      expect(getInitialState(loremData, { theme: 'light' })).toMatchObject({
+        theme: 'light'
+      });
+    });
```
ottis: Could this test and the one above be merged i.e. are they testing the same thing (mutation of expected state) e.g.

```
it('returns mutated state if prop is set', () => {
      const mutate = { visible: { themeBtn: false }, theme: 'light' };
      expect(getInitialState(loremData, mutate)).toMatchObject({
        visible: { labelBtn: true, themeBtn: false },
        theme: 'light'
      });
    });
```

src/components/app/load-data.test.js
```diff
@@ -0,0 +1,74 @@
...
+      saveState({ theme: 'light' });
+      expect(getInitialState(loremData, { theme: 'dark' })).toMatchObject({
+        theme: 'dark'
+      });
+    });
+  });
+
+  describe('loadData', () => {
+    it('returns the correct dataset when passed a dataset string', () => {
+      expect(loadData('random')).toEqual(expect.any(Object));
```
ottis: Would this test pass if the random data set was changed to 

```
loadData(name) => {
   switch(name) {
    case 'random':
      return {}
  }
}
```

? (what i'm asking is if we should be checking the format of random as well)

src/components/icon-toolbar/icon-toolbar.test.js
```diff
@@ -1,6 +1,45 @@
...
+    const onToggle = t => {
+      theme = t;
+    };
+    const wrapper = setup.shallow(ThemeButton, { theme, onToggle });
+    wrapper.find('button').simulate('click');
+    expect(theme).toBe('light');
+  });
+});
+
+describe('LabelButton', () => {
```
ottis: Do you think these components should be separated out of the icon-toolbar? Right now it's in the same file which means the tests are testing multiple components in one file rather than being focused

src/components/icon-toolbar/index.js
```diff
@@ -43,21 +43,27 @@ export const IconToolbar = ({
...
-    </li>
-    <li>
-      <LabelButton onToggle={onToggleTextLabels} textLabels={textLabels} />
-    </li>
+    {visible.themeBtn && (
+      <li>
+        <ThemeButton onToggle={onToggleTheme} theme={theme} />
+      </li>
+    )}
+    {visible.labelBtn && (
```
ottis: This is the guarding I mentioned above ☝️

src/utils/state.mock.js
```diff
@@ -11,8 +11,8 @@ import loremIpsum from './data/lorem-ipsum.mock';
  * Example state objects for use in tests of redux-enabled components
  */
 export const mockState = {
-  lorem: getInitialState(formatData(loremIpsum)),
-  animals: getInitialState(formatData(animals))
+  lorem: getInitialState(formatData(loremIpsum), {}),
+  animals: getInitialState(formatData(animals), {})
```
ottis: Are the second arguments needed here (since getInitialState has that defaulting to {})

src/components/app/index.js
richardwestenra: Nice idea, but a little too clever for my taste. This smells like premature/unnecessary abstraction to me. I'm trying to keep things simple and relatively accessible for non-expert devs.

src/components/app/load-data.test.js
richardwestenra: True, but I just wanted to add a test to guard against removing the default argument. Might be unnecessary 🤷‍♂

src/components/app/load-data.test.js
richardwestenra: 👍

src/components/app/load-data.test.js
richardwestenra: Yeah it would. I guess I was being a bit lazy with it because I didn't consider it the right place to implement a bunch of checks on the random data generator. I can expand it though

src/components/icon-toolbar/icon-toolbar.test.js
richardwestenra: I don't have a problem with multiple components in a single file. They're pretty simple, dumb components, it made sense to me to keep them together

src/utils/state.mock.js
richardwestenra: ooh good catch, I'd added these before I added the default argument. I'll remove them.

richardwestenra:
@ottis Thanks for the review! I've tried to address your comments where I thought the suggestions germane, and will keep the others in the back of my mind moving forward - they might make their way into subsequent reviews. Seeing as I've already got a ✅ from John, I'm gonna go ahead and merge - hope that's all good with you!

PR #82, opened 2019-11-28
[KED-785] Add button to allow users to export graph as SVG/PNG image

richardwestenra:
## Description

Many of our users want to be able to grab a high-fidelity transparent screenshot so they can put it in a Powerpoint slide etc. This PR adds a button that allows them to export the graph as either an SVG or PNG.
image
image

## Development nodes

- Add export button to icon toolbar
- Create separate icon-button component for icon toolbar buttons
- Install svg-crowbar, to export as SVG/PNG
- Add export modal to allow choosing between SVG/PNG
- Allow export button/modal to be hidden along with the other icon buttons
- Convert the FontLoadChecker component into a function that dispatches an action in the store. This enables it to be used in different places across the application, and to trigger updates when the font loads.
- Refactor font loading check logic a little

## Notice

## Checklist


src/components/app/check-font-loaded.js
```diff
@@ -0,0 +1,69 @@
...
+      return;
+    }
+
+    /**
+     * Check whether the webfont is ready, and show the chart if so
+     */
+    const checkIfLoaded = () => {
+      if (fonts.check(fontName)) {
+        // Add 0.1s delay because fonts.check often returns true too early
+        setTimeout(setLoaded, 100);
```
ottis: Is this delay always consistent? e.g. if it loaded after 101 ms

src/components/app/check-font-loaded.js
```diff
@@ -0,0 +1,69 @@
+import { UPDATE_FONT_LOADED } from '../../actions';
```
ottis: Does font loaded in this context mean, the font file has loaded? or the font file has been loaded and applied to the page

ottis: Did you explore using https://github.com/typekit/webfontloader? It gives you lots of events to hook into e.g.

```
Web Font Loader provides an event system that developers can hook into. It gives you notifications of the font loading sequence in both CSS and JavaScript.

loading - This event is triggered when all fonts have been requested.
active - This event is triggered when the fonts have rendered.
inactive - This event is triggered when the browser does not support linked fonts or if none of the fonts could be loaded.
fontloading - This event is triggered once for each font that's loaded.
fontactive - This event is triggered once for each font that renders.
fontinactive - This event is triggered if the font can't be loaded.
```

The active one seems like the one you're replicating below

src/components/app/check-font-loaded.js
```diff
@@ -0,0 +1,69 @@
...
+        setTimeout(setLoaded, 100);
+      }
+    };
+
+    /**
+     * Recursive requestAnimationFrame step function, which acts as a backup
+     * to the native FontFaceSet event handlers.
+     */
+    const step = () => {
+      // If it's been 8 seconds since page load then just set loaded=true
```
Huongg: Hey, any specific reason why you set 8 seconds here?

ottis:
@richardwestenra Noticed some safari specific issues, the SVG output seems broken via safari (slight issues with PNG output as well)  and the modal dialog has a slight layout issue compared to firefox and chrome:

Screenshot 2019-12-11 at 11 32 49
Screenshot 2019-12-11 at 11 33 18
Screenshot 2019-12-11 at 11 33 37

src/components/app/check-font-loaded.js
richardwestenra: It was really hard to say how consistent it is, as it was a trial and error thing - the only way of programmatically detecting it is broken, which is the reason for the delay. It seemed to come in around 50ms or less, so 100ms seemed like a safe delay - long enough to cover the issue, but short enough that it's a pretty imperceptible increase in load-time.

src/components/app/check-font-loaded.js
richardwestenra: > Does font loaded in this context mean, the font file has loaded? or the font file has been loaded and applied to the page

It *should* mean the latter. I think the problem is that this is an experimental feature and the implementation in each browser is pretty universally terrible.

richardwestenra: > Did you explore using https://github.com/typekit/webfontloader?

I didn't, because the font is added in Kedro-UI, so it would mean adding it to that package and exposing the callbacks via an API so they could be utilised in Kedro-Viz. It seemed like a lot of overhead. Seemed easier to use the built-in browser API, even if it is currently a bit inconsistent.

src/components/icons/export.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export default ({ className }) => (
+  <svg
+    className={className}
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 24 24">
+    <path d="M5 16v4h14v-4h2v4c0 1-.8 2-1.9 2H5a2 2 0 01-2-2v-4h2zm8-14v11l3.5-3.5L18 11l-6 6-6-6 1.5-1.5L11 13V2h2z" />
+  </svg>
+);
```
ottis: In create react app you can load SVGs without wrapping them in a component e.g.

```
import { ReactComponent as ExportIcon } from './export.svg';

const App = () => (
  <div>
    {/* Icon is an actual React component */}
    <ExportIcon />
  </div>
);

```

src/components/app/check-font-loaded.js
richardwestenra: It's pretty arbitrary, but 8 seconds seemed like an appropriate length of time because it should be long enough that the font should have loaded by now even on slow-ish connections, but short enough that it won't cause users massive headaches.

src/components/icons/export.js
richardwestenra: Yep that's what I had originally, but I had to remove it because it broke in Studio, which doesn't support webpack's inline SVG loader. See #75.

richardwestenra:
@ottis Safari button issue is a good catch thanks! Will fix. 

Thanks for your thorough testing of the SVG export functionality! I think I'm aware of the SVG rendering issues - did you output the SVG from Chrome then try to load it in Safari? Seems to be that when you export the SVG from one browser, it may not work in other browsers. I believe this is because SVG-Crowbar is using `getComputedStyle` to grab CSS styles and inline them, and this is including some of the browser's proprietary styling which isn't supported in other browsers. I've got [an active PR on SVG-Crowbar](https://github.com/cy6erskunk/svg-crowbar/pull/87) which should resolve this issue (though its primary intent is to reduce overall file-size.

ottis:
> @ottis Safari button issue is a good catch thanks! Will fix.
> 
> Thanks for your thorough testing of the SVG export functionality! I think I'm aware of the SVG rendering issues - did you output the SVG from Chrome then try to load it in Safari? Seems to be that when you export the SVG from one browser, it may not work in other browsers. I believe this is because SVG-Crowbar is using `getComputedStyle` to grab CSS styles and inline them, and this is including some of the browser's proprietary styling which isn't supported in other browsers. I've got [an active PR on SVG-Crowbar](https://github.com/cy6erskunk/svg-crowbar/pull/87) which should resolve this issue (though its primary intent is to reduce overall file-size.

Makes sense re: applying browser specific styles (loading the SVG, that was exported with safari, in safari renders correctly). I think the shareable nature of the format it needs to be fine regardless. Looking at that PR, how would that apply all CSS browser styling to the SVG?

richardwestenra:
> I think the shareable nature of the format it needs to be fine regardless.

Yeah I agree. I should file a bug with SVG-Crowbar about this. No idea how they'd fix it though.

> Looking at that PR, how would that apply all CSS browser styling to the SVG?

Currently, SVG-Crowbar runs `getComputedStyle` on every element in the SVG, and inlines all those styles on every element. It results in a massive CSS overhead if you've got a lot of elements in your SVG, which results in a very large file-size. But it's a very accurate representation of how the SVG looks as long as you run it in that browser.

My PR changes it to provide an option to get all the external/internal styling from `document.styleSheets` and instead drop them in a `<style>` tag. There's still a lot of unnecessary CSS added, but still less than with the current system. I can show you an example if you wanna hop over to my desk.

PR #93, opened 2020-01-22
[KED-1196] Add --pipeline flag to visualize non-default Kedro pipeline

921kiyo:
## Notice

## Motivation and Context
Why was this PR created?
- Add a command-line flag to select pipeline to visualize (e.g. kedro viz --pipeline de)
- It's needs to be bacwards compatible for all previous Kedro versions

## How has this been tested?
What testing strategies have you used?
Unit tests

## Checklist


RELEASE.md
```diff
@@ -1,8 +1,9 @@
 # Release 3.1.2:
 
 ## Major features and improvements
-- Promote viz from project specific commands to global commands.
-- Enable `kedro viz --load-file` outside of a Kedro project.
+- Promoted viz from project specific commands to global commands.
+- Enabled `kedro viz --load-file` outside of a Kedro project.
+- Added `--pipeline` flag to visualize modular pipeline.
```
lorenabalan:
```suggestion
- Added `--pipeline` option to visualize modular pipeline.
```

package/tests/test_server.py
```diff
@@ -243,6 +255,79 @@ def test_nodes_endpoint(cli_runner, client):
     assert data == EXPECTED_PIPELINE_DATA
 
 
+@pytest.mark.usefixtures("patched_get_project_context")
+def test_pipeline_flag(cli_runner, client):
+    """Test that running viz with `--pipeline` flag will return a correct pipeline"""
+    cli_runner.invoke(server.commands, ["viz", "--pipeline", "another"])
```
lorenabalan: Could you add a test for trying to load a pipeline that doesn't exist?

package/kedro_viz/server.py
```diff
@@ -229,8 +244,29 @@ def _call_viz(host=None, port=None, browser=None, load_file=None, save_file=None
...
         try:
+            # Kedro 0.15.0+
+            from kedro.context import KedroContextError
+
+            context = get_project_context()
+            try:
+                # Kedro 0.15.2+
+                pipeline = context._get_pipeline(  # pylint: disable=protected-access
+                    name=pipeline_name
+                )
```
lorenabalan: This in turn can raise `KedroContextError` if the pipeline `pipeline_name` doesn't exist. But this is caught further down and an unrelated messages is printed ("could not find a Kedro project root").

package/kedro_viz/server.py
```diff
@@ -229,8 +244,29 @@ def _call_viz(host=None, port=None, browser=None, load_file=None, save_file=None
...
+            except NotImplementedError:
+                # Kedro 0.15.0 or 0.15.1
+                if pipeline_name:
+                    raise KedroCliError(
+                        "`--pipeline` flag was provided, but there is no multiple pipelines."
```
lorenabalan: Maybe
```suggestion
                        "`--pipeline` flag was provided, but there is not support for multiple pipelines for your Kedro version."
```
or sth along those lines?

package/tests/test_server.py
921kiyo: Good point! I will add an extra test and also catch the KeyError in `except`.

package/kedro_viz/server.py
921kiyo: Thanks! I will change the message to be more general (to cover KeyError and Not ImplementedError).

package/kedro_viz/server.py
921kiyo: Thanks! Makes sense, I will raise a different message for this case. `try/except` is getting quite messy on this function to do backward-compatiblity....

package/kedro_viz/server.py
```diff
@@ -228,16 +243,43 @@ def _call_viz(host=None, port=None, browser=None, load_file=None, save_file=None
                 click.echo("Invalid file, top level key '{}' not found.".format(key))
```
DmitriiDeriabinQB: Line 241 above:
```python
for key in {"nodes", "edges", "tags"} - data.keys():
```

DmitriiDeriabinQB: I believe it's better to raise `kedro.cli.utils.KedroCliError` (or `ClickException` if you don't want to rely on kedro exception) rather than exiting yourself.

package/kedro_viz/server.py
```diff
@@ -213,12 +213,27 @@ def commands():
...
+    host=None,
+    port=None,
+    browser=None,
+    load_file=None,
+    save_file=None,
+    pipeline_name=None,
+):
     global data  # pylint: disable=global-statement,invalid-name
 
     if load_file:
```
DmitriiDeriabinQB: I think the whole section can be refactored into a separate function (like `data = _load_from_file(load_file)`)

package/kedro_viz/server.py
```diff
@@ -228,16 +243,43 @@ def _call_viz(host=None, port=None, browser=None, load_file=None, save_file=None
                 click.echo("Invalid file, top level key '{}' not found.".format(key))
                 sys.exit(1)
     else:
+        error_project_root = (
```
DmitriiDeriabinQB: This can be refactored into a module level constant

package/kedro_viz/server.py
```diff
@@ -228,16 +243,43 @@ def _call_viz(host=None, port=None, browser=None, load_file=None, save_file=None
                 click.echo("Invalid file, top level key '{}' not found.".format(key))
                 sys.exit(1)
     else:
```
DmitriiDeriabinQB: I suggest something like:

```python
try:
    context = get_project_context("context")
except KeyError:
    # Kedro <0.15.0
    try:
        pipeline = get_project_context("create_pipeline")()
        catalog = get_project_context("create_catalog")(None)
    except KeyError:
        raise KedroCliError(error_project_root)
else:
    pipeline = _get_pipe_from_ctx(context, pipeline_name)
    catalog = context.catalog

# ...
def _get_pipe_from_ctx(context, pipeline_name):  
    if pipeline_name and not hasattr(context, "_get_pipeline"):
        raise KedroCliError(
            "`--pipeline` flag was provided, but the context is old and rusty and doesn't know how to do things blah-blah"
        )
    
    return context._get_pipeline(name=pipeline_name)
    
```

package/features/environment.py
```diff
@@ -76,39 +76,22 @@ def call(cmd, verbose=False):
     path = [p for p in path if not (Path(p).parent / "pyvenv.cfg").is_file()]
     path = [p for p in path if not (Path(p).parent / "conda-meta").is_dir()]
     path = [str(bin_dir)] + path
+    # Activate environment
     context.env["PATH"] = path_sep.join(path)
 
-    # install this plugin by resolving the requirements using pip-compile
```
921kiyo: This block was not being used, so I removed it.

package/features/environment.py
```diff
@@ -39,7 +39,7 @@
 from features.steps.util import create_new_venv
 
 
-def before_all(context):
+def before_scenario(context, feature):  # pylint: disable=unused-argument
```
921kiyo: I moved all code in `before_all` into `before_scenario` (I know it's slowing down the CI), and the same for `after_all` to `after_scenario`, because Scenario 0.14.0 (or 0.14.3)  was making Scenario 0.15.0 and latest failing. In theory, each scenario should not interfere with each other, but `venv` was being shared across scenarios, which might be the reason why it was failing.

If anyone has a better idea, I'm happy to change this :)

RELEASE.md
```diff
@@ -1,16 +1,17 @@
-# Release 3.1.2:
+# Upcoming release:
 
 ## Major features and improvements
-- Promote viz from project specific commands to global commands.
-- Enable `kedro viz --load-file` outside of a Kedro project.
+- Promoted viz from project specific commands to global commands (#91)
+- Enabled `kedro viz --load-file` outside of a Kedro project (#91)
+- Added `--pipeline` option to visualize modular pipeline (#93)
+- Added `--env` option to pass configuration environment (#93)
```
921kiyo: `--env` was introduced to make Kedro-viz backward-compatible with Kedro<0.15.0.

package/kedro_viz/server.py
```diff
@@ -113,6 +118,31 @@ def run_viz(port=None, line=None) -> None:
...
+
+def _get_pipeline_from_context(context, pipeline_name):
+    if pipeline_name and hasattr(context, "_get_pipeline"):
+        # Kedro >=0.15.2
+        return context._get_pipeline(  # pylint: disable=protected-access
+            name=pipeline_name
+        )
+    if pipeline_name:
+        raise KedroCliError(
+            "`--pipeline` flag was provided, but the specified pipeline {} "
```
DmitriiDeriabinQB: I think this error message is a bit misleading - it fails not because the pipeline is not found, but rather because the context does not support named pipelines

package/kedro_viz/server.py
```diff
@@ -213,31 +244,66 @@ def commands():
     type=click.Path(dir_okay=False, writable=True),
     help="Path to save the pipeline JSON file",
 )
-def viz(host, port, browser, load_file, save_file):
+@click.option(
+    "--pipeline",
+    type=str,
+    default=None,
+    help="Name of the modular pipeline to run."
```
DmitriiDeriabinQB:
```suggestion
    help="Name of the modular pipeline to run. "
```

package/kedro_viz/server.py
```diff
@@ -213,31 +244,66 @@ def commands():
...
-            )
+            context = get_project_context("context", env=env)
+        # In Kedro<0.15.0, `get_project_context` does not have **kwargs,
+        # so `env=env` fails with TypeError. Or it fails with `context` keyword with KeyError.
+        except (KeyError, TypeError):
+            # Kedro <0.15.0
+            try:
+                pipeline = get_project_context("create_pipeline")()
+                get_config = get_project_context("get_config")
+                conf = get_config(Path.cwd(), env)
```
DmitriiDeriabinQB:
```suggestion
                conf = get_config(str(Path.cwd()), env)
```

package/kedro_viz/server.py
```diff
@@ -213,31 +244,66 @@ def commands():
...
-                "You can run `kedro viz` by either providing `--load-file` flag with a JSON file "
-                "path for your pipeline, or if the current working directory is "
-                "the root of a Kedro project."
-            )
+            context = get_project_context("context", env=env)
+        # In Kedro<0.15.0, `get_project_context` does not have **kwargs,
+        # so `env=env` fails with TypeError. Or it fails with `context` keyword with KeyError.
+        except (KeyError, TypeError):
+            # Kedro <0.15.0
+            try:
```
DmitriiDeriabinQB: Can you refactor the logic for Kedro 0.14.* in another helper function please?
So many levels of nesting are usually error prone and not good for the readability either.

package/kedro_viz/server.py
```diff
@@ -213,31 +244,66 @@ def commands():
...
-            pipeline = get_project_context("create_pipeline")()
-            catalog = get_project_context("create_catalog")(None)
-        except KedroContextError:
-            raise KedroCliError(
-                "Could not find a Kedro project root. "
-                "You can run `kedro viz` by either providing `--load-file` flag with a JSON file "
-                "path for your pipeline, or if the current working directory is "
-                "the root of a Kedro project."
-            )
+            context = get_project_context("context", env=env)
```
DmitriiDeriabinQB: A thought that just can to me - why do we need this complex logic in the first place if `get_project_context` is a part of the library? You could simply read the current version of Kedro and build the logic based on that, couldn't you?

```python
if kedro.__version__ >= "0.15.2":
    # do this
elif kedro.__version__ >= "0.15.0":
    # do that
else:
    # do something else
```

package/kedro_viz/server.py
921kiyo: Do we want this for all click options?

package/kedro_viz/server.py
921kiyo: That was the original plan, but installed kedro version in your conda env does not necessary match with the project version. Also `context.project_version` does not necessary match with the installed kedro version. So we decided to to this try/except complex logic.

package/kedro_viz/server.py
DmitriiDeriabinQB: No, help string here is split into 2 lines, the first one lacks the space, otherwise both of them will be concatenated without a space in between the sentences.

package/kedro_viz/server.py
DmitriiDeriabinQB: You are running the project with the version of Kedro installed in your current interpreter. `get_project_context` sits in the library, which is indeed updated when you update Kedro. If you updated Kedro and did not update your project, `get_project_context` will break anyways since it makes certain assumptions about the project structure.

921kiyo: Aw, I see! I will change it to the above simpler one. That's much better.

package/kedro_viz/server.py
```diff
@@ -55,6 +55,20 @@
     __name__, static_folder=str(Path(__file__).parent.absolute() / "html" / "static")
 )
 
+ERROR_PROJECT_ROOT = (
+    "Could not find a Kedro project root. "
+    "You can run `kedro viz` by either providing `--load-file` flag with a JSON file "
+    "path for your pipeline, or if the current working directory is "
```
lorenabalan: Maybe " `--load-file` flag with a filepath to a JSON pipeline representation, ..."

package/kedro_viz/server.py
```diff
@@ -55,6 +55,20 @@
...
+    "the root of a Kedro project."
+)
+
+ERROR_PIPELINE_FLAG_NOT_SUPPORTED = (
+    "`--pipeline` "
+    "flag was provided, but it is not supported in Kedro version {}".format(
```
lorenabalan: OCD comment but any reason not to make the two lines a bit more even? 😂

package/kedro_viz/server.py
```diff
@@ -213,31 +265,62 @@ def commands():
...
+            try:
+                context = get_project_context("context", env=env)
+                pipeline = _get_pipeline_from_context(context, pipeline_name)
+            except KedroContextError:
+                raise KedroCliError(ERROR_PROJECT_ROOT)
+            catalog = context.catalog
+
+        else:
+            # Kedro 0.14.*
+            pipeline, catalog = _get_pipline_catalog_from_kedro14(env, pipeline_name)
```
lorenabalan:
```suggestion
            pipeline, catalog = _get_pipeline_catalog_from_kedro14(env, pipeline_name)
```

package/kedro_viz/server.py
```diff
@@ -113,6 +127,43 @@ def run_viz(port=None, line=None) -> None:
...
+            return context._get_pipeline(  # pylint: disable=protected-access
+                name=pipeline_name
+            )
+    else:  # Kedro 0.15.0 or 0.15.1
+        if pipeline_name:
+            raise KedroCliError(ERROR_PIPELINE_FLAG_NOT_SUPPORTED)
+    return context.pipeline
+
+
+def _get_pipline_catalog_from_kedro14(env, pipeline_name):
```
lorenabalan:
```suggestion
def _get_pipeline_catalog_from_kedro14(env, pipeline_name):
```

package/kedro_viz/server.py
```diff
@@ -113,6 +127,43 @@ def run_viz(port=None, line=None) -> None:
...
+        if key not in data:
+            raise KedroCliError(
+                "Invalid file, top level key '{}' not found.".format(key)
+            )
+    return data
+
+
+def _get_pipeline_from_context(context, pipeline_name):
+    if match(kedro.__version__, ">=0.15.2"):
+        if pipeline_name:
```
lorenabalan: Is there any point in checking `pipeline_name`? `_get_pipeline()` knows how to handle the `None`s.

lorenabalan: If we indeed don't need it, the `else` will go away too, but keep the if-raise and final return.

package/tests/test_server.py
```diff
@@ -243,6 +261,128 @@ def test_nodes_endpoint(cli_runner, client):
...
+        }[key]
+
+    mock.patch("kedro_viz.server.get_project_context", new=get_project_context).start()
+    result = cli_runner.invoke(server.commands, "viz")
+    assert result.exit_code == 0, result.output
+
+
+@mock.patch("kedro.__version__", "0.14.0")
+def test_viz_kedro14_pipeline_flag(cli_runner):
+    """Test that running viz with `--pipeline` flag in Kedro 0.14.0."""
```
lorenabalan:
```suggestion
    """Test that running viz with `--pipeline` flag in Kedro 0.14.0 is not supported."""
```

package/tests/test_server.py
```diff
@@ -243,6 +261,128 @@ def test_nodes_endpoint(cli_runner, client):
...
+
+    mock.patch("kedro_viz.server.get_project_context", new=get_project_context).start()
+    result = cli_runner.invoke(server.commands, ["viz", "--pipeline", "another"])
+    assert "`--pipeline` flag was provided" in result.output
+
+
+@mock.patch("kedro.__version__", "0.14.0")
+def test_viz_kedro14_invalid(cli_runner):
+    """Test that running viz in Kedro 0.14.0,
+    and it is outside of a Kedro project root."""
```
lorenabalan:
```suggestion
    while outside of a Kedro project is not supported."""
```

package/tests/test_server.py
```diff
@@ -243,6 +261,128 @@ def test_nodes_endpoint(cli_runner, client):
...
+def test_viz_kedro15_invalid(cli_runner):
+    """Test that running viz in Kedro 0.15.0,
+    and it is outside of a Kedro project root."""
+
+    def get_project_context(key: str, **kwargs):
+        raise KedroContextError
+
+    mock.patch("kedro_viz.server.get_project_context", new=get_project_context).start()
+    result = cli_runner.invoke(server.commands, "viz")
+    assert "Could not find a Kedro project root." in result.output
```
lorenabalan: Just realised we're using `unittest` for mocking instead of `pytest`. Could we use `mocker.patch` instead of mock to make it consistent?
We start a patch but never stop it, so I assume it's fine to just do mocker.patch("this")

RELEASE.md
```diff
@@ -1,16 +1,17 @@
...
 
 ## Bug fixes and other changes
 
 # Release 3.1.0:
 
 ## Major features and improvements
 
-**BREAKING CHANGE:**  Kedro<0.15.0 no longer works with this version of Kedro-Viz(#72)
```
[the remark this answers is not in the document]
921kiyo: It was introduced in https://github.com/quantumblacklabs/kedro-viz/pull/72 (the line `catalog = get_project_context("create_catalog")(None)`), but when the PR was merged, we did not notice that it was breaking the compatibility with Kedro 0.14.*, and 3.1.0 was released (in Kedro 0.14.*, `create_catalog` needs an argument `conf`, but `None` was passed). 

So I added this line quickly 5 days ago https://github.com/quantumblacklabs/kedro-viz/pull/99 (and pinned kedro >=0.15.0). 

But I think you are right, I should not remove this line since we cannot undo the release 3.1.0. Instead I will say "the breaking change was fixed and it is not backward-compatible with Kedro 0.14.*" in the upcoming release note.

package/kedro_viz/server.py
921kiyo: Thanks! I simplified the checking and removed `else` :)

package/tests/test_server.py
921kiyo: As discussed offline, I will address mock -> mocker stuff in a separate PR :)

package/kedro_viz/server.py
```diff
@@ -213,31 +261,62 @@ def commands():
     type=click.Path(dir_okay=False, writable=True),
     help="Path to save the pipeline JSON file",
 )
-def viz(host, port, browser, load_file, save_file):
+@click.option(
+    "--pipeline",
+    type=str,
+    default=None,
+    help="Name of the modular pipeline to run. "
```
idanov:
```suggestion
    help="Name of the modular pipeline to visualize. "
```

package/kedro_viz/server.py
```diff
@@ -213,31 +261,62 @@ def commands():
...
+                pipeline = _get_pipeline_from_context(context, pipeline_name)
+            except KedroContextError:
+                raise KedroCliError(ERROR_PROJECT_ROOT)
+            catalog = context.catalog
+
+        else:
+            # Kedro 0.14.*
+            pipeline, catalog = _get_pipeline_catalog_from_kedro14(env, pipeline_name)
+
         data = format_pipeline_data(pipeline, catalog)
```
idanov: Why is the catalog needed to visualise the pipeline?

package/tests/test_server.py
```diff
@@ -140,6 +145,19 @@ def teardown_function():
...
+        mocked_context._get_pipeline = get_pipeline  # pylint: disable=protected-access
+        mocked_context.catalog = lambda x: None
+        mocked_context.pipeline = create_pipeline()
+        return {
+            "create_pipeline": create_pipeline,
+            "create_catalog": lambda x: None,
+            "context": mocked_context,
+        }[key]
+
     mock.patch("kedro_viz.server.get_project_context", new=get_project_context).start()
```
DmitriiDeriabinQB:
```suggestion
    mocker.patch("kedro_viz.server.get_project_context", new=get_project_context)
```
Just flagging - ignore it here if you plan to address in a separate PR

package/tests/test_server.py
```diff
@@ -140,6 +145,19 @@ def teardown_function():
 
 @pytest.fixture
 def patched_get_project_context():
```
DmitriiDeriabinQB:
```suggestion
def patched_get_project_context(mocker):
```

package/kedro_viz/server.py
921kiyo: @idanov It's not necessary at the moment, but catalog was added in https://github.com/quantumblacklabs/kedro-viz/pull/72

This is Gordon's comment. 

Screenshot 2020-02-06 at 14 02 26

package/kedro_viz/server.py
```diff
@@ -113,6 +124,42 @@ def run_viz(port=None, line=None) -> None:
...
+            name=pipeline_name
+        )
+    # Kedro 0.15.0 or 0.15.1
+    if pipeline_name:
+        raise KedroCliError(ERROR_PIPELINE_FLAG_NOT_SUPPORTED)
+    return context.pipeline
+
+
+def _get_pipeline_catalog_from_kedro14(env, pipeline_name):
+    if pipeline_name:
```
DmitriiDeriabinQB: I think this check can be handled outside of this function, the argument doesn't really belong here.

package/tests/test_server.py
```diff
@@ -102,6 +103,14 @@
 }
 
 
+def get_pipeline(name: str = None):
```
DmitriiDeriabinQB:
```python
def get_pipeline(name: str = "__default__"):
    if name == "__default__":
        return create_pipeline()
    if name == "another":
        return Pipeline([])
    raise KeyError("Failed to find the pipeline.")
```

package/tests/test_server.py
```diff
@@ -243,6 +261,128 @@ def test_nodes_endpoint(cli_runner, client):
...
+
+@mock.patch("kedro.__version__", "0.15.0")
+def test_viz_kedro15(cli_runner):
+    """Test that running viz in Kedro 0.15.0."""
+
+    def get_project_context(
+        key: str = "context", **kwargs  # pylint: disable=bad-continuation
+    ):  # pylint: disable=unused-argument
+        mocked_context = mock.Mock()
+        mocked_context.catalog = lambda x: None
```
DmitriiDeriabinQB: I think this one may be redundant

package/tests/test_server.py
```diff
@@ -243,6 +261,128 @@ def test_nodes_endpoint(cli_runner, client):
...
+
+@mock.patch("kedro.__version__", "0.14.0")
+def test_viz_kedro14_invalid(cli_runner):
+    """Test that running viz in Kedro 0.14.0,
+    while outside of a Kedro project is not supported."""
+
+    def get_project_context(key: str, **kwargs):
+        raise KeyError
+
+    mock.patch("kedro_viz.server.get_project_context", new=get_project_context).start()
```
DmitriiDeriabinQB:
```suggestion
    mock.patch("kedro_viz.server.get_project_context", side_effect=KeyError).start()
```

package/kedro_viz/server.py
```diff
@@ -31,19 +31,19 @@
...
 
 import click
+import kedro
 import requests
 from flask import Flask, jsonify, send_from_directory
 from IPython.core.display import HTML, display
-from kedro.cli import get_project_context
-from kedro.cli.utils import KedroCliError
-from kedro.context import KedroContextError
+from kedro.cli import get_project_context  # pylint: disable=ungrouped-imports
```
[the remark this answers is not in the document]
921kiyo: Because `import kedro` was added above, but isort wants it in line 40...

package/features/viz.feature
```diff
@@ -28,17 +28,30 @@
...
+    Background:
+        Given I have prepared a config file with example code
+
+    Scenario: Execute viz with Kedro 0.14.0
+        Given I have installed kedro version "0.14.0"
+        And I have run a non-interactive kedro new
+        When I execute the kedro viz command "viz"
+        Then kedro-viz should start successfully
+
+    # `load_context` function was in Kedro 0.14.3, but not KedroContext class
```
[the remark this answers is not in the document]
921kiyo: Sorry it wan't clear enough. I meant why I am adding a test for 0.14.3 and not 0.14.1 nor 0.14.2 (because of `load_context` was introduced in 0.14.3).

package/tests/test_server.py
921kiyo: I meant to mock the original `get_pipeline` implementation (not exactly the same though), and I think the current one looks closer to the original one :)

package/tests/test_server.py
921kiyo: Nice, I didn't know about this. Thanks :)

PR #96, opened 2020-01-28
[KED-1046] Redesign sidebar list to group nodes by type

richardwestenra:
## Description

- Group nodes in node-list by type
- Add headings for each group, with a checkbox to toggle all nodes of that type
- Add accordion/dropdown toggles to allow expanding/collapsing each section
- Remove combined/node/data view radio buttons (as they are now redundant)
- Remove parameters toggle (as it is now redundant)

This design change simplifies the sidebar logic, and should make it more intuitive and concise

image

## Notice

## Checklist


package.json
```diff
@@ -38,6 +38,7 @@
     "dagre": "^0.8.5",
     "konami-code": "^0.2.1",
     "react-custom-scrollbars": "^4.2.1",
+    "react-flip-toolkit": "^7.0.7",
```
ottis: Is this dependency needed i.e. could we do it in vanilla JS (or does this just make it super trivial to implement?) - thinking build size, giving performance over to another library etc.

src/actions/actions.test.js
```diff
@@ -1,41 +1,30 @@
...
   TOGGLE_NODE_HOVERED,
-  TOGGLE_PARAMETERS,
   TOGGLE_TAG_ACTIVE,
   TOGGLE_TAG_FILTER,
   TOGGLE_THEME,
+  TOGGLE_TYPE_DISABLED,
   TOGGLE_TEXT_LABELS,
   UPDATE_CHART_SIZE,
-  changeView,
+  UPDATE_FONT_LOADED,
```
ottis: Are the consts exported just for tests or used elsewhere (just a silly question)

src/components/app/index.js
```diff
@@ -25,7 +24,7 @@ class App extends React.Component {
 
   componentDidUpdate(prevProps) {
     if (prevProps.data.schema_id !== this.props.data.schema_id) {
-      this.resetStoreData(formatData(this.props.data));
+      this.resetStoreData(loadData(this.props.data));
```
ottis: Can you explain the change here i.e. difference between formatData and loadData?

package.json
richardwestenra: I hate to introduce a new library unnecessarily, but this takes something that'd be pretty complicated to implement, and makes it trivial. i think it's worthwhile.

src/actions/actions.test.js
richardwestenra: The consts are used in the reducer

src/components/node-list/index.js
```diff
@@ -34,10 +31,17 @@ class NodeList extends React.Component {
    * @return {object} The results array with a new field added
    */
   highlightMatch(results) {
-    return results.map(result => ({
-      highlightedLabel: getHighlightedText(result.name, this.state.searchValue),
-      ...result
-    }));
+    const obj = {};
+    Object.keys(results).forEach(key => {
+      obj[key] = results[key].map(result => ({
```
ottis: Should this guard against results[key] not being an array?

src/components/app/index.js
richardwestenra: In this case, `loadData` just calls `formatData`, so it does the same thing. But it meant I could remove an extra dependency, and make the usage more consistent

src/components/node-list/index.js
```diff
@@ -54,11 +58,26 @@ class NodeList extends React.Component {
...
+      );
+    });
+    return obj;
+  }
+
+  /**
+   * Get a list of IDs of the visible nodes
+   * @param {object} filteredNodes
+   */
+  getNodeIDs(filteredNodes) {
```
ottis: This seems more of a helper / mapper function rather than needing to belong to the class itself, have you considered moving this out of the class? (would be good to have some test coverage of this as well)

src/components/node-list/index.js
```diff
@@ -54,11 +58,26 @@ class NodeList extends React.Component {
...
+    return Object.keys(filteredNodes).reduce(
+      (nodeIDs, key) => nodeIDs.concat(filteredNodes[key].map(node => node.id)),
+      []
```
ottis: If there another way to write this or could a comment be penned on whats happening here? Think it looks quite complex with the nested concat + map 😨

src/components/node-list/index.js
```diff
@@ -83,15 +102,14 @@ class NodeList extends React.Component {
...
-      onToggleNodeDisabled,
-      nodes,
-      theme
-    } = this.props;
+    const { hasData, nodes, theme } = this.props;
     const { searchValue } = this.state;
-    const formattedNodes = this.highlightMatch(this.filterResults(nodes));
+    if (!hasData) {
+      return null;
+    }
```
ottis: Should this case be considered here i.e. could this be pushed further up the chain, or does / could this component be used outside the context of kedro viz

src/components/node-list/index.js
richardwestenra: Yeah that's fair. I could move it out of the class, but it's basically only used on the class's state, and I don't think it'd be useful to abstract it at this point. Personally I think it doesn't really matter either way whether it's a class method or a standalone function, so I kept it as a method for neatness, so that it's kept with the other related functions. Do you disagree? If so I don't mind moving it.

The test coverage is a good shout. I'll add that

src/components/node-list/index.js
richardwestenra: I don't think there's ever a case where it wouldn't be. I can add a check but I'm not sure that it's necessary

src/components/node-list/index.js
richardwestenra: I've rewritten it to separate it out into different named functions, so that there isn't so much complex logic going on in just a couple of lines

src/components/node-list/index.js
richardwestenra: I added it here intentionally, to prevent things from breaking when no data is provided, or when it's still loading. But it seems to be running fine now when I remove it? I'll doublecheck but it looks like it may no longer be necessary

src/components/app/index.js
ottis: Does load data actually request data or is it more mapping data to a structure?

ottis: (Maybe the naming could be changed what do you think?)

src/components/node-list/index.js
ottis: 👍

src/components/node-list/index.js
ottis: Maybe it's something that can still live in this file (at the top) and can be exported to have some test coverage around it

src/components/node-list/index.js
ottis: 👍

src/components/app/index.js
richardwestenra: `loadData` determines how data should be loaded (i.e. async from JSON, or imported from a JS file, or randomly-generated, or directly via props), then formats it and returns it.

Maybe `getData` (or similar) would be a better name, idk?

https://github.com/quantumblacklabs/kedro-viz/blob/develop/src/components/app/load-data.js

src/components/node-list/index.js
ottis: 🙂

src/components/node-list/node-list.scss
```diff
@@ -101,11 +101,89 @@
...
+  margin: -$size/2 0 0;
+  border-radius: 50%;
+  font-size: 1.5em;
+  background: none;
+  border: none;
+  box-shadow: none;
+  color: inherit;
+  font-family: inherit;
+  cursor: pointer;
+  transition: transform ease 0.3s;
```
ottis: Can this be structured a bit more e.g. outside in ordering

```
.declaration-order {
  /* Positioning */
  position: absolute;
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
  z-index: 10;
 
  /* Box Model */
  display: block;
  float: right;
  width: 100px;
  height: 100px;
  margin: 10px;
  padding: 10px;
 
  /* Typography */
  color: #888;
  font: normal 16px Helvetica, sans-serif;
  line-height: 1.3;
  text-align: center;
 
  /* Visual */
  background-color: #eee;
  border: 1px solid #888;
  border-radius: 4px;
  opacity: 1;
 
  /* Animation */
  transition: all 1s;
 
  /* Misc */
  user-select: none;
}
```

This can do auto done by stylelint. Would be great to structure these consistently so it's not a big chunk of styling (harder to approach when changing in future)

src/components/node-list/node-list.test.js
```diff
@@ -9,7 +10,12 @@ describe('NodeList', () => {
     const search = wrapper.find('.pipeline-node-list-search');
     const nodeList = wrapper.find('.pipeline-node-list');
     expect(search.length).toBe(1);
-    expect(nodeList.length).toBe(1);
+    expect(nodeList.length).toBeGreaterThan(0);
+  });
+
+  it('renders null if hasData is falsey', () => {
+    const wrapper = setup.mount(<NodeList />, loadData([]));
+    expect(wrapper.html()).toBeFalsy();
```
ottis: What happens if loadData is supplied with undefined or null? (same expect or throws error)

src/components/node-list/node-list.scss
richardwestenra: 🤷‍♂  I've tried to order these in a way that made sense to me, more or less. e.g. I put positioning at the top, then display, etc. I think it's broadly fine as is? Happy to introduce stylelint to do it automatically, but that can be a separate PR. In the meantime I don't think it's worth doing manually.

src/components/node-list/node-list.test.js
richardwestenra: Throws error

src/reducers/index.js
```diff
@@ -51,60 +38,54 @@ function reducer(state = {}, action) {
...
         theme: action.theme
       });
     }
 
+    case TOGGLE_TYPE_DISABLED: {
+      return updateState({
+        typeDisabled: Object.assign({}, state.typeDisabled, {
+          [action.typeID]: action.disabled
+        })
+      });
```
ottis: The object assign here makes it difficult for me to understand what you want to update, could it change to something like (or is it because you've used object assign in updateState)

```
case TOGGLE_TYPE_DISABLED: {
      return updateState({
        typeDisabled: {
          ...state.typeDisabled,
          [action.typeID]: action.disabled
        }
      })
```

src/selectors/nodes.js
```diff
@@ -113,10 +109,26 @@ export const getNodeData = createSelector(
...
 
+/**
+ * Returns formatted nodes grouped by type
+ */
+export const getGroupedNodes = createSelector(
+  [getNodeData],
+  nodes =>
+    nodes.reduce(function(obj, item) {
+      const key = item.type;
+      if (!obj.hasOwnProperty(key)) {
```
ottis: Does this need hasOwnProperty or can it simple be

```
if (!obj[key])
```

richardwestenra: It probably could, but this is safer as it will still create keys for falsey values like empty strings

src/reducers/index.js
richardwestenra: I've tried to keep to using Object.assign over spread in the reducer. Honestly, I find Object.assign slightly easier to read, but I think it doesn't matter much either way? Seeing as that's what I've used throughout the rest of the reducer, and it's more common to use it in reducers generally, I'd prefer to keep it as is for consistency.

src/components/node-list/index.js
richardwestenra: done

src/components/app/index.js
ottis: Ah ok - reading the method loadData it seems like it's doing different things based on the parameter being passed to the method (sometimes passing a string / others an array). It sometimes loads from a mock data file, other times will load from a specific path, and lastly will format the data array if you supply it that structure.

To me it seems like this method is doing a lot and isn't very specific in its use, have you considered breaking this down a bit or is it more convenience? (Just something to consider rather than a blocker on this PR)

src/components/node-list/node-list.scss
ottis: I'll create an issue to do it automatically, we should have consistency across the style ordering and structure 👍

package.json
ottis: Cool 👍 Any idea on the build size increase from this or is it minimal?

src/actions/actions.test.js
ottis: 👍

src/components/node-list/node-list.scss
ottis: Issue here - https://github.com/quantumblacklabs/kedro-viz/issues/102

src/components/node-list/node-list.test.js
ottis: 👍

package.json
richardwestenra: 22K (gzipped 7K)

It's also performing some work in lieu of `react-transition-group`, so it's got that going for it, which is nice

src/reducers/index.js
ottis: Cool cool - maybe for me it slightly breaks the focus when reading a case when the updateState has been moved to the top (mental model for nesting object assigns located in different locations)

src/components/app/index.js
richardwestenra: Yeah I get that, but it's kind of necessary to have this switch statement to determine what data source to use dependent on the argument it gets passed, as this is the way that the application chooses a data source, and while most of them are synchronous, one is async. I don't think I'd change anything but perhaps the name, though it doesn't bother me hugely.

src/selectors/nodes.js
ottis: Why is it safer though? I could see it in use if it was a class / function that had properties inside it rather than on the prototype but afaik nodes is an array of objects?:

```
function A() {
   this.x = "I'm an own property";
}

A.prototype.y = "I'm not an own property";

var instance = new A();
var xIsOwnProperty = instance.hasOwnProperty("x"); // true
var yIsOwnProperty = instance.hasOwnProperty("y"); // false
```

```
var obj = {}
if (!obj['']) console.log('hello') // logs "hello"
if (!obj[null]) console.log('hello') // logs "hello"
if (!obj[undefined]) console.log('hello') // logs "hello"
```

It works either way but just wondering 🙂

richardwestenra: I meant if the property value itself were falsey.

e.g.
```
const obj = {
 foo: 0
};
console.log(!obj['foo']); // true
console.log(!obj.hasOwnProperty('foo')); // false
```

I figured it didn't matter too much either way, but this way is slightly more explicit and potentially safer as it's only checking for whether the key is present. I don't feel strongly about it but I'm pretty sure it doesn't matter?

package.json
ottis: 👍 Nice

src/components/app/index.js
ottis: Think it's more breaking the single responsibility idea but happy to take outside this PR 👍

src/selectors/nodes.js
ottis: Cool happy to leave as it is then

ottis:
LGTM

PR #106, opened 2020-02-10
[KED-1374] Refactor store

richardwestenra:
## Description

Move store/state config from the App component to the store directory and rename some files & functions

## Development notes

Also simplifies some of the state mocking functions used in testing

## QA notes

No functionality should have changed

## Legal notice

## Checklist


src/store/index.js
```diff
@@ -1,14 +1,33 @@
 import { createStore } from 'redux';
 import reducer from '../reducers';
-import { saveState } from '../utils';
+import { resetData } from '../actions';
+import getInitialState from './initial-state';
+import loadData from './load-data';
+import { saveState } from './helpers';
+import checkFontLoaded from './check-font-loaded';
+
+export default function configureStore(props) {
```
ottis: Can you outline the methods functionality in a JSDoc (particularly outlining the expected data to be passed to the method)? Also is there any expected names in the object being passed e.g. "data" is one - I ask because maybe this method could be updated like:

```
export default function configureStore({
  data,
  someOtherProp,
  anotherAwesomeProp,
  etcProp
}) {
```

src/components/app/index.js
```diff
@@ -16,24 +14,21 @@ import './app.css';
...
 
   /**
    * Dispatch an action to update the store with new pipeline data
-   * @param {Object} formattedData Normalised state data
    */
-  resetStoreData(formattedData) {
-    this.store.dispatch(resetData(formattedData));
+  resetStoreData() {
+    const normalizedData = loadData(this.props.data);
+    this.store.dispatch(resetData(normalizedData));
```
ottis: I wonder if "normalizedData" could have a better name? I ask because normalising data is usually something like scaling a range of values but in this case loadData is just mapping the data back to a workable structure, what do you think?

src/store/index.js
```diff
@@ -1,14 +1,33 @@
...
+   * Dispatch an action to update the store with new pipeline data
+   * @param {Object} normalizedData Normalised state data
+   */
+  const resetStoreData = normalizedData => {
+    if (store) {
+      store.dispatch(resetData(normalizedData));
+    }
+  };
+
+  const pipelineData = loadData(props.data, resetStoreData);
```
ottis: It seems there's a bit of a mismatch in how data is injected into the store i.e. inside configureStore it's actually loading data via loadData in configureStore but then in the reset data it's triggering an action with some data, what i'd recommend is to consolidate the mental model with data in the store e.g.

- Remove onLoadData callback from loadData as it only happens in one case of the switch and feels like a convenience callback rather than something that adding value to the flow (I still think loadData doing very different things based on whether you pass a string, array or nothing to it needs a rethink but not for now 🙂 
- Remove loadData call from inside configureStore - It should focus purely on store configuring rather than making decisions itself
- Pass in fully formed initial data into configureStore from App i.e. loadData in app

```

```

src/store/index.js
```diff
@@ -1,14 +1,33 @@
...
-export default function configureStore(initialState) {
   const store = createStore(reducer, initialState);
 
   store.subscribe(() => {
     const { textLabels, theme, typeDisabled } = store.getState();
     saveState({ textLabels, theme, typeDisabled });
   });
 
+  checkFontLoaded(store);
+
```
ottis: This seems like a view level check rather than something that should be in the store (the store doesn't care about fonts rather the application state) - I'd recommend moving to componentDidMount on App as a place for this check

src/store/load-data.js
```diff
@@ -1,10 +1,10 @@
...
-import formatData from '../../utils/format-data';
-import loremIpsum from '../../utils/data/lorem-ipsum.mock';
-import animals from '../../utils/data/animals.mock';
-import demo from '../../utils/data/demo.mock';
+import config from '../config';
+import getRandomPipeline from '../utils/random-data';
+import normalizeData from './normalize-data';
+import loremIpsum from '../utils/data/lorem-ipsum.mock';
+import animals from '../utils/data/animals.mock';
+import demo from '../utils/data/demo.mock';
```
ottis: One thing i've noticed in this refactor PR is there is a lot of changing paths which can be hard to manage, as a recommendation I'd look to move to setting baseUrl in jsconfig.json (for CRA)

https://create-react-app.dev/docs/importing-a-component/#absolute-imports
https://code.visualstudio.com/docs/languages/jsconfig

src/components/sidebar/sidebar.test.js
```diff
@@ -9,7 +10,7 @@ const mockProps = {
 
 describe('ShowMenuButton', () => {
   it('renders without crashing', () => {
-    const wrapper = setup.shallow(ShowMenuButton, mockProps);
+    const wrapper = setup.mount(<ShowMenuButton {...mockProps} />);
```
ottis: How come this needs to be mounted now?

src/store/check-font-loaded.js
```diff
@@ -1,4 +1,4 @@
-import { UPDATE_FONT_LOADED } from '../../actions';
+import { UPDATE_FONT_LOADED } from '../actions';
 
 /**
  * Prevent chart from displaying until the webfont has loaded,
```
ottis: I feel like this file shouldn't live in the store as it's not related to the store, maybe utils would be a better place?

src/components/app/index.js
richardwestenra: I think you're talking about [normalization in statistics](https://en.wikipedia.org/wiki/Normalization_(statistics)), as opposed to [database normalization](https://en.wikipedia.org/wiki/Database_normalization)?

richardwestenra: I named it this because I roughly followed [the guidelines in the Redux docs](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape/) on state normalization when designing the state when I added a Redux store. Admittedly my state is a lot flatter than the one in the docs.

src/store/index.js
richardwestenra: JSDoc is a good shout, thanks, will do.

Easiest to keep props intact as I'm just passing the whole thing to getInitialState and I think it's less brittle not to deconstruct then immediately reconstruct it?

src/store/index.js
richardwestenra: I'd love to pair with you on this because I agree that it's a little janky how it's handled currently, but I couldn't figure out a better way to structure the code while keeping the outward behaviour as is.

Here are some stipulations/assumptions about application usage:

1. The application can use a variety of different data sources:
    - test datasets (x2), imported directly via ES6 imports
    - demo dataset, imported directly via ES6 import
    - JSON file dataset at `/public/api/nodes.json`, imported asynchronously and parsed with `d3.json`
    - randomly-generated dataset, imported directly via ES6 import and generated at runtime
    - JS dataset object literal, passed as a prop to the App component
2. The application can be run in a variety of different contexts:
    - standalone JS app in development (which can run/test any of the different available data sources)
    - as a static build (usually served via a local Python server), which imports the JSON, which is served at `/api/nodes.json`
    - as a React component, imported into another React application, and passed the data as a prop.
3. When the application is run as a React component, it should update when passed new data via the `data` prop. This should only reset the chart data - there is no need to reset the rest of the state (e.g. the stuff set in `getInitialState`) when updating the chart data.
4. When loading the data asynchronously, the state will need to be initialised immediately, then updated once the data is loaded. This update is triggered with a Promise callback.

Because it can use data from different formats and in different contexts, I want to keep it simple, so I'm just exposing a single top-level `data` prop, and the application can figure out what to do with that by evaluating the value provided. This is what `loadData` does. I could name it something else but I haven't thought of a better name.

- If you remove the `onLoadData` callback then the state won't be updated when loading the data asynchronously, which means that in the main production use-case, no data will be displayed. We can structure this differently, but ultimately the callback needs to either initialise or update the state once it's triggered.

- I can split up configureStore into multiple functions, but it's tricky because `resetStoreData` needs access to the `store` object, but it also needs to be defined before it's created. It's a bit of a Gordian Knot. I don't think splitting it makes it much better tbh!

- Passing in fully-formed data into configureStore from App is what I had it doing initially, and it felt like too much was going on in App so I wanted to move this logic into `configureStore`! I don't see how placing this logic in `configureStore` makes much difference.

Basically, I don't think I can do what you're asking me to without breaking some of the required functionality.

richardwestenra: See [this commit](https://github.com/quantumblacklabs/kedro-viz/pull/106/commits/504989be386d78c34888ecbe3e28255c1b0d7ddb)

src/store/index.js
richardwestenra: Can do - I was just hoping to keep this stuff separated from App and avoid having App do too much. To me it seems more store-related, as it's not directly affecting rendering 🤷‍♂

src/store/load-data.js
richardwestenra: I'd be happy to revisit this at a later date if the project becomes more complex and deeply nested, but at this stage I'd prefer to avoid switching to absolute paths, in the interest of keeping the code environment simpler and more approachable for newbie devs - especially as it's open-source and we're inviting contributions.

src/components/sidebar/sidebar.test.js
richardwestenra: It doesn't really, except that the child SVG component was showing as 0% coverage in the test coverage report and lowering the total coverage score, so by changing it to mount, it resolves this without adding pointless extra tests for the SVG icons. I figured it didn't much matter either way so this way seemed the easiest.

Incidentally before you ask, the SVG image needs to be a JS component instead of a regular SVG because imports need to work in the JS component context and we can't count on consumers having the webpack-inline-svg Babel transform installed.

src/store/check-font-loaded.js
richardwestenra: Sure 🤷‍♂

src/components/app/index.js
ottis: So it doesn't really look like it's normalising the data either way, example from normalzr:

```
[
  {
    id: 1,
    title: 'Some Article',
    author: {
      id: 1,
      name: 'Dan'
    }
  },
  {
    id: 2,
    title: 'Other Article',
    author: {
      id: 1,
      name: 'Dan'
    }
  }
];
```

Can be normalized to:

```
{
  result: [1, 2],
  entities: {
    articles: {
      1: {
        id: 1,
        title: 'Some Article',
        author: 1
      },
      2: {
        id: 2,
        title: 'Other Article',
        author: 1
      }
    },
    users: {
      1: {
        id: 1,
        name: 'Dan'
      }
    }
  }
}
```

But in formatData (which is returning the object) - it's more mapping the loaded data than normalizing it into relational entities e.g.

```
/**
 * Add a new Tag if it doesn't already exist
 * @param {string} name - Default node name
 */
const addTag = state => tag => {
  const { id } = tag;
  state.tags.push(id);
  state.tagName[id] = tag.name;
};
```

Might be worth having a look at the approach here https://github.com/paularmstrong/normalizr/blob/master/docs/quickstart.md and if not the direction you want i'd rename this to avoid confusion about the structure of the state 👍

src/store/index.js
ottis: Happy to pair on this, essentially it's about separation of concerns with this function - it's doing too much (loadData + store data injection). We need to separate out the applications contexts because right now it's not obvious what happens and when it happens (and you're masking it by wrapping everything in the one method). This will become brittle over time without some structure in place. Happy to chat further and not block this PR 👍

src/components/app/index.js
richardwestenra: Yeah I appreciate that:
> "Admittedly my state is a lot flatter than the one in the docs."

I'd say that my state is following the approach of the normalized one you've post above, but the only difference is that mine is less deeply nested. e.g. instead of `entities.articles[1].id` I've got something like `articleID[1]`. It's still normalized, it's just not as nested so that updating the state is easier and I don't need to use ImmutableJS because everything is max 1-2 levels deep.

richardwestenra: From the Redux docs:
> The basic concepts of normalizing data are:
> - Each type of data gets its own "table" in the state.
> - Each "data table" should store the individual items in an object, with the IDs of the items as keys and the items themselves as the values.
> - Any references to individual items should be done by storing the item's ID.
> - Arrays of IDs should be used to indicate ordering.

My approach meets all four of these criteria

src/store/index.js
richardwestenra: Thanks. Yeah tbh in refactoring all I'm doing is shifting this logic back and forth between the App component and the store function. This also works:

```
class App extends React.Component {
  constructor(props) {
    super(props);
    const pipelineData = loadData(props.data, this.resetStoreData.bind(this));
    const initialState = getInitialState(pipelineData, props);
    this.store = configureStore(initialState);
  }
 ...
}
```
but now it's back to how it was before. I started this change because I was trying to move the state initialisation out of App while allowing it access to the store (which it needs for the reset)

src/store/index.js
ottis: ## Description

It's not related to store (it only manipulates the state through actions), it's an action firing to do some async check.

Looking at checkFontLoaded it is in fact a "thunk" i.e. an action that can have side effects (e.g. triggering other actions, async things etc) Example from redux documentation:

## Redux thunk example

https://redux.js.org/advanced/async-actions/

```
export const REQUEST_POSTS = 'REQUEST_POSTS'
function requestPosts(subreddit) {
  return {
    type: REQUEST_POSTS,
    subreddit
  }
}
export const RECEIVE_POSTS = 'RECEIVE_POSTS'
function receivePosts(subreddit, json) {
  return {
    type: RECEIVE_POSTS,
    subreddit,
    posts: json.data.children.map(child => child.data),
    receivedAt: Date.now()
  }
}
export const INVALIDATE_SUBREDDIT = 'INVALIDATE_SUBREDDIT'
export function invalidateSubreddit(subreddit) {
  return {
    type: INVALIDATE_SUBREDDIT,
    subreddit
  }
}
// Meet our first thunk action creator!
// Though its insides are different, you would use it just like any other action creator:
// store.dispatch(fetchPosts('reactjs'))
export function fetchPosts(subreddit) {
  // Thunk middleware knows how to handle functions.
  // It passes the dispatch method as an argument to the function,
  // thus making it able to dispatch actions itself.
  return function(dispatch) {
    // First dispatch: the app state is updated to inform
    // that the API call is starting.
    dispatch(requestPosts(subreddit))
    // The function called by the thunk middleware can return a value,
    // that is passed on as the return value of the dispatch method.
    // In this case, we return a promise to wait for.
    // This is not required by thunk middleware, but it is convenient for us.
    return fetch(`https://www.reddit.com/r/${subreddit}.json`)
      .then(
        response => response.json(),
        // Do not use catch, because that will also catch
        // any errors in the dispatch and resulting render,
        // causing a loop of 'Unexpected batch number' errors.
        // https://github.com/facebook/react/issues/6895
        error => console.log('An error occurred.', error)
      )
      .then(json =>
        // We can dispatch many times!
        // Here, we update the app state with the results of the API call.
        dispatch(receivePosts(subreddit, json))
      )
  }
}
```

This coupled with introducing redux-thunk:

```
const store = createStore(
  rootReducer,
  applyMiddleware(
    thunkMiddleware,
  )
)
```

## Steps to change here

- Introduce thunkMiddleware
- Update createStore
- Move checkFontLoaded into actions folder (along with tests)
- Rejig it to return a function like the example above
- Call the action in App (e.g. componentDidMount)

src/store/load-data.js
ottis: Why would it be less approachable for newbies? Introducing absolute imports doesn't break relative imports and is a learn once thing (no real overhead in complexity)

src/components/sidebar/sidebar.test.js
ottis: Have you investigated the use of snapshots?

src/store/index.js
ottis: Cool, JSDoc description of the expected data structure being passed down would be great 🙂

src/components/app/index.js
ottis: Ok - must be my wrong interpretation of what's happening, resolving 🙂

src/store/load-data.js
richardwestenra: When I first encountered it, I found it a little confusing, and spent some time trying to figure out where the path rename was being configured (as it's not evident at the location you first spot it). I did some googling for pros/cons and the learning curve was something listed as one of the cons.

Seeing as it would take >0 time to configure, it wouldn't solve the problem of changing paths in any of the changed paths in this PR, and this project doesn't have deeply nested paths, it doesn't seem necessary to me. I'm not saying it's a terrible idea, I just think the minuscule pros are outweighed by the minor cons.

src/components/sidebar/sidebar.test.js
richardwestenra: Not especially - I'm fairly ambivalent on them but usually try to avoid them because of their tendency to fail on the tiniest change. I don't have strong feelings either way though

src/store/index.js
richardwestenra: As discussed on Slack: Good idea but right now it doesn't simplify the code but it does add a bunch of extra complexity, so I'm gonna shelve this for now and revisit it later if we need to add more middleware 👍

PR #124, opened 2020-03-04
[KED-1450] Move visibleNav/navWidth calculations into redux store

richardwestenra:
## Description

The visibleNav calculation affects the flowchart width and is currently a class method used in multiple places, which is getting difficult to maintain. It's handled in this way because the boolean flag is stored in the Wrapper component state, so it can only be passed down as a prop. Moving this prop to the redux store and using selectors to manage the navWidth calculations will save time and make things more maintainable.

## Development notes

I've also added/changed some of the tests to improve coverage and make them more flexible, and done some slight sidebar refactoring.

## QA notes

No functionality changes

## Checklist

## Legal notice

### DIFF
--- a/src/components/sidebar/menu-buttons.js
+++ b/src/components/sidebar/menu-buttons.js
@@ -6,15 +6,15 @@ import MenuIcon from '../icons/menu';
 /**
  * Hamburger menu button
  * @param {Function} props.onToggle Show menu on click
- * @param {Boolean} props.visible Whether nav is visible
+ * @param {Boolean} props.sidebarVisible Whether the sidebar is visible
  */
-export const ShowMenuButton = ({ onToggle, visible }) => (
+export const ShowMenuButton = ({ onToggle, visible: sidebarVisible }) => (
   <button
     aria-label="Show menu"
     className={classnames(
       'pipeline-sidebar__show-menu pipeline-sidebar__icon-button',
       {
-        'pipeline-sidebar__icon-button--visible': !visible
+        'pipeline-sidebar__icon-button--visible': !sidebarVisible
       }
     )}
     onClick={() => onToggle(true)}>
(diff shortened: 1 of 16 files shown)


921kiyo:
I think it's best for the front-end engineers to review the changes, so I will leave it to them.

src/components/sidebar/menu-buttons.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {Function} props.onToggle Show menu on click
+ * @param {Boolean} props.visible Whether nav is visible
+ */
+export const ShowMenuButton = ({ onToggle, visible }) => (
+  <button
+    aria-label="Show menu"
+    className={classnames(
+      'pipeline-sidebar__show-menu pipeline-sidebar__icon-button',
+      {
+        'pipeline-sidebar__icon-button--visible': !visible
```
limdauto: Do we have to pass `visible` in here? `onToggle` is already hard-coded to `true` right. The same applies to `HideMenuButton`. The reason I'm asking is because `'pipeline-sidebar__icon-button--visible': !visible` is a bit hard to parse for my brain, but it's not a biggie.

src/selectors/layout.js
```diff
@@ -68,35 +68,64 @@ export const getGraphSize = createSelector(
...
+    if (!width || !height) {
+      return {};
+    }
+    const sidebarWidth = getSidebarWidth(visibleSidebar, width);
+    return {
+      left,
+      top,
+      outerWidth: width,
+      outerHeight: height,
+      width: width ? width - sidebarWidth : width,
```
limdauto: This should be `sidebarWidth ? width - sidebarWidth : width` (probably)

src/components/sidebar/menu-buttons.js
richardwestenra: Yeah that's a really tricky one - I'm using 'visible' to refer to two different things - the visibility of the sidebar, and the visibility of the Show Menu button, which is only visible when the other is hidden. I think renaming the prop to be more specific will help - I'll do that. Thanks!

src/selectors/layout.js
richardwestenra: ahhhh this is a leftover relic from before I added the check for `if (!width || !height) {` on line 91. I can now remove the ternary check completely.

PR #132, opened 2020-03-15
[KED-1486] Proposal to make _call_viz a public function 

WaylonWalker:
## Description

I am wanting to use `call_viz` in a small library [kedro-static-viz](https://github.com/WaylonWalker/kedro-static-viz/blob/develop/kedro_static_viz/cli.py#L91), but am concerned about the direction of this function since it is currently private.

## Checklist

## Legal notice

- [ ] I acknowledge and agree that by checking this box and clicking "Submit Pull Request":

- I certify that (a) this contribution is my original creation and/or (b) to the extent it is not my original creation, I am authorized to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorized by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
- I am submitting this as my own thoughts.  This PR does not represent any company I work for.

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -110,7 +110,7 @@ def run_viz(port=None, line=None) -> None:
         _VIZ_PROCESSES[port].terminate()
 
     viz_process = multiprocessing.Process(
-        target=_call_viz, daemon=True, kwargs={"port": port}
+        target=call_viz, daemon=True, kwargs={"port": port}
     )
     viz_process.start()
     _VIZ_PROCESSES[port] = viz_process
@@ -300,7 +300,7 @@ def commands():
 def viz(host, port, browser, load_file, save_file, pipeline, env):
     """Visualize the pipeline using kedroviz."""
     try:
-        _call_viz(host, port, browser, load_file, save_file, pipeline, env)
+        call_viz(host, port, browser, load_file, save_file, pipeline, env)
     except KedroCliError:
         raise
     except Exception as ex:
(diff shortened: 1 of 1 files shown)


921kiyo:
Hi @WaylonWalker, I'm okay with making it public (if @DmitriiDeriabinQB also agrees), since `_call_viz` func was introduced when I added `run_viz` line magic function for notebook. 

Could you please
- Check the legal box in the PR description
- Add docstring in `call_viz()` func (since it is public now)
- Update the `test_serve.py` since it's still calling `_call_viz` and CI will break it.

DmitriiDeriabinQB:
@921kiyo I'm quite hesitant to make it public, to be perfectly honest. It means that we lock down the signature of it and have to make sure that we don't break it in the future releases. And given the volatile nature of Viz and the contract between the frontend and the backend, this very likely to happen.

WaylonWalker:
I do not believe that making it public means the api is necessarily locked, but I understand that it makes it easier to move keeping it closed.  I have always admired how focused the pandas project is to not break things.  Api breakages only occur during major releases and there was a deprecation notice in the last major release.  Upcoming changes are always very clear ahead of time with these deprecation notices.  Breaking changes only being things that would truly break existing code, not new additions/options that are added on.

I completely understand that this can slow down your development.  If you feel there are too many breaking changing coming up that you are comfortable with don't feel pressured to release something that is not ready.  I just wanted to get the discussion started here. 

I also think that it would be reasonable to keep the js and python versions in sync whether its a hard pin or a must be greater than vx.x.x.

lorenabalan:
Hi @WaylonWalker apologies for the late reply! We'd like to keep this private since we see it more as a helper function, and the contents inside are fairly easy to reproduce. Feel free to import the private function, but be aware that it could break in the future. Alternatively, if you want to lock in the implementation, you can copy-paste the contents for your own plugin.

WaylonWalker:
@lorenabalan I completely understand.  Thanks for considering this and giving suggestions.

PR #136, opened 2020-03-25
[KED-1487] Only open browser on localhost

lorenabalan:
## Description

Resolve https://github.com/quantumblacklabs/kedro-viz/issues/131
Reported issue is actually intended behaviour, but it's not entirely clear from the way the code is currently written. We should make this decision more explicit/transparent.

## Development notes

We intentionally only open Viz in a browser on localhost. Given a scenario where server is running remotely, we don't want to open web browser on that server.

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -349,6 +349,7 @@ def _call_viz(
     if save_file:
         Path(save_file).write_text(json.dumps(data, indent=4, sort_keys=True))
     else:
-        if browser:
-            webbrowser.open_new("http://127.0.0.1:{:d}/".format(port))
+        is_localhost = host == "127.0.0.1"
+        if browser and is_localhost:
+            webbrowser.open_new("http://{}:{:d}/".format(host, port))
         app.run(host=host, port=port)
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -267,7 +267,7 @@ def commands():
     "--browser/--no-browser",
     default=True,
     help="Whether to open viz interface in the default browser or not. "
-    "Defaults to True.",
+    "Browser will only be opened if host is localhost. Defaults to True.",
 )
 @click.option(
     "--load-file",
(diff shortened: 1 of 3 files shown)


WaylonWalker:
@lorenabalan It may be worth looking for a couple of aliases to localhost.

``` diff
-  is_localhost = host == "127.0.0.1"
+  is_localhost = host in ["127.0.0.1", "localhost", "0.0.0.0"]
```

lorenabalan:
> @lorenabalan It may be worth looking for a couple of aliases to localhost.
> 
> ```diff
> -  is_localhost = host == "127.0.0.1"
> +  is_localhost = host in ["127.0.0.1", "localhost", "0.0.0.0"]
> ```

Yup that was my next move! Thanks!

package/kedro_viz/server.py
```diff
@@ -349,6 +349,7 @@ def _call_viz(
     if save_file:
         Path(save_file).write_text(json.dumps(data, indent=4, sort_keys=True))
     else:
-        if browser:
-            webbrowser.open_new("http://127.0.0.1:{:d}/".format(port))
+        is_localhost = host in ("127.0.0.1", "localhost", "0.0.0.0")
+        if browser and is_localhost:
```
921kiyo: Could this change the user experience at all from the next release (when they pass --browser flag to non-local host)? If so, it's worth mentioning in the release note (but I can't think of the cases...).

package/tests/test_server.py
```diff
@@ -197,6 +197,19 @@ def test_no_browser(cli_runner):
     assert server.webbrowser.open_new.called
```
DmitriiDeriabinQB: I know it's not part of this PR, but lines 175 and 184 use `.called_with` method instead of [`assert_called_with`](https://docs.python.org/3.6/library/unittest.mock.html#unittest.mock.Mock.assert_called_with). I think `called_with` is not a standard mock method, so it will return just another mock. If this is true, then this is a bug.

Could you please try changing it to `assert_called_with`?

package/kedro_viz/server.py
DmitriiDeriabinQB: You can also update line 269 and say that it only works for local host

lorenabalan: Updated both

package/tests/test_server.py
lorenabalan: good spot, fixed

PR #142, opened 2020-04-03
(KED-1516) Stop kedro-viz from swallowing stacktraces

mzjp2:
## Description

I'm a little uneasy about my approach here, so would welcome comments. Also, I guess this only prints stacktraces for generic exceptions and not for exceptions that are already `KedroCLIError`'s (from `_call_viz`) -- should I do it there too? 

When there's an underlying generic exception (e.g in your pipeline file, missing a comma or some syntax issue) kedro-viz silently prints something like `Error: invalid syntax (pipeline.py, line 64)` without the (sometimes helpful) stacktrace.

## Development notes

Use `traceback` to print the stacktrace before continuing with the `KedroCLIError`. Sadly the usual `raise a from b` trick doesn't work here because of the way `click` handles exceptions (and our custom `KedroCLIError` inherits from `click.Exception`).

## QA notes

Added a test that makes sure that the traceback is printed when `_call_viz` returns a `ValueError`.

## Checklist

## Legal notice

### DIFF
--- a/package/tests/test_server.py
+++ b/package/tests/test_server.py
@@ -202,7 +202,9 @@ def test_no_browser_if_not_localhost(cli_runner):
     """Check that call to open browser is not performed when host
     is not the local host.
     """
-    result = cli_runner.invoke(server.commands, ["viz", "--browser", "--host", "123.1.2.3"])
+    result = cli_runner.invoke(
+        server.commands, ["viz", "--browser", "--host", "123.1.2.3"]
+    )
     assert result.exit_code == 0, result.output
     assert not server.webbrowser.open_new.called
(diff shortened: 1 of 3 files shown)


package/tests/test_server.py
```diff
@@ -385,6 +387,16 @@ def test_viz_kedro14_invalid(mocker, cli_runner):
     assert "Could not find a Kedro project root." in result.output
 
 
+def test_viz_stacktrace(mocker, cli_runner):
+    """Test that in the case of a generic exception,
+    the stacktrace is printer."""
```
DmitriiDeriabinQB, line 392:
```suggestion
    the stacktrace is printed."""
```

PR #146, opened 2020-04-14
[KED-1517] Added layers to the response

andrii-ivaniuk:
## Description

https://jira.quantumblack.com/browse/KED-1517

## Development notes

Added layers to the response.

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -198,9 +198,35 @@ def format_pipeline_data(pipeline, catalog):
     edges = []
     namespace_tags = defaultdict(set)
     all_tags = set()
+    namespace_to_layer = {}
+
+    data_set_to_layer_map = {
+        ds_name: getattr(ds_obj, "_layer", None)
+        for ds_name, ds_obj in catalog._data_sets.items()
+    }
+
+    all_layers = sorted(set(layer for layer in data_set_to_layer_map.values() if layer))
 
     for node in sorted(pipeline.nodes, key=lambda n: n.name):
         task_id = _hash(str(node))
+        all_tags.update(node.tags)
+        node_layer_id = None
+
+        for data_set in node.inputs:
+            namespace = data_set.split("@")[0]
+            namespace_to_layer[namespace] = data_set_to_layer_map.get(data_set)
+            edges.append({"source": _hash(namespace), "target": task_id})
+            namespace_tags[namespace].update(node.tags)
+
+        for data_set in node.outputs:
+            namespace = data_set.split("@")[0]
+            layer = data_set_to_layer_map.get(data_set)
+            namespace_to_layer[namespace] = layer
+            if layer and node_layer_id is None:
+                node_layer_id = all_layers.index(layer)
+            edges.append({"source": task_id, "target": _hash(namespace)})
+            namespace_tags[namespace].update(node.tags)
+
         nodes.append(
             {
                 "type": "task",
@@ -208,19 +234,14 @@ def format_pipeline_data(pipeline, catalog):
                 "name": getattr(node, "short_name", node.name),
                 "full_name": getattr(node, "_func_name", str(node)),
                 "tags": sorted(node.tags),
+                "layer": node_layer_id,
             }
         )
-        all_tags.update(node.tags)
-        for data_set in node.inputs:
-            namespace = data_set.split("@")[0]
-            edges.append({"source": _hash(namespace), "target": task_id})
-            namespace_tags[namespace].update(node.tags)
-        for data_set in node.outputs:
-            namespace = data_set.split("@")[0]
-            edges.append({"source": task_id, "target": _hash(namespace)})
-            namespace_tags[namespace].update(node.tags)
 
     for namespace, tags in sorted(namespace_tags.items()):
+        layer = namespace_to_layer[namespace]
+        layer_id = all_layers.index(layer) if layer else None
+
         is_param = bool("param" in namespace.lower())
         nodes.append(
             {
@@ -229,6 +250,7 @@ def format_pipeline_data(pipeline, catalog):
                 "name": pretty_name(namespace),
                 "full_name": namespace,
                 "tags": sorted(tags),
+                "layer": layer_id,
             }
         )
 
@@ -236,7 +258,9 @@ def format_pipeline_data(pipeline, catalog):
     for tag in sorted(all_tags):
         tags.append({"id": tag, "name": pretty_name(tag)})
 
-    return {"nodes": nodes, "edges": edges, "tags": tags}
+    layers = [{"id": idx, "name": value} for idx, value in enumerate(all_layers)]
+
+    return {"nodes": nodes, "edges": edges, "tags": tags, "layers": layers}
 
 
 @app.route("/api/nodes.json")
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -203,15 +203,21 @@ def format_pipeline_data(pipeline, catalog):
 
     data_set_to_layer_map = {
         ds_name: getattr(ds_obj, "_layer", None)
-        for ds_name, ds_obj in catalog._data_sets.items()  # pylint: disable=protected-access
+        for ds_name, ds_obj in catalog._data_sets.items()
     }
-
-    all_layers = sorted(set(layer for layer in data_set_to_layer_map.values() if layer))
 
     for node in sorted(pipeline.nodes, key=lambda n: n.name):
         task_id = _hash(str(node))
         all_tags.update(node.tags)
-        node_layer_id = None
+        nodes.append(
+            {
+                "type": "task",
+                "id": task_id,
+                "name": getattr(node, "short_name", node.name),
+                "full_name": getattr(node, "_func_name", str(node)),
+                "tags": sorted(node.tags),
+            }
+        )
 
         for data_set in node.inputs:
             namespace = data_set.split("@")[0]
@@ -221,28 +227,11 @@ def format_pipeline_data(pipeline, catalog):
 
         for data_set in node.outputs:
             namespace = data_set.split("@")[0]
-            layer = data_set_to_layer_map.get(data_set)
-            namespace_to_layer[namespace] = layer
-            if layer and node_layer_id is None:
-                node_layer_id = all_layers.index(layer)
+            namespace_to_layer[namespace] = data_set_to_layer_map.get(data_set)
             edges.append({"source": task_id, "target": _hash(namespace)})
             namespace_tags[namespace].update(node.tags)
 
-        nodes.append(
-            {
-                "type": "task",
-                "id": task_id,
-                "name": getattr(node, "short_name", node.name),
-                "full_name": getattr(node, "_func_name", str(node)),
-                "tags": sorted(node.tags),
-                "layer": node_layer_id,
-            }
-        )
-
     for namespace, tags in sorted(namespace_tags.items()):
-        layer = namespace_to_layer[namespace]
-        layer_id = all_layers.index(layer) if layer else None
-
         is_param = bool("param" in namespace.lower())
         nodes.append(
             {
@@ -251,7 +240,7 @@ def format_pipeline_data(pipeline, catalog):
                 "name": pretty_name(namespace),
                 "full_name": namespace,
                 "tags": sorted(tags),
-                "layer": layer_id,
+                "layer": namespace_to_layer[namespace],
             }
         )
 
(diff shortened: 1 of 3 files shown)


package/tests/result.json
```diff
@@ -0,0 +1,85 @@
+{
```
andrii-ivaniuk, line 1: @richardwestenra This is the response format.

package/kedro_viz/server.py
```diff
@@ -198,29 +199,50 @@ def pretty_name(name):
...
     namespace_tags = defaultdict(set)
     all_tags = set()
+    namespace_to_layer = {}
+
+    data_set_to_layer_map = {
+        ds_name: getattr(ds_obj, "_layer", None)
+        for ds_name, ds_obj in catalog._data_sets.items()  # pylint: disable=protected-access
+    }
+
+    all_layers = sorted(set(layer for layer in data_set_to_layer_map.values() if layer))
```
limdauto, line 209: I don't think alphabetically sorting layers is the desired behaviour. I'm checking with Yetu and the viz team to see where we should encode the order information.

limdauto: @idanov suggested that we will have to deduce the layer's ordering based on the pipeline's topology. An idea I have is after parsing node's layer information, we use `pipeline.grouped_nodes` to deduce a sorted list of layers. If the order of dataset's layer can't be deduced, maybe we can omit it? Not sure... What do you think? @andrii-ivaniuk -- We can also bring it up at tech design tomorrow.

andrii-ivaniuk: This ordering doesn't have any effect. If you take a look at the example response you can see that we have `layers` list which is a mapping between layer name and some ID (in this case just sequence number). Then each node refers to layer by ID (like in RDBMS). I don't see the need to do something extra like deduce a sorted list of layers.

andrii-ivaniuk: We can consider it as a relational database with two tables: `layers` and `nodes`.
`node.layer` is a foreign key on `layer.id` and that's it.

limdauto: I was thinking more like how does the frontend know if they should display the `raw` layer first or the `final` layer first (visually) but maybe the frontend works it out automatically so this ordering is indeed not an issue. @richardwestenra could you give us an advice here please?

richardwestenra: I believe the front-end does rely on the ordering, yes. `layers[0]` is the top/first one.

richardwestenra: I mean, it calculates the positioning automatically, but having them in order is important for calculating the padding between layers. I could do it automatically but it'd be trickier, so ideally it'd be better to have them in order

limdauto: I feel like calculating and returning layers in order is the backend's responsibility. Otherwise, why bother returning the list of layers at all? Why can't we just return the layer's name inside the `node` and let the frontend deduce whatever information it needs to deduce?

andrii-ivaniuk: @limdauto We have an execution order defined by edges. Having this order we know the start point and its layer. The strat point layer is the first layer.

E.g. We have two pipelines with the following layer order:
1. raw -> null -> intermediate -> null -> final
2. new -> intermediate -> pre-final
Also, we have one dataset with `unused` layer in catalog.yml but it is not used in any pipeline

What should the final order be?

limdauto: I don't know. @richardwestenra what do you think the behaviour / backend response for the above example should be?

richardwestenra: - I would probably leave out any layer that isn't used by any of the nodes, but it doesn't really matter as it'll be filtered ultimately regardless.
- `null` isn't a layer, it's the absence of a layer, so it shouldn't be in the layer order object. so `raw -> null -> intermediate -> null -> final1` becomes `[raw, intermediate, final1]`

limdauto: What about `new` and `final2`? I think @andrii-ivaniuk meant those 2 pipelines are combined into one.

andrii-ivaniuk: @richardwestenra But we have two separate pipelines. 
Let's introduce some layer naming convention according to which each layer should be prefixed with a sequence number e.g. 1_raw, ..., 5_final.

richardwestenra: oh. Are we talking about modular/sub-pipelines? The current API architecture doesn't work for that anyway - I feel like it shouldn't be reflected in this PR, and we can cross that bridge when we add modular pipeline support in KedroViz in a subsequent PR?

btw, for context: The `name` field will be displayed to the user in Kedro-Viz, so it should be readable and user-friendly. The `id` field can be used for a UID if you need to avoid confusion/duplication. If you need to add extra context (e.g. to illustrate which sub-pipeline something belongs to) then maybe that could be added to a separate property on the object?

image

limdauto: I think in @andrii-ivaniuk's example, then maybe the final order should be `["new", "raw", "intermediate", "final1", "final2"]`. The first level of ordering is based on pipeline's topology. The second level of ordering (e.g. "new", "raw") can just be alphabetical. WDYT?

andrii-ivaniuk: @richardwestenra I didn't talk about the modular pipeline or I don't understand what do you mean. I talk about the following example:
Screen Shot 2020-04-14 at 20 08 06


@limdauto Why did you make such an assumption? 
My thoughts: we have a lot of assumptions that result in a lot of edge cases.

limdauto: It wasn't really an assumption. It was more like an idea. We do know for a fact that we need to return the ordering of layers __somehow__ to the frontend. Part of this ordering is dictated by node's execution order (the pipeline's topology), so that part can't be changed. In your example, this part of ordering is `"new"` and `"raw"` needs to come before `"intermediate"` and `"intermediate"` needs to come before `"final1"` and `"final2"`. For the ordering that can't be inferred, it needs to be based on convention. Your initial implementation, ie alphabetical order, is as good a convention as any. If you have any other suggestion, would love to learn more.

I think what we should do is come up with a list of possible options here and bring it up with the larger group during tech design / post standup tomorrow.

richardwestenra: huh, ok. I mean, at the moment I think it assumes that you only have a single pipeline? I'm not sure the metaphor works with multiple unconnected pipelines displayed at once. hmm.

limdauto: I think this edge case needs product input. We will bring this up with the group tomorrow. To recap, the summary of the problem is: we are not sure where to put the logic to determine the ordering of layers. There are 3 places we can solve this problem:

* User: letting user specify the ordering of layers themselves (e.g. through a prefix convention or just stating them explicitly in catalog.yml)
* Backend: deducing the layers ordering based on node's execution order and some other criteria for the edge case presented above.
* Frontend: based on how best things look visually (?)

Let's bring it up after standup tomorrow @andrii-ivaniuk @richardwestenra

package/kedro_viz/server.py
```diff
@@ -198,9 +199,16 @@ def pretty_name(name):
     edges = []
     namespace_tags = defaultdict(set)
     all_tags = set()
+    namespace_to_layer = {}
+
+    data_set_to_layer_map = {
```
lorenabalan, line 204: Minor minor suggestion. `namespace_to_layer` is also a dict and the `a_to_b` usually clearly suggests a dictionary, so `map` seems superfluous. Plus you a get a shorted name! :D
```suggestion
    dataset_to_layer = {
```

package/kedro_viz/server.py
```diff
@@ -210,13 +218,16 @@ def pretty_name(name):
...
+
         for data_set in node.inputs:
             namespace = data_set.split("@")[0]
+            namespace_to_layer[namespace] = data_set_to_layer_map.get(data_set)
             edges.append({"source": _hash(namespace), "target": task_id})
             namespace_tags[namespace].update(node.tags)
+
         for data_set in node.outputs:
             namespace = data_set.split("@")[0]
+            namespace_to_layer[namespace] = data_set_to_layer_map.get(data_set)
```
lorenabalan, line 230: Just checking I understood correctly, so if a user assigns to different layers to a transcoded dataset (A@spark in intermediary and A@pandas - used later - in primary), is it the output one overwrites the existing entry in `namespace_to_layer`? Or sth else happens? Are they warned anywhere that they're doing sth weird with their layers for that namespace (A), or is that part of future work?

package/tests/result.json
```diff
@@ -0,0 +1,73 @@
...
+            "source": "60e68b8e",
+            "target": "3751475c"
+        },
+        {
+            "source": "3751475c",
+            "target": "37a5301a"
+        }
+    ],
+    "tags": []
+}
```
lorenabalan, line 73: Does it need an empty line at the end?

andrii-ivaniuk: Why do you think so?

lorenabalan: Nevermind, it had a github symbol for "no blank line" after, looks alright now.

package/kedro_viz/server.py
lorenabalan: Any reason why this was resolved?

package/kedro_viz/server.py
andrii-ivaniuk: @lorenabalan As I understood, `A` should be the name of a dataset, isn't it? So so we expect to have `A@spark` and `A@pandas` in one file.
I just took a look at unit tests and found the following examples: "fred_in@pandas" and "fred_out@pandas".

So maybe I don't understand the transcoded dataset correctly.

lorenabalan: So in your catalog you'd have
```yaml
A@spark:
    type: spark.SparkDataSet
    layer: de
A@pandas:
    type: pandas.PandasDataSet
    layer: ds   # bad
```
and my understanding was that, as far as the catalog knows, in `self._data_sets` you'd have both keys "A@pandas" and "A@spark".

lorenabalan: So depending on how inputs/outputs are set, you could end up with `namespace_to_layer` to look like {"A": "de"} or {"A": "ds"}. Please correct me if I'm missing sth!

limdauto: I think that might be a product question... Does kedro-viz currently visualise transcoded dataset?

yetudada: I think it does. I remember Gordon doing some hackery to show that it was a transcoded dataset.

lorenabalan: In Viz you only see "A", the two get merged into one dataset. Anything that's after @ is irrelevant. But in this case if we attach one layer over the other, it's not clear which wins (to the user). It's bad behaviour to start with anyway, to assign 2 different layers, just wondering if/how we thought about this or if this is surfaced to the user somewhere yet.

andrii-ivaniuk: @lorenabalan As you said kedro-viz will get two `A` identical nodes that according to your example related to different values (in one case it is `pandas` data in another `spark`). How pandas and spark data be merged?

andrii-ivaniuk: Let's imaging we have the following:
DE pipeline: `data1@spark` -> `node1` -> `data1@pandas`.
How it will be depicted on Kedro-Viz?

andrii-ivaniuk: Screen Shot 2020-04-21 at 15 45 30
```
def create_pipeline(**kwargs):
    return Pipeline(
        [
            node(
                report_accuracy,
                "spark_data_2@user",
                "spark_data_2@custom"
            )
        ]
    )
```

Is this the correct picture?

andrii-ivaniuk: Where catalog.yml is:

```
spark_data_2@user:
  type: spark.SparkDataSet
  filepath: /tmp/spark/test_data

spark_data_2@custom:
  type: pandas.CSVDataSet
  filepath: /tmp/spark/test_data/test.csv
```

lorenabalan: So your example pipeline is not allowed, Kedro will complain when user is trying to write to the same thing you're reading from, i.e. the same `spark_data_2` "namespace".

andrii-ivaniuk: I have another use case which we can discuss online.
I've got confused because from kedro-viz perspective I get two identical nodes.

limdauto: I think unless we start displaying the full transcoded name in viz, there is no way at the moment we can show transcoded datasets in 2 different layers. Probably worth putting it in a ticket or a TODO comment.

PR #147, opened 2020-04-16
[KED-1570] Experiment with layer ordering and assignment client-side

limdauto:
## Description

This PR introduces logic to:

* Calculate layer ordering server-side based on the layer's dependencies.
* This layer ordering is then fed forward to a toposort client-side when calculating node ranks. I left this calculation untouched as I think it's good enough.
* Remove the need to use a separate layer id and layer name. Instead, use name every where.
* Specify `layer: null` for every Kedro node in the mock data.
* Specify `layer: null` for some dataset.

The reason `layer: null` works is because a graph node is always assigned a rank, so it will get displayed based on its topological order in any case. If there is a layer assigned, it just means this layer changes the node's topological order and the node is displayed in the same rank as other nodes in the same layer. 

An interesting property of the way we add fake edges between layers is that a Kedro node will be displayed on the same layer as its output dataset if its input dataset is in the layer below, which is in line with data engineer's convention.

## Checklist

## Legal notice

### DIFF
--- a/src/store/normalize-data.js
+++ b/src/store/normalize-data.js
@@ -1,5 +1,4 @@
 import { getInitialPipelineState } from '../store/initial-state';
-import batchingToposort from 'batching-toposort';
 /**
  * Check whether data is in expected format
  * @param {Object} data - The parsed data input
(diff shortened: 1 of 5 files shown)


src/store/normalize-data.js
```diff
@@ -85,13 +76,35 @@ const formatData = data => {
...
+    const nodeDeps = {};
+    const nodeIdsMap = {};
+    for (const node of data.nodes) {
+      addNode(state)(node);
+      nodeDeps[node.id] = [];
+      nodeIdsMap[node.id] = node;
+    }
+    for (const edge of data.edges) {
+      addEdge(state)(edge);
+      nodeDeps[edge.source].push(edge.target);
```
DmitriiDeriabinQB, line 90: As far as I can see, we are not doing that artificial node insertion, which may bring us some problems.

Let's imagine the following graph:

mermaid-diagram-20200417104328

If I understand that correctly, Batching toposort will put `Raw 0` and `Int 0` into the same group (as they don't depend on each other). Then the order in which `Raw` and `Int` layers will be presented only depends on the order in which they appear in that toposort group (which is unreliable). So if I flip the branches on the graph, `Int` layer will be presented first, which is incorrect to me.

Gordon's suggestion was to add an artificial dependency node `Raw-to-Int` as in
mermaid-diagram-20200417105148
Which _depends_ on all `Raw` layer nodes and _is a dependency_ for all `Int` layer nodes. In that case toposort will put `Raw 0` and `Int 0` into different groups and `Raw` layer will go before `Int` layer.

I'm happy to discuss it over Zoom after the standup if necessary.

limdauto:
Closing in favour of https://github.com/quantumblacklabs/kedro-viz/pull/148

PR #148, opened 2020-04-18
[KED-1570] Calculate layers order and return it to the frontend

limdauto:
## Description

This PR introduces logic to:

* Calculate layer ordering server-side based on the layer's dependencies.
* Remove the need to use a separate layer id and layer name. Instead, use layer name every where.
* Specify layer: `null` for every Kedro node in the mock data.
* Specify layer: `null` for some dataset.

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -217,7 +217,7 @@ def _sort_layers(nodes: Dict[str, Dict], dependencies: Dict[str, Set[str]]) -> L
                     "layer": Optional[str]
                     ...
                 }
-        dependencies: A dictionary of {node_id -> set(child_node_ids)}
+        dependencies: A dictionary of {node_id -> set(child_ids)}
             represents the direct dependencies between nodes in the graph.
 
     Returns:
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -43,14 +43,13 @@ from typing import Dict, List, Set
 import click
 import kedro
 import requests
-from flask import Flask, jsonify, send_from_directory
 from IPython.core.display import HTML, display
+from flask import Flask, jsonify, send_from_directory
 from kedro.cli import get_project_context  # pylint: disable=ungrouped-imports
 from kedro.cli.utils import KedroCliError  # pylint: disable=ungrouped-imports
+from kedro_viz.utils import wait_for
 from semver import match
 from toposort import toposort_flatten
-
-from kedro_viz.utils import wait_for
 
 _VIZ_PROCESSES = {}  # type: Dict[int, multiprocessing.Process]
 
@@ -182,7 +181,7 @@ def _get_pipeline_catalog_from_kedro14(env):
         raise KedroCliError(ERROR_PROJECT_ROOT)
 
 
-def _sort_layers(nodes: Dict[str, Dict], node_dependencies: Dict[str, Set[str]]) -> List[str]:
+def _sort_layers(nodes: Dict[str, Dict], dependencies: Dict[str, Set[str]]) -> List[str]:
     """Given a DAG represented by a dictionary of nodes, some of which have a `layer` attribute,
     along with their dependencies, return the list of all layers sorted according to
     the nodes' topological order, i.e. a layer should appear before another layer in the list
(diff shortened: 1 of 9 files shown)


package/kedro_viz/server.py
```diff
@@ -180,6 +181,92 @@ def _get_pipeline_catalog_from_kedro14(env):
...
+    # compute the layer dependencies dictionary based on the node_layers dependencies,
+    # represented as {layer -> set(parent_layers)}
+    layer_dependencies = defaultdict(set)
+    for node_id, dependent_layers in node_layers.items():
+        node_layer = nodes[node_id].get("layer")
+
+        # add the node's layer as a parent layer for all dependent layers
+        if node_layer is not None:
+            for layer in dependent_layers:
+                layer_dependencies[layer].add(node_layer)
```
DmitriiDeriabinQB, line 263: We should only add if `layer != node_layer`. If I read everything correctly, we haven't checked for this earlier, right?

limdauto: So interestingly enough, I thought of this yesterday. If you mark a node as a dependency of itself, `toposort` still works, so I omitted it to reduce mental tax. Happy to put it back in though.

```
>>> layers = {"raw": {"raw"}, "int": {"raw", "int"}}
>>> toposort_flatten(layers)
['raw', 'int']
```

DmitriiDeriabinQB: Oh, that's cool, didn't know about that 🙂 
Could you add a note please? So that whoever reads this knows that the case is covered.

limdauto: Done

package/kedro_viz/server.py
```diff
@@ -209,46 +306,56 @@ def pretty_name(name):
     for node in sorted(pipeline.nodes, key=lambda n: n.name):
```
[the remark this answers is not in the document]
limdauto: * Not sure. It was there before. I thought of removing this sort by name but that breaks a bunch of tests so didn't change it to reduce the scope of this PR.
* That's a good question. Unfortunately, we can't rely on toposorted list of nodes alone :( There is an example here if you are interested: https://github.com/quantumblacklabs/kedro-viz/pull/147#discussion_r410119362 -- the example suggests adding transition nodes and toposort again, but adding transition nodes still require knowing layers' dependencies, so there is no avoiding toposorting layers.

package/kedro_viz/server.py
```diff
@@ -180,6 +181,94 @@ def _get_pipeline_catalog_from_kedro14(env):
...
+        node_layers[node_id] = set()
+
+        # for each dependent node of the given node_id,
+        # mark its layer and all layers that depend on it as dependent layers of the given node_id.
+        for dependent_node_id in node_dependencies[node_id]:
+            dependent_node = nodes[dependent_node_id]
+            dependent_layer = dependent_node.get("layer")
+            if dependent_layer is not None:
+                node_layers[node_id].add(dependent_layer)
+            node_layers[node_id].update(find_dependent_layers(dependent_node_id))
```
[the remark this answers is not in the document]
limdauto: Python's default recursion depth is 1000. Will put this as a comment.

src/store/normalize-data.js
```diff
@@ -91,7 +81,7 @@ const formatData = data => {
       data.tags.forEach(addTag(state));
     }
     if (data.layers) {
-      data.layers.forEach(addLayer(state));
+      state.layer.ids = data.layers;
```
richardwestenra, line 84: There are two options here:

1. Keep the layer IDs array and the layer names object intact, and populate them both with the layer name. This keeps some redundant code around as they're duplicates, but it futureproofs it if we need a separate layer ID in the future for whatever reason.

2. The alternative is to remove them both and replace the state layer property (`layer: { ids: [], name: {} }`) with `layer: []`.

Instead, in this PR you seem to be going for both options incompletely at the same time. I think it's better to commit to one of them properly. Can I please request that you either update the layer state shape everywhere it's touched (i.e. in `initial-state.js`, and in the layer selector where it creates both an id and a name field, and anywhere else I've forgotten), OR you restore the `addLayer` function in the data normalizer, and modify it so that it sets layer id and name to be equal to the layer name.

src/selectors/layers.js
```diff
@@ -41,7 +39,7 @@ export const getLayers = createSelector(
       const nextID = layerIDs[i + 1];
       return {
         id,
-        name: layerName[id],
+        name: id,
```
richardwestenra, line 42: As mentioned in another comment, I think that you should either keep `name: layerName[id],`, or delete the `id` field and just use `name` for everything.

src/store/normalize-data.js
limdauto: I will go with option 1. Thanks.

src/selectors/layers.test.js
```diff
@@ -36,6 +36,9 @@ describe('Selectors', () => {
 
       expect(
         nodes.every(node => {
+          if (node.layer === null) {
```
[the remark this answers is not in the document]
limdauto: No, this is to make it work with `null` layer. Your previous mock doesn't have any `null` layer, so it wasn't needed before. Now it does.

limdauto:
@richardwestenra Thank you for commenting. I have made the requested changes.

PR #158, opened 2020-05-05
[KED-1518] Wrap tooltip names & invert tooltip in top half of screen

richardwestenra:
## Description

1. Refactor tooltips to turn it into a new standalone component
2. Wrap tooltip text for long node names, by injecting zero-width spaces around special characters, and adding `overflow-wrap: break-word;` as a backup.

3. Make tooltips hang below their node in the top half of the screen, just as they do with inverting their positioning on the right side of the screen:

## Checklist

## Legal notice

### DIFF
--- a/src/components/tooltip/tooltip.test.js
+++ b/src/components/tooltip/tooltip.test.js
@@ -1,0 +1,98 @@
+import Tooltip, { insertZeroWidthSpace } from './index';
+import { setup } from '../../utils/state.mock';
+
+const mockProps = {
+  chartSize: {
+    height: 766,
+    left: 0,
+    outerHeight: 766,
+    outerWidth: 1198,
+    sidebarWidth: 300,
+    top: 0,
+    width: 898
+  },
+  targetRect: {
+    bottom: 341.05895,
+    height: 2.1011,
+    left: 856.19622,
+    right: 866.03076,
+    top: 338.95785,
+    width: 9.83453,
+    x: 856.19622,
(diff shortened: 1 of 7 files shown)


src/components/tooltip/tooltip.test.js
```diff
@@ -0,0 +1,98 @@
...
+    };
+    const wrapper = setup.shallow(Tooltip, { ...mockProps, targetRect });
+    const container = wrapper.find('.pipeline-tooltip--right');
+    expect(container.length).toBe(1);
+  });
+});
+
+describe('insertZeroWidthSpace', () => {
+  it('wraps special characters with zero-width spaces', () => {
+    expect(insertZeroWidthSpace('-').length).toBe(3);
```
bru5, line 78: I think it's better to compare with the exact expected result string for these rather than just length, unless there was a reason?

richardwestenra: Basically, the expected string result contains zeroWidthSpaces, which don't show up in  the text editor in a string. I can however use the same trick for this that I did in the function itself  - that is, this: 
```JavaScript
const zeroWidthSpace = String.fromCharCode(0x200b);

expect(text).toBe(`${zeroWidthSpace}${text}${zeroWidthSpace}`);
```

I'll change it now. Thanks for the review!

richardwestenra: @bru5 Fixed. Thanks!

PR #162, opened 2020-05-07
[KED-1645] Add node-list-count to node-list group headings

richardwestenra:
## Description

Add numerical counters to show the total number of nodes (and visible nodes) for each node-list group:

## Development notes

This doesn't take into account search filtering:

## Checklist

## Legal notice

### DIFF
--- a/src/components/node-list/node-list-group.js
+++ b/src/components/node-list/node-list-group.js
@@ -7,10 +7,10 @@ import { toggleTypeDisabled } from '../../actions/node-type';
 
 export const NodeListGroup = ({
   children,
+  collapsed,
   onToggleTypeDisabled,
   onToggleCollapsed,
-  type,
-  collapsed
+  type
 }) => (
   <Flipped flipId={type.id}>
     <li>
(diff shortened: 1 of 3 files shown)


src/components/node-list/node-list-group.js
```diff
@@ -20,7 +20,9 @@ export const NodeListGroup = ({
             <NodeListRow
               checked={!type.disabled}
               id={type.id}
-              label={type.name}
+              label={`${type.name} <i>(${type.nodeCount.visible}/${
```
bru5, line 23: Is there a way `label` could optionally take a render function and this be in JSX?

richardwestenra: @bru5 Yeah I absolutely could, but this is [importing a utility from Kedro-UI](https://github.com/quantumblacklabs/kedro-ui/blob/master/src/components/search-results/search-results-utils.js), which functions this way. I don't remember why it does it this way, because I thought it used to use react components and Array.reduce. I could change it in Kedro-UI, but for now I'd rather keep it as is, and look into changing that in a future release?

The alternative is to stop importing the functionality from Kedro-UI, and replicate it here but modify it so that it generates an object instead. Reckon it's necessary?

bru5: I see thanks for clarifying, sure that seems fair then to change it upstream when it becomes feasible later on.

PR #178, opened 2020-05-27
[KED-1657] Seed randomly-generated data with UUIDs

richardwestenra:
## Description

1. Generate a random UUID, and log it to the console, if random data is selected. Use it to seed all the random number generators.
2. Allow users to copy the ID from the console and add it to a query string in the URL, so that the randomly-generated data can be replicated.

## Development notes

1. This uses https://github.com/davidbau/seedrandom, and globally modifies the behaviour of `Math.random`. Since it can only be called when the app is being run in standalone mode, it shouldn't have any untoward effects, as far as I can tell.
2. Uses the URL.searchParams web API, which is not supported in IE.

## QA notes

Check the console when on `?data=random`, and try clicking the URL to keep the data as is.
Example output:
image

## Checklist

## Legal notice


src/utils/random-utils.js
```diff
@@ -0,0 +1,81 @@
...
+ * Get an array of numbers
+ * @param {number} n Length of the array
+ */
+export const getNumberArray = n => Array.from(Array(n).keys());
+
+/**
+ * Get a random number between 0 to n-1, inclusive
+ * @param {number} n Max number
+ */
+export const randomIndex = n => Math.floor(Math.random() * n);
```
bru5, line 11: Again I think it's a good idea to swap out the `Math.random` calls here to use the seeded random function directly instead.

src/utils/random-data.js
```diff
@@ -34,13 +35,38 @@ const LAYERS = [
...
+      seed = generateHash(30);
+      url.searchParams.set('seed', seed);
+    }
+    if (typeof jest === 'undefined') {
+      console.info(
+        `%cRandom data seed: ${seed}\nTo reuse this layout, visit ${url.toString()}`,
+        'font-weight: bold'
+      );
+    }
+    seedrandom(seed, { global: true });
```
bru5, line 67: I think it's better to use the seeded random function directly rather than patching the global random, as not to potentially clash with any other uses on the page.

src/utils/random-utils.js
richardwestenra: @bru5 Cheers, I've updated this, let me know what you think

PR #181, opened 2020-05-28
Update readme/contributing docs

richardwestenra:
## Description

1. Improve JS development docs
    - Fix incorrect data locations
    - Improve wording and formatting
    - Add more info about using different data sources
2. Fix poorly-formatted markdown tables
    - These were auto-ruined by prettier, but I have fixed it so it won't do this in future, so they don't have to look so ugly anymore.

## Checklist

## Legal notice

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,7 +79,11 @@ npm run lib
 
 #### Data sources
 
-Kedro-Viz uses unique identifiers to configure the data source. You can configure the data source by appending a query string to the URL, e.g. `http://localhost:4141/?data=random`. You can also set it with an environment variable when starting up the dev server, e.g. `DATA=random npm start`.
+Kedro-Viz uses a unique identifier to determine the data source. You can configure this by appending a query string to the URL, e.g. `http://localhost:4141/?data=random`. Alternatively, you can set it with an environment variable when starting up the dev server:
+
+```bash
+DATA=random npm start
+```
 
 There are several different data sources available. By default in production, the app asynchronously loads JSON from the `/api/nodes.json` endpoint, and you can replicate this in development by placing a dataset at `/public/api/nodes.json`. Alternatively, you can use one of the mock unit-testing/demo datasets, or pseudo-random data procedurally-generated on page load, which is often useful for local development.
 
@@ -92,7 +96,7 @@ There are several different data sources available. By default in production, th
 | `animals` | `/src/utils/data/animals.mock.js` |
 | `layers` | `/src/utils/data/layers.mock.js` |
 
-Randomly-generated data can be seeded with a hash string, which will allow you to replicate a generated layout. You can supply a seed with a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`. When using randomly-generated data without a seed, the app will generate a new pseudo-random one every time, and will output it to the browser console in case you wish to reuse it.
+Randomly-generated data can be seeded with a hash string, which will allow you to replicate a generated layout. You can supply a seed with a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`. If you do not supply a seed, the app will generate a new pseudo-random one every time, and will output it to the browser console in case you wish to reuse it.
 
 ### Branching conventions
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,12 +98,12 @@ Randomly-generated data can be seeded with a hash string, which will allow you t
 
 We use a branching model that helps us keep track of branches in a logical, consistent way. All branches should have the hyphen-separated convention of: `<type-of-change>/<short-description-of-change>` e.g. `feature/awesome-new-feature`
 
-| Types of changes | Description                                                                 |
-| ---------------- | --------------------------------------------------------------------------- |
-| `docs`           | Changes to the documentation of the plugin                                  |
-| `feature`        | Non-breaking change which adds functionality                                |
-| `fix`            | Non-breaking change which fixes an issue                                    |
-| `tests`          | Changes to project unit (`tests/`) and / or integration (`features/`) tests |
+| Types of changes | Description |
+| ---------------- | ----------- |
+| `docs`    | Changes to the documentation of the plugin |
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -59,35 +59,51 @@ def count_truthy(elements: List[Any]) -> int:
...
 
-- `DATA=random npm start` --> Serve randomly-generated data
-- `DATA=animals npm start` --> Serve example test data, from `/src/utils/data.mock.js`
-- `DATA=lorem npm start` --> Serve example test data, from `/src/utils/data.mock.js`
-- `npm start` --> Serve data loaded from `/public/logs/nodes.json`
+```bash
+npm start
+```
 
 This will serve the app at [localhost:4141](http://localhost:4141/), and watch files in `/src` for changes. It will also update the `/lib` directory, which contains a Babel-compiled copy of the source. This directory is exported to `npm`, and is used when importing as a React component into another application. It is updated automatically on save in case you need to test/debug it locally (e.g. with `npm link`). You can also update it manually, by running
```
stichbury, line 74:
```suggestion
This will serve the app at [localhost:4141](http://localhost:4141/), and watch files in `/src` for changes. It will also update the `/lib` directory, which contains a Babel-compiled copy of the source. This directory is exported to `npm`, and is used when importing as a React component into another application. It is updated automatically when you save in case you need to test/debug it locally (e.g. with `npm link`). You can also update it manually, by running
```

CONTRIBUTING.md
```diff
@@ -59,35 +59,51 @@ def count_truthy(elements: List[Any]) -> int:
...
 
 ```bash
 npm run lib
 ```
 
+#### Data sources
+
+Kedro-Viz uses unique identifiers to configure the data source. You can configure the data source by appending a query string to the URL, e.g. `http://localhost:4141/?data=random`. You can also set it with an environment variable when starting up the dev server, e.g. `DATA=random npm start`.
```
stichbury, line 82:
```suggestion
Kedro-Viz uses unique identifiers to configure the data source. You can configure the data source by appending a query string to the URL, e.g. `http://localhost:4141/?data=random`. Alternatively, you can set it by configuring an environment variable that is used when the dev server starts up: 

```bash
DATA=random npm start
```

stichbury: My suggestion is a bit passive tense-y. You could also say "Alternatively, you can set it by configurating an environment variable that the dev server uses when it starts up:"

CONTRIBUTING.md
```diff
@@ -59,35 +59,51 @@ def count_truthy(elements: List[Any]) -> int:
...
+| Identifier | Data source |
+|------------|-------------|
+| `json` (default) | `/public/api/nodes.json` |
+| `random` | Randomly-generated data |
+| `demo` | `/src/utils/data/demo.mock.js` |
+| `lorem` | `/src/utils/data/lorem-ipsum.mock.js` |
+| `animals` | `/src/utils/data/animals.mock.js` |
+| `layers` | `/src/utils/data/layers.mock.js` |
+
+Randomly-generated data can be seeded with a hash string, which will allow you to replicate a generated layout. You can supply a seed with a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`. When using randomly-generated data without a seed, the app will generate a new pseudo-random one every time, and will output it to the browser console in case you wish to reuse it.
```
stichbury, line 95:
```suggestion
If you do not use a seed, the app will generate new pseudo-random every time, and will output it to the browser console in case you wish to reuse it. You can seed the randomly-generated data with a hash string, which will allow you to replicate a generated layout. You can supply a seed with a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`. 
```

CONTRIBUTING.md
richardwestenra: To me, this is less clear than the original phrasing: What I am trying to communicate is that you should set the environment variable _when you are starting the dev server_, rather than just that the environment variable is used when the server starts up, which is less specific.

How about something like this? 
> Alternatively, you can set it by configuring an environment variable when starting up the dev server: 

Also, I noticed when re-reading this paragraph that the word `configures` is repeated quite a lot. So maybe we could replace one or two of the uses of that word with a synonym? How about this:

> Kedro-Viz uses a unique identifier to determine the data source. You can configure this by appending a query string to the URL, e.g. `http://localhost:4141/?data=random`. Alternatively, you can set it with an environment variable when starting up the dev server:

CONTRIBUTING.md
richardwestenra: This phrasing discusses the how to generate a seed, before explaining what the seed is used for, which is confusing as you would start the paragraph confused about why you are being told how to generate a "seed" when you don't yet know what it is. Reordering the sentences makes more sense to me:

> You can seed the randomly-generated data with a hash string, which will allow you to replicate a generated layout. You can supply a seed with a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`. If you do not supply a seed, the app will generate new pseudo-random every time, and will output it to the browser console in case you wish to reuse it.

CONTRIBUTING.md
stichbury: Ah, I see. OK, yes that makes sense. I like your final suggestion.

CONTRIBUTING.md
stichbury: Hmm, I thought seed and randomness would be broadly recognised by the readership, but I guess that's an assumption too far. I like your version, but I think it makes sense to go with the default (ie, no seed) first, and then explain how to use a seed if you want to.

Howsabout

By default, the app will generate new pseudo-random every time, and will output the seed it used to generate it to the browser console.  You can use the output hash string to replicate a generated layout, by using the output as a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`.

You can also use your own preferred hash string as a seed to generate random data in order to replicate a generated layout.

stichbury: However, we are at the point of splitting hairs, so I'm happy with whatever you ultimately choose.

richardwestenra: That works. Thanks! :)

PR #184, opened 2020-06-03
[KED-1735] Parameterise run_viz with project_path

limdauto:
## Description

This is the first part of the fix for KED-1735.

A summary of the problem and the fix:

* After we release Kedro 0.16.1, the `run_viz` line magic has stopped working because we stop changing user's working directory directly, so the line magic isn't being called with the correct `project_path`.
* To fix this problem, this PR adds a backward-compatible way to call `run_viz` and subsequently `_call_viz` with a `project_path` parameter. For `run_viz`, this parameter is provided via the `local_ns` parameter, which is provided by IPython line magic's local scope functionality.
* Another follow-up PR in Kedro will make sure that the line magic has the correct local scope, which contains the `project_path` value, as well as parameterise `get_project_context` to use it.

## Checklist

## Legal notice

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,7 +15,7 @@ Please follow the established format:
 
 ## Bug fixes and other changes
 
-<!-- Add release notes for the upcoming release here -->
+* Fix `%run_viz` line magic to display kedro viz inside Jupyter notebook. Note that the fix needs `kedro>=0.16.2` to take effect.
 
 # Release 3.3.0:
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,7 +15,7 @@ Please follow the established format:
 
 ## Bug fixes and other changes
 
-* Fix `%run_viz` line magic to display kedro viz inside Jupyter notebook. Note that the fix needs `kedro>=0.16.2` to take effect.
+* Fix `%run_viz` line magic to display kedro viz inside Jupyter notebook, previously broken for kedro 0.16.0/0.16.1. Note that the fix needs `kedro>=0.16.2` to take effect.
 
 # Release 3.3.0:
 
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -125,15 +125,13 @@ def run_viz(port=None, line=None, local_ns=None) -> None:
         _VIZ_PROCESSES[port].terminate()
 
     if local_ns is not None and 'project_path' in local_ns:
-        viz_process = multiprocessing.Process(
-            target=partial(_call_viz, project_path=local_ns['project_path']),
-            daemon=True,
-            kwargs={"port": port}
-        )
+        target = partial(_call_viz, project_path=local_ns['project_path'])
     else:
-        viz_process = multiprocessing.Process(
-            target=_call_viz, daemon=True, kwargs={"port": port}
-        )
+        target = _call_viz
+
+    viz_process = multiprocessing.Process(
+        target=target, daemon=True, kwargs={"port": port}
+    )
 
     viz_process.start()
     _VIZ_PROCESSES[port] = viz_process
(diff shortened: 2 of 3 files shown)


limdauto:
@mzjp2 @lorenabalan Could I get a blessing here as well please 🙏

package/kedro_viz/server.py
```diff
@@ -104,14 +105,17 @@ def _check_viz_up(port):
 
 
 # pylint: disable=unused-argument
-def run_viz(port=None, line=None) -> None:
+def run_viz(port=None, line=None, local_ns=None) -> None:
```
921kiyo, line 108: Would users need to pass `local_ns` variable when using this line magic in the notebook?

limdauto: No, IPython will do that automatically for us thanks to https://github.com/quantumblacklabs/private-kedro/pull/639

limdauto:
@921kiyo 
> Since you added unit tests, could you also delete https://github.com/quantumblacklabs/kedro-viz/blob/develop/package/features/jupyter.feature please?

Actually we still need to fix that :( The unit test mocked out the Process underneath, so it wasn't testing the whole thing. I had to test it manually end-to-end.

package/kedro_viz/server.py
```diff
@@ -120,9 +124,15 @@ def run_viz(port=None, line=None) -> None:
     if port in _VIZ_PROCESSES and _VIZ_PROCESSES[port].is_alive():
         _VIZ_PROCESSES[port].terminate()
 
+    if local_ns is not None and 'project_path' in local_ns:
```
lorenabalan, line 127:
```suggestion
    if local_ns is not None and "project_path" in local_ns:
```
to stay consistent with using double quotes in our codebase

package/kedro_viz/server.py
```diff
@@ -120,9 +124,15 @@ def run_viz(port=None, line=None) -> None:
...
+        target = partial(_call_viz, project_path=local_ns['project_path'])
```
lorenabalan, line 128:
```suggestion
        target = partial(_call_viz, project_path=local_ns["project_path"])
```

RELEASE.md
```diff
@@ -15,7 +15,7 @@ Please follow the established format:
 
 ## Bug fixes and other changes
 
-<!-- Add release notes for the upcoming release here -->
+* Fix `%run_viz` line magic to display kedro viz inside Jupyter notebook. Note that the fix needs `kedro>=0.16.2` to take effect.
```
lorenabalan, line 18: This is super minor, just to clarify that if they're not on 0.16.* already they're alright. But I don't have a strong stance on it.
```suggestion
* Fix `%run_viz` line magic to display kedro viz inside Jupyter notebook, previously broken for kedro 0.16.0/0.16.1. Note that the fix needs `kedro>=0.16.2` to take effect.
```

package/kedro_viz/server.py
```diff
@@ -124,8 +124,8 @@ def run_viz(port=None, line=None, local_ns=None) -> None:
     if port in _VIZ_PROCESSES and _VIZ_PROCESSES[port].is_alive():
         _VIZ_PROCESSES[port].terminate()
 
-    if local_ns is not None and 'project_path' in local_ns:
-        target = partial(_call_viz, project_path=local_ns['project_path'])
+    if local_ns is not None and "project_path" in local_ns:
+        target = partial(_call_viz, project_path=local_ns["project_path"])
```
mzjp2, line 128: do we not run `black` on `viz`?

limdauto: Yea I was wondering the same thing. It looks like we don't. Only `pylint`. Gonna log a ticket to fix that.

mzjp2: ty

PR #185, opened 2020-06-04
[KED-1650] New graphing algorithm & feature flags

bru5:
## Description

Improves pipeline visualisation using a new DAG graphing algorithm approach.

comparison

The algorithm is focused on improving the quality and clarity of output for both node layout and edge routing while maintaining  good performance.

Goals include balancing the regularity, legibility, symmetry, spacing and compactness of layout and edges while scaling to graphs with a large number of nodes and edges, for which the existing solution does not handle well enough for this purpose.

These changes are delivered alongside a minimal feature flag system to support toggling of experimental new features such as this one until their general release.

**The feature flag must be enabled in your browser before use and testing (see below).**

## Development notes

As there is currently no GUI for flags, they must be enabled and disabled using parameters in the URL. The status of all flags will persist in the user's local storage config and the console shows a message regarding the status of all flags on load.

Screenshot 2020-06-04 at 12 18 30

The flag for the new graphing algorithm is `newgraph`, it is disabled by default, you can use the URL parameters `newgraph=true` or `newgraph=false`.

e.g.`http://localhost:4141/?data=demo&newgraph=true`.

The implementation itself is self contained and includes a selector that connects it to the rest of the system in the same way as the existing solution, such that it works as a drop-in replacement.

It currently depends on the `kiwi.js` solver, though there is potential that this may not eventually be required given some upgrades to the main solver.

## QA notes

The feature must be enabled by the user before testing, as described above. Try the following test data sets. Side-by-side comparison is possible by enabling or disabling the flag:

__Demo__
http://localhost:4141/?data=demo&newgraph=true
http://localhost:4141/?data=demo&newgraph=false

__Random__ (includes layers)
http://localhost:4141/?data=random&newgraph=true
http://localhost:4141/?data=random&newgraph=false

_(see console to get a reusable seed for direct comparison)_

__Animals__
http://localhost:4141/?data=animals&newgraph=true
http://localhost:4141/?data=animals&newgraph=false

__Nodes.json__ (as generated by Kedro)
http://localhost:4141/?newgraph=true
http://localhost:4141/?newgraph=false

It has been tested on both real samples as well as varied pipelines generated by the existing random generator, both ranging so far from around 10 nodes to around 1000, all showing some strong improvement, both when labels are visible and when not.

It has support for and has been tested with pipelines that include layers.

There are also unit tests covering the majority of the changes, though they are not sufficient to test the actual quality of graphing output which must be done manually. For this I have been considering some metrics for more objective evaluation but more exploration is needed.

While the algorithm appears to work well on most pipelines seen so far, wider beta testing is encouraged before the flag is removed for general release.

## Checklist

## Legal notice


src/selectors/layers.js
```diff
@@ -10,41 +10,35 @@ const getLayerName = state => state.layer.name;
...
   (nodes, layerIDs, layerName, { width }) => {
-    // Get list of layer Y positions from nodes
-    const layerY = nodes.reduce((layerY, node) => {
-      if (!layerY[node.layer]) {
-        layerY[node.layer] = [];
-      }
-      layerY[node.layer].push(node.y);
-      return layerY;
-    }, {});
+    const bounds = {};
```
richardwestenra, line 13: What's the difference here?

src/utils/graph/index.js
```diff
@@ -0,0 +1,104 @@
...
+
+const defaultOptions = {
+  layout: {
+    spaceX: 16,
+    spaceY: 110,
+    basisX: 1500,
+    padding: 100
+  },
+  routing: {
+    spaceX: 26,
```
richardwestenra, line 13: 🚀

src/utils/graph/index.js
```diff
@@ -0,0 +1,104 @@
+import { offsetNode, offsetEdge } from './common';
+import { layout } from './layout';
+import { routing } from './routing';
+
+const defaultOptions = {
```
richardwestenra, line 5: I wonder whether this object should go in /src/config.js? 🤔

src/utils/graph/solver.js
```diff
@@ -0,0 +1,146 @@
+import * as kiwi from 'kiwi.js';
```
richardwestenra, line 1: So I just looked up kiwi.js and the [full library](https://github.com/gamelab/kiwi.js/blob/master/build/kiwi.min.js) is 322kb minified. Guess it's not possible to just pick and choose what we need, huh? I ask only because if/when we web-workerize this layout function, any libraries imported will need to be bundled into the web worker.

src/utils/graph/routing.js
```diff
@@ -0,0 +1,250 @@
...
+    // Sort the node's target edges by the angle between source and target nodes
+    // Break ties by alternating increasing X offsets per row
+    node.targets.sort((a, b) =>
+      compare(
+        Math.atan2(
+          b.sourceNode.y - b.targetNode.y,
+          b.sourceNode.x - b.targetNode.x !== 0
+            ? b.sourceNode.x - b.targetNode.x
+            : (b.targetNode.row % 2 === 0 ? -1 : 1) *
+                Math.pow(b.targetNode.row - b.sourceNode.row, 3)
```
richardwestenra, line 58: Is it possible to simplify this? Maybe add some variables? Same for the other one below.

src/selectors/layers.js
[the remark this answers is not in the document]
bru5: While implementing layers in the new layout algorithm there were some cases the original function appeared to not correctly handle causing overlaps. This replacement function didn't seem to have those problems I was having at the time,  but it may well have just been that those were invalid cases anyway, it's hard to tell.

Either way I think this version might be a little more performant than the original even if equivalent, since it does min/max once per layer vs twice and uses less memory (`O(layers * 2)` vs `O(nodes)`), if I understand both correctly?

Might be worth keeping if you agree, otherwise I can easily revert it unless I see the original problem again.

src/selectors/layout.js
```diff
@@ -11,41 +14,69 @@ const getNodeLayer = state => state.node.layer;
...
-    edges.forEach(edge => {
-      graph.setEdge(edge.source, edge.target, edge);
-    });
+    // Use experimental graph rendering if flag enabled
+    if (flags.newgraph) {
+      const result = graph(nodes, edges, showLayers && layers);
 
-    // Run Dagre layout to calculate X/Y positioning
-    dagre.layout(graph);
+      return {
```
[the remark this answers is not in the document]
bru5: Yeah for sure, this is only here as an adapter so they both use the same interface.

src/utils/graph/common.js
```diff
@@ -0,0 +1,166 @@
...
+ * Returns the point on the line segment `ax, ay, bx, by` closest to point `x, y`
+ * @param {number} x The test point x
+ * @param {number} y The test point y
+ * @param {number} ax The start of the line segement x point
+ * @param {number} ay The start of the line segement y point
+ * @param {number} bx The end of the line segement x point
+ * @param {number} by The end of the line segement y point
+ * @returns {object} An object with the closest point and both line segment points
+ */
+export const nearestOnLine = (x, y, ax, ay, bx, by) => {
```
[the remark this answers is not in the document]
bru5: I think that just ends up being extra overhead here though, in typing and objects, since all arguments are required by definition?

src/utils/graph/layout.js
```diff
@@ -0,0 +1,319 @@
...
+
+    // Clear separation constraints from previous iteration
+    separationConstraints.length = 0;
+
+    // For each row
+    for (let l = 0; l < rows.length; l += 1) {
+      const rowNodes = rows[l];
+
+      // Sort rows in order of X position. Break ties with ids for stability
+      rowNodes.sort((a, b) => compare(a.x, b.x, a.id, b.id));
```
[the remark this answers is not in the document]
bru5: Necessary as it stands. But I think this part should be complexity `<= O(sort(nodes))`, since `flat(rows) === nodes`, so not relatively that bad.

I've not looked at optimising anywhere much yet as overall as performance seems sufficient for an initial flagged release?

When I get on to this I'm sure there will be some simple improvements to find first, then after that the bigger wins will probably take some special techniques.

src/utils/graph/solver.js
bru5: Oh that's not the right `kiwi.js` actually, it's [this one](https://github.com/IjzerenHein/kiwi.js/), which seems to be about 14kb minified which I think is fine?

src/utils/graph/index.js
bru5: These are the default values for the function / module, rather than a user config, so I'd say it's best they stay with the module if that makes sense.

src/utils/flags.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Returns an object with flags as set in given or current URL using `enable` and `disable` parameters
+ * @param {string=} url The URL (optional, default current location)
+ * @returns {object} An object with flags and their values
+ */
+export const getFlagsFromUrl = url => {
+  const urlParams = new URL(url || document.location).searchParams;
+  const enableNames = (urlParams.get('enable') || '').split(/\W/g);
+  const disableNames = (urlParams.get('disable') || '').split(/\W/g);
```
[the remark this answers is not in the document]
bru5: Yeah I had some reasoning behind this approach, but I don't mind going with this alternative too, so I've made the change as noted.

src/utils/graph/layout.js
```diff
@@ -0,0 +1,319 @@
...
+    if (
+      edge.sourceNode.targets.length === 1 &&
+      edge.targetNode.sources.length === 1
+    ) {
+      parallelDoubleConstraints.push(parallelConstraint);
+    }
+  }
+
+  // Solving loop for constraints in X
+  for (let i = 0; i < 20; i += 1) {
```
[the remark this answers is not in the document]
bru5: This has been moved to a parameter called `iterations`.

src/utils/graph/layout.js
```diff
@@ -0,0 +1,319 @@
...
+    // Minimise crossing
+    solve(crossingConstraints, 1);
+
+    // Minimise edge length
+    solve(parallelConstraints, 1);
+
+    // Minimise edge length specifically for low-degree edges more strongly
+    solve(parallelSingleConstraints, 10);
+    solve(parallelDoubleConstraints, 10);
```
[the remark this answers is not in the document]
bru5: This is now derived from a new parameter called `iterations`.

src/utils/graph/routing.js
bru5: I've actually simplified this and removed the extra steps, since it looks like they are not all that effective, so this one should now be fine I think.

src/utils/graph/common.js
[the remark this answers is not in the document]
richardwestenra: Fair enough!

src/utils/graph/solver.js
richardwestenra: ahhhhh phew. Ok that should be sweet then, cheers 👍

src/utils/graph/layout.js
```diff
@@ -0,0 +1,319 @@
...
+      weightA: () => 0.5,
+      weightB: () => 0.5
+    };
+
+    parallelConstraints.push(parallelConstraint);
+
+    // Collect edges connected to single-degree nodes at either end
+    if (
+      edge.sourceNode.targets.length === 1 ||
+      edge.targetNode.sources.length === 1
```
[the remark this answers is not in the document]
bru5: I've given these some names, naturally it's a little confusing without a diagram still but I think it does help a bit.

src/utils/flags.js
```diff
@@ -0,0 +1,75 @@
...
+ * @param {string=} url The URL (optional, default current location)
+ * @returns {object} An object with flags and their values
+ */
+export const getFlagsFromUrl = url => {
+  const urlParams = new URL(url || document.location).searchParams;
+  const flags = {};
+
+  [...urlParams].forEach(([name, value]) =>
+    Flags.isDefined(name)
+      ? (flags[name] = value === 'true' || value === '1')
```
richardwestenra, line 43: I feel like it should probably check for `value === ''` and evaluate that to `true` too, no? Because I tend to regard valueless query string keys as being truthy even though technically their value is `''`, and I feel like it should be okay to just add `?newgraph` to turn the feature on. What do you think?

src/utils/flags.js
```diff
@@ -0,0 +1,75 @@
...
+      const isEnabled = flagsEnabled[name];
+      const status = isEnabled ? 'Enabled' : 'Disabled';
+      const statusIcon = isEnabled ? '🟢' : '⚪️';
+      const icon = flagsConfig[name].icon;
+      const description = flagsConfig[name].description;
+      info += `\n${statusIcon} ${icon} "${name}" · ${description} · ${status}`;
+    });
+
+    info += `\n\nSee docs on flags for more info 📖`;
+    info += `\nhttps://github.com/quantumblacklabs/kedro-viz#flags`;
```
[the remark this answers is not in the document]
bru5: Not yet, shall I go ahead and update them myself in this PR?

src/utils/flags.js
bru5: Good catch, I've added this one too, thanks.

src/utils/flags.js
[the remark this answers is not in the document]
richardwestenra: Yeah I think it's a good idea to do it sooner rather than later so it's not forgotten, but I don't mind whether it's part of this PR. If you don't make it part of this PR then definitely add a new ticket for this sprint so it's not missed out of the next release.

Also we should probably ask @stichbury to review the wording once it's ready.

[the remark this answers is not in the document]
stichbury: Yep, just give me a shout when you want a swift review of the docs @bru5!

README.md
```diff
@@ -100,6 +100,24 @@ As a JavaScript React component, the project is designed to be used in two diffe
...
+
+To enable or disable a flagged feature, add the flag as a parameter with the value `true` or `false` to the end of the URL in your browser when running Kedro-Viz, e.g.
+
+`http://localhost:4141/?data=demo&newgraph=true`
+
+This provided setting will then persist for all sessions on your machine, until it is otherwise changed.
+
+### Viewing flags
+
+See the message in your browser's console for info regarding the available flags and their values as currently set on your machine.
```
stichbury, line 119: I'm not sure I follow this. What message? What console?

README.md
```diff
@@ -100,6 +100,24 @@ As a JavaScript React component, the project is designed to be used in two diffe
...
+
+- `newgraph` - Improved graphing algorithm (default `false`).
+
+### Setting flags
+
+To enable or disable a flagged feature, add the flag as a parameter with the value `true` or `false` to the end of the URL in your browser when running Kedro-Viz, e.g.
+
+`http://localhost:4141/?data=demo&newgraph=true`
+
+This provided setting will then persist for all sessions on your machine, until it is otherwise changed.
```
stichbury, line 115:
```suggestion
The setting you provide persists for all sessions on your machine, until you change it.
```

README.md
```diff
@@ -100,6 +100,24 @@ As a JavaScript React component, the project is designed to be used in two diffe
 
    The React component exposes props that can be used to supply data and customise its behaviour. For information about the props, their expected prop-types and default values, see [/src/components/app/index.js](https://github.com/quantumblacklabs/kedro-viz/blob/master/src/components/app/index.js). For examples of the expected data input format, see the mock data examples in [/src/utils/data](https://github.com/quantumblacklabs/kedro-viz/tree/master/src/utils/data), and compare the [resulting demo](https://quantumblacklabs.github.io/kedro-viz/).
 
+## Flags
+
+The following flags are available in the latest release to toggle experimental features:
```
stichbury, line 105:
```suggestion
The following flags are available to toggle experimental features:
```

README.md
```diff
@@ -100,6 +100,24 @@ As a JavaScript React component, the project is designed to be used in two diffe
...
+
+- `newgraph` - Improved graphing algorithm (default `false`).
```
stichbury, line 107:
```suggestion
- `newgraph` - From release vX.x. Improved graphing algorithm (defaults to `false`).
```

My thoughts are that it's best to say which release they are introduced from individually as you add them, since "latest" is/will go out of date.

README.md
bru5: I've reworded with some extra information including a link to a Mozilla article explaining the [developer console](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_are_browser_developer_tools#The_JavaScript_console), does that help?

README.md
bru5: I've updated this line, thanks.

README.md
bru5: Sounds good, I've updated this to include the version.

921kiyo:
Legal headers LGTM :)

src/selectors/layers.js
```diff
@@ -13,14 +13,17 @@ export const getLayers = createSelector(
...
-      const bound =
-        bounds[node.layer] || (bounds[node.layer] = [Infinity, -Infinity]);
-      if (node.y - node.height < bound[0]) bound[0] = node.y - node.height;
-      if (node.y + node.height > bound[1]) bound[1] = node.y + node.height;
+      const layer = node.nearestLayer || node.layer;
+
+      if (layer) {
+        const bound = bounds[layer] || (bounds[layer] = [Infinity, -Infinity]);
+        if (node.y - node.height < bound[0]) bound[0] = node.y - node.height;
+        if (node.y + node.height > bound[1]) bound[1] = node.y + node.height;
```
richardwestenra, line 25: I'd prefer these not to be inline if statements and to add braces, just for improved readability? Is that ok with you?

README.md
```diff
@@ -102,21 +102,21 @@ As a JavaScript React component, the project is designed to be used in two diffe
 
 ## Flags
 
-The following flags are available in the latest release to toggle experimental features:
+The following flags are available to toggle experimental features:
 
-- `newgraph` - Improved graphing algorithm (default `false`).
+- `newgraph` - From release v3.3.2. Improved graphing algorithm (default `false`).
```
richardwestenra, line 107: Yeah I'd say `3.4.0`, though I doubt it's necessary to specify as it'll be in the release notes, and this will only be visible on branches that contain this flag, no?

package/features/environment.py
```diff
@@ -26,6 +26,14 @@
...
 # limitations under the License.
 
+#------------------------------------------------------------------------------
+# Copyright (c) 2013, Nucleic Development Team & H. Rutjes.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#------------------------------------------------------------------------------
+
```
richardwestenra, line 36: It'd make a lot more sense to me for us to only put this legal header on the one JS file that includes kiwi.js, and none of the Python files, as they don't even include kiwi.js.

src/selectors/layers.js
bru5: Good point I've added these.

README.md
bru5: Yeah I was thinking on the same lines but this was a suggestion from @stichbury earlier in this thread and it seems a good thing to keep - I've updated this to read `3.4.0`.

package/features/environment.py
bru5: Yeah I thought this seemed a little unusual too, after discussing with @yetudada here are the changes:

- reverted the notice from Python files (see in f92d68f) 
- added the notice as a special `@license` comment where it is imported in the JS source (see [here](https://github.com/quantumblacklabs/kedro-viz/pull/185/commits/f92d68fbf2d07e8d82ac3ce55e12274d05269259#diff-506cc426df42a78b3776a93989f292c6)) so this way it should also be retained in build files.

Does this suit all?

src/utils/graph/solver.js
```diff
@@ -0,0 +1,155 @@
+/** @license kiwi.js v1.1.2
```
yetudada, line 1: Kiwi.js is only used here?

README.md
stichbury: It does. So much better 🚀

stichbury:
Looks good to me -- the changes in ReadMe are spot on.

src/utils/graph/solver.js
bru5: Yeah this is the only file it is used in, just inside the function `solveStrict` below.

PR #186, opened 2020-06-04
[KED-1730] Improve rendering performance

richardwestenra:
## Description

Kedro-Viz often runs very slowly on large pipelines. While there's only a limited amount that we can do to fix this (as the layout algorithm is intrinsically expensive), I have tried to find ways to reduce unnecessary overhead where possible.

## Development notes

1. Prevent layout calc from running twice on pageload: Though I had tried to avoid from being layout being called until after fontLoaded was ready, I missed that it was being called by the export-graph function, which meant that it was called twice on page load (i.e. on first load and then again when the font loads). I have changed this so that it now explicitly listens to fontLoaded, and won't return if fontLoaded is false. The same is now true for some of the more expensive node selectors.

2. Optimise getNodeTextWidth selector: This selector is a relatively expensive one, as it involves some hefty DOM operations. It turns out (from benchmarking tests) that using D3's selectAll approach is faster than the approach that I was using.

## Checklist

## Legal notice

### DIFF
--- a/src/selectors/layout.js
+++ b/src/selectors/layout.js
@@ -9,6 +9,7 @@ const getHasVisibleLayers = state =>
 const getNodeType = state => state.node.type;
 const getNodeLayer = state => state.node.layer;
 const getVisibleSidebar = state => state.visible.sidebar;
+const getFontLoaded = state => state.fontLoaded;
 
 /**
  * Calculate chart layout with Dagre.js.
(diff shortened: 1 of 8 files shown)


src/selectors/layout.js
```diff
@@ -17,9 +18,9 @@ const getVisibleSidebar = state => state.visible.sidebar;
  * which don't affect layout.
  */
 export const getGraph = createSelector(
-  [getVisibleNodes, getVisibleEdges, getHasVisibleLayers],
-  (nodes, edges, hasVisibleLayers) => {
-    if (!nodes.length || !edges.length) {
+  [getVisibleNodes, getVisibleEdges, getHasVisibleLayers, getFontLoaded],
+  (nodes, edges, hasVisibleLayers, fontLoaded) => {
+    if (!fontLoaded || !nodes.length || !edges.length) {
```
bru5, line 23: Can the check on `fontLoaded` be pushed higher up? I know it's an optimisation, but as it doesn't seem to be this function's responsibility to know about this normally, so it would be cleaner to move it.

richardwestenra: I hear you, and I'm not 100% happy with this approach. However, I previously had this check way higher up elsewhere, and then I made a change which called this function elsewhere, and then it started being called prematurely again. So I think the most sensible thing to do is to put the check right here (in addition to some of the higher up functions), so that it's _explicit_ about not being called, and won't be accidentally affected by a future change.

bru5: Yeah I suspected it might be something like this, it's not ideal but if that's the most practical approach then seems fine with me, thanks for confirming.

PR #191, opened 2020-06-09
[KED-1833] Hook implementation for kedro-viz

MerelTheisenQB:
## Description
POC: Adding in an implementation for hooks, exactly like in Kedro, to allow plugins using the hook to add additional data to kedro-viz. Frontend changes to pick up this extra data and show it will follow.

This idea came from wanting to integrate kedro-ge with kedro-viz. Kedro-ge generates validation and profiling reports. It would be amazing to have those linked inside kedro-viz, so users don't have to spin up yet another UI for these results. See this PR on the kedro-ge side that shows how the hook is used: https://github.com/quantumblack/kedro-great-expectations/pull/76 

The hook spec is registered within the Kedro-GE mixin, so inside the kedro project the hook itself can just be added as any Kedro hook. 

`
hooks = (
        KedroGEVizHooks(),
    )
`
## Development notes
* Added `specs.py` to hold the kedro-viz hook specs.
* Refactored `_call_viz` inside `server.py` and added call to hook right after the Pipeline has been formatted.

## QA notes
I ran a toy kedro project to make sure the additional data shows up in `nodes.json` when running kedro viz. 

## Checklist

## Legal notice

- [X] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -500,7 +500,7 @@ def _call_viz(
 
     enhanced_data = None
     if KEDRO_VERSION.match(">=0.16.0"):
-        enhanced_data = _get_context(project_path, env)._hook_manager\
+        enhanced_data = context._hook_manager\
             .hook.after_api_data_ready(api_data=data)  # pylint: disable=protected-access
 
     if save_file:
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -487,22 +487,8 @@ def _call_viz(
         data = _load_from_file(load_file)
     else:
         if KEDRO_VERSION.match(">=0.15.0"):
-            # pylint: disable=import-outside-toplevel
-            if KEDRO_VERSION.match(">=0.16.0"):
-                from kedro.framework.context import KedroContextError
-            else:
-                from kedro.context import KedroContextError
-
-            try:
-                if project_path is not None:
-                    context = get_project_context("context", project_path=project_path, env=env)
-                else:
-                    context = get_project_context("context", env=env)
-                pipeline = _get_pipeline_from_context(context, pipeline_name)
-            except KedroContextError:
-                raise KedroCliError(ERROR_PROJECT_ROOT)
-            catalog = context.catalog
-
+            pipeline = _get_pipeline_from_context(_get_context(project_path, env), pipeline_name)
+            catalog = _get_context(project_path, env).catalog
         else:
             # Kedro 0.14.*
             if pipeline_name:
@@ -511,10 +497,30 @@ def _call_viz(
 
         data = format_pipeline_data(pipeline, catalog)
 
+    enhanced_data = _get_context(project_path, env)._hook_manager\
+        .hook.after_pipeline_formatted(formatted_pipeline=data)  # pylint: disable=protected-access
+
     if save_file:
+        if enhanced_data:
+            data = enhanced_data
         Path(save_file).write_text(json.dumps(data, indent=4, sort_keys=True))
     else:
         is_localhost = host in ("127.0.0.1", "localhost", "0.0.0.0")
         if browser and is_localhost:
             webbrowser.open_new("http://{}:{:d}/".format(host, port))
         app.run(host=host, port=port)
+
+
+def _get_context(project_path=None, env=None):
+    # pylint: disable=import-outside-toplevel
+    if KEDRO_VERSION.match(">=0.16.0"):
+        from kedro.framework.context import KedroContextError
+    else:
+        from kedro.context import KedroContextError
+
+    try:
+        if project_path is not None:
+            return get_project_context("context", project_path=project_path, env=env)
+        return get_project_context("context", env=env)
+    except KedroContextError:
+        raise KedroCliError(ERROR_PROJECT_ROOT)
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -501,7 +501,7 @@ def _call_viz(
     enhanced_data = None
     if KEDRO_VERSION.match(">=0.16.0"):
         enhanced_data = _get_context(project_path, env)._hook_manager\
-            .hook.after_pipeline_formatted(formatted_pipeline=data)  # pylint: disable=protected-access
+            .hook.after_api_data_ready(api_data=data)  # pylint: disable=protected-access
 
     if save_file:
         if enhanced_data:
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -487,8 +487,9 @@ def _call_viz(
         data = _load_from_file(load_file)
     else:
         if KEDRO_VERSION.match(">=0.15.0"):
-            pipeline = _get_pipeline_from_context(_get_context(project_path, env), pipeline_name)
-            catalog = _get_context(project_path, env).catalog
+            context = _get_context(project_path, env)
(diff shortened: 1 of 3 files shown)


package/kedro_viz/server.py
```diff
@@ -511,10 +498,32 @@ def _call_viz(
...
 
+    enhanced_data = None
+    if KEDRO_VERSION.match(">=0.16.0"):
+        enhanced_data = _get_context(project_path, env)._hook_manager\
+            .hook.after_pipeline_formatted(formatted_pipeline=data)  # pylint: disable=protected-access
+
     if save_file:
+        if enhanced_data:
+            data = enhanced_data
         Path(save_file).write_text(json.dumps(data, indent=4, sort_keys=True))
```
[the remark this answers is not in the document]
MerelTheisenQB: @tsanikgr I had this, but it doesn't work. I think there might be some FE magic happening that checks what's in `data` and it probably doesn't like the added validation and profiling fields.

package/kedro_viz/server.py
```diff
@@ -511,10 +498,32 @@ def _call_viz(
 
         data = format_pipeline_data(pipeline, catalog)
 
+    enhanced_data = None
+    if KEDRO_VERSION.match(">=0.16.0"):
+        enhanced_data = context._hook_manager\
```
DmitriiDeriabinQB, line 503: Also, if there are multiple registrations of this hook implementation, hook call will [return a list of results](https://pluggy.readthedocs.io/en/latest/index.html#collecting-results), which, I believe, will break the frontend.

limdauto:
Thanks @DmitriiDeriabinQB for your detailed comment. I have been thinking a lot about this. I have been refraining from commenting further because I would like to see the outcome of the product discussion with the viz team first. cc @yetudada 

As @MerelTheisenQB has demonstrated, linking kedro-ge and kedro-viz is possible. As @DmitriiDeriabinQB has pointed out, there are a couple of different ways to do this with tradeoffs.

My biggest concern, however, is about the high-level vision for this suite of products, which should influence how we engineer this. From a product point of view, this essentially wants 2 independent web pages having hyperlinks between each other. In this PR, it's linking kedro-viz to kedro-ge, but I'm sure the need to link kedro-ge back to kedro-viz is only a matter of time. Or kedro-ge to kedro-pai, kedro-pai to kedro-viz, etc. 

We _could_ go with a model where every plugin has its own hook manager and a set of specs. Other plugins are free to implement these specs. The upside is less burden on Kedro core and more flexibility, as @DmitriiDeriabinQB has pointed out. However, the downside is that the dependencies between these plugins will grow exponentially (2^n, with n=number of plugins). This is similar to a microservice situation where every service can call every other service freely. Maybe that won't become any problem, but I think it's very likely that users might find some plugin not working at some point because another plugin is interfering with it.

Alternatively, let's take a step back and apply the principle of inversion of control here. What we want to give people is a coherent and rich experience of using Kedro. Maybe there should be something sitting one layer up that knows about all plugins and direct users accordingly. Right now, Kedro core is the only place that can do it. However, we could also promote kedro-viz to be this coordinating layer, in which case making kedro-viz extensible with its own hook mechanism makes sense. But it needs to be done deliberately with this longer-term vision in mind, because there might be other easier options to serve as this layer too. Why not studio, for example?

cc @tsanikgr @idanov

MerelTheisenQB:
Thanks so much for your review @DmitriiDeriabinQB ! And thanks @limdauto for chipping in and explaining a lot better than I could have done 😉 

I think the question about whether the kedro-viz should have its own hook manager or not is very valid. This PR is a POC to show that hooks for viz are relevant and offer a lot of opportunities. I agree with Lim that it's a question of how we envision these products and future products to be linked and share data on a higher level. 

I personally understand the pros+cons of both approaches and don't have a very strong preference. This seems to be a decision on Kedro/Kedro-Viz's architecture, kedro-ge is merely a 'consumer' here and it's internals shouldn't matter to Kedro. I'm happy to setup a meeting if that makes it easier to discuss this.

DmitriiDeriabinQB:
@limdauto I personally am okay with having this ge -> viz, viz -> ge, viz -> pai, I also don't see a big problem in such individual connections as currently they are always unidirectional. Host plugin exposes a hook that whoever concerned may implement and register using the entry point system. I don't see much of an immediate value in "centralising" such communication through Kedro (or Kedro-Viz) as there is no real associated cost (like http contract, queue service to care about, latency SLAs etc.), on the contrary I like those connections to be logically isolated. If we, however, decided to make Kedro responsible for hosting such communication (acting like a message bus between its plugins), I would prefer having a clear logical separation - something like `get_hook_manager("kedro-viz")`.

@MerelTheisenQB Yeah, we probably need to discuss it online 😄 Thanks a lot for bringing up this topic 👍  This is a very good question that we haven't thought of much ourselves. Would also be great to include @idanov in the discussion.

limdauto:
@DmitriiDeriabinQB Thanks, I think your reasoning makes sense.

The centralised model I have in mind is quite similar to `pytest`. `pytest` allows plugins to declare their own hooks and add them to pytest's plugin manager: https://docs.pytest.org/en/2.7.3/plugins.html#_pytest.hookspec.pytest_addhooks. It's also similar to what @MerelTheisenQB implemented here. If we decide to do this, we will need better support on the Kedro side through the entrypoint system for sure. 

The biggest advantage of this centralised approach, IMO, is that on the consumer / plugin side (the GE side), users should be able to provide as many hook implementations as they want without needing to explicitly check whether the host (Kedro-Viz) is available. If the host is available, Kedro will automatically load their specs and register them with the global hook manager. 

But I agree this might not be as big of a problem as I'm making it out to be in my head, and an explicit model where plugins are aware of each other is the way to go here.

idanov:
@MerelTheisenQB thank you for creating the POC, I think there's a lot of benefit in connecting **kedro-ge** and **kedro-viz**. As this is a POC, I am not sure whether we can merge it at the moment before we have a bigger discussion on the architecture of **kedro** and **kedro-viz** to allow for such integrations. I agree with @limdauto that this needs a bigger product discussion, and I believe an ad-hoc integration now can limit us in what we can do in the future, especially when we need to preserve backwards compatibility, which is something **kedro** needs to account for due to all the stuff people build on top of it (**kedro-ge** is just one of our client projects). However this PR is a great conversation starter and a good prompt that we need to have a better product backlog alignment in QB Labs.

package/kedro_viz/server.py
```diff
@@ -487,22 +487,9 @@ def _call_viz(
...
-
-            try:
-                if project_path is not None:
-                    context = get_project_context("context", project_path=project_path, env=env)
-                else:
-                    context = get_project_context("context", env=env)
-                pipeline = _get_pipeline_from_context(context, pipeline_name)
-            except KedroContextError:
-                raise KedroCliError(ERROR_PROJECT_ROOT)
+            context = _get_context(project_path, env)
```
idanov, line 490: Nice 👍

yetudada:
This is exciting work to see! I've been able to read all perspectives for this and ultimately I've enjoyed how you've thought about the user experience in this PR and in #194. We do need to have a conversation about how best to build this, I'll get on organising that session. And @MerelTheisenQB thank you so much for triggering this conversation.

yetudada:
Same as my comment on #194, we'll pick up this work when we look into the Kedro-GE integration with Kedro-Viz. Thank you so much for making this PR, it's given us bigger aspirations.

PR #203, opened 2020-07-02
[KED-1729] Add minimap

bru5:
## Description

Adds an interactive minimap to the toolbar.

Screenshot 2020-07-02 at 09 45 20

## Development notes

The PR adds a new minimap and toolbar component. 

It also makes some changes and improvements to the main flowchart component that were added for performance, so it should now be more efficient in making drawing calls.

## QA notes

- The minimap is opened by clicking the new map button on the sidebar.
- The size is dynamic depending on the graph size and viewport size.
- The main view can be panned by dragging the viewport around in the minimap.
- The minimap is animated and shows highlighting similar to the main view.
- Toggling labels will also apply to the minimap.
- Light and dark themes are implemented.

## Checklist

## Legal notice


src/components/minimap-toolbar/minimap-toolbar.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../styles/variables';
+@import '../../styles/extends';
+
+.pipeline-minimap-toolbar {
+  display: block;
+  padding: 0;
+  list-style: none;
+
+  .pipeline-icon-toolbar__button {
```
richardwestenra, line 9: This nesting is unnecessary - that's why we use BEM

src/components/minimap/draw.js
```diff
@@ -0,0 +1,88 @@
...
+  this.el.viewport
+    .attr('transform', `translate(${minX}, ${minY})`)
+    .attr('width', Math.max(0, maxX - minX))
+    .attr('height', Math.max(0, maxY - minY));
+};
+
+/**
+ * Render nodes
+ */
+export const drawNodes = function() {
```
richardwestenra, line 31: Is there no way to reuse the other drawNodes function from the main flowchart but customise it a little?

src/components/minimap/index.js
```diff
@@ -0,0 +1,422 @@
...
+  /**
+   * Handle pointer wheel
+   * @param {Object} event Event object
+   */
+  onPointerWheel = event => {
+    const { scale = 1 } = this.props.chartZoom;
+
+    // Change zoom based on wheel velocity
+    this.props.onUpdateChartZoom({
+      scale: scale - (event.deltaY || 0) * scale * 0.0025,
```
richardwestenra, line 195: Where does this number come from?

src/components/sidebar/sidebar.scss
```diff
@@ -2,12 +2,12 @@
 
 .pipeline-sidebar {
   position: absolute;
-  top: 0;
+  top: -1px; /* Avoids pixel rounding gaps */
+  bottom: -1px;
```
richardwestenra, line 6: 😮

src/components/minimap/styles/minimap.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+.pipeline-minimap-container {
+  $offsetBottom: 2px; /* Sub-pixel adjustment */
+  position: absolute;
+  top: calc(100% + #{$offsetBottom});
+  left: 100%;
+  width: 500px;
+  height: 222px;
+  box-shadow: 0 0 10px rgba(0, 0, 0, 0.07);
```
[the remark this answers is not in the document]
bru5: I see what you mean, though I do think for a minimap it's a good idea for it to be relatively unobtrusive as otherwise it's tough to focus on the main screen.

We should check with @GabrielComymQB on this, but I think if we did want to separate it more then maybe just a subtle pixel border might be cleaner than a shadow? I think potentially in the designs there might not be any shadow actually, I was just naturally following from the sidebar but that might be the intended to keep a crisp contrast?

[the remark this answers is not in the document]
richardwestenra: The CSS I supplied is definitely subtle - it's only 2px and virtually invisible:
image 
image

But you could definitely try changing it to `0 0 0 1px`, I just thought this was subtler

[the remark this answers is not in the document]
bru5: Yeah I'd tried it, that drop shadow is subtler but since it's still soft it doesn't give a crisp edge which I think might be important for contrast?

src/components/sidebar/sidebar.scss
bru5: Yeah while working on the minimap I'd noticed there's a missing pixel gap around the toolbar, so you could see a bit of the diagram under it.

I think it's usually due to transforms and GPU rounding type issues, so this is a good idea to take care of it as it can be a little distracting, if that makes sense?

src/components/app/app.scss
```diff
@@ -10,4 +10,18 @@
   .kui-theme--dark & {
     fill: #fff;
   }
+
+  &.pipeline-icon-map,
+  &.pipeline-icon-plus,
+  &.pipeline-icon-minus {
```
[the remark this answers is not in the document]
bru5: I've added the latter for now, since we'll be updating the icons later anyway and can remove once they're in I think.

src/components/minimap/index.js
bru5: I've added a name for this, but essentially just trial and error to get a reasonable speed.

src/components/sidebar/sidebar.scss
richardwestenra: Yup! Interesting. Thanks for fixing 👍

src/components/app/app.scss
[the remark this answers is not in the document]
richardwestenra: Good call, I'd have done the same.

src/components/minimap/styles/minimap.scss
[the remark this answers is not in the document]
richardwestenra: Happy to yield to @GabrielComymQB here, but my main point is that I think it should be a white glow/outline rather than a black shadow, to make it easier to identify the edge of the box.

src/components/minimap/draw.js
richardwestenra: Makes sense

src/components/minimap-toolbar/minimap-toolbar.scss
richardwestenra: We should never be nesting classes if we don't have to, because it unnecessarily increases CSS specificity. This is the main motivation behind BEM: To allow class combinations without increasing specificity. The above would output:
```css
.pipeline-minimap-toolbar { ... }
.pipeline-minimap-toolbar .pipeline-icon-toolbar__button { ... }
```
which is more specific than the following, without giving any benefit:
```css
.pipeline-minimap-toolbar { ... }
.pipeline-icon-toolbar__button { ... }
```

The theme classes (i.e. `.kui-theme--light & {}`) are also nested, which is unfortunate but necessary, given the way Kedro-UI applies theming. I'd like to fix this with CSS custom properties in future, but for now it'll have to stay that way.

bru5: The increased specificity was intended in this case because that descendent button is inside another component whose style is not owned by the host component, but instead of this I was able to remove the nesting by using a unique classname instead.

bru5:
> I'd still like for @GabrielComymQB to weigh in on the CSS shadow/highlight, but other than that, nice work!
> 
> :shipit:

Oh I'd actually added these earlier but looks like it wasn't pushed, they're there now.

PR #209, opened 2020-07-06
[KED-1859] Increase width of layer rects

richardwestenra:
## Description

This is to avoid the ends of the layer rects being visible on tall thin graphs, e.g.
http://localhost:4141/?data=random&seed=jZkSoKJvyzJVQjELcepsJfWRVEWcr5&newgraph
image

## Development notes

The ultimate fix would be to remove the layer rects from the scale/pan container and just change their y/height on zoom, while keeping their x/width consistent. This is what currently happens for the labels. However, this would require the DOM attributes to be updated on zoom tick, which might be worse for performance. So for now I reckon it's safer/better to just ensure that the widths are much wider than necessary.

## Checklist

## Legal notice

### DIFF
--- a/src/selectors/layers.js
+++ b/src/selectors/layers.js
@@ -9,7 +9,7 @@ const getLayerName = state => state.layer.name;
  */
 export const getLayers = createSelector(
   [getLayoutNodes, getVisibleLayerIDs, getLayerName, getGraphSize],
-  (nodes, layerIDs, layerName, { width }) => {
+  (nodes, layerIDs, layerName, { width, height }) => {
     if (!nodes.length) {
       return [];
     }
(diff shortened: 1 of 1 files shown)


src/selectors/layers.js
```diff
@@ -44,13 +44,14 @@ export const getLayers = createSelector(
       ];
       const start = (prevBound[1] + currentBound[0]) / 2;
       const end = (currentBound[1] + nextBound[0]) / 2;
+      const rectWidth = Math.max(width, height) * 5;
 
       return {
         id,
         name: layerName[id],
-        x: -width / 2,
+        x: (rectWidth - width) / -2,
```
bru5, line 52: Not an issue but I think `(width - rectWidth) / 2` might work here too and avoid that negative denominator.

richardwestenra: Ah interesting, good shout thanks! Probably won't bother now but good to bear in mind.

PR #215, opened 2020-07-08
[KED-1874] Move data loader into standalone-app entry point

richardwestenra:
## Description

- **Move data-source logic from initial-state:** The test datasets aren't necessary in the exported component lib, and are only required when running the full app. So I've moved the data source loading functionality out of the App/initial-state logic
- **Replace window.location with document.location:** This allows us to remove the typeof window checks, as document is available in build while window isn't.
- **Remove conditional window checks on svg-crowbar:** These are no longer necessary as of [svg-crowbar v0.4.0](https://github.com/cy6erskunk/svg-crowbar/releases/tag/v0.4.0), updated in #208
- **Load json in src/index.js instead of App component:** This simplifies the App component and moves unnecessary logic out of the exported library component
- **Allow an argument to be passed to loadJsonData:** This will allow external users to load and run it if they wish. e.g.
```JSX
import KedroViz from '@quantumblack/kedro-viz`;
import loadData from '@quantumblack/kedro-viz/lib/store/load-data';

loadData('/path/to/data.json').then(data => {
  ReactDOM.render(<KedroViz data={data} />, document.getElementById('root'));
});
```

## QA notes

I've changed the `componentDidUpdate` check to not pay attention to the schema_id. Hopefully this won't change anything. Not sure why I didn't do this before.

## Checklist

## Legal notice

### DIFF
--- a/src/utils/data-source.test.js
+++ b/src/utils/data-source.test.js
@@ -37,7 +37,7 @@ describe('getDataValue', () => {
     expect(getDataValue('json')).toEqual('json');
   });
 
-  it("should return random data when passed 'random'", () => {
+  it("should return a dataset object when passed 'random'", () => {
     expect(getDataValue('random')).toEqual(
       expect.objectContaining({
         edges: expect.any(Array),
--- a/src/utils/data-source.test.js
+++ b/src/utils/data-source.test.js
@@ -1,6 +1,61 @@
-import getDataSource from './data-source';
+import getPipelineData, { getSourceID, getDataValue } from './data-source';
+import animals from './data/animals.mock';
+import demo from './data/demo.mock';
 
-describe('getDataSource', () => {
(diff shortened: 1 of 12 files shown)


src/utils/data-source.test.js
```diff
@@ -1,6 +1,61 @@
...
+  it('should return the correct dataset when passed a dataset string', () => {
+    expect(getDataValue('animals')).toEqual(animals);
+    expect(getDataValue('demo')).toEqual(demo);
+  });
+
+  it("should return the string 'json' when passed 'json'", () => {
+    expect(getDataValue('json')).toEqual('json');
+  });
+
+  it("should return random data when passed 'random'", () => {
```
bru5, line 40: I think this should probably say `"should return a dataset object when passed 'random'"` since this check seems to only be about the type.

PR #220, opened 2020-07-24
[KED-1926] Refactor initial state setup, and store disabled nodes in localStorage

richardwestenra:
## Description

1. The initial state setup was getting pretty unmanageable, so I've used `deepmerge` to simplify it. Now, when you start up the app, it loads the initial state, normalises the data into it, merges the localStorage data into it, then applies some manual overrides. But it should be much simpler and easier to read.
2. I've added `state.node.disabled` to localStorage, so now when you disable an individual node, it will stay disabled unless you change the name.

## Development notes

I've added a wee utility to prune falsey keys from the localStorage when saving it, so that it doesn't add useless values and clutter the UI unnecessarily.

## QA notes

Two mild concerns:
1. Increased onload performance cost from the initial deepmerge. In tests it was 1-50ms, but usually approx 5ms.
2. If node names change frequently then the store might get rapidly bigger (as users won't be able to delete old data via the UI), but I doubt it?

## Checklist

## Legal notice


src/store/initial-state.js
```diff
@@ -28,7 +28,23 @@ export const createInitialState = () => ({
 });
 
 /**
- * Add state overrides from props, URL flags, etc
+ * Load values from localStorage and combine with existing state,
+ * but filter out any unused values from localStorage
+ * @param {object} state Initial/extant state
+ * @return {object} Combined state from localStorage
+ */
+export const applyLocalStorage = state => {
```
bru5, line 36: I think `merge` might be clearer than `apply` in the name here

PR #221, opened 2020-07-27
[KED-1829] Add top-level pipeline selection UI

richardwestenra:
## Description

Add Kedro-UI dropdown to select top-level modular pipelines.

## Development notes

WIP

## Checklist

## Legal notice

### DIFF
--- a/src/components/pipeline-list/index.js
+++ b/src/components/pipeline-list/index.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Dropdown from '@quantumblack/kedro-ui/lib/components/dropdown';
+import MenuOption from '@quantumblack/kedro-ui/lib/components/menu-option';
+import { updateActivePipeline } from '../../actions';
+import './pipeline-list.css';
+
+/**
+ * A Dropdown displaying a list of selectable pipelines
+ * @param {Object} pipeline Pipeline IDs, names, and active pipeline
+ * @param {Function} onUpdateActivePipeline Handle updating the active pipeline
+ * @param {string} theme Kedro UI light/dark theme
+ */
+export const PipelineList = ({ onUpdateActivePipeline, pipeline, theme }) => {
+  if (!pipeline.ids.length) {
+    return null;
+  }
+  return (
+    <div className="pipeline-list">
+      <Dropdown
+        theme={theme}
+        width={null}
+        onChanged={onUpdateActivePipeline}
+        defaultText={pipeline.name[pipeline.active]}>
+        {pipeline.ids.map(id => (
+          <MenuOption
+            key={`pipeline-${id}`}
+            value={id}
+            selected={id === pipeline.active}
+            primaryText={pipeline.name[id]}
+          />
+        ))}
+      </Dropdown>
+    </div>
+  );
+};
+
+export const mapStateToProps = state => ({
+  pipeline: state.pipeline,
+  theme: state.theme
+});
+
+export const mapDispatchToProps = dispatch => ({
+  onUpdateActivePipeline: event => {
+    dispatch(updateActivePipeline(event.value));
+  }
+});
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(PipelineList);
--- a/src/components/pipeline-list/pipeline-list.scss
+++ b/src/components/pipeline-list/pipeline-list.scss
@@ -1,0 +1,11 @@
+@import '../../styles/extends';
+@import '../../styles/variables';
+
+.pipeline-list {
+  padding-bottom: 1.8em;
+
+  .kui-dropdown {
+    display: block;
+    margin: 0 $sidebar-padding;
+  }
+}
--- a/src/components/pipeline-list/pipeline-list.test.js
+++ b/src/components/pipeline-list/pipeline-list.test.js
@@ -1,0 +1,124 @@
+import React from 'react';
+import TagList, {
+  TagList as UnconnectedTagList,
+  mapStateToProps,
+  mapDispatchToProps
+} from './index';
+import { mockState, setup } from '../../utils/state.mock';
+
+describe('TagList', () => {
+  it('renders without crashing', () => {
+    const wrapper = setup.mount(<TagList />);
+    const container = wrapper.find('.pipeline-tags');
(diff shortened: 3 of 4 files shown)


bru5 (approved):

src/selectors/pipeline.js
```diff
@@ -0,0 +1,50 @@
...
+);
+
+/**
+ * Get IDs of tags used in the current pipeline
+ */
+export const getPipelineTagIDs = createSelector(
+  [getTagIDs, getPipelineNodeIDs, getNodeTags],
+  (tagIDs, nodeIDs, nodeTags) => {
+    const visibleTags = {};
+    for (let tagID of tagIDs) {
```
bru5, line 40: This function appears to be something like `O(ts·ns·k)` i.e. `O(n^3)` worst case, is that right? I have a feeling this might be refactored to be linear instead?

src/selectors/pipeline.test.js
```diff
@@ -0,0 +1,120 @@
...
+  describe('getPipelineTagIDs', () => {
+    it('returns an array of tag IDs', () => {
+      expect(getPipelineTagIDs(mockState.animals)).toEqual(
+        expect.arrayContaining(getTagIDs(mockState.animals))
+      );
+    });
+
+    it('does not contain any tags that are not in the current pipeline', () => {
+      // Tag 'huge' is not in pipeline 'de'
+      const activePipeline = 'de';
```
bru5: Seems like this test could be too data set specific and therefore unreliable? Ideally the set up for it would be isolated in here.

src/store/initial-state.js
```diff
@@ -44,24 +44,16 @@ export const mergeLocalStorage = state => {
...
-  if (props.visible) {
-    state.visible = Object.assign(state.visible, props.visible);
-  }
+export const preparePipelineState = props => {
+  // Normalize raw data, and apply saved state from localStorage
+  const state = mergeLocalStorage(normalizeData(props.data));
   // Turn layers off if there are no layers present:
   if (!state.layer.ids.length) {
-    state.visible.layers = false;
+    state.visible = Object.assign({}, state.visible, { layers: false });
```
bru5, line 56: Could be written as `state.visible = { ...state.visible, layers: false }` which I feel shows the intention of creating a new object more clearly, assuming that was the reason for the change?

src/store/initial-state.js
```diff
@@ -71,21 +63,35 @@ export const overrideInitialState = (state, props) => {
...
- * Configure the redux store's initial state
+ * Prepare the non-pipeline data part of the state. This part is separated so that it
+ * will persist if the pipeline data is reset.
+ * Merge local storage and add custom state overrides from props etc
+ * @param {object} props Props passed to App component
+ */
+export const prepareNonPipelineState = props => {
+  const state = mergeLocalStorage(createInitialState());
+  // Override flag defaults with URL values (on page load only)
+  state.flags = Object.assign({}, state.flags, getFlagsFromUrl());
```
bru5, line 74: Same as above for the `Object.assign` usage in these kinds of case?

src/store/initial-state.test.js
```diff
@@ -51,56 +52,49 @@ describe('mergeLocalStorage', () => {
...
-      ids: []
-    },
-    visible: {
-      miniMap: true,
-      sidebar: true
-    }
-  };
+describe('preparePipelineState', () => {
+  it('disables layer button if there are no layers present', () => {
+    const data = Object.assign({}, animals, { layers: undefined });
```
bru5, line 57: Same as above with `Object.assign({}, ...`

bru5 (changes requested):

src/selectors/pipeline.js
richardwestenra: I'd welcome your suggestions. This is actually what I came up with after refactoring my initial attempt, which was something like this:
```es6

    const visibleTags = {};
    for (let nodeID of nodeIDs) {
      for (let tagID of nodeTags[nodeID]) {
        if (!visibleTags[tagID]) {
          visibleTags[tagID] = true;
        }
      }
    }
```
which would keep going through all the nodes even after checking all the tags. I think the one I came up with is better, and it seems to be faster in testing (seems to usually take about half as many operations).

src/selectors/pipeline.test.js
richardwestenra: Yes I agree, but I couldn't think of a good way of doing it while essentially replicating the selector. unless.. hmm, maybe I could programmatically add an extra tag/node/pipeline without manually adding them to the dataset?

src/store/initial-state.js
richardwestenra: I have a slight preference for Object.assign here (which in practice is more often than not used to create a new object). Also taking transpiled size into account, there's a decent chance that Object.assign will be smaller. But overall I really don't see much significant difference between the two.

src/selectors/pipeline.js
bru5: Yeah I realise that it's fast enough in practice for now, but I think it's a good idea to avoid this anyway if there's an equally simple linear version.

Would this do the trick?

```js
export const getPipelineTagIDs = createSelector(
  [getPipelineNodeIDs, getNodeTags],
  (nodeIDs, nodeTags) => {
    const tagIds = {};

    for (const nodeID of nodeIDs) {
      for (const tagID of nodeTags[nodeID]) {
        tagIds[tagID] = true;
      }
    }

    return Object.keys(tagIds);
  }
);
```

src/selectors/pipeline.test.js
bru5: Sounds fine to me, I think it would be a good idea as currently the test is not going to be as robust.

src/store/initial-state.js
bru5: Yeah it's not a big deal for sure, I know what you mean as I've seen `Object.assign` being used for cloning like this before object spread was a thing, but since we have support for it I feel it's a more natural way to express this.

src/selectors/pipeline.js
richardwestenra: yep that's pretty much Option B, but without the filter (which I was undecided on). Honestly it depends on the data shape and how many tags there are vs the number of nodes. But yeah I might as well use that one. done.

PR #231, opened 2020-08-05
[KED-1522] Expose node and dataset metadata in "api/nodes/" endpoint

921kiyo:
## Description

Exposing Kedro node and dataset metadata

### Node metadata: 
- `code_location`: location of a Kedro node function
- `code`: the entire code snippet of a Kedro node function (including docstring, if any)
- `docstring`: Docstring of a node function, if exists.

### DataSet metadata:
- `dataset_location`: location of input/output dataset
- `dataset_type`: full module of the dataset class

UI of metadata side base looks like this (exposed metadata in this PR are highlighted in red)
Screenshot 2020-08-05 at 12 49 14

### Metadata JSON example

These are examples of what each JSON will look like. 

#### `/api/nodes/task_id`

```json
// 20200805124556
// http://127.0.0.1:4141/api/nodes/27087ee9

{
  "code": "def split_data(data: pd.DataFrame, example_test_data_ratio: float) -> Dict[str, Any]:\n ...",
  "code_location": "hello/src/hello/pipelines/data_engineering/nodes.py",
  "docstring": "Node for splitting the classical Iris data set into training and test\n ...."
}
```

If users didn't write a docstring, `docstring` won't show up in JSON (instead of returning `{"docstring": null}`).

#### `/api/nodes/dataset_id` (for both input and output dataset)

```json
// 20200805124617
// http://127.0.0.1:4141/api/nodes/7fe1d536

{
  "dataset_location": "/Users/kiyohito_kunii/Desktop/hello/data/01_raw/iris.csv",
  "dataset_type": "kedro.extras.datasets.pandas.csv_dataset"
}
```

*Note for back-end dev team*: The value of `dataset_type` is full module path for now. While UI example of this is not a full module but rather the exact definition of catalog like `pandas.ParquetDataSet`, we are returning the module path in the first iteration, because it's not possible to distinguish between a user's custom dataset (in this case we should return full module path), and a built-in dataset class (in this case we should return `pandas.CSVDataSet`).  

We will improve this once we have https://jira.quantumblack.com/browse/KED-986 implemented on Kedro side, so we can deserialise DataSet into a yaml definition. 

#### `/api/nodes/param_id`

Since there is no metadata for parameter, the output is an empty JSON.

```json
// 20200805125148
// http://127.0.0.1:4141/api/nodes/d577578a

{
  
}
```

#### Invalid endpoint (e.g `/api/nodes/invalid_id`)

Both invalid ID and paramter ID will return an empty JSON now. 

```json
// 20200805125708
// http://127.0.0.1:4141/api/nodes/invalid_id

{
  
}

```

## Checklist

## Legal notice


package/tests/test_server.py
```diff
@@ -46,117 +47,20 @@
 from kedro_viz.server import _allocate_port, _sort_layers, format_pipelines_data
 from kedro_viz.utils import WaitForException
 
-EXPECTED_PIPELINE_DATA = {
```
921kiyo, line 49: I had to move this somewhere because the linter was complaining that `test_server.py` exceeded 1000 lines. Nothing in `EXPECTED_PIPELINE_DATA` was changed, it's just copy/paste.

package/kedro_viz/server.py
```diff
@@ -474,7 +477,61 @@ def format_pipeline_data(
...
+    return jsonify(_DATA)
+
+
+@app.route("/api/nodes/<string:node_id>")
+def nodes_metadata(node_id):
+    """Serve the metadata for node and dataset."""
+    if _PIPELINES:
+        for pipeline in _PIPELINES.values():
+            for node in sorted(pipeline.nodes, key=lambda n: n.name):
+                task_id = _hash(str(node))
```
limdauto, line 489: In `format_pipeline_data`, can I suggest we keep a global dictionary of `node_id -> node` so we don't have to keep searching for the node every time by looping through all pipelines? That way we also don't have to perform extra logic to match dataset with node id.

package/kedro_viz/server.py
```diff
@@ -56,18 +57,18 @@
...
 
 
 _VIZ_PROCESSES = {}  # type: Dict[int, multiprocessing.Process]
 
 _DEFAULT_KEY = "__default__"
 
-data = None  # pylint: disable=invalid-name
+_DATA = None
+_PIPELINES = None
+_CATALOG = None
```
limdauto, line 71: Could you type annotate these so know what data types are they?

package/kedro_viz/server.py
```diff
@@ -474,7 +483,40 @@ def format_pipeline_data(
...
+    node = _NODES.get(node_id)
+    if node and node.get("type") == "task":
+        task_metadata = {
+            # pylint: disable=protected-access
+            "code": inspect.getsource(node["obj"]._func)
+        }
+        # pylint: disable=protected-access
+        code_full_path = Path(inspect.getfile(node["obj"]._func)).expanduser().resolve()
+
+        code_location = Path(code_full_path).relative_to(Path.cwd().parent)
```
921kiyo, line 500: I used `Path.cwd()` to get the project root, given that this endpoint is only accessible when users run `kedro viz` from the project root (this endpoint is not reachable if they do `kedro viz --load-file=data.json`). 

Another way to get the project root is to store context object in a global `_CONTEXT` variable (or store a dictionary if they use Kedro<0.15.* since context was introduced in 0.15.0). If we prefer this approach, I can change that too :)

921kiyo:
> Should we be including the full path on the user's OS in the code/data location string, or just the path relative to the project directory? Seems a bit weird to be including the `/Users/kiyo/home/Desktop/...` bit.

@richardwestenra Thank you, that's a good point! I changed `code_location` to exclude `/Users/kiyo/home/Desktop/...` bit :) 
For `data_location` however, we need a full path because the data could come from cloud storage (e.g AWS S3 bucket, Azure Blob etc)

package/kedro_viz/server.py
```diff
@@ -452,10 +452,25 @@ def format_pipeline_data(
...
     for namespace, tag_names in sorted(namespace_tags.items()):
         is_param = bool("param" in namespace.lower())
         node_id = _hash(namespace)
+        if not is_param:
+
+            if KEDRO_VERSION.match(">=0.16.0"):
+                try:
+                    # pylint: disable=protected-access
+                    dataset = _CATALOG._get_dataset(namespace)
+                    _NODES[node_id] = {"type": "data", "obj": dataset}
```
DmitriiDeriabinQB, line 465: This line and line 472 are complete duplicates. I think it can fairly easily be refactored to avoid the repetition.

package/kedro_viz/server.py
```diff
@@ -474,7 +489,39 @@ def format_pipeline_data(
...
     """Serve the pipeline data."""
-    return jsonify(data)
+    return jsonify(_DATA)
+
+
+@app.route("/api/nodes/<string:node_id>")
+def nodes_metadata(node_id):
+    """Serve the metadata for node and dataset."""
+    node = _NODES.get(node_id)
+    if node and node.get("type") == "task":
```
DmitriiDeriabinQB, line 499: Why do we need `node.get("type")`? Aren't `type` guaranteed to be in there?

package/kedro_viz/server.py
DmitriiDeriabinQB: Why is it relative to the _parent_ of the current directory?

DmitriiDeriabinQB: Also, `code_full_path` is an instance of `Path` already, no need to wrap it the second time.

package/kedro_viz/server.py
```diff
@@ -474,7 +489,39 @@ def format_pipeline_data(
...
+        task_metadata = {
+            # pylint: disable=protected-access
+            "code": inspect.getsource(node["obj"]._func)
+        }
+        # pylint: disable=protected-access
+        code_full_path = Path(inspect.getfile(node["obj"]._func)).expanduser().resolve()
+        code_location = Path(code_full_path).relative_to(Path.cwd().parent)
+        task_metadata.update({"code_location": str(code_location)})
```
DmitriiDeriabinQB, line 507:
```suggestion
        task_metadata["code_location"] = str(code_location)
```

package/kedro_viz/server.py
```diff
@@ -474,7 +489,39 @@ def format_pipeline_data(
...
+
+        # pylint: disable=protected-access
+        docstring = inspect.getdoc(node["obj"]._func)
+        if docstring:
+            task_metadata.update({"docstring": docstring})
```
DmitriiDeriabinQB, line 512:
```suggestion
            task_metadata["docstring"] = docstring
```

package/kedro_viz/server.py
```diff
@@ -56,18 +59,17 @@
...
     from kedro.cli.utils import KedroCliError  # pragma: no cover
 
 
 _VIZ_PROCESSES = {}  # type: Dict[int, multiprocessing.Process]
 
 _DEFAULT_KEY = "__default__"
 
-data = None  # pylint: disable=invalid-name
+_DATA = None  # type: Dict
+_CATALOG = None  # type: DataCatalog
```
limdauto, line 71: You can do `_DATA: Optional[Dict] = None` so you don't have to disable unused-import. Also `node` (lowercase) isn't a type, is it?

package/kedro_viz/server.py
```diff
@@ -452,10 +452,25 @@ def format_pipeline_data(
...
+
+            if KEDRO_VERSION.match(">=0.16.0"):
+                try:
+                    # pylint: disable=protected-access
+                    dataset = _CATALOG._get_dataset(namespace)
+                    _NODES[node_id] = {"type": "data", "obj": dataset}
+                except DataSetNotFoundError:
+                    pass
+            else:
+                dataset = _CATALOG._data_sets.get(namespace)
```
DmitriiDeriabinQB, line 469: can you use `_CATALOG.datasets`?

package/kedro_viz/server.py
```diff
@@ -474,7 +489,39 @@ def format_pipeline_data(
...
+
+        # pylint: disable=protected-access
+        docstring = inspect.getdoc(node["obj"]._func)
+        if docstring:
+            task_metadata.update({"docstring": docstring})
+        return jsonify(task_metadata)
+    if node and node.get("type") == "data":
+        dataset = node["obj"]
+        dataset_metadata = {
+            "dataset_type": dataset.__module__,
```
DmitriiDeriabinQB, line 517: I think you need `dataset.__class__.__module__` and `dataset.__class__.__qualname__`

package/kedro_viz/server.py
```diff
@@ -474,7 +489,39 @@ def format_pipeline_data(
...
+    return jsonify(_DATA)
+
+
+@app.route("/api/nodes/<string:node_id>")
+def nodes_metadata(node_id):
+    """Serve the metadata for node and dataset."""
+    node = _NODES.get(node_id)
+    if node and node.get("type") == "task":
+        task_metadata = {
+            # pylint: disable=protected-access
```
limdauto, line 501: let's move this disable to the top of the file so you don't have to repeat it.

package/kedro_viz/server.py
limdauto: For simplicity, can't we just return `code_full_path`? If not, maybe it's worth putting this github comment into the code as a comment.

package/kedro_viz/server.py
```diff
@@ -452,10 +452,25 @@ def format_pipeline_data(
                 edges_list.append(edge)
```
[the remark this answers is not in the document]
921kiyo: I agree with the size of `format_pipeline_data` being too big, I will minimize increasing the size of this function in this PR, but we will need to do the major refactoring in a separate PR I think.

package/kedro_viz/server.py
921kiyo: Yes, I can change it to `node["type"] == "task"`:)

limdauto, line 499: I'd prefer if we return a 404 if the node doesn't exist.

package/tests/test_server.py
```diff
@@ -338,22 +267,148 @@ def test_no_load_file(cli_runner):
...
+    """Test `/api/nodes.json` endpoint is functional and returns a valid JSON."""
     cli_runner.invoke(server.commands, ["viz", "--port", "8000"])
     response = client.get("/api/nodes.json")
     assert response.status_code == 200
     data = json.loads(response.data.decode())
     assert data == EXPECTED_PIPELINE_DATA
 
 
+@pytest.mark.usefixtures("patched_get_project_context")
+def test_node_metadata_endpoint_task(cli_runner, client, mocker, tmp_path):
```
limdauto, line 287: nice test 👍

package/kedro_viz/server.py
921kiyo: This is to address https://github.com/quantumblacklabs/kedro-viz/pull/231#pullrequestreview-461701527. 
For example, 
Code full path would be `/Users/kiyohito_kunii/Desktop/hello/src/hello/pipelines/data_engineering/nodes.py`

and cwd is `/Users/kiyohito_kunii/Desktop/hello/src/hello/` (so the parent of this is `/Users/kiyohito_kunii/Desktop/hello/src/`) 
but we want to only add  `hello/src/hello/pipelines/data_engineering/nodes.py` in JSON,

921kiyo: @limdauto I used to return the full path, but @richardwestenra also had a point about this in https://github.com/quantumblacklabs/kedro-viz/pull/231#pullrequestreview-461701527 

We have to agree on either format: either simple code full path or only showing from the project root (e.g `hello/src/hello/pipelines/data_engineering/nodes.py`).

package/kedro_viz/server.py
921kiyo: https://github.com/quantumblacklabs/kedro-viz/pull/231/files#r466326373 fixed the unused-import thing :)

package/kedro_viz/server.py
921kiyo: Unfortunately, it doesn't exist in Kedro 0.14.0... https://github.com/quantumblacklabs/private-kedro/blob/release/0.14.0/kedro/io/data_catalog.py#L112

DmitriiDeriabinQB: We should drop the support for such old versions

921kiyo: Yeah, we have a ticket for dropping Kedro 0.14 in KED-1696, and will simplify the backend.

package/kedro_viz/server.py
921kiyo: Thanks! I tried different combinations of class modules, and got this result (very interesting!!)

Screenshot 2020-08-06 at 14 57 46


So seems like we need `dataset.__module__` and `dataset.__qualname__`

921kiyo: Never mind, you are right. The above examples are from class, not an instance of the class. 


Screenshot 2020-08-06 at 15 53 05

package/kedro_viz/server.py
```diff
@@ -471,10 +472,74 @@ def format_pipeline_data(
...
-    return jsonify(data)
+    return jsonify(_DATA)
+
+
+@app.route("/api/nodes/<string:node_id>")
+def nodes_metadata(node_id):
+    """Serve the metadata for node and dataset."""
+    node = _NODES.get(node_id)
+    if not node:
+        abort(404, description="Invalid node ID.")
```
921kiyo, line 497: Following this for 404 error https://flask.palletsprojects.com/en/1.1.x/patterns/errorpages/#returning-api-errors-as-json

package/kedro_viz/server.py
```diff
@@ -452,10 +448,15 @@ def format_pipeline_data(
...
             node_dependencies[task_id].add(namespace_id)
-
+    # Parameters and data
     for namespace, tag_names in sorted(namespace_tags.items()):
         is_param = bool("param" in namespace.lower())
         node_id = _hash(namespace)
+        if is_param:
+            _NODES[node_id] = {"type": "parameters", "obj": None}
+        else:
+            _get_dataset_node(node_id, namespace)
```
DmitriiDeriabinQB, line 458: Can we make the API symmetric? It's not a good idea for a "getter" to have side effects.
```python
node_data = {"type": "parameters", "obj": None} if is_param else _get_dataset_node(node_id, namespace)
_NODES[node_id] = node_data
```

package/kedro_viz/server.py
DmitriiDeriabinQB: Don't you need an `@app.errorhandler(404)` then? As `abort` [raises an `HTTPException`](https://flask.palletsprojects.com/en/1.1.x/api/#flask.abort).

package/tests/test_server.py
```diff
@@ -338,22 +261,151 @@ def test_no_load_file(cli_runner):
...
+    assert not data
+
+
+@pytest.mark.usefixtures("patched_get_project_context")
+def test_node_metadata_endpoint_invalid(cli_runner, client):
+    """Test `/api/nodes/invalid_id` endpoint returns an empty JSON."""
+    cli_runner.invoke(server.commands, ["viz", "--port", "8000"])
+    param_id = "invalid"
+    response = client.get(f"/api/nodes/{param_id}")
+    assert response.status_code == 404
```
DmitriiDeriabinQB, line 406: You should also check the body of the response, I suspect it might not be formatted as a JSON

package/kedro_viz/server.py
```diff
@@ -471,10 +472,74 @@ def format_pipeline_data(
...
+    Example:
+        'code_full_path':   'path-to-project/project_root/path-to-code/node.py'
+        'Path.cwd().parent':'path-to-project/'
+        'code_location':    'project_root/path-to-code/node.py''
+
+    """
+    task_metadata = {"code": inspect.getsource(node["obj"]._func)}
+
+    code_full_path = Path(inspect.getfile(node["obj"]._func)).expanduser().resolve()
+    code_location = code_full_path.relative_to(Path.cwd().parent)
```
DmitriiDeriabinQB, line 523: I still don't get why it's a parent of the working dir and not the working dir itself 😄

921kiyo: Let me rephrase this, as the line itself doesn't explain much. 

Does this example help?

Screenshot 2020-08-07 at 11 07 27

and this. Without `parent`, it starts from `src`, but we want do display from `new_api/src/`
Screenshot 2020-08-07 at 11 08 13

DmitriiDeriabinQB: Why do we need to show `new_api`? As to me, `src/pipelines/...` is exactly what we need.

package/kedro_viz/server.py
```diff
@@ -415,11 +409,13 @@ def format_pipeline_data(
...
     namespace_to_layer = {}
 
-    dataset_to_layer = _construct_layer_mapping(catalog)
+    dataset_to_layer = _construct_layer_mapping()
 
+    # Nodes and edges
     for node in sorted(pipeline.nodes, key=lambda n: n.name):
         task_id = _hash(str(node))
         tags.update(node.tags)
+        _NODES[task_id] = {"type": "task", "obj": node}
```
DmitriiDeriabinQB, line 418: We have both `_NODES` and `nodes` variables in this function. Can we rename one of them as this is really confusing? Maybe `_NODES` -> `_ALL_NODES`?

package/kedro_viz/server.py
921kiyo: Okay, I can do that :) If someone (or Yetu) asks us to show from the project root, we can easily add `parent` later.

package/kedro_viz/server.py
921kiyo: Agree. In Viz, `_NODES` means a graph node/edge in JSON, whereas Kedro nodes mean nodes/pipelines. I can change it to `_JSON_NODES` or something.

PR #235, opened 2020-08-10
Improve performance of `getLinkedNodes`

bru5:
## Description

Improves performance of `getLinkedNodes`.

## Development notes

The original implementation appears to be `O(n^2)` or worse, causing high CPU usage on hover and freezing with large numbers of nodes. This replaces it with an `O(n)` implementation that will scale better with more edges.

## QA notes

No changes in behaviour.

## Checklist

## Legal notice

### DIFF
--- a/src/selectors/linked-nodes.js
+++ b/src/selectors/linked-nodes.js
@@ -1,8 +1,6 @@
 import { createSelector } from 'reselect';
+import { getVisibleEdges } from './edges';
 
-const getEdgeIDs = state => state.edge.ids;
-const getEdgeSources = state => state.edge.sources;
-const getEdgeTargets = state => state.edge.targets;
 const getClickedNode = state => state.node.clicked;
 const getHoveredNode = state => state.node.hovered;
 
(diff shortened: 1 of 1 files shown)


src/selectors/linked-nodes.js
```diff
@@ -17,36 +15,79 @@ export const getCentralNode = createSelector(
...
+      targetEdges[edge.source] = targetEdges[edge.source] || [];
+      targetEdges[edge.source].push(edge);
+    }
+
+    return { sourceEdges, targetEdges };
+  }
+);
+
+/**
+ * Finds all visible successors for the given nodeID in given direction
```
richardwestenra, line 42: So is the problem with the old one, that it will revisit ascendants/descendants of nodes that are linked to twice, e.g. in situations like this?
```
2     2
 \   /
   2
 /   \
1     1
 \   /
   A
```
(where A is the original hovered node, and the numbers are nodes with the number referring to how many times each one gets visited)

src/selectors/linked-nodes.js
```diff
@@ -17,36 +15,79 @@ export const getCentralNode = createSelector(
...
+
+    for (const edge of edges) {
+      sourceEdges[edge.target] = sourceEdges[edge.target] || [];
+      sourceEdges[edge.target].push(edge);
+    }
+
+    for (const edge of edges) {
+      targetEdges[edge.source] = targetEdges[edge.source] || [];
+      targetEdges[edge.source].push(edge);
+    }
```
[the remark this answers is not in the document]
bru5: Good idea thanks, I've made this change

src/selectors/linked-nodes.js
bru5: If I understand it correctly, the original version appears to loop over every edge for every edge visited, so it degrades to `O(edges^2)` in the worst case (e.g. highly connected graphs). It can also end up traversing the same edges multiple times (since paths may intersect). Both of these problems should now be avoided with this approach, hopefully that makes sense?

richardwestenra: ~I think only the second case is true~

omg you're right, it does! For some reason I thought it only traversed to linked nodes but you're right, I'd forgotten about the loop. Terrible. I'd been meaning to optimise this but it hadn't occurred to me quite how bad this part was.

src/selectors/linked-nodes.js
```diff
@@ -17,36 +15,83 @@ export const getCentralNode = createSelector(
...
+        direction,
+        visited,
+        false,
+        empty
+      )
+    );
+  }
+
+  return visited;
+};
```
richardwestenra, line 78: I'd prefer to avoid this many arguments, because I find more than three a little confusing to read. 

Personally I'd recommend currying instead and placing most of the variables (e.g. `edgesByNode`, `direction`, `visited`, `empty`) in the parent scope, because they are consistent for each run. e.g. something like

```es6
const findLinkedNodes = (edgesByNode, direction, visited) => {
  let empty; //...
  return (nodeId) => (function(nodeID){ /*...*/ })(nodeID);
};

const findLinkedSourceNodes = findLinkedNodes(sourceEdges, 'source', linkedNodes);
findLinkedSourceNodes(nodeID);
```

Else, I'd use an object so that they're named.

But this is mostly a personal preference thing, so I'm not gonna block this PR because of it.

src/selectors/linked-nodes.js
```diff
@@ -17,36 +15,83 @@ export const getCentralNode = createSelector(
...
+);
+
+/**
+ * Finds all visible successors for the given nodeID in given direction
+ * @param {string} nodeID
+ * @param {Object} edgesByNode
+ * @param {string} direction
+ * @param {?object} visited
+ * @param {?boolean} isStart
+ * @param {?Array} empty
```
richardwestenra, line 52: Can you please add descriptions for each of these arguments?

src/selectors/linked-nodes.js
```diff
@@ -17,36 +15,83 @@ export const getCentralNode = createSelector(
...
+ */
+const findLinkedNodes = (
+  nodeId,
+  edgesByNode,
+  direction,
+  visited = {},
+  isStart = true,
+  empty = []
```
richardwestenra, line 60: What is `empty`? It gets set once on first run, then checked each time, but never modified. Is it even necessary?

src/selectors/linked-nodes.js
bru5: Sure I've added some info here

src/selectors/linked-nodes.js
bru5: Ah yes fair point, originally I was returning and joining arrays but decided to optimise this with a single array to avoid GC thrashing since it's a lot of calls. I just now made another change since you're right that the cached empty array is no longer needed this way.

src/selectors/linked-nodes.js
bru5: I see what you're saying, I've now added comments, removed the empty array and generalised `isStart` to `depth` - hopefully all of that makes it much clearer?

PR #236, opened 2020-08-10
[KED-1887] Sidebar update

bru5:
## Description

Improves the sidebar behaviour, styling and UX.

Screenshot 2020-08-25 at 17 14 51

Screenshot 2020-08-25 at 17 15 08

## Development notes

- Changes colors, spacing, typography, icons in both themes
- Adds hover and active states
- Moves tags into the tree list
- Changes tree list behaviour for tags
- Moves tag-filtered items to the bottom of each section
- Search works across all items
- Removes collapse transitions
- Moves the tags list into the main tree with a new interaction mode

__Todo__

- Secondary level for 'Elements' if necessary
- Refactoring
- Tests

## QA notes

As per development notes.

Interaction with tag tree items differs from other element tree items, it works very similarly to the previous checkbox list.

## Checklist

## Legal notice


src/components/icons/index.js
```diff
@@ -0,0 +1,27 @@
...
+  theme: ThemeIcon,
+  export: ExportIcon,
+  map: MapIcon,
+  plus: PlusIcon,
+  minus: MinusIcon,
+  reset: ResetIcon,
+  indicator: IndicatorIcon,
+  visible: VisibleIcon,
+  invisible: InvisibleIcon
+};
```
richardwestenra, line 27: I would consider using non-default exports instead, so that you could import individual icons,
e.g.
```es6
import { download } from '../icons';
```
Note that I had to rename 'export' to 'download' because `export` is a reserved keyword.

```suggestion
export const label = LabelIcon;
export const layers = LayersIcon;
export const menu = MenuIcon;
export const theme = ThemeIcon;
export const download = ExportIcon;
export const map = MapIcon;
export const plus = PlusIcon;
export const minus = MinusIcon;
export const reset = ResetIcon;
export const indicator = IndicatorIcon;
export const visible = VisibleIcon;
export const invisible = InvisibleIcon;
```

src/components/icons/indicator.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export default ({ className }) => (
+  <svg
+    className={className}
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 24 24">
+    <rect x="8" y="9" width="7.5" height="7.5" />
```
richardwestenra, line 8: image

Overall I think this icon is pretty good, but could use some work to make it more visually consistent with the 👁️  icon. e.g.
- Make it white rather than blue?
- Use icon shape and not just opacity to indicate checked vs unchecked

cc @GabrielComymQB

src/components/icons/node-icon.js
```diff
@@ -17,9 +17,13 @@ export const paths = {
...
-  </svg>
+  <>
+    {paths[type] && (
+      <svg className={className} viewBox="0 0 24 24">
+        {paths[type].map(d => (
+          <path key={d} d={d} />
+        ))}
+      </svg>
+    )}
+  </>
```
richardwestenra, line 28: This might be better:
```es6
export default ({ className, type }) => paths[type] && (
  <svg className={className} viewBox="0 0 24 24">
    {paths[type].map(d => (
      <path key={d} d={d} />
    ))}
  </svg>
);
```

richardwestenra (changes requested):
Overall, looking great. I've not dug deeply into the code yet so this isn't a final review. Sorry about the occasional code nitpicks that snuck into this review

src/components/icons/index.js
```diff
@@ -0,0 +1,27 @@
+import LabelIcon from '../icons/label';
```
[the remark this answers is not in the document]
bru5: This is actually just a refactoring of what already existed inside `IconButton`, I just moved it out into a file so it can be reused and follow the same pattern again in `NodeListRow` for consistency.

If you're still keen to change it though, I guess it should be possible to just pass the icon component directly or as a render prop instead of a string?

src/components/node-list/styles/_search.scss
```diff
@@ -1,8 +1,104 @@
...
+  .kui-theme--dark & .kui-searchbar.kui-searchbar--focused .kui-icon svg {
+    fill: #c8c8c8;
+  }
+
+  .kui-theme--light & .kui-searchbar.kui-searchbar--focused .kui-icon svg {
+    fill: #262626;
+  }
+
+  .kui-input {
+    padding: 8px 0 0 0;
```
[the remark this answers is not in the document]
bru5: I've added some extra padding on the inside, let me know if that was it?

src/components/icons/node-icon.js
bru5: Yeah unfortunately React complains if you don't return something, so I took the suggestion but tweaked it to return an empty node to satisfy it.

src/components/icons/index.js
bru5: Would that still work as a string id mapping object though as that was the reason for this? Maybe?

src/components/icons/indicator.js
bru5: We've made some more changes here so please take a look now it's updated. Not the colour yet though, will need to see what @GabrielComymQB is thinking.

src/components/node-list/styles/_search.scss
```diff
@@ -1,8 +1,109 @@
...
+  box-shadow: 0 14px 18px $color-bg-light-3;
+
+  .kui-theme--dark & {
+    box-shadow: 0 14px 18px $color-bg-dark-3;
+  }
+
+  .pipeline-nodelist-search-inner {
+    margin-bottom: -6px;
+    padding: 2px $sidebar-padding 6px $sidebar-padding;
+  }
```
richardwestenra, line 14: nesting

src/selectors/disabled.js
```diff
@@ -34,18 +33,23 @@ export const getNodeDisabledPipeline = createSelector(
...
+    return arrayToObject(
+      nodeIDs,
+      nodeID =>
+        someEnabled &&
+        (nodeTags[nodeID].length === 0 ||
+          nodeTags[nodeID].every(
+            tag =>
+              typeof tagEnabled[tag] === 'undefined' ||
+              tagEnabled[tag] === false
+          ))
```
richardwestenra, line 50: honestly, this is pretty difficult to read, so I'd prefer to break it up a bit, even if that means adding some if statements like it was before

src/selectors/node-types.js
```diff
@@ -40,3 +41,20 @@ export const getNodeTypes = createSelector(
...
+ * Get formatted list of sections
+ */
+export const getNodeSections = createSelector(
+  [getNodeTypeSections, getNodeTypes],
+  (sections, types) => {
+    const result = [];
+    for (const name of Object.keys(sections)) {
+      result.push({
+        name,
+        types: sections[name].map(id => types.find(type => type.id === id))
```
richardwestenra, line 55: Will this not include some undefined items?

src/store/initial-state.js
```diff
@@ -27,11 +27,16 @@ export const getInitialPipelineState = () => ({
     hovered: null
   },
   nodeType: {
-    ids: ['task', 'data', 'parameters'],
+    ids: ['tag', 'task', 'data', 'parameters'],
```
richardwestenra, line 30: ⚠️⚠️⚠️  'tag' isn't a node-type. I think this might lead to some undesirable outcomes. Is this only used in the sidebar? If so, maybe it should be renamed, but I suspect it might be used elsewhere, in which case I'm amazed that it hasn't caused issues.

richardwestenra (changes requested):

src/components/node-list/styles/_filter.scss
```diff
@@ -0,0 +1,115 @@
+@import '../../../styles/extends';
+@import '../../../styles/mixins';
+@import '../../../styles/variables';
+@import './variables';
+
+.pipeline-nodelist__item--is-filter {
+  .kui-theme--dark & .pipeline-nodelist__row__visibility-icon {
```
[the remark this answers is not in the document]
bru5: I see what you mean but I'd suggest looking at the whole file as that would require a lot of duplication, the modifier would need to be added in a lot of places, which could also break separation of concerns.

The selectors would become much longer too since a lot of them are already modifiers, so you'd still end up with two classes anyway. If it's the nesting you don't like I could manually add the prefix class in everywhere, but I feel that would be very bad for readability.

src/components/node-list/styles/_filter.scss
```diff
@@ -0,0 +1,115 @@
...
+    &--unset {
+      fill: none;
+      stroke: $filter-indicator-off-dark;
+    }
+  }
+
+  // Always show row parent text
+  .pipeline-nodelist__heading
+    > .pipeline-nodelist__row
+    .pipeline-nodelist__row__text {
```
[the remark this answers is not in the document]
bru5: I could expand out some of the nesting selectors, at the cost of making them longer, if that helps?

src/components/node-list/styles/_filter.scss
```diff
@@ -0,0 +1,115 @@
...
+    opacity: 0.55;
+
+    &:hover {
+      opacity: 0.75;
+    }
+
+    // Show unchecked indicators as solid blue when all unset
+    .pipeline-nodelist__row__visibility-icon--checked,
+    .pipeline-nodelist__row__visibility-icon--unchecked,
+    .pipeline-nodelist__row__visibility-icon--unset {
```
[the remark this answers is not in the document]
bru5: A lot of this is to override the base styles, so some parts may appear redundant but the specificity wouldn't be enough as it stands.

src/components/node-list/styles/_group.scss
```diff
@@ -15,27 +15,23 @@
...
 
 .pipeline-nodelist__heading {
   margin: 0;
+
+  .pipeline-nodelist__row__label {
+    font-weight: bold;
+  }
 }
 
 .pipeline-type-group-toggle {
```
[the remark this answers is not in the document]
bru5: I'm not seeing that, can you check if it's still an issue?

src/components/icons/index.js
richardwestenra: I think so, e.g.
```es6
import * as icons from '../icons';

const MapIcon = icons['map'];
```

src/components/icons/node-icon.js
richardwestenra: ok in that case I'd typically use a ternary and return null if `!paths[type]`

src/components/node-list/styles/_group.scss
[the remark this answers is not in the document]
richardwestenra: It is. You need to tab onto it with keyboard only, avoiding the mouse entirely.

src/components/node-list/styles/_filter.scss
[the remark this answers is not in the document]
richardwestenra: It's not the nesting that's the problem: As far as I'm concered, these are equivalent:
```scss
.parent .child {
  ...
}

.parent {
  .child {
    ...
  }
}
```

The problem is the increased CSS specificity, which makes it much harder to maintain if you want to override CSS rules later without escalating specificity. I think it's worth the longer class names - if you want, you can reduce this `pipeline-nodelist__row__visibility-icon` down to something like `pipeline-visibility-icon` if it helps.

src/components/node-list/styles/_filter.scss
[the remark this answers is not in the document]
richardwestenra: If I understand you correctly, this is exactly why I want to reduce the specificity of the base styles

src/components/node-list/styles/_filter.scss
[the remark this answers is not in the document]
richardwestenra: as above. It's about specificity, not readability

src/components/icons/index.js
bru5: Since they are string references, I don't think the bundler would be able to tree shake them though as it will be determined at run-time, though we can instead switch to passing the actual icon components in the prop rather than strings if you like?

src/store/initial-state.js
bru5: Yeah I thought that could be the case which is why I was asking about this before, though no problems showed up strangely. Anyway I was planning to move it to the config instead, which has now been completed.

src/components/icons/node-icon.js
bru5: Sure I've done this, I guess React is fine with `null` but not `undefined`.

src/components/node-list/index.js
```diff
@@ -1,40 +1,247 @@
...
+      invisibleIcon: 'indicatorOff',
+      active: false,
+      selected: false,
+      faded: false,
+      visible: true,
+      disabled: false,
+      unset: typeof tagsEnabled[tag.id] === 'undefined',
+      checked: tagsEnabled[tag.id] === true
+    })),
+    ...Object.keys(filteredNodes).reduce((result, type) => {
```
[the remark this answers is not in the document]
bru5: Yeah exactly, I've spent some time refactoring this into selectors which hopefully should help with this.

src/components/node-list/index.js
```diff
@@ -1,40 +1,247 @@
...
-        autoHide
-        hideTracksWhenNotNeeded>
-        <NodeListToggleAll nodeIDs={nodeIDs} />
-        <NodeListGroups nodes={filteredNodes} />
-      </Scrollbars>
-    </React.Fragment>
+    <NodeList
+      items={items}
+      sections={sections}
+      groups={groups}
```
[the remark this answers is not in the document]
bru5: I know what you mean, I've added a description to this component now which hopefully should help explain how these are being used. I've tried to keep this fairly generic though so that future sections can easily be added to the sidebar.

src/components/node-list/node-list-toggle.js
```diff
@@ -1,36 +0,0 @@
-import React from 'react';
-import { connect } from 'react-redux';
-import { toggleNodesDisabled } from '../../actions/nodes';
-import VisibleIcon from '../icons/visible';
-import InvisibleIcon from '../icons/invisible';
-
-export const NodeListToggleAll = ({ onToggleNodesDisabled, nodeIDs }) => (
```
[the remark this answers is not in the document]
bru5: Yeah I think the intention is that it will come back in another form soon though (e.g. multi select etc.)?

src/selectors/disabled.js
bru5: Yeah it's strange, prettier decided to make it into a staircase which doesn't help. Anyway, I've actually reverted this back to the original method now since it ended up coming full circle in the end.

src/selectors/node-types.js
bru5: Now removed so I'll resolve this one.

src/components/node-list/styles/_group.scss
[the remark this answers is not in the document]
bru5: I can see it now, looks like a Chrome update introduced the problem, you can also see it on the current main branch. I've fixed it here by applying a `font-size` which seems to resolve it.

src/components/node-list/styles/_filter.scss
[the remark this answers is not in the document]
bru5: That works for some of the selectors in this file but not all, e.g.

`.pipeline-nodelist__item--is-filter .pipeline-nodelist__row__visibility-icon--checked`

will become:

`.pipeline-nodelist__row__visibility-icon--is-filter.pipeline-nodelist__row__visibility-icon--checked`

Which I think are the same specificity, so we'd have to add combined versions to get lower:

`.pipeline-nodelist__row__visibility-icon--is-filter--checked`

Is that what you mean?

src/components/node-list/styles/_filter.scss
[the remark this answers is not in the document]
bru5: Looks like this one was fine to just remove, as it was no longer necessary.

src/components/icons/index.js
bru5: All icons are now passed as components instead of strings, seems to work well.

src/components/node-list/styles/_filter.scss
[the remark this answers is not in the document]
bru5: Looks like I was able to simplify this selector as you mentioned actually, thanks.

src/components/flowchart/styles/_layers.scss
```diff
@@ -69,8 +69,16 @@
...
   font-size: 1.6em;
   white-space: nowrap;
+
+  .kui-theme--dark & {
+    color: #b5bdc2;
+  }
+
+  .kui-theme--light & {
+    color: #4e5861;
+  }
```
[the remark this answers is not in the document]
bru5: I've gone through and refactored colours to use opacity instead then, as well as moving some in to variables.

src/styles/_variables.scss
```diff
@@ -1,29 +1,32 @@
...
+$color-bg-dark-alt: $color-light;
 
 //-- Layout --//
 
 $sidebar-padding: 1.8em;
 
 // Also set in src/config.js:
 $sidebar-width-open: 400px;
-$sidebar-width-closed: 60px;
+$sidebar-width-closed: 56px;
```
richardwestenra, line 31: If you change this value in the scss you also need to change it in config.js:
```es6
export const sidebarWidth = {
  open: 400,
  closed: 60
};
```
I tried to add a comment to mention this but I guess it needs to be more explicit. I'll add a suggestion.

src/styles/_variables.scss
```diff
@@ -1,29 +1,32 @@
...
+$color-bg-dark-2: #031a26;
+$color-bg-dark-3: #061c28;
+$color-bg-dark-4: #0e2330;
+$color-bg-dark-alt: $color-light;
 
 //-- Layout --//
 
 $sidebar-padding: 1.8em;
 
 // Also set in src/config.js:
```
richardwestenra, line 29:
```suggestion
// These values are used in both SCSS and JS, and we don't have variable-sharing
// across Sass and JavaScript, so they're defined in two places. If you update their
// value here, please also update their corresponding value in src/config.js
```


would you mind also updating the comment in config.js too please?

richardwestenra (changes requested):

src/components/icons/indicator.js
richardwestenra: Overall I think the new icons look good. Ideally I'd make them a little bigger, again so they're a comparable size/opacity to the 👁️ icon. Some design feedback:

- I really like the show/hide all icon. However I don't like that it often looks the same as the rest of the icons, so it's unclear that it behaves differently:
image
- I think the show/hide icon is a bit small, especially when it's showing hide all - it wasn't clear to me that it was a minus at first:
image
- The icons are 100% opacity by default and have reduced opacity on hover. I'd have thought it should be the opposite, especially to be consistent with our other icons:
image
- I'd like there to be a tooltip on the show/hide icon, to indicate what will happen when you toggle it. This could be part of a separate PR. If you agree, let's add a ticket for it.

src/components/flowchart/styles/_layers.scss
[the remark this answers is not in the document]
richardwestenra: Thanks! I've added a comment about this on `src/components/app/app.scss`, but basically, as much as possible I'd recommend using the `opacity` css rule rather than `fill`/`color`/`stroke` etc, so that these icons can inherit colours via the cascade and just modify their root colour, and reduce the number of places where it might need to be changed, and the amount of overrides.

src/components/node-list/styles/_filter.scss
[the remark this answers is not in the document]
bru5: Sure I've applied this and simplified it a bit further. This does come at a cost of increasing coupling on the JS side, but in this case that seems natural.

src/components/node-list/styles/_row.scss
```diff
@@ -100,32 +107,92 @@
...
+    background: linear-gradient(
+      90deg,
+      rgba(0, 0, 0, 0) 0%,
+      $row-selected-dark 100%
+    );
+  }
+
+  .pipeline-nodelist__row.pipeline-nodelist__row--selected &:before {
+    opacity: 1;
+  }
```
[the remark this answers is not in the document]
bru5: I often use nesting to help keep things 'dry' and for logical grouping of styles, which probably was the reason originally. In any case though I've added more modifier classes so this file is now much flatter anyway.

src/components/app/app.scss
```diff
@@ -3,23 +3,45 @@
...
+
+  .kui-theme--light .pipeline-icon-toolbar__button:hover & {
+    fill: $color-dark;
+  }
+
+  .kui-theme--dark .pipeline-icon-toolbar__button:hover & {
+    fill: $color-light;
+  }
+}
+
```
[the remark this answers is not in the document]
bru5: Yeah good spot this is done, I think this was just the result of various refactoring, that definitely makes sense here and helps simplify.

src/components/app/app.scss
```diff
@@ -3,23 +3,45 @@
...
+.pipeline-icon--stroke .pipeline-icon {
```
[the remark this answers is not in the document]
bru5: Yeah I don't think I named this well originally as part of the minimap PR, so I've refactored this as suggested, thanks.

src/components/node-list/node-list-groups.js
```diff
@@ -1,122 +1,83 @@
...
-      </NodeListGroup>
-    );
-  };
-
-  return (
-    <Flipper flipKey={collapsed}>
-      <ul className="pipeline-nodelist">{types.map(renderTypeGroup)}</ul>
-    </Flipper>
-  );
+  return sections.map(section => (
```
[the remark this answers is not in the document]
bru5: I think it's clearer to read inline like this as it stands, as there's not really any extra logic going on in the map functions, but if it gets more complex down the line we could do it then?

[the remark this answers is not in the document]
richardwestenra: well okay then I guess so

richardwestenra (approved):
Approving the code changes, but pending resolution on the discussion around tag icon styling. But I'm happy for that to be updated in a subsequent PR, too

GabrielComymQB:
Hey guys, this is awesome work!

We have to have in mind that Tags behave differently from all the other elements (Nodes, Datasets, Parameters). Tags are a single cell that goes on or off. There’s no multiple hit areas like common elements divided into cell and icon. That’s the mindset. Knowing that, here we go.

When we first land on Viz. All the tags are ON. However, I don’t see the led “markers” on - as a hint. And that’s what is confusing for me from the beginning cause the creates a disconnection between label and “marker” - and remember, on Tags there’s no such disconnection. It’s like one big cell going on and off. That’s also why the blue “markers” are very delicate and gentle to the eyes, differently from all the other interface icons we have around. The marker is not doing the indicative function all alone, but also with the support of the label changing in colour when it goes off.

My suggestion is uploaded to Zeplin.

https://app.zeplin.io/project/5d5ea7a05efca76a74f7d0ea/dashboard?seid=5f75f5181cdc948272aacd15

I’m thinking we should show the “hint” 
I’ve uploaded 2 other boards. A table showing the different states and also 6 different examples of the different states.

To Richard’s comments:

- On Looking the same as the rest of the icons.
Not sure if I understand precisely what you mean, however the confusion might be there as I tried to explain above. We should have the mindset that this is not an icon like we have for the other elements. This is a “led” that goes on and off. In any case, this exploration was already validated during all our user testing series, remember? Let’s go with that cause we’ve observed no one had any problems in understanding it :)

- On Icons being small.
I think is great as it is and the intention was exactly to make it delicate and not chunky as the icons, also because it’s not an icon as I mentioned and it has the support of the different states of the labels and the cell as a whole. However, I agree and I’ve uploaded a new version of the “minus” marker which is wider than the current one and it brings more balance/volume to it when compared to others. This might help.

- On Icons having 100% opacity by default
I do agree with Richard that it should be the opposite. Full opacity on hover. I also agree that we have to make use of a tooltip on timed hover to indicate the action you are about to take.


Hope that clarifies. I'm happy to jump on a call if we still have details to discuss. I know these tiny details are sometimes very difficult to discuss by text.

Cheers!

richardwestenra:
Thanks @GabrielComymQB! Repeating my replies from slack here (let's keep the discussion on Github).

> On Looking the same as the rest of the icons.
> Not sure if I understand precisely what you mean

What I meant is that the icon for the button to toggle all tags on/off looks the exact same as the ones to toggle individual tags in some contexts but not others - it looks different when it’s a minus, but otherwise it’s the same. I thought there should be some other way to distinguish that it behaves differently, because it looks identical which I found confusing as it made it less clear what it would do when i clicked it

> On Icons being small

We should also pay attention to tap area. I need to doublecheck this but I suspect for these icons the entire row is the tap area so it's fine, but if we think there's any chance of the row being used for something else then it might be worth increasing the icon size

src/components/app/app.scss
```diff
@@ -3,23 +3,30 @@
...
+    opacity: 1;
   }
+}
 
-  .pipeline-icon--stroke & {
-    .kui-theme--light & {
-      fill: none;
-      stroke: #333;
-    }
+.pipeline-icon.pipeline-icon--stroke {
```
richardwestenra, line 22: Could this not just be `.pipeline-icon--stroke {`? Why does it need the additional class? The whole point is to avoid needing to compound the classes in order to avoid the increased specificity

src/components/app/app.scss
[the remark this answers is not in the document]
richardwestenra: I see that it's now 
```css
.pipeline-icon.pipeline-icon--stroke {
```
Could this not just be a single class?
```css
.pipeline-icon--stroke {
```
Why does it need the additional class? As mentioned previously, the whole point of using BEM is to avoid needing to compound the classes in order to avoid the increased specificity. It's not clear from reading it that the additional specificity is necessary.

src/components/icons/indicator.js
bru5: I've updated the partial icon to be thinner and wider as per @GabrielComymQB's suggestion, I've also made the indicators brighter on hover by making them dimmer in their on state.

src/components/app/app.scss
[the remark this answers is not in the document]
bru5: This was needed as the default selector had higher specificity due to properties being set through theme classes. I've instead duplicated the `fill` property here to both theme classes instead so this can work as suggested.

src/components/node-list/styles/_filter.scss
```diff
@@ -0,0 +1,95 @@
...
+  &.pipeline-nodelist__row__visibility-icon--unset {
+    opacity: 0.55;
+  }
+
+  // Bright indicator when checked or hovered
+  &.pipeline-nodelist__row__visibility-icon--checked,
+  &.pipeline-nodelist__row__visibility-icon--unchecked:hover,
+  &.pipeline-nodelist__row__visibility-icon--unset:hover {
+    opacity: 1;
+  }
```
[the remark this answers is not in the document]
bru5: Agreed yeah I've renamed these classes to something shorter now.

src/components/node-list/node-list-groups.js
```diff
@@ -1,122 +1,84 @@
...
+              childCount={group.count}
+              allUnset={group.allUnset}
+              visibleIcon={group.visibleIcon}
+              invisibleIcon={group.invisibleIcon}
+              collapsed={Boolean(searchValue) ? false : collapsed[group.id]}
+              onToggleCollapsed={onToggleGroupCollapsed}
+              onToggleChecked={onToggleGroupChecked}>
+              <ul className="pipeline-nodelist pipeline-nodelist--nested">
+                {(items[group.id] || []).map(item => (
+                  <li key={item.id}>
```
richardwestenra, line 52: I noticed that this results in the following HTML structure:
image

Why not change the parent element in `NodeListRow` from a `<div>` to a `<li>`, and get rid of this parent element, to avoid having an unnecessary container dom node?

src/components/node-list/node-list-groups.js
```diff
@@ -1,122 +1,84 @@
...
-    );
-  };
-
-  return (
-    <Flipper flipKey={collapsed}>
-      <ul className="pipeline-nodelist">{types.map(renderTypeGroup)}</ul>
-    </Flipper>
-  );
+  return sections.map(section => (
+    <nav className="pipeline-nodelist-section kedro" key={section.name}>
```
richardwestenra, line 31: Why are the sections `<nav>` elements? You've got multiple navs, which feels odd to me.
image

I'd be more inclined to have something like this:
image

richardwestenra: Happy to entertain discussion about what should or should not be a `<nav>`, but these definitely feel like `<section>`s to me.

src/components/node-list/node-list-groups.js
```diff
@@ -1,122 +1,84 @@
...
+              kind={group.kind}
+              checked={group.checked}
+              childCount={group.count}
+              allUnset={group.allUnset}
+              visibleIcon={group.visibleIcon}
+              invisibleIcon={group.invisibleIcon}
+              collapsed={Boolean(searchValue) ? false : collapsed[group.id]}
+              onToggleCollapsed={onToggleGroupCollapsed}
+              onToggleChecked={onToggleGroupChecked}>
+              <ul className="pipeline-nodelist pipeline-nodelist--nested">
```
richardwestenra, line 50: This element seems like it could be merged with its parent `pipeline-nodelist__children` div, so that it's just a single `<ul>`:
image

I'd probably get rid of the `pipeline-nodelist pipeline-nodelist--nested` classes and just apply that styling to the `pipeline-nodelist__children` class, which currently has no styling.

src/components/node-list/node-list-group.js
```diff
@@ -1,74 +1,63 @@
...
 
-export const mapDispatchToProps = dispatch => ({
-  onToggleTypeDisabled: (typeID, disabled) => {
-    dispatch(toggleTypeDisabled(typeID, disabled));
-  }
-});
+    <div
+      className={classnames({
+        'pipeline-nodelist__children': true,
+        'pipeline-nodelist__children--open': !collapsed,
```
richardwestenra, line 55: This class has no styling attached. Remove?

src/components/node-list/node-list-group.js
```diff
@@ -1,74 +1,63 @@
...
+    </h3>
 
-export const mapDispatchToProps = dispatch => ({
-  onToggleTypeDisabled: (typeID, disabled) => {
-    dispatch(toggleTypeDisabled(typeID, disabled));
-  }
-});
+    <div
+      className={classnames({
+        'pipeline-nodelist__children': true,
```
richardwestenra, line 54: Technically you can do this like this, though it's not a big deal either way:
```js
classnames('pipeline-nodelist__children', {
```

richardwestenra (changes requested):
Noticed a few HTML issues, sorry fam

src/components/node-list/node-list-groups.js
bru5: I'm not sure that's a good idea personally: creating components that use `<li>`as their root limits them to only being used inside list elements. I've already noticed the same problem with the `IconButton` component, which has no real reason to be a list element. As this is already pre-existing in the codebase too, my preference is to avoid tightly coupling components to their parent's markup.

src/components/node-list/node-list-groups.js
bru5: Do you mean they're not `<nav>` since they're not links? As far as I know multiple `<nav>` are normal in principle, though you could argue that the whole sidebar is just one `<nav>`. The `<section>` element is usually intended for blocks of content, like sections of an article or a book. I didn't choose it here as navigation isn't usually classed as content, so I wouldn't say this change is needed.

src/components/node-list/node-list-groups.js
bru5: Yeah I see what you mean, I think this one is also pre-existing too. I do still think that tying separate components to a markup structure is generally not a great idea since it limits re-use, as mentioned in my other comment.

Actually, I think this already applies here: the reason this might have been done originally is that `NodeListRow` is used for both for the list items and the group heading which is placed inside a `<h3>`?

src/components/node-list/node-list-groups.js
richardwestenra: It was more just that they didn't seem to me like separate nav panels, they're just different sections in the same nav panel. Overall not a big issue, I just felt like `<nav><section/><section/></nav>` made more sense than `<div><nav/><nav/></div>`. I believe you can use a section in a nav, but `<nav><nav/><nav/></nav>` might make more sense, idk?

src/components/node-list/node-list-groups.js
richardwestenra: As above, I'd use a fragment instead of a div if possible? Can the classes be moved from the div to the ul?

src/components/node-list/node-list-groups.js
bru5: Fragments won't work here as classnames are needed on the container unfortunately. I'm with you here though that this is a good idea to reduce overhead since there can be a lot of rows.

So Iistead I've added a prop `container` that allows optionally passing the tag name to `NodeListRow` and something similar for the children container of `NodeListGroup` (which is needed for the open/close functionality).

This approach then allowed those extra containers to be removed, while still maintaining separation of concerns and the flexibility of these components as is currently needed.

src/components/node-list/node-list-groups.js
bru5: I think it works fine as it is personally, but if you like we could change the root element to `nav` and also the `NodeListGroup` too, then it would be `nav` all the way down 🐢

src/components/node-list/node-list-groups.js
bru5: Yeah again fragments won't work here as classnames are needed on the container unfortunately, but I was able to apply the `container` prop here and removed the extra container.

richardwestenra (approved):

PR #239, opened 2020-08-12
Rename default json endpoint

MerelTheisenQB:
## Description

https://jira.quantumblack.com/browse/KED-1969

## Development notes

I renamed the endpoint from `/api/nodes.json` to `/api/main`. 

## QA notes

Updated tests and ran kedro-viz locally to see if everything still works. 

## Checklist

## Legal notice

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,11 +85,11 @@ Kedro-Viz uses an unique identifier to determine the data source. You can config
 DATA=random npm start
 ```
 
-There are several different data sources available. By default in production, the app asynchronously loads JSON from the `/api/nodes.json` endpoint, and you can replicate this in development by placing a dataset at `/public/api/nodes.json`. Alternatively, you can use one of the mock unit-testing/demo datasets, or pseudo-random data procedurally-generated on page load, which is often useful for local development.
+There are several different data sources available. By default in production, the app asynchronously loads JSON from the `/api/main` endpoint, and you can replicate this in development by placing a dataset at `/public/api/main`. Note that `main` is the name of the json file, but should not have an extension. Alternatively, you can use one of the mock unit-testing/demo datasets, or pseudo-random data procedurally-generated on page load, which is often useful for local development.
 
 | Identifier | Data source |
 |------------|-------------|
-| `json` (default) | `/public/api/nodes.json` |
+| `json` (default) | `/public/api/main` |
 | `random` | Randomly-generated data |
 | `demo` | `/src/utils/data/demo.mock.js` |
 | `animals` | `/src/utils/data/animals.mock.js` |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,7 @@ Kedro-Viz uses an unique identifier to determine the data source. You can config
 DATA=random npm start
 ```
 
-There are several different data sources available. By default in production, the app asynchronously loads JSON from the `/api/main` endpoint, and you can replicate this in development by placing a dataset at `/public/api/main`. Note that `main` is the name of the json file, but should not have an extension. Alternatively, you can use one of the mock unit-testing/demo datasets, or pseudo-random data procedurally-generated on page load, which is often useful for local development.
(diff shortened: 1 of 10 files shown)


RELEASE.md
```diff
@@ -20,6 +20,7 @@ Please follow the established format:
 - Move data source loading into standalone-app entry point (#215)
 - Allow an argument to be passed to loadJsonData (#215)
 - Add information for multiple pipelines in `nodes.json` (#192)
+- Rename default endpoint from `/api/nodes.json` to `/api/nodes` (#239)
```
MerelTheisenQB, line 23: Not entirely sure if this is major or minor change 🤔 

It also looks like @921kiyo 's changes in https://github.com/quantumblacklabs/kedro-viz/pull/231 aren't added here?

richardwestenra:
@MerelTheisenQB My argument against calling it nodes is that:
1. It’s not just nodes: It’s also edges, tags, layers, pipelines, etc.
2. It’s not all nodes: It’s only the nodes for the first/main/default/initial pipeline that you see when you first load the page.
3. It’s not all the data for a given node: Just the basic/essential data

We're not really separating things in a classic RESTful way, it's more separating them into first-load vs subsequent load. So I think the names that most make sense for this file are things like main/default/initial/root

RELEASE.md
richardwestenra: I'm inclined to say major but not necessarily breaking (though it's skirting that line as people will need to rename their test json files)

richardwestenra:
oh also, can you please update the documentation for this feature? https://github.com/quantumblacklabs/kedro-viz/blob/develop/CONTRIBUTING.md#data-sources

MerelTheisenQB:
> oh also, can you please update the documentation for this feature? https://github.com/quantumblacklabs/kedro-viz/blob/develop/CONTRIBUTING.md#data-sources
> 
> You might need to make a note that the JSON file will need to have its extension removed for it to work.

So if I understand it correctly, a user will now need to add a JSON file named `main` in `/public/api` ?

richardwestenra:
@MerelTheisenQB Correct. Your update looks good to me, thanks!

Might be worth getting @stichbury to double-check it?

CONTRIBUTING.md
```diff
@@ -85,11 +85,11 @@ Kedro-Viz uses an unique identifier to determine the data source. You can config
 DATA=random npm start
 ```
 
-There are several different data sources available. By default in production, the app asynchronously loads JSON from the `/api/nodes.json` endpoint, and you can replicate this in development by placing a dataset at `/public/api/nodes.json`. Alternatively, you can use one of the mock unit-testing/demo datasets, or pseudo-random data procedurally-generated on page load, which is often useful for local development.
+There are several different data sources available. By default in production, the app asynchronously loads JSON from the `/api/main` endpoint, and you can replicate this in development by placing a dataset at `/public/api/main`. Note that `main` is the name of the json file, but should not have an extension. Alternatively, you can use one of the mock unit-testing/demo datasets, or pseudo-random data procedurally-generated on page load, which is often useful for local development.
```
yetudada, line 88:
```suggestion
There are several different data sources available. By default in production, the app asynchronously loads JSON from the `/api/main` endpoint. You can replicate this in development by placing a dataset in `/public/api/main`, using `main` as the name of the JSON file, without an extension. Alternatively, you can use one of the mock unit-testing/demo datasets, or pseudo-random data procedurally-generated on page load, which is often useful for local development.
```

RELEASE.md
921kiyo: Hi, I'm back! 
I didn't add the change https://github.com/quantumblacklabs/kedro-viz/pull/231 in release note because it's a back-end change. If we expect new API endpoints to be used by users (as a React component I guess?), then it makes sense to add it in the release note? Otherwise probably not necessary.

MerelTheisenQB: ah okay! I wasn't sure if release notes were only for users or also for ourselves to keep track of the changes we've made 🙂

richardwestenra: @921kiyo I don't expect the extra endpoints to be used in the React component library, as users would need to provide their own endpoints. Instead, I think that if they want to provide this data, they can include it in the main data prop. Haven't discussed this much but idk how it could work otherwise.

@MerelTheisenQB A bit of both, but I see the release notes as mainly for users.

PR #252, opened 2020-08-26
[KED-1422] Expose an endpoint to query pipeline-specific node

limdauto:
## Description

This PR exposes an endpoint for the FE to query pipeline-specific nodes information. At first, I tried refactoring the backend to make it more robust (i.e. calculating the response from Kedro-native objects), but went down a rabbit hole, so I opt-ed for this implementation instead (filtering the global `_DATA` object). 

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -369,8 +369,9 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
     sorted_layers = _sort_layers(nodes, node_dependencies)
 
     default_pipeline = {"id": _DEFAULT_KEY, "name": _pretty_name(_DEFAULT_KEY)}
-    selected_pipeline = default_pipeline if default_pipeline in pipelines_list \
-        else pipelines_list[0]
+    selected_pipeline = (
+        default_pipeline if default_pipeline in pipelines_list else pipelines_list[0]
+    )
 
     return {
         "nodes": nodes_list,
@@ -378,7 +379,7 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
         "tags": sorted_tags,
         "layers": sorted_layers,
         "pipelines": pipelines_list,
-        "selected_pipeline": selected_pipeline
+        "selected_pipeline": selected_pipeline,
     }
 
 
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -44,7 +44,7 @@ from typing import Dict, List, Set, Tuple, Union
 import click
 import kedro
 import requests
-from flask import Flask, jsonify, send_from_directory, abort
+from flask import Flask, abort, jsonify, send_from_directory
 from IPython.core.display import HTML, display
 from kedro.io import AbstractDataSet, DataCatalog, DataSetNotFoundError
 from kedro.pipeline.node import Node
@@ -515,13 +515,15 @@ def pipeline_data(pipeline_id):
         if edge["source"] in pipeline_node_ids and edge["target"] in pipeline_node_ids:
             pipeline_edges.append(edge)
 
-    return jsonify({
-        "nodes": pipeline_nodes,
-        "edges": pipeline_edges,
-        "tags": _DATA["tags"],
-        "layers": _DATA["layers"],
-        "pipelines": pipelines_list,
-    })
+    return jsonify(
+        {
+            "nodes": pipeline_nodes,
+            "edges": pipeline_edges,
+            "tags": _DATA["tags"],
+            "layers": _DATA["layers"],
+            "pipelines": pipelines_list,
+        }
+    )
 
 
 @app.route("/api/nodes/<string:node_id>")
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -488,8 +488,40 @@ def _get_dataset_node(node_id, namespace):
 
 @app.route("/api/main")
 def nodes_json():
-    """Serve the pipeline data. This includes basic node data amongst others edges, tags, and layers."""
+    """Serve the data from all Kedro pipelines in the project.
+    This includes basic node data amongst others edges, tags, and layers.
+    """
     return jsonify(_DATA)
+
+
+@app.route("/api/pipelines/<string:pipeline_id>")
+def pipeline_data(pipeline_id):
+    """Serve the data from a single pipeline in a Kedro project."""
+    current_pipeline = {"id": pipeline_id, "name": _pretty_name(pipeline_id)}
+    pipelines_list = _DATA["pipelines"]
+    pipelines_list.remove(current_pipeline)
+    pipelines_list.insert(0, current_pipeline)
+
+    pipeline_node_ids = set()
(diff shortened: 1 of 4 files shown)


package/kedro_viz/server.py
```diff
@@ -488,10 +488,42 @@ def _get_dataset_node(node_id, namespace):
...
     return jsonify(_DATA)
 
 
+@app.route("/api/pipelines/<string:pipeline_id>")
+def pipeline_data(pipeline_id):
+    """Serve the data from a single pipeline in a Kedro project."""
+    current_pipeline = {"id": pipeline_id, "name": _pretty_name(pipeline_id)}
+    pipelines_list = _DATA["pipelines"]
+    pipelines_list.remove(current_pipeline)
+    pipelines_list.insert(0, current_pipeline)
```
921kiyo, line 503: Why do we need to move the pipeline in the first elemennt in the list?

package/kedro_viz/server.py
```diff
@@ -488,10 +488,42 @@ def _get_dataset_node(node_id, namespace):
...
 def nodes_json():
-    """Serve the pipeline data. This includes basic node data amongst others edges, tags, and layers."""
+    """Serve the data from all Kedro pipelines in the project.
+    This includes basic node data amongst others edges, tags, and layers.
+    """
     return jsonify(_DATA)
 
 
+@app.route("/api/pipelines/<string:pipeline_id>")
+def pipeline_data(pipeline_id):
```
921kiyo, line 498: Should we also return 404 Json for invalid pipeline ID, similar to how it's done for `api/nodes/id`?

package/tests/test_server.py
```diff
@@ -277,6 +277,36 @@ def test_nodes_endpoint(cli_runner, client):
     assert data == EXPECTED_PIPELINE_DATA
```
921kiyo, line 279: And test for invalid case as well?

921kiyo (approved):
LGTM 👍

package/kedro_viz/server.py
limdauto: 👍 very good point

package/kedro_viz/server.py
limdauto: It's in the ticket. The FE takes the first element as the selected pipeline.

package/kedro_viz/server.py
```diff
@@ -488,10 +488,42 @@ def _get_dataset_node(node_id, namespace):
...
+    pipeline_nodes = []
+
+    for node in _DATA["nodes"]:
+        if pipeline_id in node["pipelines"]:
+            pipeline_node_ids.add(node["id"])
+            pipeline_nodes.append(node)
+
+    pipeline_edges = []
+    for edge in _DATA["edges"]:
+        if edge["source"] in pipeline_node_ids and edge["target"] in pipeline_node_ids:
```
DmitriiDeriabinQB, line 515:
```suggestion
        if {edge["source"], edge["target"]} <= pipeline_node_ids:
```

package/tests/test_server.py
```diff
@@ -277,6 +277,36 @@ def test_nodes_endpoint(cli_runner, client):
...
 
 
+@pytest.mark.usefixtures("patched_get_project_context")
+def test_pipelines_endpoint(cli_runner, client):
+    """Test `/api/main` endpoint is functional and returns a valid JSON."""
```
DmitriiDeriabinQB, line 282: I guess it's a leftover from the copy-paste 🙂

package/tests/test_server.py
```diff
@@ -277,6 +277,36 @@ def test_nodes_endpoint(cli_runner, client):
...
+    cli_runner.invoke(server.commands, ["viz", "--port", "8000"])
+    selected_pipeline_id = "third"
+    response = client.get(f"/api/pipelines/{selected_pipeline_id}")
+    assert response.status_code == 200
+    data = json.loads(response.data.decode())
+
+    # make sure the selected pipeline is the first on the list
+    # and the list of all pipelines are returned
+    assert data["pipelines"][0]['id'] == selected_pipeline_id
```
DmitriiDeriabinQB, line 291: It needs to be blacked

package/tests/test_server.py
```diff
@@ -277,6 +277,36 @@ def test_nodes_endpoint(cli_runner, client):
...
+    assert {p["name"] for p in data["pipelines"]} == \
```
DmitriiDeriabinQB, line 292: And this too 🙂

DmitriiDeriabinQB (commented):
I agree that the backend looks rather messy now: 
* a number of global objects with unclear dependency
* functions like `format_pipelines_data` and `format_pipeline_data` and the latter having some side effects and not returning anything
* `_DATA` containing the _list_ of pipelines, the order of which is mutated in 2 different places

I think we are in need of a major Viz backend refactoring task/sprint...

DmitriiDeriabinQB (approved):

921kiyo:
> I agree that the backend looks rather messy now:
> 
> * a number of global objects with unclear dependency
> * functions like `format_pipelines_data` and `format_pipeline_data` and the latter having some side effects and not returning anything
> * `_DATA` containing the _list_ of pipelines, the order of which is mutated in 2 different places
> 
> I think we are in need of a major Viz backend refactoring task/sprint...

Totally agree with refactoring!

limdauto:
> I think we are in need of a major Viz backend refactoring task/sprint

@DmitriiDeriabinQB  Yep, I will try to organise it off the back of the meeting next week.

package/kedro_viz/server.py
DmitriiDeriabinQB: Can't the FE store the selected pipeline in the local storage? Relying on the order in thee list is very hacky to me.

921kiyo: Actually we do this in https://github.com/quantumblacklabs/kedro-viz/blob/main/package/kedro_viz/server.py#L374-L375 as well .. 
@limdauto I remember that you suggested adding a new entry `selected_json` the JSON a while ago instead of using the first index in the list. Maybe something we could implement as a follow-up PR?

limdauto: I did propose adding a new field called `selected_pipeline`. Can't remember why it was rejected. Let me check with Richard again.

921kiyo: I don't think it was rejected. I remember he was happy with it?

richardwestenra: 1. Yeah I'm happy either way - you can either use the first pipeline in the list to indicate the default one (i.e. the one that should display when you start the server), or you can set an extra prop with that value. I just figured that using the first one is easier and avoids having an extra prop, but maybe it's good to be explicit about it. Don't mind.
2. We actually are already using localStorage to save the value of the last pipeline the user had open and reuse it where possible. However we need to know which pipeline to display in cases where the user does not already have anything stored in localStorage, or in cases where the stored value does not match any of the existing pipelines. By default the front-end will just serve up the first pipeline in the list - see point 1.

limdauto: 👍 thanks @richardwestenra I will add another prop then.

921kiyo (approved):
LGTM :)

PR #254, opened 2020-08-28
[KED-2034] Put multi-pipeline selector behind a flag

richardwestenra:
## Description

Put multi-pipeline selector behind a flag, to make it safe to release even while it's still a work-in-progress.

Now, even if you have multiple pipelines available, the selector UI will not be displayed unless you enable the flag.

image

## Checklist

## Legal notice


bru5:
If you try enabling the flag, choosing a pipeline other than default, then disable the flag I think I'd expect it to reset back to the default pipeline. It doesn't seem to do that though, so there might be something missing here from the implementation?

bru5 (changes requested):
If you try enabling the flag, choosing a pipeline other than default, then disable the flag I think I'd expect it to reset back to the default pipeline. It doesn't seem to do that though, so there might be something missing here from the implementation?

bru5 (approved):

PR #268, opened 2020-09-15
[KED-1857] Refactor and optimise flowchart performance

bru5:
## Description

Improves performance of the flowchart rendering to reduce jank on hover, while also improving the clarity of the drawing code.

Improves label transitions and exit transitions for cases where opacity is applied by CSS (e.g. faded out).

## Development notes

The flowchart component now passes the `changed` function to the draw functions so they can be much more granular. This minimises the amount of work done during a change, especially for hover interactions. In particular, previously edge path curves were calculated on every update, which appeared to significantly hit performance.

## QA notes

Behaviour should be identical, flowchart functionality should be tested through by toggling elements in various ways.

Performance improvements are most pronounced on the sidebar when hovering and moving the cursor on large sized graphs that have many edges. This is also more noticeable when applied to the coming #236 sidebar updates branch.

## Checklist

## Legal notice

### DIFF
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -137,7 +137,7 @@ export const drawNodes = function(changed) {
     exitNodes
       .transition('exit-nodes')
       .duration(this.DURATION)
-      .attr('opacity', 0)
+      .style('opacity', 0)
       .remove();
 
     this.el.nodes = this.el.nodeGroup.selectAll('.pipeline-node');
@@ -226,7 +226,7 @@ export const drawEdges = function(changed) {
     exitEdges
       .transition('exit-edges')
       .duration(this.DURATION)
-      .attr('opacity', 0)
+      .style('opacity', 0)
       .remove();
 
     allEdges
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -4,6 +4,11 @@ import { select } from 'd3-selection';
 import { curveBasis, line } from 'd3-shape';
 import icon from './icon';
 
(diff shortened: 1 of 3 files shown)


src/components/flowchart/draw.js
```diff
@@ -83,141 +88,165 @@ export const drawNodes = function() {
...
+      .data(nodes, node => node.id);
+  }
+
+  if (!this.el.nodes) {
+    return;
+  }
+
+  const enterNodes = this.el.nodes.enter().append('g');
+  const exitNodes = this.el.nodes.exit();
+  const allNodes = this.el.nodes.merge(enterNodes).merge(exitNodes);
```
richardwestenra, line 103: Does allNodes need to include exitNodes?

richardwestenra (approved):
👍

src/components/flowchart/draw.js
bru5: Yeah so I tried to find out if they're already included but I guessed they might not be since the enter nodes are not, so I just added that to be explicit about that either way for consistency. It's pretty subtle too but I think it's worth updating those nodes just before they exit to be sure we avoid visual glitches.

richardwestenra: Yeah that's what I was wondering about. Guess there's the potential for minor visual glitches, but balanced against the potential for marginal performance hits from updating those nodes even as they're fading out anyway. Guess it probably doesn't matter much either way.

Anyway, nice work, thanks for this!

richardwestenra (approved):

PR #269, opened 2020-09-21
[KED-1857] Fix interrupted chart transitions

bru5:
## Description

Fixes chart transitions when they are interrupted mid-transition.

## Development notes

Adds extra routines to chart drawing to cancel exit transitions when a node re-enters before its in-progress exit transition completes.
This also fixes a problem introduced by #268 where the above may cause invisible nodes.

Also changes the node tab order sorting routine to avoid a rare error sometimes seen during rapid transition changes.

## QA notes

Behaviour is the same.

To reproduce the issue try changing the selected tags rapidly in random orders, in order to cause overlapping exit and enter transitions on the chart. The issue is seen on the current main branch, this branch should fix that.

## Checklist

## Legal notice

### DIFF
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -76,6 +76,17 @@ export const drawLayerNames = function() {
 };
 
 /**
+ * Sets the size and position of the given node rects
+ */
+const updateNodeRects = nodeRects =>
+  nodeRects
+    .attr('width', node => node.width - 5)
+    .attr('height', node => node.height - 5)
+    .attr('x', node => (node.width - 5) / -2)
+    .attr('y', node => (node.height - 5) / -2)
+    .attr('rx', node => (node.type === 'task' ? 0 : node.height / 2));
+
+/**
  * Render node icons and name labels
  */
 export const drawNodes = function(changed) {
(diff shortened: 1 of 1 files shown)


src/components/flowchart/draw.js
```diff
@@ -98,6 +109,7 @@ export const drawNodes = function(changed) {
     return;
   }
 
+  const updateNodes = this.el.nodes;
```
richardwestenra, line 112: Why introduce this variable instead of just using `this.el.nodes`? Just for clarity/consistency/explicitness?

src/components/flowchart/draw.js
```diff
@@ -164,22 +179,18 @@ export const drawNodes = function(changed) {
       .transition('update-nodes')
       .duration(this.DURATION)
       .attr('transform', node => `translate(${node.x}, ${node.y})`)
-      .end()
-      .catch(() => {})
-      .finally(() => {
+      .on('end', () => {
```
richardwestenra, line 182: Didn't realise you could do this, nice nice 👍

richardwestenra (approved):
👍  :shipit:

src/components/flowchart/draw.js
bru5: Yeah exactly this is for all of those things as it's not obvious unless you know d3 well.

PR #270, opened 2020-09-23
[KED-1979] Retrieve modular pipeline data asynchronously from different endpoints

richardwestenra:
## Description

After fetching data from api/main, check the list of pipelines, then load multi-pipeline data (when required) from api/pipelines/[id] and add it to state.

## Development notes

This should only occur when the data source is 'json'. If using other sources then it should just load all the data on first run, then apply filtering as necessary

## QA notes

Check all of the following use cases:

- [ ] Data sources: `json` (prod async load) and `demo`/`animals` (test sync load)
- [ ] With another pipeline set in localStorage, vs clean first load (in incognito mode or using `window.localStorage.clear()`
- [ ] First page load vs changing the pipeline via the dropdown
- [ ] Using a dataset with support for pipelines vs an older legacy one which doesn't include them
- [ ] When pipelines flag is set to false
- [ ] When the requested endpoint is the 'main' one, or not.

## Checklist

## Legal notice


src/actions/pipelines.js
```diff
@@ -0,0 +1,129 @@
...
+    return getUrl('main');
+  }
+  return getUrl('pipeline', pipeline.active);
+};
+
+/**
+ * Check whether to make another async data request
+ * @param {object} pipeline Pipeline state
+ */
+export const requiresSecondRequest = (flags, pipeline) => {
```
bru5, line 65: Is there a more specific name possible here? Something like `shouldRequestXXX`. If there are multiple things in the second request, you could add multiple functions with this naming style to make it clear what's needed per-case, even if they all share this same base function.

src/components/icons/loading.scss
```diff
@@ -18,12 +18,10 @@ $path2-gap: $section-length + $section-gap-2x;
...
   width: 30px;
   height: auto;
-  opacity: 0;
-  transition: all ease 0.3s;
-  pointer-events: none;
+  visibility: hidden;
 
   &--visible {
-    opacity: 1;
+    visibility: visible;
```
bru5, line 24: This may trigger jank on chart animations because visibility can trigger layout, paint and composite where opacity is only paint, see https://csstriggers.com/visibility vs https://csstriggers.com/opacity.

I think it's best to revert this if possible?

src/config.js
```diff
@@ -1,5 +1,20 @@
-export const dataPath = './api/main';
-export const fullDataPath = `/public${dataPath.substr(1)}`;
+/**
+ * Determine where to load data from
+ * @param {string} type Data type
+ * @param {string=} id
+ */
+export const getUrl = (type, id) => {
```
bru5, line 6: We should probably avoid functions in config, ideally break out the constants here and put the function elsewhere I think, especially since it isn't a pure function.

src/store/__mocks__/load-data.js
```diff
@@ -0,0 +1,40 @@
...
+  return dataCopy;
+};
+
+/**
+ * Mock asynchronously loading/parsing data
+ * @param {string} path JSON file location. Defaults to main data url from config.js
+ * @return {function} A promise that will return when the file is loaded and parsed
+ */
+const loadJsonData = async (path = getUrl('main')) => {
+  await Promise.resolve();
```
bru5, line 25: Might be worth introducing a small timeout to simulate real world use, which should help catch any race conditions.

src/store/initial-state.js
```diff
@@ -43,16 +47,22 @@ export const mergeLocalStorage = state => {
 };
 
 /**
- * Prepare the pipeline data part of the state. This part is separated so that it
- * can be reset without overriding user settings.
- * @param {object} props Props passed to App component
+ * Prepare the pipeline data part of the state by normalizing the raw data,
```
bru5, line 50: In docs like this I'd consider splitting these comments over multiple more direct bullet points, rather than using full sentences in a block, to make each stand out and easier to read quickly.

src/store/initial-state.js
```diff
@@ -62,40 +72,32 @@ export const preparePipelineState = props => {
...
-    state.theme = props.theme;
-  }
-  // Override button visibility if set in props
-  if (props.visible) {
-    state.visible = Object.assign({}, state.visible, props.visible);
-  }
-  return state;
+
+  // Override with URL values
+  const flags = Object.assign({}, state.flags, getFlagsFromUrl());
```
bru5, line 81: In all of these `Object.assign` cases, spread could be used to make this clearer e.g.

```
const flags = { ...state.flags, ...getFlagsFromUrl() };
```

src/store/initial-state.js
```diff
@@ -62,40 +72,32 @@ export const preparePipelineState = props => {
...
- * @param {Object} props App component props
- * @return {Object} Initial state
+ * @param {object} props App component props
+ * @return {object} Initial state
  */
-const getInitialState = (props = {}) => {
-  // Perform 2 deepmerges seperately because it performs much faster
-  const state = deepmerge(
+const getInitialState = (props = {}) =>
+  Object.assign(
```
bru5, line 97: Could use spread here too.

bru5 (changes requested):

src/actions/pipelines.js
richardwestenra: In this case, it's just about whether it needs to make another pipeline endpoint request, i.e. to an endpoint other than 'main'. I don't have a specific name for these other endpoints so I avoided referring to them specifically because it would be more confusing? like, the best I can come up with is `shouldRequestAnotherEndpoint` which is garbage. I think it's implied that it's referring to pipeline endpoints because of what file it's in and which function is calling it. I think if anything, the best option here would be to add more detail to the JSDoc comment?

src/components/icons/loading.scss
richardwestenra: Will this be the case even if there's no transition happening here? But yeah sure.

src/config.js
richardwestenra: okay fair enough, I'll move it to `utils/index.js` I guess, and split out the string parts into an object and put them in config.

src/store/initial-state.js
richardwestenra: Personally I'm probably more pro-`Object.assign` than pro-spread, but yeah sure

src/actions/pipelines.js
bru5: Yeah I see how it could be difficult to make this code more self describing, so a comment is fine to avoid too much refactoring. Though I do think this updated comment could be simplified / split up to make it clearer, it's not a huge deal.

bru5 (approved):

src/actions/pipelines.js
richardwestenra: Cool, I've updated the comments again

src/components/app/app.test.js
```diff
@@ -67,4 +68,22 @@ describe('App', () => {
...
+    const activePipeline = animals.pipelines.find(
+      pipeline => !demo.pipelines.map(d => d.id).includes(pipeline.id)
+    );
+    const { container, rerender } = render(<App data={animals} />);
+    const pipelineDropdown = container.querySelector('.pipeline-list');
+    const menuOption = within(pipelineDropdown).getByText(activePipeline.name);
+    const pipelineDropdownLabel = pipelineDropdown.querySelector(
+      '.kui-dropdown__label > span:first-child'
+    );
+    expect(pipelineDropdownLabel.innerHTML).toBe('Default');
```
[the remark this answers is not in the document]
richardwestenra: Totally agree with you: God I tried to use something else, I really did. It's something I wanted to flag: This dropdown has no label, and few roles, etc. I'm trying to select the element and enquire about what its text is, so I can't select it by its text. I tried using `innerText` but that returned `undefined` for some reason, but `innerHTML` worked so I went with it.

I just worked out that I could do this instead:
```es6
    const pipelineDropdownLabel = within(pipelineDropdown)
      .getByRole('button')
      .querySelector('span:first-child');
    expect(getNodeText(pipelineDropdownLabel)).toBe('Default');
    fireEvent.click(menuOption);
    expect(getNodeText(pipelineDropdownLabel)).toBe(activePipeline.name);
    rerender(<App data={demo} />);
    expect(getNodeText(pipelineDropdownLabel)).toBe('Default');
  });
```
but it's pretty much the same anyway? Happy to change it though.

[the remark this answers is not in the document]
bru5: Fair enough, text nodes can make things tricky. It looks like https://github.com/testing-library/jest-dom is the recommended way in this case, which might be worth trying in future if we don't have it ready, but for now this is fine since it's in a test script only.

As for the label, I'd assume [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) will be the cleanest way to add that?

PR #273, opened 2020-09-30
[KED-2135]Fix side-effects in prepareState caused by pipeline flag setup

### LINKED ISSUE #272: importing Mockstate into tests will cause tests to fail
## Description
Currently importing and using mockState in tests will cause the 'announceFlags' test to fail. This is because the `pipeline` field within `flags` in the mock state are assigned to true which does not match the default state as set up in the config file. ( this is previously set up to enable other tests.). To solve this, we would need to add checks within the config file to assign different default values for flags under testing scenarios. 

## Context
This bug has caused test to fail while working on KED-1941 when we needed to import and use `mockState` within the tests

## Steps to Reproduce

1. import `mockState` into `app.test.js`
2. run tests
3. the `it announces flags` test will fail

## Expected Result
On fixing the bug, different default values will be assigned for flags which would not cause any tests to fail. 

## Actual Result
The `it announces flags` test fails ( see below)

image

## Your Environment

* NodeJS version used: 12.18.4
* Kedro version used: 3.5.1

studioswong:
## Description

This is a fix to a failing test that was uncovered while working on KED-1941. Currently `mockState` would assign the `pipeline` field within `flags` as true to accomodate some tests, which in turn causes discrepancies with the defaule value for `flags` within the app thereby causing the test for `'it announces flags'` to fail. 

## Development notes

I have added a simple check to assign different default values for `pipeline` under testing scenerios. With this check, we can also delete the assignment of the default value of `flags` within the `mockState` file.

merging this in will resolve #272

P.S Special thanks to the amazing @richardwestenra for going through this problem and pairing with me on this ticket! 

## QA notes

N/A

## Checklist

## Legal notice

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -18,7 +18,7 @@ export const flags = {
   },
   pipelines: {
     description: 'Select from multiple pipelines',
-    default: false,
+    default: typeof jest !== undefined,
     icon: '🔀'
   }
 };
--- a/src/utils/state.mock.js
+++ b/src/utils/state.mock.js
@@ -18,8 +18,6 @@ import { saveState } from '../store/helpers';
  * @param {Object} props
  */
 export const prepareState = (...props) => {
-  // Set pipeline flag to true:
-  saveState({ flags: { pipelines: true } });
   const initialState = getInitialState(...props);
   const actions = [
     // Set fontLoaded = true:


richardwestenra (approved):
Nice one! :shipit:

PR #275, opened 2020-10-06
[KED-2035] Expose parameter metadata 

921kiyo:
## Description

As an extension for metadata side panel, we are exposing parameters metadata to Viz. Here are examples of the endpoints and and outputs. 

`http://127.0.0.1:4141/api/nodes/d577578a` (endpoint for `params:xxx`)

```json
{
  "parameters": 0.2
}
```

`http://127.0.0.1:4141/api/nodes/f1f1425b` (endpoint for `parameters`)

```json
{
  "parameters": {
    "example_learning_rate": 0.01,
    "example_num_train_iter": 10000,
    "example_test_data_ratio": 0.2
  }
}
```

## QA notes

Modified unit tests
## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -455,12 +455,11 @@ def format_pipeline_data(
         is_param = bool("param" in namespace.lower())
         node_id = _hash(namespace)
 
-        node_data = (
-            _get_parameter_node(node_id, namespace)
-            if is_param
-            else _get_dataset_node(node_id, namespace)
-        )
-        _JSON_NODES[node_id] = node_data
+        node_data = _get_dataset_data_params(node_id, namespace)
+        _JSON_NODES[node_id] = {
+            "type": "parameters" if is_param else "data",
+            "obj": node_data,
+        }
 
         if node_id not in nodes:
             nodes[node_id] = {
@@ -477,20 +476,15 @@ def format_pipeline_data(
             nodes[node_id]["pipelines"].append(pipeline_key)
 
 
-def _get_parameter_node(node_id, namespace):
-    parameters = _CATALOG._get_dataset(namespace)
-    return {"type": "parameters", "obj": parameters}
-
-
-def _get_dataset_node(node_id, namespace):
+def _get_dataset_data_params(node_id, namespace):
     if KEDRO_VERSION.match(">=0.16.0"):
         try:
-            dataset = _CATALOG._get_dataset(namespace)
+            node_data = _CATALOG._get_dataset(namespace)
         except DataSetNotFoundError:
-            dataset = None
+            node_data = None
     else:
-        dataset = _CATALOG._data_sets.get(namespace)
-    return {"type": "data", "obj": dataset}
+        node_data = _CATALOG._data_sets.get(namespace)
+    return node_data
 
 
 @app.route("/api/main")
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -457,7 +457,7 @@ def format_pipeline_data(
         is_param = bool("param" in namespace.lower())
         node_id = _hash(namespace)
 
-        node_data = _get_dataset_data_params(node_id, namespace)
+        node_data = _get_dataset_data_params(namespace)
         _JSON_NODES[node_id] = {
             "type": "parameters" if is_param else "data",
             "obj": {namespace.replace("params:", ""): node_data}
@@ -480,7 +480,7 @@ def format_pipeline_data(
             nodes[node_id]["pipelines"].append(pipeline_key)
 
 
-def _get_dataset_data_params(node_id, namespace):
+def _get_dataset_data_params(namespace: str):
     if KEDRO_VERSION.match(">=0.16.0"):
         try:
             node_data = _CATALOG._get_dataset(namespace)
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -546,14 +546,14 @@ def nodes_metadata(node_id):
 
     parameters = node["obj"]
     if isinstance(parameters, dict):
-        # In case of `params:` prefix
+        # In case of 'params:' prefix
         parameters_metadata = {
             "parameters": {
                 next(iter(parameters)): next(iter(parameters.values())).load()
             }
         }
         return jsonify(parameters_metadata)
-    # In case of `parameters`
+    # In case of 'parameters'
     parameters_metadata = {"parameters": node["obj"].load()}
     return jsonify(parameters_metadata)
 
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -545,11 +545,9 @@ def nodes_metadata(node_id):
     if node["type"] == "data":
         dataset_metadata = _get_dataset_metadata(node)
         return jsonify(dataset_metadata)
-    if node["type"] == "parameters":
-        parameters_metadata = {"parameters": node["obj"].load()}
-        return jsonify(parameters_metadata)
-    # return empty JSON for parameters type
-    return jsonify({})
+
+    parameters_metadata = {"parameters": node["obj"].load()}
+    return jsonify(parameters_metadata)
 
 
 @app.errorhandler(404)
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -541,14 +541,13 @@ def nodes_metadata(node_id):
         dataset_metadata = _get_dataset_metadata(node)
         return jsonify(dataset_metadata)
 
+    parameter_values = node["obj"].load()
     if "parameter_name" in node:
         # In case of 'params:' prefix
-        parameters_metadata = {
-            "parameters": {node["parameter_name"]: node["obj"].load()}
-        }
+        parameters_metadata = {"parameters": {node["parameter_name"]: parameter_values}}
     else:
         # In case of 'parameters'
-        parameters_metadata = {"parameters": node["obj"].load()}
+        parameters_metadata = {"parameters": parameter_values}
     return jsonify(parameters_metadata)
 
 
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -68,7 +68,7 @@ _DEFAULT_KEY = "__default__"
 
 _DATA = None  # type: Dict
 _CATALOG = None  # type: DataCatalog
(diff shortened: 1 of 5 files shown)


richardwestenra:
This is great, thank you! So if I were to click on a single parameter node on the chart, which type of endpoint would I get? I assume it'd be the single `params:xxx` one, right? Can you please give a bit more info about the difference between these two?

921kiyo:
> This is great, thank you! So if I were to click on a single parameter node on the chart, which type of endpoint would I get? I assume it'd be the single `params:xxx` one, right? Can you please give a bit more info about the difference between these two?

Yes! In Kedro side, there are 2 ways to define parameters. Let's say in my `parameters.yml` in Kedro side, I have the following parameter definition.

```yaml
example_test_data_ratio: 0.2
example_num_train_iter: 10000
example_learning_rate: 0.01
```

If I want to access all of them in one of my Kedro nodes, I can use a keyword `parameters`, which returns all of them as a dictionary like

```
"parameters": {
    "example_learning_rate": 0.01,
    "example_num_train_iter": 10000,
    "example_test_data_ratio": 0.2
  }
```
But image if I have 1000 parameters in my `parameters.yml`, I probably don't want to grab all of them at once. If I only need to access just one of them, I can use `params:example_test_data_ratio`

From Kedro Viz UI, both `params:xxx` and `parameters` show up as a parameter node

Screenshot 2020-10-07 at 10 46 38

Hope this helps :)

limdauto (approved):
👏

richardwestenra:
ah okay interesting, thanks! Looks like you can access them both from clicking nodes on the graph then. It'd be great for these to have the same structure if possible, but I can't think of a way I'd do it better. I guess when receiving the data, we can just use `typeof parameters === 'object`', as long we the content of the single-param property `parameters` will only ever be either an object literal or a number/string/boolean. If it's possible for it to be something else (e.g. an array or object literal) then maybe we should use a different structure.

To explain better what I mean:
```json
// params:xxx
{
  // these are fine:
  "parameters": 0.2
  "parameters": 'this is a string'
  "parameters": true
  // these would cause problems, because I can't distinguish them from the full dictionary:
  "parameters": [0, 1, 2, 3]
  "parameters": { a: 1, b: 2 }
}
```
If we want to be more clear about distinguishing them, maybe we should use a different property name, e.g. `parameter` vs `parameters`?

921kiyo:
To give you more example, parameters can be nested. 
If I have the following paramter definitions in `parameters.yml`, 
```yaml
example_test_data_ratio:
  nested1: 0.2
  nested2: 0.3
  nested3: 0.4
example_num_train_iter: 10000
example_learning_rate: 0.01
```
Hitting `params:example_test_data_ratio` will return 
```json
{
  "parameters": {
    "nested1": 0.2,
    "nested2": 0.3,
    "nested3": 0.4
  }
}
```

and `parameters` endpoint will return

```json
{
  "parameters": {
    "example_learning_rate": 0.01,
    "example_num_train_iter": 10000,
    "example_test_data_ratio": {
      "nested1": 0.2,
      "nested2": 0.3,
      "nested3": 0.4
    }
  }
}
```

richardwestenra:
@921kiyo cool in that case, as mentioned on Slack, I propose that for single params, you just include a single key/value pair. instead of this:
```js
{
  parameters: 0.01,
}
```
do this instead:
```js
{
  parameters: {
    "example_learning_rate": 0.01,
  }
}
```
So that way, the formatting is consistent and we can treat them the same way.

package/kedro_viz/server.py
```diff
@@ -475,15 +476,15 @@ def format_pipeline_data(
             nodes[node_id]["pipelines"].append(pipeline_key)
 
 
-def _get_dataset_node(node_id, namespace):
+def _get_dataset_data_params(node_id, namespace):
```
DmitriiDeriabinQB, line 479: Why `node_id` argument is not used in this function?

package/kedro_viz/server.py
```diff
@@ -539,8 +540,8 @@ def nodes_metadata(node_id):
         dataset_metadata = _get_dataset_metadata(node)
         return jsonify(dataset_metadata)
 
-    # return empty JSON for parameters type
-    return jsonify({})
+    parameters_metadata = {"parameters": node["obj"].load()}
```
DmitriiDeriabinQB, line 543: I guess you are going to refactor this to cater for cases like `params:foo.bar`, correct?

package/kedro_viz/server.py
921kiyo: Mmm, seems like `node_id` argument was not used before this PR as well. I've removed it.

package/kedro_viz/server.py
921kiyo: Yep, I've addressed that in https://github.com/quantumblacklabs/kedro-viz/pull/275/commits/cfa570355fcd58d39adf143f8d076b8a7e05fbf8

package/kedro_viz/server.py
```diff
@@ -539,8 +544,18 @@ def nodes_metadata(node_id):
...
         return jsonify(dataset_metadata)
 
-    # return empty JSON for parameters type
-    return jsonify({})
+    parameters = node["obj"]
+    if isinstance(parameters, dict):
+        # In case of 'params:' prefix
+        parameters_metadata = {
+            "parameters": {
+                next(iter(parameters)): next(iter(parameters.values())).load()
```
DmitriiDeriabinQB, line 552: The readability is not great here. Can you reduce the nesting by doing

```python
key, value = next(iter(parameters.items()))
```

package/kedro_viz/server.py
```diff
@@ -539,8 +544,18 @@ def nodes_metadata(node_id):
         dataset_metadata = _get_dataset_metadata(node)
         return jsonify(dataset_metadata)
 
-    # return empty JSON for parameters type
-    return jsonify({})
+    parameters = node["obj"]
+    if isinstance(parameters, dict):
```
DmitriiDeriabinQB, line 548: I'm not a big fan of such `isinstance` here. Can we maybe explicitly save parameter name in the dict we put in `_JSON_NODES`?
Something like:
```python
if "parameter_name" in node:
    # handle "params:..."
else:
    # handle "parameters"
```
@limdauto wdyt?

package/kedro_viz/server.py
```diff
@@ -453,12 +457,13 @@ def format_pipeline_data(
...
-            {"type": "parameters", "obj": None}
-            if is_param
-            else _get_dataset_node(node_id, namespace)
-        )
-        _JSON_NODES[node_id] = node_data
+        node_data = _get_dataset_data_params(namespace)
+        _JSON_NODES[node_id] = {
+            "type": "parameters" if is_param else "data",
+            "obj": {namespace.replace("params:", ""): node_data}
+            if is_param and "params:" in namespace
```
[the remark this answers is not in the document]
921kiyo: Thanks! Removed `is_param` checking :)

package/kedro_viz/server.py
921kiyo: Yeah, I can do that but how do you get "parameter_name" in this function? we can only access to `parameter_name` from `next(iter(parameters.items()))` correct?

DmitriiDeriabinQB: I meant I don't particularly like the fact that instead of the object we put the dictionary with name and object. Parameter name can have its own top-level key rather than being squeezed into the `obj` key.

richardwestenra: @DmitriiDeriabinQB Hope I understand you correctly, but it sounds like you're asking for the single-parameter endpoint to have a different structure from the multiple-parameter endpoint, is that right? Kiyo had it like that originally, but [I asked him to change it](https://github.com/quantumblacklabs/kedro-viz/pull/275#issuecomment-705170566).

If you think about this from the front-end perspective, we might want to separate out different keys to display them differently in the viz meta panel when clicking on a 'Parameters' node - e.g.:
image
but if you click on a single node, we might want to display it the same way, but if it's not in the same format we might not be able to recognise it as a distinct type just from the structure, because a parameter can be one of lots of different types. So we're using the same structure to keep it simple on the front end. Does that help? Hope I haven't misunderstood you

921kiyo: @richardwestenra I'm still keeping the format you asked, his suggestion was more about how to generate it in code :)

package/kedro_viz/server.py
```diff
@@ -539,8 +541,15 @@ def nodes_metadata(node_id):
         dataset_metadata = _get_dataset_metadata(node)
         return jsonify(dataset_metadata)
 
-    # return empty JSON for parameters type
-    return jsonify({})
+    if "parameter_name" in node:
+        # In case of 'params:' prefix
+        parameters_metadata = {
+            "parameters": {node["parameter_name"]: node["obj"].load()}
```
limdauto, line 547: Sorry could you remind me what does `node["obj"].load()` do?

921kiyo: `node["obj"]` is MemoryDataSet, so it will load the value of params:key parameter (e.g. `0.2`)

package/kedro_viz/server.py
DmitriiDeriabinQB: @richardwestenra yeah, I read through your discussion with Kiyo above and understand the solution. I wasn't talking about the format it is exposed to the frontend but rather the way the data stored in the backend. I'm moderately cautious about the complexity of the data wrangling logic on Python side (while the data itself is very trivial).
TLDR: It's purely the discussion about the Python side of things and should not affect the API contract.

package/kedro_viz/server.py
```diff
@@ -539,8 +541,15 @@ def nodes_metadata(node_id):
         dataset_metadata = _get_dataset_metadata(node)
         return jsonify(dataset_metadata)
 
-    # return empty JSON for parameters type
-    return jsonify({})
+    if "parameter_name" in node:
```
DmitriiDeriabinQB, line 544: This is much better (to me personally at least), thanks @921kiyo 👍

package/kedro_viz/server.py
DmitriiDeriabinQB, line 547: Maybe do `param_data = node["obj"].load()` once before `if "parameter_name" in node` since it's used in both cases

DmitriiDeriabinQB (approved):

PR #276, opened 2020-10-09
[KED-1423] Feature/metadata panel

bru5:
## Description

Adds a metadata side panel. 

Screenshot 2020-10-29 at 16 41 18

## Development notes

This first version is behind flag `meta` and only shows existing data, a later update will add data from further APIs.

The responsive approach has been adapted to account for the additional sidebar.

## QA notes

The metadata flag must be enabled before testing. To see inputs and outputs, the new graph flag must also be enabled (at least for the initial implementation). Here is an example:

`http://localhost:4141/?data=demo&newgraph=true&meta=true`

Then to see the metadata panel, you should click on any node either in the sidebar or on the chart.

Note that different node types show different information, so test each kind.

To test the responsive approach, test with each sidebar individually and with both open, at sizes below `800px`.

## Checklist

## Legal notice


src/components/metadata/metadata-row.js
```diff
@@ -0,0 +1,46 @@
...
+  kind = 'text',
+  empty = '-',
+  visible = true,
+  inline = true,
+  commas = true,
+  children
+}) => {
+  const showElements = React.isValidElement(children);
+  const showList = !showElements && Array.isArray(children) && children.length;
+  const showValue = !showList && !showElements;
```
richardwestenra, line 21:
```suggestion
  const showValue = !showList;
```
`!showList` implies `!showElements`

src/components/metadata/metadata-row.js
```diff
@@ -0,0 +1,46 @@
...
+            inline={inline}
+            commas={commas}
+            kind={kind}
+            empty={empty}>
+            {children}
+          </MetaDataList>
+        )}
+        {showValue && (
+          <MetaDataValue value={children} kind={kind} empty={empty} />
+        )}
```
richardwestenra, line 39:
```suggestion
        {showList ? (
          <MetaDataList
            property={property}
            inline={inline}
            commas={commas}
            kind={kind}
            empty={empty}>
            {children}
          </MetaDataList>
        ) : (
          <MetaDataValue value={children} kind={kind} empty={empty} />
        )}
```

src/config.js
```diff
@@ -6,6 +6,10 @@ export const localStorageName = 'KedroViz';
 // These values are used in both SCSS and JS, and we don't have variable-sharing
 // across Sass and JavaScript, so they're defined in two places. If you update their
 // value here, please also update their corresponding value in src/styles/_variables.scss
+export const metaSidebarWidth = {
+  open: 400,
+  closed: 0
+};
```
richardwestenra, line 12: It looks like the CSS is just using the `sidebarWidth` values. If it's worth adding different ones for the JS, to be consistent you should use different variables for the CSS too.

src/selectors/layout.js
```diff
@@ -38,26 +39,34 @@ export const getSidebarWidth = (visibleSidebar, outerChartWidth) => {
   return sidebarWidth.closed;
 };
 
+/**
+ * Return the displayed width of the meta sidebar
+ */
+export const getMetaSidebarWidth = visibleMetaSidebar =>
+  visibleMetaSidebar ? metaSidebarWidth.open : metaSidebarWidth.closed;
```
richardwestenra, line 46: This might help with the mobile screen width issues that I mentioned previously:
```suggestion
export const getMetaSidebarWidth = (visibleMetaSidebar, outerChartWidth) => {
  if (visibleMetaSidebar && outerChartWidth > sidebarBreakpoint) {
    return metaSidebarWidth.open;
  }
  return metaSidebarWidth.closed;
};
```

src/selectors/layout.js
```diff
@@ -38,26 +39,34 @@ export const getSidebarWidth = (visibleSidebar, outerChartWidth) => {
...
       left,
       top,
       outerWidth: width,
       outerHeight: height,
-      width: width - sidebarWidth,
+      width: width - currentSidebarWidth - currentMetaWidth,
       height,
-      sidebarWidth
+      sidebarWidth: currentSidebarWidth,
+      metaSidebarWidth: currentMetaWidth
```
richardwestenra, line 69: I don't understand why you've renamed these to add 'current' to the names, as they just get renamed again in the function output anyway? Can you please explain the distinction?

src/selectors/metadata.js
```diff
@@ -0,0 +1,67 @@
...
+    const metadata = {
+      node,
+      tags: [...nodeTags[node.id]]
+        .map(tagId => tagNames[tagId])
+        .sort(sortAlpha),
+      pipeline: pipeline.name[pipeline.active],
+      runCommand: getRunCommand(node)
+    };
+
+    // Note: node.sources node.targets require newgraph enabled
```
richardwestenra, line 55: That's interesting. It's probably achievable relatively easily with the old Dagre setup, but I imagine it's not worth it as we'll be deprecating that soon anyway

src/selectors/nodes.js
```diff
@@ -16,12 +16,16 @@ const getNodeTags = state => state.node.tags;
...
-const getClickedNode = state => state.node.clicked;
 const getTagActive = state => state.tag.active;
 const getTextLabels = state => state.textLabels;
 const getFontLoaded = state => state.fontLoaded;
 const getNodeTypeDisabled = state => state.nodeType.disabled;
 
+/**
+ * Gets the currently clicked node if any
+ */
+export const getClickedNode = state => state.node.clicked;
```
richardwestenra, line 27: This doesn't need to be exported from this file and imported in the other, as it's not a true complex selector, it's just a simple prop selector function like the ones on lines 12-22. I'd put it with those and replicate it in the files where it's imported. The reason I say this is that it's annoyingly easy to introduce circular dependencies in the selector files and I'd prefer not to risk this more than necessary. I know this file's functions are already being imported in `src/selectors/metadata.js` but I'd still prefer to avoid unnecessary imports tbh. Does this make sense?

src/selectors/nodes.js
```diff
@@ -49,6 +53,18 @@ export const getNodeSelected = createSelector(
     )
 );
 
+/**
+ * Gets a map of nodeIds to nodes
+ */
+export const getNodesById = createSelector(
+  [state => state.graph.nodes],
```
richardwestenra, line 60: Conversely, I'd make this a named function (`getGraphNodes`?) and put it up top

src/selectors/nodes.js
```diff
@@ -49,6 +53,18 @@ export const getNodeSelected = createSelector(
...
+  (nodes = []) =>
+    nodes.reduce((result, node) => {
+      result[node.id] = node;
+      return result;
+    }, {})
+);
+
```
richardwestenra, line 67: tbh if this selector is only used for metadata you could put it in the metadata selector file if you like. This file's already pretty long and it's already starting to get broken up with some other selectors moving to other files to resolve dependency issues etc.

richardwestenra (changes requested):
👍 looks really good, but I've added quite a lot of comments sorry!

src/components/flowchart/index.js
```diff
@@ -243,7 +243,10 @@ export class FlowChart extends Component {
         });
       });
 
-    this.el.svg.call(this.zoomBehaviour);
+    this.el.svg
+      .call(this.zoomBehaviour)
+      // Prevent default double click zoom.
+      .on('dblclick.zoom', null);
```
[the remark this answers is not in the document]
bru5: Yeah the issue here is that it interferes with the metadata panel movement / zoom on click. I'd like to keep this too ideally, but was waiting for design input on the sidebar button which would avoid the need for any fix here.

But we can either wait for that (probably a later PR) or I can look at implementing a workaround now (potentially reimplementing the double click zoom feature and accounting for the metadata panel). I'm not sure how long that would take, but it might be worth it if we are considering releasing this unflagged (else I could put a flag on this line for now).

src/components/metadata/index.js
```diff
@@ -0,0 +1,87 @@
...
+ * Shows node meta data
+ */
+const MetaData = ({ visible = true, metadata }) => {
+  const [showCopied, setShowCopied] = useState(false);
+
+  const isTaskNode = metadata?.node.type === 'task';
+
+  const runCommandText = !showCopied
+    ? metadata?.runCommand
+    : 'Copied to clipboard.';
```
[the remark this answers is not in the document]
bru5: I think that could work, but changing the height would cause the content below to jump down and up which could be a bit jarring? (that said, I still need to add something for that here for multi-line cases)

src/components/metadata/metadata-row.js
```diff
@@ -0,0 +1,46 @@
...
+  children
+}) => {
+  const showElements = React.isValidElement(children);
+  const showList = !showElements && Array.isArray(children) && children.length;
+  const showValue = !showList && !showElements;
+
+  return (
+    visible && (
+      <div className="pipeline-metadata__row">
+        <h3 className="pipeline-metadata__label">{label}</h3>
```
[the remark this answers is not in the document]
bru5: Yeah I was wondering this, I didn't think `dt / dd / dfn` were appropriate here initially since it's not a `term: definition` relationship, but looking closer at the spec it does seem like `dd` is intended for that.

src/components/flowchart/index.js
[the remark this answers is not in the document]
richardwestenra: Okay fair enough. If it's necessary to keep this as is for now, I'd recommend adding to the comment, to explain why it's disabled, so that the next dev doesn't come along and wonder why and remove this line.

Also maybe create a new ticket for implementing the workaround or design change etc if necessary, just so we don't lose this through the cracks?

src/components/metadata/index.js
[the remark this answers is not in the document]
richardwestenra: Ah yeah fair. In that case feel free to keep it, but it sounds like you might need to figure out the height issue anyway with some sort of fixed-height message in separate element. I'll ping Gabriel on slack for his thoughts.

src/components/metadata/metadata-row.js
bru5: Yeah it is redundant but I left it in to make it easier to read, but it seems clear enough then so I'll remove it.

src/components/metadata/metadata-row.js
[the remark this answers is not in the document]
richardwestenra: Yeah I had that thought too, but I think it should be relatively straightforward to change the styling from `display: flexbox` to `display: block` and use different margin/padding top/bottom for the different rows. Each component could be just something like 
```es6
return (
  <>
    <dt>{label}</dt>
    <dd>{value or whatever}</dd>
  </>
)
```

src/components/metadata/metadata-row.js
bru5: Again I wrote it this way as I felt it was clearer to read, since with both of these changes you now need to follow the logic through to understand all the potential cases here, rather then it being self describing. But I can still make the change if you think it's clear enough?

richardwestenra: Yeah I hear you - I just found that it was relatively straightforward to understand that if there's a list, show the list, else just show a value. The extra steps actually felt a bit less clear tbh as you need to puzzle out the extra expressions. I think it would make sense to include only if `showValue` were more complex than just `!showList`

src/selectors/layout.js
bru5: Ah yes this can now just be reverted as it's no longer necessary.

src/selectors/metadata.js
bru5: Yeah so ideally this logic should be refactored into the selectors, which I started doing to avoid this but there were a lot of complications due to some dependencies and exactly we're due to deprecate so it's not worth that effort at this time, but I'll likely revisit the idea later.

src/selectors/nodes.js
bru5: I think ideally even these should be shared from a maintainability point of view, but if this causes circular dependencies then sure I'll make a copy instead. Going forwards, it might be possible to refactor all of these into a separate shared file which might avoid those too, but for now I've done this.

src/selectors/nodes.js
bru5: I think since this is such a general selector, one that will likely be reused in future, it feels more natural for it to be in this layer?

src/selectors/nodes.js
richardwestenra: I hear you, I think I could go either way on them. Tbh though I see that as so simple that it doesn't seem hugely worthwhile to abstract them? But yeah if we were going to, then putting them in a single share file would make the most sense. But that should be a separate PR. Thanks!

src/selectors/nodes.js
richardwestenra: Fair enough, I think there's decent arguments either way, feel free to leave it

src/components/metadata/metadata-row.js
bru5: Actually trying this again I remember why I needed to do this originally, because there's a subtle requirement that this fails on, that a value or list should not be shown if there are React children.

I think there might be a better way to avoid this though by changing around some props instead, so I'll see if that works better.

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,181 @@
+@import '../../../styles/variables';
+
+.pipeline-metadata {
+  position: absolute;
+  top: -1px; /* Avoids pixel rounding gaps */
+  right: 0;
+  bottom: -1px;
+  z-index: 1;
+  display: flex;
+  flex-direction: column;
```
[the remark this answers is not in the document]
bru5: Yeah I think this was no longer necessary, so I've removed it then.

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,181 @@
...
+    background: $color-bg-dark-2;
+    border-bottom: 1px solid rgba(0, 0, 0, 0.2);
+    box-shadow: 0 0 10px rgba(0, 0, 0, 0.3);
+  }
+}
+
+.pipeline-metadata__row {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
```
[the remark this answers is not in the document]
bru5: No longer necessary, so I've removed it too.

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,181 @@
...
+    background: $color-bg-light-3;
+    border-bottom: 1px solid rgba(0, 0, 0, 0.05);
+    box-shadow: 0 0 10px rgba(0, 0, 0, 0.07);
+  }
+
+  .kui-theme--dark & {
+    background: $color-bg-dark-2;
+    border-bottom: 1px solid rgba(0, 0, 0, 0.2);
+    box-shadow: 0 0 10px rgba(0, 0, 0, 0.3);
+  }
```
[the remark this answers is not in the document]
bru5: Sure I've refactored these into a mixin which is now shared for both sidebars.

src/config.js
bru5: Good point, I've added the new constants in CSS too.

src/selectors/nodes.js
bru5: Okay I think that works if we make the distinction, since that seems to be the case elsewhere, so made the change.

src/components/metadata/metadata-row.js
bru5: I've made that refactoring so now this should be avoided anyway.

src/components/metadata/metadata-row.js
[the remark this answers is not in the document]
bru5: Yeah I've managed to wrangle this to work for now then, assuming we allow `dd` to have any kind of children. 

Though it isn't ideal from a flexibility and reusability point of view and may cause problems in future. Just as an example of what I'm getting at, think about trying to add sub-components into this sidebar that didn't fit the `dt / dd` model. 

I could certainly use the `<Container>` prop pattern again here if needed, but depending on sidebar design, we'd need to potentially split out multiple `<dl>` lists. Doing this with dynamic rows, like sorting or hide / show based on type, it may get complicated to support this markup.

src/components/flowchart/index.js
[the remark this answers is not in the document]
bru5: Added a better comment here, will add a ticket too.

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,181 @@
...
+  padding: 12px 0 24px 0;
+  overflow-y: auto;
+  transform: translateX(100%);
+  transition: transform ease 0.4s;
+
+  @media (min-width: 900px) {
+    &--visible {
+      transform: translateX(0);
+    }
+  }
```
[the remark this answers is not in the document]
bru5: Yeah the responsive approach here needed some work. I've made some changes that affect both sidebars and the chart zoom approach, but I think it works well.

The best way is to try it out first, but here are roughly how it works (this sounds more complicated then it really is):

- the main sidebar
  - covers the screen at `viewport <= 400px`

- the metadata sidebar 
  - now has a close button
  - covers the screen at `viewport <= 400px`
  - begins to overlay main sidebar at `viewport <= 800px`

- the chart zoom
  - moves and zooms the chart to fit in the space between either sidebar at `viewport > 400px`
  - as above but excludes sidebars at `viewport <= 400px` (such that they just overlay)

The implementation of this is simple just using `min/max-width` on the sidebars and allowing them to flex naturally, rather than specific media queries.

src/components/metadata/index.js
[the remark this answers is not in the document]
bru5: This has been improved to handle multi-line cases.

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,234 @@
...
+
+.pipeline-metadata__run-command-value--no-visible {
+  opacity: 0;
+}
+
+.pipeline-metadata__toolbox .pipeline-metadata__copy-button {
+  width: 1.65em;
+  height: 1.65em;
+  opacity: 0.55;
+}
```
richardwestenra, line 207: This should probably be outside of the `<code>` element so that it doesn't inherit the font-family

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,234 @@
...
+  top: 0;
+  left: 0;
+
+  .kui-theme--light & {
+    color: rgba($color-text-light, 0.8);
+  }
+
+  .kui-theme--dark & {
+    color: rgba($color-text-dark, 0.8);
+  }
```
richardwestenra, line 220: Again, this is probably best replaced with `opacity: 0.8`

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,234 @@
...
+.pipeline-metadata__copy-message--no-visible {
+  display: none;
+}
+
+.pipeline-metadata__copy-button:hover {
+  opacity: 1;
+}
+
+.pipeline-metadata__copy-button:active {
+  transform: translate(0, 1px);
```
richardwestenra, line 232: Love it! `translateY(1px)` might be better though but doesn't matter at all

src/config.js
```diff
@@ -35,3 +45,8 @@ export const sidebar = {
     Parameters: 'parameters'
   }
 };
+
+export const runCommandTemplates = {
+  data: name => `kedro run --to-inputs ${name}`,
+  task: name => `kedro run --to-nodes ${name}`
+};
```
richardwestenra, line 52: I don't personally mind, but I thought you said that config.js shouldn't have variables in it? Should this be a utility? Though personally I'm happy for it to be in the meta sidebar component, because it's not very universal.

src/styles/_variables.scss
```diff
@@ -20,6 +20,8 @@ $color-bg-dark-1: #001521;
 $color-bg-dark-2: #031a26;
 $color-bg-dark-3: #061c28;
 $color-bg-dark-4: #0e2330;
+$color-bg-dark-5: #2a3b46;
```
richardwestenra, line 23: This is going to cause conflicts with #286, unfortunately. One of us is going to have to sort out these colours and their order. We should come up with an official shared colour palette

richardwestenra (changes requested):

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,234 @@
...
+    color: rgba($color-text-light, 0.8);
+  }
+
+  .kui-theme--dark & {
+    color: rgba($color-text-dark, 0.8);
+  }
+}
+
+.pipeline-metadata__list {
+  margin: 0;
```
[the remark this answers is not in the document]
bru5: The `dl` element has a margin top and bottom by default where I wanted `0` in this case, so I left this in to be explicit rather than just relying on margin collapse to do that, which is context dependent. I can remove this still though if you'd really prefer?

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,234 @@
...
+  font-weight: normal;
+  user-select: none;
+
+  .kui-theme--light & {
+    color: rgba($color-text-light, 0.55);
+  }
+
+  .kui-theme--dark & {
+    color: rgba($color-text-dark, 0.55);
+  }
```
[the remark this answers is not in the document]
bru5: Okay I've converted all of these to use the opacity approach instead then.

src/components/metadata/styles/metadata.scss
bru5: That was intentional as I found the font changing this way to be jarring, I've made the change though if you want to try it out.

src/config.js
bru5: Ah yeah so the thought is that configs shouldn't have things that:

  - have side effects
  - have conditionals
  - are non deterministic

So in other words they should be as static as possible, to avoid trouble.

Templates themselves are static, the only reason for the function here is because JS string templating needs it (I think), vs. e.g. `printf` style templates which don't.

Hopefully that makes sense as a compromise, but I could easily convert this to just use a string concat.

src/selectors/layout.js
```diff
@@ -31,33 +32,46 @@ export const getGraphInput = createSelector(
...
       top,
       outerWidth: width,
       outerHeight: height,
-      width: width - sidebarWidth,
       height,
-      sidebarWidth
+      width: excludeSidebars ? width : chartWidth,
+      sidebarWidth: excludeSidebars ? 0 : sidebarWidth,
+      metaSidebarWidth: excludeSidebars ? 0 : metaSidebarWidth
     };
```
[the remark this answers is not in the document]
bru5: Yeah the idea behind that was to make the best use of available space and to keep the chart centered, given any configuration of sidebars.

Maybe that's a little too jarring in some cases then, I've changed it to always ignore both sidebars for screens `< 1200px` let me know if you think that works?

src/components/metadata/styles/metadata.scss
```diff
@@ -0,0 +1,234 @@
...
+  line-height: 1.5;
+  word-break: break-word;
+
+  .kui-theme--light & {
+    color: rgba($color-text-light, 0.8);
+  }
+
+  .kui-theme--dark & {
+    color: rgba($color-text-dark, 0.8);
+  }
```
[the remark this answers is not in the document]
bru5: I changed these to use `0.85` is that what you meant?

src/components/metadata/styles/metadata.scss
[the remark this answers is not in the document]
richardwestenra: okay fair enough!

src/selectors/layout.js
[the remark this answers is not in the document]
richardwestenra: Much better. However I'd have it still respect the main sidebar when the sidebar doesn't go full-width - otherwise at anything between 400-1200px width, the layer labels and part of the graph just sits mostly under the main sidebar:
image

also, the layer labels are never visible at <1200px widths, because they sit under the sidebar even when it's closed:
image

Basically, I think that with respect to the main sidebar, anytime you have the graph and sidebar visible at the same time, the graph should default to the remaining screen width, because otherwise you won't see the layer labels and it looks off-centered and frankly a bit weird. Then once it hits a width where that's untenable (usually ~700px), just make the sidebar go full-width because you might as well rather than keeping it to max 400px because the graph underneath is unusable either way. This is mainly because you typically keep the main sidebar open or closed, you don't toggle it all the time.
Then for the meta panel, it should sit atop the graph content at anything below about 1200px width or so. This is because it's typically toggled often, and only temporarily open.

I might have a look into it and see if I can come up with a real code recommendation, if that's ok?

src/components/metadata/styles/metadata.scss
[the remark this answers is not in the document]
richardwestenra: That's better, but what I meant was, why do they need reduced opacity at all? I'm not sure that it improves things. But this goes against what I was saying in the meeting we just had, about quibbling over small design details. Personally I'd be inclined to make it 100% opacity (for consistency and also readability) if I were designing it, but I think it's better to just leave it and we can revisit if necessary

src/components/metadata/styles/metadata.scss
richardwestenra: Thanks, yeah I definitely find it better with the changed font, and I think Gabriel was on board with that too. Cheers!

src/config.js
richardwestenra: ok sure thing. I'd still probably put it in the component (because it's not likely to be reused elsewhere in the app, and it's effectively just copy text), but no biggie

src/styles/_variables.scss
richardwestenra: That PR has now been merged, so you're gonna have to rename some variables, sorry! btw are the shades still all in order from darkest to lightest? If not we should do that or else maybe come up with a new naming schema.

src/selectors/layout.js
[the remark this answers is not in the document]
richardwestenra: @bru5 I've [created a new branch with some changes](https://github.com/quantumblacklabs/kedro-viz/commits/feature/pipeline-selection-styling-responsive-proposal) to show you what I'm proposing. Feel free to merge if you think this works.

I think the JS changes in particular work much better. However I've also reverted the main sidebar width breakpoint, because many datasets have nodes with long names, and on smaller screen widths (<700px) you can't see the graph properly anyway when the sidebar is open, and you don't have a mouse to hover over the node names anyway to see their titles, so why not just make it full-width anyway?
I've also removed the min-widths because they don't appear to serve any useful purpose, to me. If the screen is that small, might as well just let them shrink? Better to be useful than perfect.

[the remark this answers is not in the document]
bru5: Okay thanks, I've merged your suggest changes here.

src/components/metadata/styles/metadata.scss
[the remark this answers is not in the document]
bru5: Yeah the approach to colours in both code and design on this I'm finding to be pretty confusing, so apologies for mixing any up. I agree it would be a good idea to consolidate and improve the approach in a later update rather than blocking here?

src/config.js
bru5: Ah I've removed it from the config then.

src/styles/_variables.scss
bru5: I've not tried to keep them in any kind of order, so maybe not sorry. Honestly that seems like it would require additional refactoring every time we add new colours in between?

src/selectors/layout.js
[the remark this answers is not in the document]
richardwestenra: Okay thanks! btw I hadn't yet fixed the tests that I broke by making those changes, because I wasn't sure whether you'd want to use them - if you like I can fix them, seeing as it was me who broken them?

src/components/metadata/styles/metadata.scss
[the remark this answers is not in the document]
richardwestenra: Totally agree. Let's park this for now and discuss later if necessary.

src/styles/_variables.scss
richardwestenra: I've found the order pretty useful personally (especially when trying to tweak the colours myself), but I agree that it might be a lot of work to refactor when adding new colours between. We might want to rename them to something more flexible (e.g. `$color-bg-dark-05, $color-bg-dark-10, $color-bg-dark-15`, or `$color-bg-very-dark, $color-bg-darker, $color-bg-darkest`, or whatever).

You'll need to rename the new ones in this case to resolve the merge conflict, but I'm happy to park the reordering for now, consider it tech debt, and revisit it later when we revisit the colour schema.

src/selectors/layout.js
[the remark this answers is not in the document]
bru5: Just to check, with your changes it looks like the metadata panel now always overlays? So for example if you click a node on the right side of the screen, it will then be covered by the panel. Just checking that's what you intended?

src/styles/_variables.scss
bru5: Ok I'll handle the merge conflict and revisit the approach to colours later, since that seems to be an app-wide problem.

richardwestenra (approved):
👉😎👉 zoop

PR #279, opened 2020-10-13
[KED-2099] Use the same JSON mock data files for JS+Python end unit/e2e tests

studioswong:
## Description
This PR aims to create a common test dataset that will be used in tests on both the Kedro-viz and Kedro-core projects.

## Development notes
The following are introduced in this PR:

1. Set up of the dataset (`test-data.json`) to replace the original data set (`animals.mock.js`) that was used throughout the tests for the FE.The dataset is based on the `animals` dataset, with the only difference in the `id` field under `nodes`, as well as the `edges` field in hash data format instead of string format. This is needed to allow the same dataset to be used by the Kedro-core tests.

The same data set has also been passed to Kiyo to test with the BE and could be subjected to revisions.

2. Update mockState to reference the `test-data.json` file in constructing the mock data.

3. Update of all tests to replace the usage of the `animals` dataset to use the new `test-data` dataset.

4. Revision of existing tests that references and tests with the `id` field of `nodes` to use hash data rather than the string data as setup previously.

I have still kept the `animals` dataset given it is still used as a sample data set under `data-source.js`. I have also kept the reference to the animals dataset within `data-source.js` and its relevant test. 

I have also kept the documentation as is with the reference to the `animals` dataset - @richardwestenra please let me know your thoughts, and we could revert to replacing the `animals` dataset entirely with the `test-data` dataset and I could make the relevant changes in `data-source.js` and the contribution docs. 

Many thanks to @921kiyo and @richardwestenra for your help and collaboration on this ticket!

## QA notes

N/A

## Checklist

## Legal notice


richardwestenra:
@studioswong Pro tip: Instead of adding "WIP" to the PR title, you can [open a draft pull request instead](https://github.blog/2019-02-14-introducing-draft-pull-requests/)!

studioswong:
> @studioswong Pro tip: Instead of adding "WIP" to the PR title, you can [open a draft pull request instead](https://github.blog/2019-02-14-introducing-draft-pull-requests/)!

Thanks for the pro tip Richard 👍 ! Yea I originally thought of opening this as a draft PR as well, but on second thought I think it might be more straight forward to just set this up as an actual PR so we can just merge this in directly on obtaining confirmation from Kiyo with the test data file on their side - my attempts to avoid another PR :D  ( The WIP is indicative that I might push new changes for the test data file) 

Sorry I should've mentioned this in my description! I can also close this and open another draft one if this turns out to be more more upcoming changes than intended ;)

src/utils/data/input.json
```diff
@@ -0,0 +1,168 @@
+{
+  "selected_pipeline": "__default__",
```
921kiyo, line 2: I've just removed `input.json` in https://github.com/quantumblacklabs/kedro-viz/pull/281, and started using animal test data, so you can delete this file :)

src/selectors/linked-nodes.test.js
```diff
@@ -5,20 +5,20 @@ import reducer from '../reducers';
...
+    expect(linkedNodes['4ffcf321']).toBe(true);
+    expect(linkedNodes['2cd4ba93']).toBe(true);
+    expect(linkedNodes['7196f150']).toBe(true);
+    expect(linkedNodes['1769e230']).toBe(true);
+    expect(linkedNodes['091b5035']).toBe(true);
+    expect(linkedNodes['6525f2e6']).toBe(true);
+    expect(linkedNodes['9d989e8d']).toBe(true);
+    expect(linkedNodes['e4951252']).toBe(true);
+    expect(linkedNodes['f1f1425b']).toBe(true);
+    expect(linkedNodes['2a31900d']).toBe(true);
```
richardwestenra, line 22: This is tricky, because these are no longer very identifiable. I would be tempted to put them into  a  lookup table of names and IDs so you can easily see from the test which is which, e.g.
```es6
describe('getLinkedNodes function', () => {
  const { nodes } = mockState.animals.graph;
  const nodeID = nodes.find(d => d.id.includes('salmon')).id;
  const newMockState = reducer(mockState.animals, toggleNodeClicked(nodeID));
  const linkedNodes = getLinkedNodes(newMockState);

  it('should return an object', () => {
    expect(linkedNodes).toEqual(expect.any(Object));
  });

  describe('should return true for ancestor/descendant nodes', () => {
    test.each([
      ['salmon', '4ffcf321'],
      ['pig', '2cd4ba93'],
      ['trout', '7196f150'],
      ['whale', '1769e230'],
      ['horse', '091b5035'],
      ['sheep', '6525f2e6'],
      ['cat', '9d989e8d'],
      ['dog', 'e4951252'],
      ['parameters', 'f1f1425b'],
      ['parameters_rabbit', '2a31900d']
    ])('node %s should be true', (name, id) => {
      expect(linkedNodes[id]).toBe(true);
    });
  });
});
```

It might be worth adding a new separate test for the 'false' case too. I'm sorry, it's not a very good test and that's my fault. I'm happy to work with you on this!

src/utils/data/test-data.json
```diff
@@ -0,0 +1,250 @@
+{
```
richardwestenra, line 1: Have you deleted `animals.mock.js` yet? Or changed `demo.mock.js` to a json file?

richardwestenra (changes requested):
Nice work! A couple of notes, that I'm happy to pair with you and discuss if you like

src/utils/data/input.json
studioswong: awesome, thanks Kiyo!

src/components/node-list/node-list-items.test.js
```diff
@@ -23,8 +23,8 @@ const ungroupNodes = groupedNodes =>
 
 describe('node-list-selectors', () => {
   describe('getFilteredNodes', () => {
-    const nodes = getGroupedNodes(mockState.animals);
-    const searchValue = 'e';
+    const nodes = getGroupedNodes(mockState.testData);
+    let searchValue = 'bear';
```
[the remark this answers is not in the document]
studioswong: I'm glad you asked - nice spot 👍 I introduced this change given the current new data arrangement where the `id` in the nodes is a hash that does not contain any reference to the `name` value of `nodes` - by setting the search term as a general `e`, currently it will return nodes with ids that contains 'e' which has no correlation with the string of the `name` and hence causing the tests to break. ( under the previous `animals` arrangement the `id` field is correlated to the `name` field (e.g `id` is something like `task/bear` with name as `bear`, while under the new data type it would be `id: 09f5edeb` and `name: bear`). 

By setting the search term as a more specfic term, it will ensure that only the node with the exact matching term will be returned. I understand this might not allow the testing for a more general search, but this new search term still allows us to test the original search feature of filtering the nodes that contains the search term.

src/components/node-list/node-list-items.test.js
```diff
@@ -45,6 +45,8 @@ describe('node-list-selectors', () => {
     });
 
     describe('nodeIDs', () => {
+      let searchValue = '09f5edeb';
```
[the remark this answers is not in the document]
studioswong: gotcha - thanks for this comment, I completely agree with this approach - will delete this test in this case 👍

src/selectors/linked-nodes.test.js
studioswong: great call, I have the same feeling when I was modifying the test as well, currently it's extremely unreadable and those without context of this test will have no clue on what's going on - I will modify it to set it into a table :)

src/utils/data/test-data.json
studioswong: I haven't deleted `animals.mock.js` yet as I saw that it is still referenced and used in `data-source.js` and wanted to double check with you first - I will delete it and replace the reference in `data-source.js`. 

In terms of `demo.mock.js`, I have kept it as is as this file is only reference within our tests and are not used for BE tests - i can change it to a json file as well if that's preferred?

studioswong:
As @richardwestenra and I discussed, I will construct a separate PR for changing `demo.mock.js` into a json file and updating relevant tests after this PR is merged.

src/utils/data-source.js
```diff
@@ -37,9 +37,9 @@ export const getSourceID = () => {
  */
 export const getDataValue = source => {
   switch (source) {
-    case 'animals':
-      // Use data from the 'animals' test dataset
-      return animals;
+    case 'testData':
+      // Use data from the 'testData' test dataset
+      return testData;
```
richardwestenra, line 42: hmm. I'm not stoked that this has a capital letter in it - I think ideally we'd want to keep URLs in lowercase. Anyway I'd be inclined to use 'test' as the datasource name anyway.

It's something I've been pondering through this renaming process - i.e. whether I'd prefer for it to just be called 'test' rather than 'testData', because that should be reasonably clear already as it's in the 'data' directory. e.g. 'demo' doesn't need to be called 'demoData', because it's clear that 'data/demo.json' is a data file. Furthermore, 'testData' gets confusing when you have multiple test data files, or if we want to add another in the future. And 'testData' doesn't describe the contents. Finally, changing the URL in the source effectively means we're changing the external-facing developer experience - it's a very minor breaking change. Not something to be done lightly.

Let's discuss this tomorrow and see what we think.

src/actions/pipelines.test.js
```diff
@@ -126,7 +126,7 @@ describe('pipeline actions', () => {
 
       it('should request data from a different dataset if the active pipeline is set', async () => {
         const { pipeline } = mockState.testData;
-        const active = pipeline.ids.find(id => id !== pipeline.main);
+        const active = pipeline.ids.find(id => id == pipeline.main);
```
richardwestenra, line 129: Why is this test being changed? Also, why is this using `==` instead of `===`?

src/utils/data-source.js
studioswong: I agree with your concerns - definitely on board with your comment on keeping URLs in lowercase. While I was modifying this file it did come to my mind that calling this as the `test-data` dataset is a bit redundant as everything we refer to here are 'data' - I would say within the tests, it does make sense to call this as `testData` as calling it as `test` within tests is confusing. 

Perhaps the best way forward is to call this as `test` instead of `test-data` within `data-source.js`, `app.js` and the contributing docs, while keeping it as `testData` within the tests for clarify purposes?

src/actions/pipelines.test.js
richardwestenra: Resolved: The test was changed. Have reverted these changes.

richardwestenra (approved):

PR #282, opened 2020-10-16
[KED-2186] Fade graph opacity when loading

richardwestenra:
## Description

I've noticed that often, when loading the graph while changing the pipeline, it's not sufficiently obvious that the chart is still loading. This is because the sidebar updates before the graph does, which gives the impression that it's ready when it isn't yet. This change aims to prevent this by fading out the graph a bit when loading, to make it more obvious that it's still displaying old data.
loading-fade

## Development notes

I haven't added a transition to this, because it wasn't very performant and I wanted it to be snappy

## QA notes

Best observed when changing pipelines, ideally with large pipelines.

## Checklist

## Legal notice


bru5 (changes requested):
I tried this out and when the loading is quick, for example toggling tags on the demo, the result is that all nodes appear to flicker every time.

So I think there needs to be a delay before this appears e.g. 500ms and it needs to show for at least e.g. 300ms to avoid too much flicker.

The same idea could also be applied to the loading spinner, though it is less of a problem there.

As for the effect, I think we can do something less obtrusive but I agree with the reasoning behind it.

src/components/flowchart/index.js
```diff
@@ -433,12 +434,14 @@ export class FlowChart extends Component {
...
    */
   render() {
-    const { chartSize, layers } = this.props;
+    const { chartSize, isLoading, layers } = this.props;
     const { outerWidth = 0, outerHeight = 0 } = chartSize;
 
     return (
       <div
-        className="pipeline-flowchart kedro"
+        className={classnames('pipeline-flowchart kedro', {
```
bru5, line 442: Every time this class is added then removed, there is potential for creating a CPU spike, which will create jank for the graph animation, so I'd suggest to instead apply the opacity directly to a SVG `g` element instead (such as the zoom wrapper).

richardwestenra: Ok cool, I've moved it to the zoom wrapper

richardwestenra:
Closing in favour of #296: After a discussion with @bru5, we feel that when switching pipeline, hiding the graph completely (via a state reset) will accomplish the same effect with fewer negative side-effects.

PR #288, opened 2020-10-21
Fix chart rendering edge cases and hover styles

bru5:
## Description

Fixes two rare errors with chart rendering, that are likely due to timing edge cases.

Also fixes sidebar hover styling for disabled buttons.

## Development notes

As above.

## QA notes

The errors themselves are hard to reproduce, but they seem to mostly occur in Safari, after rapidly applying filters on the sidebar in random order.

To test the sidebar issue, load a graph with no layers such that the layers button is disabled. Hovering should no longer cause this or other disabled buttons to highlight.

## Checklist

## Legal notice

### DIFF
--- a/src/components/app/app.scss
+++ b/src/components/app/app.scss
@@ -13,10 +13,6 @@
   .kui-theme--dark & {
     fill: $color-light;
   }
-
-  .pipeline-icon-toolbar__button:hover & {
-    opacity: 1;
-  }
 }
 
 .pipeline-icon--stroke {
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -180,8 +180,12 @@ export const drawNodes = function(changed) {
       .duration(this.DURATION)
       .attr('transform', node => `translate(${node.x}, ${node.y})`)
       .on('end', () => {
-        // Sort nodes so tab focus order follows X/Y position
-        allNodes.sort((a, b) => a.order - b.order);
+        try {
+          // Sort nodes so tab focus order follows X/Y position
+          allNodes.sort((a, b) => a.order - b.order);
+        } catch (ex) {
+          // Avoid rare DOM errors thrown due to timing issues
+        }
       });
 
     enterNodes.select('rect').call(updateNodeRects);
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -266,8 +266,10 @@ export const drawEdges = function(changed) {
   if (changed('edges', 'centralNode', 'linkedNodes')) {
     allEdges.classed(
       'pipeline-edge--faded',
-      ({ source, target }) =>
-        centralNode && (!linkedNodes[source] || !linkedNodes[target])
+      edge =>
+        edge &&
+        centralNode &&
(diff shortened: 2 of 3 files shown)


src/components/app/app.scss
```diff
@@ -13,10 +13,6 @@
   .kui-theme--dark & {
     fill: $color-light;
   }
-
-  .pipeline-icon-toolbar__button:hover & {
-    opacity: 1;
-  }
```
richardwestenra, line 19: Why this change? Will this not totally remove all hover styling? If you want to avoid having hover states on disabled buttons, I think this would be a better approach:
```scss
.pipeline-icon-toolbar__button:not(:disabled):hover & {
  opacity: 1;
}
```

src/components/flowchart/draw.js
```diff
@@ -180,8 +180,12 @@ export const drawNodes = function(changed) {
       .duration(this.DURATION)
       .attr('transform', node => `translate(${node.x}, ${node.y})`)
       .on('end', () => {
-        // Sort nodes so tab focus order follows X/Y position
-        allNodes.sort((a, b) => a.order - b.order);
+        try {
+          // Sort nodes so tab focus order follows X/Y position
+          allNodes.sort((a, b) => a.order - b.order);
+        } catch (ex) {
```
richardwestenra, line 186: `ex`? Haven't seen that before, would've thought `err` usually, but whatever it really doesn't matter

src/components/app/app.scss
richardwestenra: ah I see, it's added elsewhere. Fair enough. I'm not stoked about the styles being added in two different places, but maybe we should address that in a separate PR. Nevermind :)

richardwestenra (approved):
Added a couple of comments, feel free to ignore them though

src/components/app/app.scss
bru5: I originally wrote the same thing, but then realised that there was no longer a need for that styling to be inside of `app.css` anymore and moved it to the icon toolbar styles where it would normally be.

There's no duplication, I'm not sure what you mean?

src/components/flowchart/draw.js
bru5: I just tend to write `ex` for exceptions that must be handled, `err` for anything else to be clear. Anyway, I changed it to `err` then.

src/components/app/app.scss
richardwestenra: you're right, no duplication, more that the intricately bound styles are being handled in two places. Made me think that it could be done in a more obvious manner. But it's not related to your PR so ignore for now

src/components/flowchart/draw.js
richardwestenra: aha. ok no worries, thanks!

richardwestenra (approved):
Once more with feeling

src/components/app/app.scss
bru5: I agree looks like it should be elsewhere too, I imagine the move to CSS variables will likely avoid what ever reason it was put here originally when it was first implemented.

PR #290, opened 2020-10-26
[KED-2189] SVG rendering performance

bru5:
## Description

Improves the performance of SVG rendering, frame rates with less jank during zoom, pan and transitions.

Large scale graphs with many edges benefit most, that in some cases originally pan and zoom at ~5 FPS,  
may now often reach ~60 FPS with these improvements.

## Development notes

A few techniques were used here, each made in a separate commit in this PR, roughly in order of effectiveness:

- ea03a4f34a731fa6020df5c5d30705e9bf97d325 adaptive quality during pan and zoom
  - the chart renders in a lower quality mode (no AA on shapes, but still on text)
  - very subtle appearance, in most cases not visible due to scaling and motion blur
  - uses `shape-rendering: optimizeSpeed` and `text-rendering: optimizeSpeed`
- cecab9ed0844cf4bc1236d63ceefe6b6bd4541b7 remove nested icon SVG, simplify icon paths, transforms for icon positions
  - icons are now single `path` elements rather than embedded SVGs
  - a further boost may be possible using pre-rendered PNG `image` elements but has limitations
- bb189a2f77b5006fe86595701dbf7bbb0b6f98ce synchronised icon and text transitions on GPU
  - text and icons are now synchronised by triggering both using d3 after layout finishes
  - both transitions moved to CSS and run on the GPU instead
- 28a3bb94cbf2d9b45c56cccab49399c6140ef415 removed text shadows on labels
  - no noticeable difference or change in readability as these were extremely subtle
- 5d7b2a78251130dc84d754d42c733fa32d5951c8 limit edge path string precision
  - generated edges now use a single decimal point precision to reduce SVG parsing time and export size

The following techniques were tested in various places but did not appear to give any noticeable improvement for these purposes (which is probably a good sign):

- `clip`
- `will-change`
- `position: absolute`
- `transform: translate(0, 0)`
- `-webkit-font-smoothing: none`

A further technique that improves the performance in cases using opacity is also possible, but this will come as a separate PR as it requires some additional refactoring.

## QA notes

The random generator settings can be tweaked to create a large graph for testing:

```js
const MAX_EDGE_COUNT = 800;
const MIN_EDGE_COUNT = 800;
```

After that change, visit this seeded example:

http://localhost:4141/?data=random&seed=qioxrqxCABRCLcTe0Eqscl4v1zKNpw

First test on the main branch and note the FPS of pan and zoom, then switch to this branch and compare.

Smaller graphs will not show much improvement, since they already tend to reach 60 FPS.

Most browsers see some improvement, with Chrome showing the most significant boost.

Most interactions that trigger a zoom or a pan should be tested, in multiple browsers.

## Checklist

## Legal notice


src/components/flowchart/styles/_node.scss
```diff
@@ -4,15 +4,13 @@
...
     font-size: 1.4em;
-    transition: opacity ease 0.3s 0.2s;
     pointer-events: none;
 
     .kui-theme--light & {
       fill: $node-labeltext-fill-light;
     }
 
     .kui-theme--dark & {
-      text-shadow: 0 0 6px black;
```
richardwestenra, line 15: I am a little disappointed to remove this because I did think it helped legibility, but if it hurt performance then I think this is a good call

richardwestenra (approved):
Outstanding work! 👏

PR #293, opened 2020-10-30
[KED-2180] retrieve node metadata asynchronously from different endpoints

studioswong:
## Description

This PR is created to allow retrieving node metadata asynchronously from different node endpoints to be consumed by the FE for the node metadata panel.

Please note there are currently tests failing which is subject to further modification, just hoping to set this PR up to get more comments and eyes on the set up given this does impact quite a bit on governing the naming of the node metadata panel.

## Development notes

The following have been added / modified:

1. Addition of new fields within the pipeline state to store node metadata information (`code`, `codeLocation`, `docString`, `parameters`, `dataset_location` and `dataset_type`. Currently the naming of those fields are adopted exactly as the fields returned by the API, which could be different from the naming of the field to be ultimately displayed on the FE ( this will subject to discussions between the design team and the BE team when aligning the information to be displayed on the metdata panel.)
2. Addition of the `loadNodeData` action to fetch information from the API. **please note this action will replace the `toggleClickedNode` action as node data will be fetched on clicking a node. **
3. Setup of test suite for nodes actions and all relevant tests for `loadNodeData`
4. Replacing the usage of `toggleClickedNode` to `loadNodeData` within flowchart.js and other parts of the app 
5. Addition of reducers to process and store the fetched node data into the state 
6. Tests to test the newly added reducer function
7. modification of `getUrl` within utils to construct url for nodes api endpoint
8. relevant tests for `getURL` in utils

## QA notes

This PR is best to be QAed alongside with development of the node metadata panel later on. 

## Checklist

## Legal notice


src/components/flowchart/index.js
```diff
@@ -391,7 +391,7 @@ export class FlowChart extends Component {
     const ENTER = 13;
     const ESCAPE = 27;
     if (event.keyCode === ENTER) {
-      this.handleNodeClick(node);
+      this.onLoadNodeData(node);
```
richardwestenra, line 394: I don't think this line should have been renamed - `onToggleNodeClicked` is different from `handleNodeClick`. This will cause issues and is ~likely~ breaking the keydown behaviour.
image

richardwestenra: We should add some tests to ensure that this issue does not occur again in future

src/components/flowchart/index.js
```diff
@@ -354,15 +354,15 @@ export class FlowChart extends Component {
    * @param {Object} node Datum for a single node
    */
   handleNodeClick = node => {
-    this.props.onToggleNodeClicked(node.id);
+    this.props.loadNodeData(node.id);
```
richardwestenra, line 357: This should be `this.props.onLoadNodeData(node.id);`. It currently throws an error.
image

src/actions/nodes.js
```diff
@@ -38,3 +41,52 @@ export function toggleNodeHovered(nodeHovered) {
...
+ * @return {function} A promise that resolves when the data is loaded
+ */
+export function loadNodeData(nodeID) {
+  return async function(dispatch, getState) {
+    const { asyncDataSource, node } = getState();
+
+    dispatch(toggleNodeClicked(nodeID));
+
+    if (asyncDataSource) {
+      if (!node.fetched[nodeID]) {
```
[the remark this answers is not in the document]
richardwestenra: This function also doesn't handle cases where `nodeID` is undefined, and will throw errors when closing the metadata panel and toggling `nodeClicked` to `null`. This should have been caught during testing, tbh.
I'd change it to something like the following:
```
if (asyncDataSource && nodeID && !node.fetched[nodeID]) {
```

src/selectors/loading.js
```diff
@@ -3,13 +3,14 @@ import { createSelector } from 'reselect';
...
 /**
  * Determine whether to show the loading spinner
  */
 export const isLoading = createSelector(
-  [getGraphLoading, getPipelineLoading, getFontLoading],
-  (graphLoading, pipelineLoading, fontLoading) => {
-    return graphLoading || pipelineLoading || fontLoading;
+  [getGraphLoading, getPipelineLoading, getFontLoading, getNodeLoading],
+  (graphLoading, pipelineLoading, fontLoading, nodeLoading) => {
+    return graphLoading || pipelineLoading || fontLoading || nodeLoading;
```
richardwestenra, line 14: Good! 👍

src/reducers/nodes.js
```diff
@@ -43,6 +44,34 @@ function nodeReducer(nodeState = {}, action) {
       });
     }
 
+    case ADD_NODE_METADATA: {
+      const {id, data} = action.data;
+      console.log('id', id)
```
richardwestenra, line 49: leftover console.log

src/actions/nodes.js
```diff
@@ -38,3 +41,52 @@ export function toggleNodeHovered(nodeHovered) {
...
+        dispatch(toggleNodeDataLoading(true));
+        const url = getUrl('nodes', nodeID);
+        const nodeData = await loadJsonData(url);
+        dispatch(addNodeMetadata({ id: nodeID, data: nodeData }));
```
richardwestenra, line 87: I don't think the nesting here is necessary. Also I would be inclined to have it pass `nodeID` instead of `id` - it's mostly a convention used across a lot of the app, and makes the reducer code a bit more clear. Hence you could do something like this:
```es6
dispatch(addNodeMetadata({ ...nodeData, nodeID }));
```

src/store/__mocks__/load-data.js
```diff
@@ -37,6 +38,11 @@ const loadJsonData = async (path = getUrl('main')) => {
     return mockAPIFeatureSupport(animals);
   }
 
+  // Use nodes_task dataset in place of 'main' endpoint
+  if (path.includes('nodes')) {
+    return mockAPIFeatureSupport(node_task);
+  }
+
```
richardwestenra, line 45: Great :)

It might be worth seeing if you can find a way to let it return any one of the three mock node data types, depending on the URL that's passed to it?

richardwestenra: Also, I don't think this one needs to be wrapped in `mockAPIFeatureSupport`, as that won't be used for this endpoint, it's just for pipeline data.

src/store/normalize-data.test.js
```diff
@@ -1,4 +1,6 @@
-import normalizeData, { createInitialPipelineState } from './normalize-data';
+import normalizeData, {
+  createInitialPipelineState,
+} from './normalize-data';
```
richardwestenra, line 3: Why has this been changed? That's odd 🤔

richardwestenra (changes requested):
Great work! However I noticed a few bugs that urgently need fixing before we can get this merged. I recommend giving this a solid browser test with async data loading

yetudada (approved):
Approved on my end. The mock data uses the Iris dataset example.

src/reducers/nodes.js
studioswong: aargh oops my bad, thanks for the great spot!

src/store/normalize-data.test.js
studioswong: yea it's a leftover styling change from importing another function ( which was later deleted ) - will revert this back to the old style:)

src/actions/nodes.js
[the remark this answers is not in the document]
studioswong: sounds good! I will add in a new test to capture that condition too!

src/actions/nodes.test.js
```diff
@@ -0,0 +1,84 @@
...
+        const store = createStore(reducer, mockState.json);
+        const { dispatch, getState, subscribe } = store;
+        const node = { id: 'f1f1425b' };
+        const storeListener = jest.fn();
+
+        subscribe(storeListener);
+        await loadNodeData(node.id)(dispatch, getState);
+
+        // we have set it at 5 times given the set of calls for the first listener, but this test is subject to change to a simplier test once we have set up a 
+        // way to directly set the 'fetched' value of the store
```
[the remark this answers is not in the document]
studioswong: that's a great question @richardwestenra ! It's expected for the store to have been called 5 times due to the current setup of the `loadNodeData` action - each call of `loadNodeData` will trigger 4 dispatch actions, which in the case of this test, the 4 actions are triggered on the first call of `loadNodeData` in this test to fetch the nodeData.

The second call to `loadNodeData` will still trigger the `toggleNodeClicked` action ( since the check for `asyncDataSource` occurs after `toggleNodeClicked`), making the 5th call to the store. There are no further calls to the store to fetch data from the API given the data is already fetched in the first round. 

I do think the set up of this test to call 5 times is quite confusing for anyone coming into this test without prior knowledge of the set up of `loadNodeData`, I will update with better notes to explain about the set up of the test :D

src/components/flowchart/index.js
studioswong: great spot, thank you!

src/components/flowchart/index.js
studioswong: yes indeed, great spot - we should definitely look into adding some tests for `handleNodeKeyDown` to ensure it won't break.

src/components/flowchart/flowchart.test.js
```diff
@@ -180,28 +180,22 @@ describe('FlowChart', () => {
   it('maps dispatch to props', () => {
     const dispatch = jest.fn();
 
-    mapDispatchToProps(dispatch).onToggleNodeClicked('123');
-    expect(dispatch.mock.calls[0][0]).toEqual({
-      nodeClicked: '123',
-      type: 'TOGGLE_NODE_CLICKED'
-    });
-
```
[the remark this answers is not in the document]
studioswong: sounds good Richard! I've added a new test specifically to test the async `onLoadNodeData` action - unfortunately with this test, I cannot exactly use the same setup as the `nodes` action test that we used when we paired given the action dispatched in this instance is another async action ( for the nodes action test suite, we are testing the B->C scenario while the setup here is testing the A->C scenario) which requires a middleware. I've set up the test with `redux-mock-store` and `redux-thunk` to resolve this issue to extract the action for the test, which you can see from my latest commit :)

src/actions/nodes.js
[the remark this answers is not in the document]
studioswong: added a new test to capture the test case in my latest commit

richardwestenra (approved):
Approved, but I've suggested a change

PR #295, opened 2020-11-02
[KED-2166] Add additional fields to meta data panel

bru5:
## Description

Adds additional fields to the metadata panel:

For parameter type nodes:
- file path
- parameters

For data type nodes:
- dataset type
- file path

For task type nodes:
- file path
- parameters (data appears when API finished)
- docstring

## Development notes

Parameters for function nodes will appear automatically once the backend API update is completed.

## QA notes

Check the above fields are correct for each type of node.

Test a real project through the proxy e.g. [spaceflights](https://github.com/quantumblacklabs/kedro-starter-spaceflights), `pip install` your local viz repo, then run `kedro viz --port 4142` alongside `npm run start` without any `api/main` file.

Any fields with missing or empty values will show as `-` (e.g. actually no value or not yet returned by API).

## Checklist

## Legal notice

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "lib"
   ],
   "homepage": ".",
+  "proxy": "http://localhost:4142/",
   "scripts": {
     "build": "npm run build:css && react-scripts build",
     "build:css": "node-sass src/ -o src/",
--- a/src/components/metadata/index.js
+++ b/src/components/metadata/index.js
@@ -21,6 +21,7 @@ const MetaData = ({ visible = true, metadata, onToggleNodeSelected }) => {
   const [showCopied, setShowCopied] = useState(false);
 
   const isTaskNode = metadata?.node.type === 'task';
+  const isDataNode = metadata?.node.type === 'data';
 
   const onCopyClick = () => {
     window.navigator.clipboard.writeText(metadata.runCommand);
@@ -55,6 +56,18 @@ const MetaData = ({ visible = true, metadata, onToggleNodeSelected }) => {
       </div>
       <dl className="pipeline-metadata__list">
         <MetaDataRow label="Type:" value={metadata.node.type} />
+        <MetaDataRow
+          label="File Path:"
+          visible={isDataNode}
+          value={metadata.filePath}
+        />
+        <MetaDataRow
+          label={`Parameters (${metadata.parameters.length}):`}
+          visible={isTaskNode}
+          commas={false}
+          inline={false}
+          value={metadata.parameters}
+        />
         <MetaDataRow
           label="Inputs:"
           property="name"
@@ -109,6 +122,12 @@ const MetaData = ({ visible = true, metadata, onToggleNodeSelected }) => {
             )}
           </div>
         </MetaDataRow>
+        <MetaDataRow
+          label="Description (docstring)"
+          visible={isTaskNode}
+          value={metadata.description}
+        />
+        <MetaDataRow label="Location of code in Kedro:" value={metadata.path} />
       </dl>
     </div>
   );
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
     "lib"
   ],
   "homepage": ".",
+  "proxy": "http://localhost:4142/",
```
studioswong, line 9: I'm confused.... isn't this already merged in main? (https://github.com/quantumblacklabs/kedro-viz/pull/318)

studioswong:
Thanks for the work @bru5 , good stuff. Please add some tests for the new node metadata fields that you have introduced for the `metadataRow` and `metadata` component to ensure they get displayed properly?

package.json
bru5: Yeah it is now, I just needed it here before it was merged a while back. Should just merge on top no problem I think.

bru5:
For sure, I've expanded out the tests for each type now. The tests check the fields are shown, though some of them only return a blank value as the test data set needs updating. 

That part is not straightforward though because of the new additional API calls not yet existing in the test data approach, so I think we might need to check with @richardwestenra on how best to do that. 

Not ideal but don't think it needs to block this PR though, see what you think.

src/components/metadata/index.js
```diff
@@ -55,6 +57,20 @@ const MetaData = ({ visible = true, metadata, onToggleNodeSelected }) => {
...
+          value={metadata.datasetType}
+        />
+        <MetaDataRow label="File Path:" kind="path" value={metadata.filepath} />
+        <MetaDataRow
+          label={`Parameters (${metadata.parameters?.length || '-'}):`}
+          visible={isParametersNode || isTaskNode}
+          commas={false}
+          inline={false}
+          value={metadata.parameters}
+        />
```
richardwestenra, line 73: Design question: Are we better off showing these fields regardless of whether they're populated, or only showing them if populated?
image

src/components/metadata/index.js
```diff
@@ -109,6 +125,11 @@ const MetaData = ({ visible = true, metadata, onToggleNodeSelected }) => {
             )}
           </div>
         </MetaDataRow>
+        <MetaDataRow
+          label="Description (docstring):"
+          visible={isTaskNode}
+          value={metadata.docstring}
+        />
```
richardwestenra, line 132: image

Additional question: If we are going to always show these fields regardless of whether they're populated, as long as the node is the right corresponding type, then can we we please add dummy data to the demo dataset as well? We want the demo to always look as good as possible for obvious reasons.

Also, potentially the animals one as well? What do you think?

This can be part of a separate PR if you prefer.

richardwestenra (approved):

src/components/metadata/index.js
bru5: Yeah it's a good point and I did this intentionally, as I think it's important to show empty values in technical use cases like this. We can discuss and improve it later though if needed I think.

src/components/metadata/index.js
bru5: Yeah for sure the included data sets should be updated with the new fields. I was planning on getting your input on how to do that, as now we have two API calls, I wasn't sure how best to implement that. I think we should do it in a separate PR though from the functionality.

src/components/metadata/index.js
richardwestenra: Sweet as! cc @GabrielComymQB let's come back to this in 2021

src/components/metadata/index.js
richardwestenra: ok cool. yeah I think for the demo dataset we can definitely inline it. Once we add the code snippets it'll get trickier and we may need to revisit this whole data loading thing from the ground up

PR #297, opened 2020-11-05
[KED-2253] Fix pipeline selector flicker when switching pipelines

richardwestenra:
## Description

This PR resolves an issue that arose as of #296, which introduced a loading flicker when switching pipeline with async-loaded data, because the pipeline would momentarily disappear then reappear:

pipeline-shift

To address this issue, I've implemented the following changes:

1. Don't hide the pipeline selector when loading asynchronously. Now, if loading async, the selector will be visible at all times, even when empty.
2. Change the default pipeline selector text to 'Default'. This makes the loading effect more pleasant.
3. Hide the graph (via opacity) when switching pipeline. Now, instead of resetting the entire pipeline, a new action toggles the opacity on the graph container, so that we don't need to rerender everything in the meantime. This should be less disruptive and more performant, but it required adding a new action.
4. Set the pipeline selector to 'disabled' when there are no pipelines in the dataset.

## Development notes

The only icky thing is that the action to hide the graph needs to be toggled on in the pipeline loading action, and toggled off in the graph calculation selector, otherwise it will be shown before it's ready. However I don't foresee this being a problem.

The disabled state for the pipeline selector is a superficial change that only applies to mouse, not keyboard. I'll need to fix this properly in a subsequent update to Kedro-UI.

## QA notes

Set to async loading mode (i.e. data=json), and toggle between pipelines. The selector should stay put, and the graph should hide while the new pipeline loads.

## Checklist

## Legal notice

### DIFF
--- a/src/components/pipeline-list/index.js
+++ b/src/components/pipeline-list/index.js
@@ -13,12 +13,13 @@ import './pipeline-list.css';
  * @param {string} theme Kedro UI light/dark theme
  */
 export const PipelineList = ({
+  asyncDataSource,
   onUpdateActivePipeline,
   pipeline,
   theme,
   onToggleOpen
 }) => {
-  if (!pipeline.ids.length) {
+  if (!pipeline.ids.length && !asyncDataSource) {
     return null;
   }
   return (
@@ -29,7 +30,9 @@ export const PipelineList = ({
         theme={theme}
         width={null}
         onChanged={onUpdateActivePipeline}
-        defaultText={pipeline.name[pipeline.active]}>
+        defaultText={
+          pipeline.name[pipeline.active] || 'No pipelines available'
+        }>
         {pipeline.ids.map(id => (
           <MenuOption
             key={`pipeline-${id}`}
@@ -43,6 +46,7 @@ export const PipelineList = ({
 };
 
 export const mapStateToProps = state => ({
+  asyncDataSource: state.asyncDataSource,
   pipeline: state.pipeline,
   theme: state.theme
 });
--- a/src/components/pipeline-list/index.js
+++ b/src/components/pipeline-list/index.js
@@ -13,13 +13,12 @@ import './pipeline-list.css';
  * @param {string} theme Kedro UI light/dark theme
  */
 export const PipelineList = ({
-  asyncDataSource,
   onUpdateActivePipeline,
   pipeline,
   theme,
   onToggleOpen
 }) => {
-  if (!pipeline.ids.length && !asyncDataSource) {
+  if (!pipeline.ids.length) {
     return null;
   }
   return (
@@ -44,7 +43,6 @@ export const PipelineList = ({
 };
 
 export const mapStateToProps = state => ({
-  asyncDataSource: state.asyncDataSource,
   pipeline: state.pipeline,
   theme: state.theme
 });
--- a/src/components/pipeline-list/index.js
+++ b/src/components/pipeline-list/index.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import classnames from 'classnames';
 import { connect } from 'react-redux';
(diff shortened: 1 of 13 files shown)


src/components/pipeline-list/index.js
```diff
@@ -29,7 +30,7 @@ export const PipelineList = ({
         theme={theme}
         width={null}
         onChanged={onUpdateActivePipeline}
-        defaultText={pipeline.name[pipeline.active]}>
+        defaultText={pipeline.name[pipeline.active] || ''}>
```
bru5, line 33: While this seems to solve the issue, I don't think just an empty label works here at all, as it appears broken and functions weirdly when you click on it.

My suggestion is when there's a single pipeline is to always show `Default` or similar, with a corresponding `MenuOption` too for consistency (@GabrielComymQB is probably best to confirm)

I think that is easily achieved in here as a default label (i.e. placeholder) option in this component, rather than needing any further business logic.

bru5 (changes requested):

src/components/pipeline-list/index.js
richardwestenra: 1. Why do you say it appears broken/functions weirdly? I mean, granted, if you click on it then it will fade the main sidebar bg without displaying any results, but that can't happen anyway unless the results don't load or there aren't any pipelines in the dataset, which Yetu assures us can't happen. But I don't think it looks weird, I think it looks like an empty visual placeholder.
2. Due to the second set of changes, here, you will only ever see the blank pipeline selector for less than half a second on first page-load, while the JSON file is being loaded and parsed. It is not shown when switching pipeline - I've changed it so it still shows the old dataset in the sidebar when you're switching pipeline.
3. I oppose using 'Default' as the placeholder, because if you are running a different pipeline as main, OR if you have a different active pipeline saved in localstorage (e.g. because you just refreshed the page), then on load it will flash 'Default' briefly then switch to your actual pipeline name.

This is how it looks on page-load:
onload-pipeline-selector

bru5: Ah well here's what I'm seeing when I load a pipeline JSON file with only one pipeline and it stays this way permanently:

pipeline-selector-blank

So #1 seems to happen at least in this case? Or do the new API changes mean this won't happen in practice (what will it say in that case)?

> I don't think it looks weird, I think it looks like an empty visual placeholder.

It's not really a placeholder if it's empty though surely? A placeholder is meant to fill in gaps like these?

> I oppose using 'Default' as the placeholder, because if you are running a different pipeline as main

Could it just say `Main` then?

If you really think blank is the right thing here, then I'd say also hide the down arrow icon and disable and gray out the field so it can't be interactive?

richardwestenra: ok do you mean one that does not have any pipelines in the dataset? Because if it has only one it should look like this (when open):
image
Whereas I'm able to replicate your case if I remove all pipelines from the dataset.

I changed it to always show whenever data is being loaded asynchronously, only because, as Yetu said on slack, due to recent API changes, it should never be the case that the JSON data doesn't include the pipelines field with at least one pipeline in it, which is why I changed this setting. If we want to support async datasets with 0 pipelines then we should revert this change. Personally, that's what I would recommend, and it's the way I had it set up before, because it is less likely to look 'broken'.

So to summarise:
I misunderstood what you were talking about before - I thought you were just talking about the brief flash on load, where it functions as a placeholder. It seems like you're talking about a zero-pipeline async dataset, which is a case that I was previously was supporting, but was instructed to stop supporting.

I am going to revert this change in the interests of prioritising platform stability across incomplete datasets, over smooth page-loads.

bru5 (approved):

bru5:
Thanks for addressing the layout issues, sorry for any confusion I didn't realise there were multiple cases of this with different causes earlier on, looks good to merge now I think!

PR #298, opened 2020-11-05
[KED- 2185] Unify backend and frontend test data - fix FE tests

studioswong:
## Description

This PR mainly aims to fix the failing FE tests that was caused by the newly proposed test data file.

## Development notes

The following has been introduced:

1. A new test data file generated from `Kedro-animal` test repo. I had also introduced minor changes to ensure all names of the nodes follows the casing for the `name` field ( uppercase for the first letter) across all types of nodes. 
2. Changes to various FE tests to match the new dataset. I try to stick to the original intent of the test and only update the tests according to the new setup of the data. 

## QA notes

The tests should test and pass as they should

## Checklist

## Legal notice


studioswong:
Currently the python tests are failing which is likely due to the minor changes (changing the casing of the `name` field for some nodes) introduced to the data file  - since this PR is done against @921kiyo 's feature branch, the failing python tests could be fixed in one go once it's merged in? cc. @921kiyo

src/components/node-list/node-list-items.test.js
```diff
@@ -50,7 +50,7 @@ describe('node-list-selectors', () => {
 
     it('returns expected number of tags', () => {
       expect(filteredTags.length).not.toBe(tags.length);
-      expect(filteredTags).toHaveLength(2);
+      expect(filteredTags).toHaveLength(1);
```
bru5, line 53: Could we use a search with 2 or more results ideally?

src/components/node-list/node-list-items.test.js
```diff
@@ -97,7 +97,7 @@ describe('node-list-selectors', () => {
 
     it('filters expected number of items', () => {
       expect(filteredTagItems.length).not.toBe(tags.length);
-      expect(filteredTagItems).toHaveLength(2);
+      expect(filteredTagItems).toHaveLength(1);
```
bru5, line 100: Could we use a search with 2 or more results ideally?

src/components/node-list/node-list.test.js
```diff
@@ -204,24 +204,38 @@ describe('NodeList', () => {
     it('selecting tags enables only elements with given tags', () => {
       const wrapper = setup.mount(<NodeList />);
 
-      changeRows(wrapper, ['huge'], true);
-      expect(elementsEnabled(wrapper)).toEqual([['whale', true]]);
+      changeRows(wrapper, ['Small'], true);
+      expect(elementsEnabled(wrapper)).toEqual([
+        ['salmon', true],
```
bru5, line 209: Where these results (and all the others) manually verified to be correct e.g. using the browser?

src/selectors/layers.test.js
```diff
@@ -35,7 +35,7 @@ describe('Selectors', () => {
 
       nodes.forEach(node => {
         // we don't need to check y/height positions if the layer isn't there.
-        if (node.layer === null) {
+        if (!node.layer || node.layer === null) {
```
bru5, line 38: If this is due to `undefined` then check for that instead, since this will fail on valid but falsy values e.g. `0`

bru5 (changes requested):

richardwestenra (commented):
Great work 👍

src/components/node-list/node-list.test.js
studioswong: good point Liam - will check these specific cases on the browser again and flag them if I spot any edge cases.

src/selectors/layers.test.js
studioswong: I agree with you on the check, unfortunately this isn't due to `node.layer` being `undefined` but rather the field being non-existent in the actual data ( this is part of the discussion with kiyo where the `dataset` node would not return any data as we discussed on slack), which checking for `undefined` in this case would not capture that the `layer field` non-existent scenerio ( I had also tried modifying the test and it doesn't work.) 

that said, I had also added in the check for `undefined` to ensure a more robust test.

src/components/node-list/node-list-items.test.js
```diff
@@ -24,7 +24,7 @@ const ungroupNodes = groupedNodes =>
 describe('node-list-selectors', () => {
   describe('getFilteredNodes', () => {
     const nodes = getGroupedNodes(mockState.animals);
-    let searchValue = 'bear';
+    let searchValue = 'Bear';
```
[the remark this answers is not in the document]
studioswong: It is currently capitalised to match the format of the actual data returned - currently `node.name` for `data` type nodes is formatted to start in caps ( however this only applies to `data` and `parameter` type nodes and not for `task` type nodes - don't ask me why, I have asked Kiyo about this and tried to modify the `animals` data file instead but this file has to match the actual data that was returned from the BE). 

This is one very good example to show the small discrepancies of the expectations of the dataset format - it might br worth to also raise the question to the BE team with regards to this inconsistent casing of the node names.

src/components/node-list/node-list-items.test.js
studioswong: This is set up to match the results from the searchvalue that was set up in the test - I had modified the search term to generate more filtered search results.

studioswong:
Currently the e2e tests are failing on the CI which has been flagged to the core team - will rerun the tests again once the issue has been resolved

src/components/node-list/node-list-items.test.js
studioswong: updated the search value to generate more results for this test!

src/components/node-list/node-list-items.test.js
```diff
@@ -97,17 +97,28 @@ describe('node-list-selectors', () => {
...
-        expect(tagItem.name).toEqual(tags[index].name);
-        expect(tagItem.id).toEqual(tags[index].id);
+    it('returns the filtered items that contains the search value', () => {
+      filteredTagItems.forEach(tagItem => {
+        expect(tagItem.name).toContain(searchValue);
+        expect(tagItem.id).toContain(searchValue);
+      });
+    });
+
+    it('filtered tagItems names and ids should match that of original tag', () => {
```
bru5, line 114: What's the need to test this, do we think there's a chance they could get changed?
Also I thought previously we were testing the output of a specific input search, that seems missing now?

src/selectors/layers.test.js
```diff
@@ -35,7 +35,11 @@ describe('Selectors', () => {
 
       nodes.forEach(node => {
         // we don't need to check y/height positions if the layer isn't there.
-        if (!node.layer || node.layer === null) {
+        if (
+          !node.layer ||
```
bru5, line 39: If this includes `!node.layer` then `... || node.layer === null || typeof node.layer === 'undefined'` is not necessary? 

It should be only one of the above expressions - I'm inclined to say don't use `!node.layer` here at all, so that `0` is allowable as a layer id (I hope this is the case elsewhere in the code, let me know if not).

Also note that when checking typeof for undefined you must always quote it 
as in `typeof x === 'undefined'`.

bru5 (changes requested):

921kiyo:
Shouldn't we merge this agains https://github.com/quantumblacklabs/kedro-viz/pull/291, not against `main` ? So that we know that we have unified the front-end and back-end tests, and CI will tell us whether it is successful or not.

studioswong:
Thanks for your comment @921kiyo  - I had switched it to merge into `main` other than the other feature branch more so for clarify purposes mainly because github isn't picking up the right diffs even after merging the latest main into your feature branch. ( this is a strange bug on github that has happened to me on other instances, and usually when you switch between different based branches it will solve itself)

This change is a temporary change for now to provide clarity for whoever that is reviewing this PR.

921kiyo:
Aw, okay, I can see the diff in `package/tests/test_server.py` in this PR, so if CI is green in this PR, we have successfully unified the test data between front-end and back-end then :) I've just closed mine.

921kiyo (approved):
Approving for the back-end change :)

src/selectors/layers.test.js
studioswong: sure I agree - we can take our the `node.layer === null || typeof node.layer === 'undefined'` cases since `!node.layer` does cover them all

src/components/node-list/node-list-items.test.js
studioswong: Here is a bit of history behind the changes in this test and further updates I am making after discussions with @bru5 on slack:
1. I had removed the test (`'returns items for each tag'`) - the test uses the `index` of each run to reference the items in the tags array for comparison with the relative `tagItem` of that run, which does not seem to be testing anything. On discussion with @bru5 , we decided that this is a bunk test which could be deleted, while further adding modifications to allow a more comprehensive set of tests for testing the filtered tagItems.
2. I have added more tests on top of existing `expect` statements under the `'filters expected number of items'` test to ensure the right tags ( both 'small' and 'medium') gets returned.
3. further tests to ensure the filtered item contains the search term.

src/selectors/layers.test.js
```diff
@@ -34,8 +34,8 @@ describe('Selectors', () => {
       }, {});
 
       nodes.forEach(node => {
-        // we don't need to check y/height positions if the layer isn't there.
-        if (node.layer === null) {
+        // Wwe don't need to check y/height positions if the layer isn't there or if the 'layer' field does not exists ( this is the case for 'task' type nodes ).
+        if (!node.layer) {
```
bru5, line 38: This comment change seems to just repeat what it already said? Is that needed?

Also the test `if (!node.layer)` still fails for the case of `node.layer = 0`.

For that reason I'd prefer `if (node.layer === null || typeof node.layer === 'undefined')` in cases like this.

bru5 (changes requested):

src/selectors/layers.test.js
bru5: That's fine, since non-existent in this case will return `undefined`.

src/selectors/layers.test.js
studioswong: thanks for the comments!  I will modify the comment change.

In terms of the check, currently on just using `(node.layer === null || typeof node.layer === 'undefined')` as a test would cause this test to fail given that `node.layer` field does not exist for some nodes - it is not simply testing for the case where `node.layer` as either assigned as `null` or `undefined` as previously set up in the animals dataset. And as we mentioned above, since `!node.layer` covers all negative case, hence we would not need to add in any extra test to specifically test for the `null` and `undefined` cases.

While I understand your concern for the `node.layer = 0` case, since this check is only within a test case where we are testing against a specfici dataset with no cases of `node.layer = 0`, hence there isn't any particular need to cater for this case.

studioswong: Thanks @bru5 - I've changed the comment and modified the check. ( the current check is only used within a test against a specific dataset without any cases of `node.layer = 0` hence there isn't a particular need to cater for this case, but nonetheless I have changed the check to your suggested format just in case.)

bru5 (approved):

src/utils/data/animals.mock.json
```diff
@@ -1,250 +1,232 @@
...
+      "name": "Default"
     },
     {
-      "target": "2cd4ba93",
-      "source": "4f90af66"
-    },
+      "id": "empty",
+      "name": "Empty"
+    }
+  ],
```
richardwestenra, line 216: @921kiyo Is there any way that we can keep the full pipeline display names here? e.g. `De` -> `Data engineering`, as it was?

richardwestenra (approved):
Approved but with a comment for @921kiyo

src/utils/data/animals.mock.json
921kiyo: @richardwestenra Unfortunately, it doesn't happen at the moment. `name` value is a prettified version of `id`. So `id` would have to be `data engineering` if we wanted `Data engineering` for `name`

richardwestenra: gotcha. ok no biggie! thanks for the quick explanation :)

PR #301, opened 2020-11-11
[KED-2134] Use CSS custom property for theme colours

richardwestenra:
## Description

Using CSS custom properties instead of Sass variables + parent classes will allow us to define our colour variables once, and decrease the specificity in our CSS declarations.

The only real downside of this is that it might slightly reduce browser support, but I don't think it will affect any browsers that we don't already support.

## Development notes

I've included the site-wide definitions in app.scss, so that they're only included once (and not repeated every time that _variables.scss is imported), and so they're included in both lib and app styles. For any CSS variables that only affect a single component, I've defined them in that component, so that if the component is deleted/changed then they'll be changed/updated also.

I've changed a `@mixin` to an `@extend`, and created a new mixin for defining transparent versions of theme colours, [which is required for CSS gradients to transparent](https://css-tricks.com/thing-know-gradients-transparent-black/).

## QA notes

Hopefully nothing has changed. The most likely thing to break is styles being overridden by other styles across the two themes, as the specificity has been changed. The main bit that I struggled with was the node-list styles, which are pretty complicated, so I might have missed some things.

## Checklist

## Legal notice


src/components/minimap-toolbar/minimap-toolbar.scss
```diff
@@ -1,6 +1,14 @@
 @import '../../styles/variables';
 @import '../../styles/extends';
 
+.kui-theme--light {
+  --minimap-toolbar-active: $color-bg-light-3;
```
bru5, line 5: Will this be needed for each component with custom colours?

src/components/node-list/styles/_filter.scss
```diff
@@ -68,12 +55,11 @@
   .pipeline-nodelist__row:hover
```
bru5, line 55: The background highlight when hovering on sidebar rows seems to be missing on both themes now, might be an issue?

src/components/node-list/styles/_filter.scss
```diff
@@ -3,29 +3,16 @@
 @import '../../../styles/variables';
 @import './variables';
 
-.kui-theme--dark .pipeline-row__toggle-icon--kind-filter {
+.pipeline-row__toggle-icon--kind-filter {
   &.pipeline-row__toggle-icon--checked {
```
bru5, line 7: The tag row indicators are showing up as black when active on both themes now, instead of blue as in the design.

src/components/metadata/styles/metadata.scss
```diff
@@ -1,8 +1,18 @@
...
+}
+
+.kui-theme--dark {
+  --color-metadata-bg: #{$color-bg-dark-5};
+  --color-metadata-kind-token-bg: #{$color-bg-dark-6};
+}
 
 .pipeline-metadata {
-  @include sidebarBase;
+  @extend %sidebar;
```
bru5, line 15: The metadata panel font colour is now a dimmer than before, such that it appears different than the font colour of the main sidebar. Seems to be in both themes, more noticeable in the dark theme. I think these should both be the same colour (i.e. brighter font on metadata panel for dark theme, and a lighter font on dark theme)?

src/components/minimap-toolbar/minimap-toolbar.scss
richardwestenra: yes I tried to put them in each specific component so that if that component were deleted then the custom property would be deleted also.

but also AHHH i forgot to escape the variable! fixed.

src/components/app/app.scss
```diff
@@ -1,28 +1,48 @@
 @import '../../styles/variables';
 
+//-- Theme colours as CSS Custom Properties --//
+
+.kui-theme--light {
+  --color-default: #{$color-light};
```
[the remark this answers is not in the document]
richardwestenra: idk, I guess it could, but I don't think it'd be simpler to read or understand. My inclination is not to over-optimise here.

src/components/flowchart/styles/_variables.scss
```diff
@@ -1,67 +1,50 @@
...
+    @include export;
+  }
+
+  @media print {
+    @include export;
+  }
+}
+
+.kui-theme--dark {
+  --node-icon-fill: #{$color-text-dark};
```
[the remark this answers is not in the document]
richardwestenra: This is a pretty standard Sass thing ([discussed in the Sass docs](https://sass-lang.com/documentation/style-rules/declarations#custom-properties)), because Custom Properties can have any value, so Sass requires you to escape anything that is not intended to be treated as a regular string. I could comment it but idk if necessary?

src/components/icon-button/icon-button.scss
```diff
@@ -76,17 +76,9 @@
     left: -$triangle-size + 0.5;
     width: 0;
     height: 0;
-    border-color: transparent;
+    border-color: transparent var(--color-bg-alt) transparent transparent;
```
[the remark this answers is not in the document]
richardwestenra: Perhaps but I just figured I'd make it more succinct and put it on one line? honestly not too fussed:
```css
border-color: transparent;
border-right-color: var(--color-bg-alt);
// or
border-color: transparent var(--color-bg-alt) transparent transparent;
```

src/components/node-list/styles/_filter.scss
richardwestenra: Yeah I think that was due to a last-minute change where I forgot to escape the custom property variables. Fixed!

src/components/node-list/styles/_filter.scss
richardwestenra: Looks like I used the wrong syntax in _row.scss on lines 11 & 17. clearly i was tired when i did this last batch 🙈. fixed.

src/components/metadata/styles/metadata.scss
richardwestenra: Good spot. Fixed

src/components/flowchart/styles/_variables.scss
```diff
@@ -1,67 +1,50 @@
...
-
-$edge-dark: #666;
-$edge-arrowhead-dark: #cacaca;
-
-// Layer
-
-$layer-dark: mix($color-bg-dark-1, $color-bg-dark-2);
+.kui-theme--light {
+  --node-icon-fill: #{$color-text-light};
+  --node-fill-default: #f0f1f3;
```
studioswong, line 9: In following the pattern of referring to colors using variables, this color should be referred by the variable`$color-bg-light-3`?

src/components/app/app.scss
```diff
@@ -1,28 +1,48 @@
...
+  --color-default-alt: #{$color-dark};
+  --color-text: #{$color-text-light};
+  --color-text-alt: #{$color-text-light-alt};
+  --color-bg-1: #{$color-bg-light-1};
+  --color-bg-2: #{$color-bg-light-2};
+  --color-bg-3: #{$color-bg-light-3};
+  --color-bg-4: #{$color-bg-light-4};
+  --color-bg-alt: #{$color-bg-light-alt};
+  --color-sidebar-background: #{$color-bg-light-3};
+  --color-sidebar-border: rgba(0, 0, 0, 0.05);
```
studioswong, line 16: any reason this color is declared here instead of as a variable like the other colors above?

src/styles/_extends.scss
```diff
@@ -75,15 +75,13 @@
...
-  }
-
-  .kui-theme--dark & {
-    color: $color-text-dark-alt;
-    background: $color-bg-dark-alt;
-    box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1);
-  }
+  color: var(--color-text-alt);
+  background: var(--color-bg-alt);
+  box-shadow: 0 1px 3px rgba(black, 0.2);
```
studioswong, line 80: just a small observation, currently it doesn't look like we have any defined variables for spacing - will adding spacing variables be something that we will explore in the future?

src/styles/_variables.scss
```diff
@@ -1,29 +1,31 @@
 //-- Colours --//
 
-$color-dark: #000;
-$color-light: #fff;
+$color-dark: black;
```
studioswong, line 3: love this 👍

studioswong:
Great work @richardwestenra  - I imagine it must feel so great to delete all those code 👍

src/components/app/app.scss
richardwestenra: Basically just because it wasn't necessary for anything - I want to move towards having just CSS custom properties and not Sass variables eventually, but I'm using both for now where they already existed. This one just wasn't a Sass variable previously so it didn't seem necessary to introduce one!

src/styles/_extends.scss
richardwestenra: I don't think we need to unless they're explicitly reused and we want to be able to batch update them.

src/components/flowchart/styles/_variables.scss
richardwestenra: ah yeah interesting, I hadn't caught that some of these are duplicates of the core ones. It's a preexisting issue so slightly out of scope of this PR but I don't mind fixing it here 🌞

src/components/app/app.scss
studioswong: I see, sounds good!

src/components/flowchart/styles/_variables.scss
richardwestenra: fixed

src/components/metadata/styles/metadata.scss
```diff
@@ -165,34 +152,33 @@ $list-inline-spacing: 0.2em;
...
+    opacity: 0;
+  }
 }
 
-.pipeline-metadata__toolbox .pipeline-metadata__copy-button {
-  width: 1.2em;
-  height: 1.2em;
+.pipeline-metadata__copy-button {
+  width: auto;
+  height: auto;
```
studioswong, line 163: minor question - are we taking the width and height conditions off for the `copy-button` inside `metadata_toolbox`?

src/components/node-list/styles/_row.scss
```diff
@@ -114,39 +95,16 @@
...
-  }
-
-  .kui-theme--light &--disabled {
-    color: rgba($color-dark, 0.25);
-  }
-
-  .kui-theme--dark & b {
-    color: $color-primary;
+  &--faded {
+    opacity: 0.65;
```
studioswong, line 99: is the color setting for the `--faded` condition replaced by the opacity in this case?

src/components/node-list/styles/_row.scss
```diff
@@ -114,39 +95,16 @@
...
   }
 
-  .kui-theme--light & b {
-    color: darken($color-primary, 10%);
+  &--disabled {
+    opacity: 0.3;
```
studioswong, line 103: same as the above - is the `color` setting replaced by opacity?

src/components/node-list/styles/_row.scss
richardwestenra: yeah I've changed it to inherit the default text color, which has 0.85 opacity. So converting 0.55 / 0.85 = 0.65

studioswong (approved):

src/components/metadata/styles/metadata.scss
richardwestenra: check the output in the browser - the width/height is set on the `svg` icon element in the original component, but for some reason a smaller width/height is set on the parent button in this case, which messed up how the focus state looks without changing the actual visible width/height. I think it's a mistake:
image

src/components/node-list/styles/_row.scss
richardwestenra: yup, 0.25 / 0.85 = 0.3

src/components/metadata/styles/metadata.scss
studioswong: good spot, glad that you fixed this in this PR!

src/components/app/app.scss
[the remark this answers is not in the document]
bru5: Okay fine for now, but I think it's worth automating this eventually for ease of updating colours and to avoid them both potentially diverging accidentally as they're very easy to mix up.

src/components/flowchart/styles/_variables.scss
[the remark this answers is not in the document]
bru5: Ah yeah that's fine, I think I remember this being a problem now from a while back.

src/components/icon-button/icon-button.scss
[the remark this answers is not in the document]
bru5: I think the former is definitely clearer for me, but yeah it's not a big deal.

src/components/minimap-toolbar/minimap-toolbar.scss
bru5: Ah makes sense. I think my comment about automating this with a little function would be reusable here as well, to make this a bit more fool proof and explicit.

src/components/icon-button/icon-button.scss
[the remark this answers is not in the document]
richardwestenra: okay cool, make sense. sorry forgot to change it but I really don't think it's a big deal either way honestly

src/components/app/app.scss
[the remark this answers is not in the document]
richardwestenra: Sure. Can you give me an example of how you'd do it? I've not come up with any ideas that are simpler than the status quo. Everything I can think of just feels like a cumbersome abstraction that doesn't actually simplify things much

PR #307, opened 2020-11-16
[KED-1923] Add LazyList component using IntersectionObserver on sidebar

bru5:
## Description

Adds lazy list rendering and memoization to the sidebar. 

Improves time to first paint and first interactive when loading a pipeline with a large number of elements, may also improve general sidebar interaction performance such as search and hover.

### Memoization

The `NodeListRow` component is now memoised using `React.memo`, checking minimal props that may change after first render, giving a performance boost on rendering each row that doesn't change (so most of them).

### LazyList

Adds a new `LazyList` component that renders only the children that are currently visible on screen at any time.

#### Motivation

Some pipelines may have hundreds of elements and tags, which means a lot of sidebar elements to render, causing a slow time to interactive.

I found most components available in this space appear to be optimised for a specific kind of layout and interaction model, usually assuming a single flat list inside a single scrolling container. Even a specialised tree component such as [react-vtree](https://github.com/Lodin/react-vtree) would likely limit potential future design directions.

In other cases significant reworking of our existing components and data structures would be needed, requiring special handling of many existing and future interactions in the sidebar.

#### An automatic `LazyList` implementation using `IntersectionObserver`

The `LazyList` component dynamically renders only the children currently visible at any time. Most user interactions should  therefore work automatically, such as scroll, resize, but the key difference being that pretty much any kind of layout change, in any part of the app, at any time, will also be handled too with no extra code.

This is currently achieved using `IntersectionObserver`, with no other handlers being required (e.g. no scroll or resize, so far anyway). 

It also does not require any virtual scrolling or component pooling to get full windowing from this either, as React's virtual DOM diffing and memoization appears to handle that fast enough by default, with a little buffering. But in practice, we have no need for this in our case so disposal has been switched off for smoother results.

The same approach can be extended to work with grids or any other kind of layout assuming the user could provide dimension / selection functions for it.

The actual implementation here is a little more involved than expected due to the details of how `IntersectionObserver` works and the React render cycle itself, but even with special care to avoid render thrashing, it still turns out to be really compact when using hooks (~200 lines total).

### Caveats

- requires `IntersectionObserver`
  - but is [well supported](https://caniuse.com/intersectionobserver) by now
  - has an automatic fallback to full-list rendering (done)

- scrolling fast briefly leaves a blank space
  - this is mostly mitigated by adding a buffer at both ends (done)
  - option to avoid disposal of rows helps avoid this too (done)
  - using a gradient fade on the ends of the list smooths this over visually (done)
  - can add a placeholder loading effect on the end too (later)
  
- blocking the main thread leaves a blank space
  - we should aim to avoid doing this anyway
  - would likely be an issue for most other implementations too
  - currently hovering nodes causes the main thread to block
    - mitigated by disabling thread blocking hover effects (done for demo only)
    - a better solution is really needed here

### Flag

This feature will be included behind a flag until it is well proven.

## QA notes

All of the sidebar features should be tested thoroughly, in all supported browsers, in combination with switching pipelines.
Checking closely for any missing rows is important.

Some console messages are currently left in for demonstration and debugging.

## Checklist

## Legal notice


src/components/node-list/node-list-group.js
```diff
@@ -51,14 +55,77 @@ export const NodeListGroup = ({
...
+              faded={item.faded}
+              visible={item.visible}
+              selected={item.selected}
+              unset={item.unset}
+              visibleIcon={item.visibleIcon}
+              invisibleIcon={item.invisibleIcon}
+              onClick={() => onItemClick(item)}
+              // Disabled to avoid unrelated hover lag for now.
+              // onMouseEnter={() => onItemMouseEnter(item)}
+              // onMouseLeave={() => onItemMouseLeave(item)}
```
richardwestenra, line 121: `// @TODO fix this`?

src/components/node-list/node-list-row.js
```diff
@@ -1,122 +1,145 @@
...
-  onMouseLeave,
-  onChange,
-  onClick,
-  selected,
-  type,
-  visibleIcon = VisibleIcon,
-  invisibleIcon = InvisibleIcon
-}) => {
-  const VisibilityIcon = checked ? visibleIcon : invisibleIcon;
+export const nodeListRowHeight = 36.59375;
```
richardwestenra, line 10: It might be worth getting this value from the DOM, or else add a comment in both the CSS and JS to ensure that updates to one get reflected in the other

src/components/node-list/node-list-group.js
```diff
@@ -51,14 +55,77 @@ export const NodeListGroup = ({
...
-  </Container>
+    <LazyList
+      name={name}
+      height={(start, end) => (end - start) * nodeListRowHeight}
+      total={items.length}
+      onChange={({ start, end, total }) =>
+        console.log(
+          `${group.name} ${start} to ${end} (${end - start} of ${total})`
+        )
+      }>
```
richardwestenra, line 66: Adding a note about this to remind us both that it should be removed before merging

Might also be worth a comment in the code:`// @TODO remove this`

richardwestenra (changes requested):
Overall, great. These notes are mostly for things that I think you're already aware of - I'm mainly adding comments just to remind us to check them so we don't forget.

src/components/node-list/styles/_group.scss
```diff
@@ -16,12 +16,46 @@
...
+  transition: opacity ease 0.3s;
+  content: ' ';
+  pointer-events: none;
+}
+
+.pipeline-nodelist__placeholder-lower:after {
+  top: -$placeholder-fade;
+  bottom: initial;
+  background: linear-gradient(0deg, #f0f1f3 0%, transparent 100%);
+}
```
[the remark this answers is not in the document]
bru5: Ah thanks yeah this was just for quick testing, this should now be implemented.

src/components/node-list/node-list-group.js
bru5: Sure I'll update this part once I figure out next steps.

src/components/node-list/node-list-row.js
bru5: Yeah for sure, it just takes a little DOM gymnastics to get hold of the size. For now I've added some comments, but hopefully will add this later!

src/components/node-list/node-list-group.js
bru5: I have added a flag and enabled hover highlighting again, but without the linked nodes search when the flag is enabled. That should work well enough until a non-blocking solution for linked nodes is implemented.

src/components/node-list/node-list-group.js
bru5: Yeah it's in just to make this PR easier to test and review, once approved it will be removed.

richardwestenra (approved):
This PR reduced the NodeList render time on my large test dataset from ~1690ms to ~30ms. Exceptional work!

Please don't forget to remove the console.logs before merging! :shipit:

richardwestenra:
oh also, you'll need to fix the test-lib tests unfortunately @bru5

README.md
```diff
@@ -104,6 +104,7 @@ The following flags are available to toggle experimental features:
 
 - `newgraph` - From release v3.4.0. Improved graphing algorithm. (default `false`)
 - `meta` - From release v3.7.0. Show node metadata panel on click. (default `false`)
+- `lazy` - From release vx.x.x. Improved sidebar performance. (default `false`)
```
studioswong, line 107: It would be worth to actually put in `v.3.8.0` given that we have decided to release it in the next release to prevent this being forgotten to be updated in the release?

src/components/lazy-list/index.js
```diff
@@ -0,0 +1,327 @@
...
+    height,
+    range,
+    total
+  ]);
+
+  // Skipped if not enabled or supported
+  if (active) {
+    // Allows an update only once per frame
+    const requestUpdate = useRequestFrameOnce(
+      // Memoise the frame callback
```
studioswong, line 64: small typo: should be memorize?

richardwestenra: I'm gonna jump in here and say no, because it's the verb form of memo, not memory: https://en.wikipedia.org/wiki/Memoization

studioswong:
I've tested the lazyload sidebar on both chrome and safari with a large dataset, looks good! @bru5 looks like the test richard mentioned above still needs to be fixed?

src/components/lazy-list/index.js
```diff
@@ -0,0 +1,327 @@
...
+  total,
+  children,
+  lazy = true,
+  dispose = false,
+  buffer = 0.5,
+  onChange,
+  container = element => element?.offsetParent
+}) => {
+  // Required feature checks
+  const supported = typeof window.IntersectionObserver !== 'undefined';
```
richardwestenra, line 26: What happens if it's not supported?

src/utils/changed.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Returns true if any of the given props are different between given objects.
+ * Only shallow changes are detected.
+ */
+export default (props, objectA, objectB) => {
+  return (
+    objectA && objectB && props.some(prop => objectA[prop] !== objectB[prop])
+  );
+};
```
richardwestenra, line 9: I would've been more inclined to put this in the main `utils/index.js` file but it honestly doesn't matter

richardwestenra (approved):
Fantastic work!

src/components/lazy-list/index.js
bru5: The line below hopefully makes that clear:

```
// Active only if enabled by prop and features detected
const active = lazy && supported;
```

Then a few lines further down:

```
  // Skipped if not enabled or supported
  if (active) {
```

Basically, it will just 'pass through' and render the whole list. Shall I add that in the description?

richardwestenra: ah thanks 🙈! yeah not a bad idea to add that as a comment but it looks like it's already made fairly clear

richardwestenra (approved):

PR #312, opened 2020-11-23
[KED-2226] Align node API endpoint field names to Kedro Core naming schema

studioswong:
## Description

related ticket: https://jira.quantumblack.com/browse/KED-2226

This PR relates to updating our store and actions to renames node data fields to match the latest naming of the fields in the API. 

## Development notes

The main changes in the name of the fields returned from the API are as follows:

- `code_location` to `filepath`
- `dataset_location` to `filepath`
- `dataset_type` to `type`

I have made the following changes to match the above API changes:
- deleting the `codeLocation` and `dataset_location` fields in the store and replacing those fields with the `filepath` field - the filepath field will now store the filepath data for both `task` and `dataset` type nodes
- updating the reference of dataset_type within the action - I had kept the field `dataset_type` within the store as we also have another field named as `type`, moreover allowing more clarity with the data as this field refers specifically to the dataset type. 

## QA notes

The QA of this feature needs to be done by running Kedro-viz against this feature branch as the changes of the API has not been deployed in the latest kedro-viz package yet. 

## Checklist

## Legal notice


src/store/normalize-data.js
```diff
@@ -122,10 +121,9 @@ const addNode = state => node => {
   state.node.tags[id] = node.tags || [];
   // supports for metadata in case it exists on initial load
   state.node.code[id] = node.code;
-  state.node.codeLocation[id] = node.codeLocation;
   state.node.docString[id] = node.docString;
   state.node.parameters[id] = node.parameters;
-  state.node.dataset_location[id] = node.dataset_location;
+  state.node.filepath[id] = node.filepath;
   state.node.dataset_type[id] = node.dataset_type;
```
richardwestenra, line 127: I didn't pick up on this last time, but it'd be great for us to be consistent in our filename schema when adding things to the store, i.e. consistently using camelCase over underscores. This need not apply to the API endpoint itself necessarily, but it'd be great to align the variable names we're using in the main store.
```suggestion
  state.node.filepath[id] = node.filepath;
  state.node.datasetType[id] = node.dataset_type;
```
However I note that the API appears to be using both `docString` and `dataset_type` which isn't ideal

richardwestenra (approved):
Overall great work, but I'd love it if you could make the naming consistent while you're at it!

src/store/normalize-data.js
studioswong: yea good call about this, I also thought about changing this earlier too - I'll make the change 👍

richardwestenra (approved):

PR #316, opened 2020-11-24
Change tag icon hover state styling

studioswong:
## Description

related PR: https://jira.quantumblack.com/browse/KED-2183

This PR is to fix the styling of the hover state of the icon next to each tag on the sidebar so that the blue dot does not change its styling on hover when there are no tags checked on the sidebar. 

Change in colr for the blue dot on toggle:
before-toggle

## Development notes

This change in color on hover of the blue dot is caused by a css opacity setting for hover of `node-list-group` - on inspecting various part of the sidebar and the codebase, this hover css setting does not have any effect any other use cases of `node-list-group` other than causing this styling issue ( i.e there are seperate hover states defined for the icons of both dataset and parameter icons which have overwritten the hover state defined in `node-list-group`), hence it could be deleted to fix this styling issue. 

Toggleing the blue dot on tags will not cause any change in opacity after the fix: 
after-toggle-2

## QA notes

This can be tested on toggling the blue dot next to tags. 

## Checklist

## Legal notice


richardwestenra (changes requested):
This is not exactly as I'd imagined/suggested we do it - I was thinking that the 0.75 opacity would apply on hover when you hover over the entire row. However this would kinda work for the inactive hover state, if we used the 0.75 opacity for active, and 1 for active+hover. That way the state would changes when you change it active, which works as a way to indicate that the state has changed from hover+inactive to hover+active.

Additionally, this PR does not fix the issue for hover state when a tag is already active, as when you hover over an active row, it gets less opaque, except when you hover over the icon itself, which is the issue we were trying to avoid:
tag-hover

Aside from the different hover state for the icon vs the row, I think an exacerbating factor is that the design does not offer a visible distinction between the hover state and the active state, and there should be. I honestly think we might need to use something other than opacity to make it more clear. cc @GabrielComymQB 

Finally, I would probably use the hand cursor icon for the entire row, _but only for tag rows_, as discussed.

GabrielComymQB:
I agree with Richard, the indication of the hand cursor should apply to the entire hit area of the cell - no only to the blue indicator area.

I'm happy for you guys to play with the opacities on your side and see how it renders. What does feel more comfortable to you? The most important thing is not to break the experience and understanding at this point. We can always make it better later.

Another thing is on my side I'm seeing some weird colours. You guys already said that might be a bug :)


Screenshot 2020-11-26 at 14 44 56

Screenshot 2020-11-26 at 14 49 30

richardwestenra:
Thanks Gabriel! Yeah that colour issue is my fault, I might as well fix it in a separate PR 😬

studioswong:
Thanks both! As after our great discussion on friday, I will be adding in the following 2 changes:

1. changing the opacity of the 'blue dot' under the 'selected tag' state with opacity of 1
2. taking out the 'hand' cursor near the blue dot to allow the indication of the entire tag row as one button.

bru5 (changes requested):
The grey states for indicators instead look black or white depending on the theme. That might be something @richardwestenra has fixed and requires a merge here?

Also I think ideally the indicator would get brighter once clicked to make active? I know that means clashing with a hover state potentially at first.

Either that or no longer show the indicator at all unless active? @GabrielComymQB

richardwestenra:
> The grey states for indicators instead look black or white depending on the theme. That might be something @richardwestenra has fixed and requires a merge here?

Yeah those were fixed in #319.

studioswong:
@GabrielComymQB @richardwestenra @bru5 I made implemented the changes in consolidation of the various changes we discussed over the past few days about the behaviour of the sidebar - please see below gif for the demo in how it would behave in both the unchecked and checked situations.

Similar to how Liam suggested, quite a few of the changes I have here is a result of iterate and test to get to what I think works, and as Liam and I have previously discussed, it might be easier to simplify the opacity of the dot down to 2 states: a lighter opacity of 0.55 during the unchecked and hover state, and a consistent opacity of 0.75 for the checked / checked+hovered states ( i.e no change in opacity for the dot once it's checked.) 

In short, here are the new changes I implemented:
1. removed the 'hand' cursor for the blue dot
2. remove the changes in opacity on hover of the blue dot under both unchecked and checked states (i.e the blue dot looks consistent on hover)
3. moved the hover settings for the toggle icon for the eye out into a seperate CSS class so the hover settings will still apply for the 'eye' icon for the nodes row on  the sidebar

for the unchecked sidebar:
sidebar-demo-unchecked

for the checked sidebar:
sidebar-demo-checked

richardwestenra (changes requested):
Things still needing to be changed:

1. When you hover the icon for an inactive tag when another tag is active (grey outline icon), then opacity changes. It should be the same opacity change for hovering the whole row: 
image
2. When visible, the toggle-all icon at the top (square/minus) decreases in opacity when you hover over it. It should increase in opacity on hover. 
image
3. When hovering over an active row, the icon doesn't increase in opacity. An active hovered row icon should be 1, not 0.75.

Reference from our slack convo, for icon opacities:
- :no_entry_sign:hover + :no_entry_sign:active = opacity 0
- :white_check_mark:hover + :no_entry_sign:active = opacity 0.55
- :no_entry_sign:hover + :white_check_mark:active = opacity 0.75
- :white_check_mark:hover + :white_check_mark:active = opacity 1
image

studioswong:
Thanks for the comment @richardwestenra :)  for no. 2, I just tested out the toggle-all icon and currently it doesn't decrease in opacity on hover rather it just staying the same in opacity. I can tweak it to increase in opacity to `1`. I wasn't aware of the need to change the other dots as mentioned in no. 1, but good spot and happy to put in a fix for that. 

In terms of point 3, I recall from our last discussion that, to simplify things, we will simply set the opacity of the blue-dot of the selected tag to stay the same once the sidebar enters into the 'selected-tag' phrase? ( i.e it will stay with the same opacity once it's selected regardless of hovering over the dot to indicate the 'on' state and no need for any further change on hover.) As I mentioned above in my last comment, I had a discussion with Liam and it seems to be much more straight forward to break it down to 2 states: hover state when not selected (0.55), and selected state (0.75). This execution follows the principles as outlined to simply setting an 'off' and 'on' state for the selected button, where on hovering over the dot during the 'on' state shouldn't change the appearance given that the dot is already in an 'on' state. 

Visually, I also support this simpler execution given that the row already changes to a 'highlighted' background, where there aren't any particular value-added to change the opacity of the blue dot, while it looks more neat to just follow the logic of the text that doesn't change in opacity. (I had experimented with increasing the opacity on hover for the blue dot to 1 and the effect isn't that apparent with the highlighted background). 

Again, I am happy to revisit this although I personally support the simplier execution of this given this further 'on + hover' state adding an extra layer of complexity that could cause more problems to maintain down the line.

richardwestenra:
> for no. 2, I just tested out the toggle-all icon and currently it doesn't decrease in opacity on hover rather it just staying the same in opacity. I can tweak it to increase in opacity to 1.

Try toggling it on and hovering on it. This issue only occurs when at least one node is selected. If you're still not seeing it, let's have a call and I'll show you.

>  I wasn't aware of the need to change the other dots as mentioned in no. 1, but good spot and happy to put in a fix for that.

Thanks, please do!

> In terms of point 3, I recall from our last discussion that, to simplify things, we will simply set the opacity of the blue-dot of the selected tag to stay the same once the sidebar enters into the 'selected-tag' phrase?

That was not my understanding of how the conversation went at all. I believe I repeated that it should have 4 opacity states to reflect the number of possible outcomes arising from the combination of the 2 different hover/active states. However I am totally fine with the blue icon not changing opacity on hover when active if you think that the row background change is sufficient, as long as it's consistent. In that case it will only need 3 different states, as you won't need a separate hover+active state.

studioswong:
@richardwestenra @bru5 I've made further following changes in my latest commits:

1. added an increase in opacity from 0.75 to 1 for the '-' icon next to the 'tags' row on hover
2. added an increase in opacity from 0.75 to 1 for the blue dot on hover of the entire selected row

sidebar-1208

3. refactored the classes for the toggle interactions of the other toggle icon ( the 'eye' toggle icon) so that it will continue to behave in the same way as set up before ( this is needed due to the CSS changes required for 1 and 2)

sidebar-1208-eye

Please definitely do click around and let me know if anything doens't look right / breaks/ etc!

src/components/node-list/node-list-row.js
```diff
@@ -88,7 +88,8 @@ const NodeListRow = ({
           `pipeline-row__toggle--kind-${kind}`,
           {
             'pipeline-row__toggle--disabled': disabled,
-            'pipeline-row__toggle--selected': selected
+            'pipeline-row__toggle--selected': selected,
+            'pipeline-row__toggle--cursor': type !== 'tag'
```
bru5, line 92: Not sure if we should use `cursor: pointer` on tags, as I think it would need to be all or nothing. Using it for all would then mean the whole row, which I don't think will feel right. If it's about consistency with the eye button behaviour, I'd probably remove `cursor: pointer` from them too (but don't think that change is really needed).

src/components/node-list/styles/_filter.scss
```diff
@@ -63,18 +63,24 @@
...
+    .pipeline-nodelist__row--visible:hover & {
+      opacity: 1;
+    }
   }
+}
 
-  // Hover states when checked
-  &.pipeline-row__toggle-icon--checked:hover {
+// Hover states for non-filter type icons
+.pipeline-row__toggle-icon--visible:hover {
```
bru5, line 82: Might be better to move this to `_row.scss`

src/components/node-list/styles/_row.scss
```diff
@@ -119,6 +119,9 @@
 .pipeline-row__toggle {
   position: absolute;
   right: $row-padding-x;
+}
+
+.pipeline-row__toggle--cursor {
   cursor: pointer;
```
bru5, line 125: See comment above about using pointer, but otherwise I think this might better with a semantic class name. Would the existing class `pipeline-nodelist__row--kind-toggle` as used below work here? Otherwise maybe something like `.pipeline-row__toggle--not-tag`.

bru5 (changes requested):

src/components/node-list/node-list-row.js
studioswong: thanks for raising this point! 

We actually discussed about this in our earlier discussion between Gabriel, Richard and I, and we decided to keep the 'hand' cursor (the pointer cursor) for the eye icon given that that icon is in fact an actual toggle button that the user can click/toggle, and the hand cursor within our app will be an indication for that type of 'toggle' button. ( This is also one of the thinking behind this ticket, which is to extract this type of interactive toggle elements out between the filters and nodes on the side bar).

Having played around with both scenerio when I was testing with the sidebar, I also personally think having the pointer cursor for the eye icon to indicate to the user that they can  toggle the button is helpful too :) but again, happy to discuss otherwise :D

src/components/node-list/styles/_filter.scss
studioswong: good spot - I've moved it across!

src/components/node-list/styles/_row.scss
studioswong: yea this is a fair point - The main reason I separated this out as a new class and named it under `--cursor` instead of directly applying it under the `-toggle` class is mainly for readability, to allow a clear indication of this class simply applying a cursor interaction to the element. 

Imo the css class naming should be separated from the logic that governs the application of this class ( i.e the logic, `not-tag`, is applied on the JS side), which allows flexibility in case the logic of applying the cursor is changed in the future, while the semantics of this naming allows it to be easily understood as applying a cursor css element to whatever logic we applied on the JS side. 

again, happy to discuss and let me know your thoughts otherwise :D

bru5: I think at the moment the approach in use is semantic BEM style names with modifiers, rather than utility classes like `--cursor`, but @richardwestenra can confirm. At least for consistencies sake, I think this should come under `pipeline-nodelist__row--kind-toggle` but check with him I'd say.

src/components/node-list/node-list-row.js
bru5: The cursor on the eye is fine, I meant on tags but it's a minor detail so not a big deal.

studioswong: I don't think the `cursor:pointer` is applied on tags, hence this new class to check and only apply to non-tag rows?

src/components/node-list/styles/_row.scss
studioswong: sure I can move this under the `--toggle` class for simplicity

studioswong: @bru5 On having moved the `pointer:cursor` setting under `pipeline-nodelist__row--kind-toggle`, unfortunately the cursor effect does not apply, and on having played around with the other classes, the cursor classes cannot be applied either unless it's an additional add-on class in the execution that I did.

In aligning to your comment, I had changed the class name to `-not-tag` as per your suggestion.

richardwestenra (approved):
Nice one, thanks! Might be worth getting @bru5 to take a quick look before merging if you don't mind

bru5 (approved):

PR #320, opened 2020-11-26
Quickfix: update naming of docstring field within codebase to match API 

studioswong:
## Description

While going through the codebase I realize that the `docstring` field within our codebase was named under JS camel case, which we should follow the python naming within our codebase. 

## Development notes

I have changed the naming of the field from `docString` to `docstring` within our store, reducers and tests.

## QA notes

N/A

## Checklist

## Legal notice

### DIFF
--- a/src/reducers/nodes.js
+++ b/src/reducers/nodes.js
@@ -56,8 +56,8 @@ function nodeReducer(nodeState = {}, action) {
         filepath: Object.assign({}, nodeState.filepath, {
           [id]: data.filepath
         }),
-        docString: Object.assign({}, nodeState.docString, {
-          [id]: data.docString
+        docstring: Object.assign({}, nodeState.docstring, {
+          [id]: data.docstring
         }),
         parameters: Object.assign({}, nodeState.parameters, {
           [id]: data.parameters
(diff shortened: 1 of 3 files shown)


bru5 (approved):

richardwestenra (approved):
Sounds good - is this the case used in the API output JSON? Just want to doublecheck that we're matching up here.

studioswong:
Thanks for raising this @richardwestenra ! For the API output JSON, do you mean the case of running viz with the JSON data file? This change wouldn't affect JSON API output, given that this only refers to the data returned from the `nodes` endpoint from a running Kedro project and would only need to refer to the data format returned from the endpoint (which, on implementing ticket(KED-2309), viz would simply not fire any API calls on selecting a node).

Speaking of, on running viz from a JSON data file, what will the metadata panel show for the dynamic data fields that should be fetched from the API? @bru5

richardwestenra:
No I meant 'case' as in camelCase vs snake_case vs lowercase vs UPPERCASE. So I was asking, what case is the API output in? Because we should be matching that in normalize-data.js when reading it, even if it's not case-sensitive. e.g. if it's in camelCase then we should do something like
```js
state.node.docstring[id] = node.docString;
```

richardwestenra:
Update: I just manually checked and yeah it's all lowercase in the API output 👍 

as you were 😄

PR #321, opened 2020-11-30
[KED-2309] Apply the 'meta' flag for requests to the nodes endpoint

studioswong:
## Description

related ticket: https://jira.quantumblack.com/browse/KED-2309

this ticket is to hide the calls to the node API endpoint behind the `meta` flag so calls will only be fired when the meta flag is true. 

## Development notes

I had added an extra check for the `meta` field within the `loadNodeData` action, as well as updated the tests to check for this condition.  

## QA notes

Clicking the nodes will not trigger any calls to the nodes endpoint unless the meta flag is true. 

## Checklist

## Legal notice

### DIFF
--- a/src/actions/nodes.js
+++ b/src/actions/nodes.js
@@ -79,12 +79,7 @@ export function loadNodeData(nodeID) {
 
     dispatch(toggleNodeClicked(nodeID));
 
-    if (
-      asyncDataSource &&
-      nodeID &&
-      !node.fetched[nodeID] &&
-      flags.meta === true
-    ) {
+    if (asyncDataSource && nodeID && !node.fetched[nodeID] && flags.meta) {
       dispatch(toggleNodeDataLoading(true));
       const url = getUrl('nodes', nodeID);
       const nodeData = await loadJsonData(url);
--- a/src/actions/nodes.js
+++ b/src/actions/nodes.js
@@ -75,11 +75,16 @@ export function addNodeMetadata(data) {
  */
 export function loadNodeData(nodeID) {
   return async function(dispatch, getState) {
-    const { asyncDataSource, node } = getState();
+    const { asyncDataSource, node, flags } = getState();
 
     dispatch(toggleNodeClicked(nodeID));
 
-    if (asyncDataSource && nodeID && !node.fetched[nodeID]) {
+    if (
+      asyncDataSource &&
+      nodeID &&
+      !node.fetched[nodeID] &&
+      flags.meta === true
+    ) {
       dispatch(toggleNodeDataLoading(true));
       const url = getUrl('nodes', nodeID);
       const nodeData = await loadJsonData(url);
--- a/src/actions/nodes.test.js
+++ b/src/actions/nodes.test.js
@@ -1,6 +1,7 @@
 import { createStore } from 'redux';
 import reducer from '../reducers';
 import { mockState } from '../utils/state.mock';
+import { changeFlag } from '../actions';
 import {
   TOGGLE_NODE_DATA_LOADING,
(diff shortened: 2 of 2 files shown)


studioswong:
looks like the `pylint` tests are failing on the CI - @limdauto wondering if you know what this could be? 
image

src/actions/nodes.js
```diff
@@ -75,11 +75,16 @@ export function addNodeMetadata(data) {
...
+    const { asyncDataSource, node, flags } = getState();
 
     dispatch(toggleNodeClicked(nodeID));
 
-    if (asyncDataSource && nodeID && !node.fetched[nodeID]) {
+    if (
+      asyncDataSource &&
+      nodeID &&
+      !node.fetched[nodeID] &&
+      flags.meta === true
```
richardwestenra, line 86: Could this be changed to just `flags.meta` (instead of `flags.meta === true`), or does it need to be this specific?

src/actions/nodes.test.js
```diff
@@ -42,24 +43,35 @@ describe('node actions', () => {
       jest.resetModules();
     });
 
+    describe('if meta flag is false', () => {
+      it('should not make any API calls if meta is false', () => {
```
richardwestenra, line 47: This is unnecessary repetition, as it will ultimately read:
```
loadNodeData > if meta flag is false > should not make any API calls if meta is false
```
I'd suggest something like this instead:
```suggestion
    describe('if meta flag is false', () => {
      it('should not make any API calls', () => {
```

src/actions/nodes.test.js
```diff
@@ -42,24 +43,35 @@ describe('node actions', () => {
...
+        const store = createStore(reducer, mockState.json);
+        loadNodeData('parametersID')(store.dispatch, store.getState);
+        expect(store.getState().loading.node).toBe(false);
```
richardwestenra, line 50: It might be worth explicitly setting the flag to false here, so that the test will still make sense even if the default is changed later upstream.

richardwestenra (changes requested):
Great work! A couple of very minor things that might be worth addressing:

src/actions/nodes.js
studioswong: sure, we can just use `flags.meta` :)

src/actions/nodes.test.js
studioswong: sounds good!

src/actions/nodes.test.js
studioswong: yea great point that you raised here - originally I thought of setting it to false here for the test could be more explicit, but given the default setting of meta is false and we have explicitly mentioned about the condition we are testing in this case, it doesn't really make sense to particularly set the flag to false here?

richardwestenra: Yeah I get that. My point was mainly that if someone changed the default upstream, the test might fail when it's not supposed to, or else if the test were broken then it might pass when it's not supposed to. But I'm not going to press the issue if you strongly feel that it shouldn't be set explicitly, because I understand your position!

richardwestenra (approved):

richardwestenra:
Lim has fixed the Python linter in #323 so if you merge main it should fix the broken CI on this PR

src/actions/nodes.test.js
bru5: Yeah it doesn't look like the test guarantees that the flag is false, which is what the description implies, so it's probably worth ensuring that or changing the describe to handle the default case (which may be true or false).

richardwestenra: @studioswong ☝️

PR #325, opened 2020-12-02
Update Kedro-UI to v1.1.4 & implement relevant changes

richardwestenra:
## Description

I've added 3 new minor features to [Kedro-UI](https://github.com/quantumblacklabs/kedro-ui) in versions 1.1.3 and 1.1.4, that will allow us to implement or improve some new/existing features:

1. [Replace check-font-loaded](https://github.com/quantumblacklabs/kedro-ui/pull/76) with [Google Webfont Loader](https://github.com/typekit/webfontloader). This will replace [a weird and unreliable bit of the codebase](https://github.com/quantumblacklabs/kedro-viz/blob/main/src/actions/check-font-loaded.js) with a proper callback, which is all we ever really wanted to begin with.
2. [Add transition-out for closing Dropdown](https://github.com/quantumblacklabs/kedro-ui/pull/78). No changes required, other than to update the version.
3. Add active pipeline menu-option border-left colour. This uses [the new className prop](https://github.com/quantumblacklabs/kedro-ui/pull/80).
4. Use [Dropdown disabled prop](https://github.com/quantumblacklabs/kedro-ui/pull/79) for pipeline selector

## Development notes

I've deleted check-font-loaded.js and its tests, as it's no longer needed. This should improve cross-platform stability. I've checked the load-time in Chrome and it appears to be about the same.

## QA notes

1. Test loading the webfont with a slower connection.
2. Try closing the dropdown
3. Try changing the active pipeline and observe the new blue left border on the menu-options, on both themes.
3. See Dropdown when disabled - e.g. when loading, or when there are no pipelines in the dataset. It should no longer be keyboard-navigable.

## Checklist

## Legal notice


src/components/pipeline-list/pipeline-list.scss
```diff
@@ -14,21 +14,12 @@
   --pipeline-list-shadow: rgba(black, 0.1);
   --pipeline-list-label-bg: #{$color-bg-dark-5};
   --pipeline-list-options-bg: #{$color-bg-dark-1};
-  --pipeline-list-option-border-width: 0 0 1px;
+  --pipeline-list-option-border-width: 0 0 1px 2px;
```
studioswong, line 17: This new border setting doesn't seem to be related with the relevant changes with kedro-UI - is this a small visual enhancement that you noticed during development?

studioswong:
great work Richard! It might be worth to put in a small UI test to test the newly implemented blue left border, as well as another test to ensure that the pipeline dropdown gets disabled during loading and when there are no pipelines in the dataset?

richardwestenra:
Thanks Susanna! I can't test for the styling, but I could test for the presence of the class. I'm not sure that it's a worthwhile substitute, but I'll give it a look. Thanks!

studioswong:
one thing I notice on testing the transition-out effect that might be worth revisiting - the background of the dropdown doesn't seem to transition at the same time with the text, hence some of the dropdown text will overlap with the text of the sidebar during the transition ( see below gif) which seems a bit unnatural? 
dropdown_transition

Personally I prefer the original dropdown transition that we have which to me seems more natural - @bru5 @GabrielComymQB  not sure what you guys think? 
dropdown_transition_original

Regardless @richardwestenra I admire the effort brought into introducing the transition animation 👍

richardwestenra:
@studioswong ooh good catch: If you slow down the transition to 25% or 10% [in devtools using the Animations tab](https://developers.google.com/web/tools/chrome-devtools/inspect-styles/animations), you'll see that it actually flickers on hiding the dropdown. This seems to be caused by the underlay, for some reason, which is super weird because it's just an opacity change. I'll investigate and see if I can figure out a good resolution that doesn't rely on updating Kedro-UI again.

bru5 (changes requested):
All looks good other than this appears to introduce a drop shadow when hovering the pipeline selector that I don't think fits the design (it's most visible on the light theme), so I think that should be removed before merge.

richardwestenra:
@bru5 Great catch, thanks! Looks like the new Kedro-UI CSS changes increased the specificity and overrode the existing styles. I've changed it to use `!important` here to make it more bulletproof and be explicit that it's overriding external styles.

studioswong (approved):
LGTM 👍 approved on my side unless @bru5 have any comments to your proposed changes?

src/components/pipeline-list/pipeline-list.test.js
```diff
@@ -31,6 +37,20 @@ describe('PipelineList', () => {
...
+    'should apply an active class to a pipeline row only if it is active (id: %s, index: %s)',
+    (id, i) => {
+      const wrapper = setup.mount(<PipelineList onToggleOpen={jest.fn()} />);
+      const { active } = wrapper.find('PipelineList').props().pipeline;
+      const isActive = active === id;
+      const hasClass = wrapper
+        .find('MenuOption')
+        .at(i)
+        .hasClass('pipeline-list__option--active');
+      expect(hasClass).toBe(isActive);
```
bru5, line 50: I think this test could be ideally be written in a much more explicit way as it is potentially a little brittle without having constants for each potential case. Consider breaking this into two tests, one where active is explicitly set as true (e.g. by prop in mount) and one where it is false.

bru5 (approved):

src/components/pipeline-list/pipeline-list.test.js
richardwestenra: I did it this way because I figured it'd be less brittle, as it doesn't rely on knowing which item is active. I could split it into two but I'm not sure that it improve anything, but yeah I'll give it a go!

richardwestenra: @bru5 done

PR #328, opened 2020-12-08
Add graceful fallback response for API calls

bru5:
## Description

This PR adds an explicit fallback response to avoid page crashing errors on failed API requests, such that the app can handle them more gracefully. For the `main` API route, the previous behaviour remains for compatibility.

## Development notes

An example use case is to avoid page crashes when loading JSON files in place of a real API, without needing to export all of the files for each possible API request. In practice this then allows the metadata panel to be used with a single `main` JSON file for testing while falling back gracefully for missing data. 

All request errors are still visible in the console logs as normal, we could potentially add more information through an additional `console.error` if needed. Later on if needed, we could also add an `error` flag to the fallback response.

We could also consider extending this behaviour to the `main` route too, but for now it remains a special case to retain the previous behaviour (an informative error), avoiding this being a breaking change for that route.

## QA notes

Using a `public/main` JSON test file, test all features that use API requests (first load, pipeline select, metadata panel).

## Checklist

## Legal notice

### DIFF
--- a/src/store/load-data.js
+++ b/src/store/load-data.js
@@ -2,16 +2,25 @@ import { json } from 'd3-fetch';
 import { getUrl } from '../utils';
 
 /**
- * Asynchronously load and parse data from json file using d3-fetch
+ * Asynchronously load and parse data from json file using d3-fetch.
+ * Throws an error if the request for `main` fails.
+ * For requests other than `main`, returns the given or default fallback response.
  * @param {string} path JSON file location. Defaults to main data url from config.js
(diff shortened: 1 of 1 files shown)


studioswong:
Great stuff @bru5 !

tested on running viz on a JSON file with setting the `meta` flag to true ( which is the scenario this fix will cater for, which is enabling the metadata panel while running viz against a JSON file instead of a Kedro project), and it works well!

studioswong (approved):

PR #334, opened 2020-12-14
[KED-2182]Replace newgraph feature flag with a new 'oldgraph' flag

studioswong:
## Description

related ticket: https://jira.quantumblack.com/browse/KED-2182

This PR replaces the `newgraph` flag to the new `oldgraph` flag, where on setting it as 'true' will display the previous version of the graph (Dagre graphing algorithm). 

## Development notes

The following have been done:

1. Renamed the 'newgraph' flag as 'oldgraph' and set the default to false
2. Update all instances of `newgraph` to `oldgraph` within the codebase and update the logic if neccesary
3. Update tests
4. update readme docs (@richardwestenra @yetudada let me know what you think about the description of the renaming of the flag in the docs)

## QA notes

This feature can be tested by triggering the 'oldgraph' flag to true. Please note that you will need to clear your localstorage before testing this feature as sometimes the browser would not pick up the new flag setting ( depending on the version of your browser.) 

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

## Legal notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #329, opened 2020-12-11
Add params to task node's metadata

limdauto:
## Description

This PR exposes a `parameters` field to a node's metadata. It contains the names and values of the `parameters` inputs to a node.

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

## Legal notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #327, opened 2020-12-08
[KED-2314] Support launching a development server against a real pipeline

limdauto:
## Description

Support launching a development server against a real Kedro pipeline by adding a `start:api` command to npm. Feel free to add it to the main `npm start` if you'd like :) 

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

## Legal notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #326, opened 2020-12-07
[KED-2319] Check None on loading params

limdauto:
## Description

Sometimes the `/node` endpoint returns 500 due to the node object being None. Will chat to @bru5 to see why it is the case. In the mean time, I have added some defensive None checking to the endpoint.

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

## Legal notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #336, opened 2020-12-16
3.8.0

richardwestenra:
# Release 3.8.0

- Finish the new node metadata side panel, which allows you to switch between different modular pipelines. This feature is no longer hidden behind a flag, and the 'meta' flag has been removed. (#293, #309, #312, #320, #321, #326, #295, #329, #333)
- Enable the new graphing layout algorithm by default, and remove the 'newgraph' feature flag. If necessary, you can still revert back to the old layout algorithm by enabling the 'oldgraph' flag. (#334, #335)
- Add experimental flagged feature to allow lazy-loading of sidebar node-list rows on scroll. This improves performance by not forcing the app to render the entire node-list on larger graphs. This feature is disabled by default, but can be enabled using the 'lazy' feature flag. (#307)
- Use CSS custom properties for theme colours (#301, #319)

- Update Kedro-UI to v1.1.4. This enables us to improve webfont-loading detection, add a transition-out for closing the pipeline dropdown, add an active pipeline menu-option border-left colour, and improve accessibility when disabling the pipeline dropdown. (#325)
- Support launching a development server against a real pipeline. This is still a work-in-progress, and we will announce when it can be used. (#318, #327)
- Unify backend and frontend test data, to help prevent bugs appearing in future due to mismatched API schemas between frontend & backend (#298)
- Fix tag list icon hover state styling (#316)
- Update various dependency versions via Dependabot (#315, #330, #331)
- Fix linters (#323)
- Add default fallback response for non-main API calls (#328)
- Remove get_project_context(), now that we no longer support old versions of Kedro (<0.15) (#324)
- Add a 'private' flag prop, to hide flags for in-development features from the flags console announcement (#322)
- Investigate the root cause of nodes being undefined in Safari (#310)
- Fix bug that caused missed click on the flowchart (#317)
- Change demo.mock to a .json file and update tests (#314)
- Disable Python 3.6/3.7 jobs in daily CI workflow (#313)
- Batch tag actions to improve performance when toggling multiple tags (#308)

## Legal notice

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -23,13 +23,13 @@ Please follow the established format:
 
 - Finish the new node metadata side panel, which allows you to switch between different modular pipelines. This feature is no longer hidden behind a flag, and the 'meta' flag has been removed. (#293, #309, #312, #320, #321, #326, #295, #329, #333)
 - Enable the new graphing layout algorithm by default, and remove the 'newgraph' feature flag. If necessary, you can still revert back to the old layout algorithm by enabling the 'oldgraph' flag. (#334, #335)
-- Lazy-load sidebar node-list rows on scroll. This will improve performance by not forcing the app to render the entire node-list on larger graphs. This feature is disabled by default, but can be enabled using the 'lazy' feature flag. (#307)
-- Support launching a development server against a real pipeline (#318, #327)
+- Add experimental flagged feature to allow lazy-loading of sidebar node-list rows on scroll. This improves performance by not forcing the app to render the entire node-list on larger graphs. This feature is disabled by default, but can be enabled using the 'lazy' feature flag. (#307)
 - Use CSS custom properties for theme colours (#301, #319)
 
 ## Bug fixes and other changes
 
 - Update Kedro-UI to v1.1.4. This enables us to improve webfont-loading detection, add a transition-out for closing the pipeline dropdown, add an active pipeline menu-option border-left colour, and improve accessibility when disabling the pipeline dropdown. (#325)
+- Support launching a development server against a real pipeline. This is still a work-in-progress, and we will announce when it can be used. (#318, #327)
 - Unify backend and frontend test data, to help prevent bugs appearing in future due to mismatched API schemas between frontend & backend (#298)
 - Fix tag list icon hover state styling (#316)
 - Update various dependency versions via Dependabot (#315, #330, #331)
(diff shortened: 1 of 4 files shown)


RELEASE.md
```diff
@@ -17,6 +17,32 @@ Please follow the established format:
...
 <!-- Add release notes for the upcoming release here -->
 
+# Release 3.8.0
+
+## Major features and improvements
+
+- Finish the new node metadata side panel, which allows you to switch between different modular pipelines. This feature is no longer hidden behind a flag, and the 'meta' flag has been removed. (#293, #309, #312, #320, #321, #326, #295, #329, #333)
+- Enable the new graphing layout algorithm by default, and remove the 'newgraph' feature flag. If necessary, you can still revert back to the old layout algorithm by enabling the 'oldgraph' flag. (#334, #335)
+- Lazy-load sidebar node-list rows on scroll. This will improve performance by not forcing the app to render the entire node-list on larger graphs. This feature is disabled by default, but can be enabled using the 'lazy' feature flag. (#307)
+- Support launching a development server against a real pipeline (#318, #327)
```
studioswong, line 27: I don't think we should mention this as a feature since this isn't tried and tested within our team yet - we are still investigating the error that were returned from running the dev server using the new bash command, it is definitely worth more testing, and if successful, more detailed documentation on the readme / contribution docs before we announce this as part of the major feature.

richardwestenra: I could move it to the minor feature list or mention that it's still a work-in-progress, but I think it should be mentioned, because as a general rule we always mention things that are included in a release, even works-in-progress and unfinished things, because I see the release notes partly as a first port of call for someone trying to debug a breakage.

yetudada: Agreed with Susanna on this.

yetudada (approved):
I can't believe this is happening. 🏆

RELEASE.md
studioswong: sure thing, moving it to minor feature would be good - it might also be worth to mention that it is WIP?

richardwestenra: @yetudada I still think we should mention it (for the above reasons), so would it be okay to just edit it to mention that it's a WIP?

yetudada, line 27: Would something like this work? 
```suggestion
- Support launching a development server against a real pipeline (#318, #327). This development server is still a work-in-progress, and we will announce when it can be used. 
```

richardwestenra: Done, I've added something very close to this

richardwestenra: Done

studioswong (approved):
awesome 💯  - let's ship it!

PR #337, opened 2020-12-17
Disable internal CSS in exported SVGs to fix error

richardwestenra:
## Description

Due to a browser CORS rule, stylesheet cssRules cannot be read from external stylesheets, meaning that the google webfont stylesheet throws an error and breaks when exporting images.

I have [an open PR on svg-crowbar to fix this issue](https://github.com/cy6erskunk/svg-crowbar/pull/276), but in the meantime I am temporarily disabling this feature to prevent the error. This will signficantly increase exported SVG filesizes, but it will at least work without throwing an error.

## Development notes

I've temporarily changed this line. Once the linked PR is merged, I will revert this change, but will probably set it so that it uses `css: 'internal'` for SVGs only, and will use `css: 'inline'` for PNGs.

## QA notes

Export SVGs/PNGs

## Checklist

- [ ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

## Legal notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #339, opened 2020-12-18
Update SVG-Crowbar to fix errors

richardwestenra:
## Description

We were originally using SVG-Crowbar v0.6.0, but started experiencing a critical CORS error with internal CSS in exported images, which was patched in #337. This PR updates the dependency version and reverts the patch.

## Development notes

SVG-Crowbar v0.6.3 [fixes the CORS issue](https://github.com/cy6erskunk/svg-crowbar/pull/276), so we can revert the fix in #337.
However v0.6.2 introduced [a call stack size error](cy6erskunk/svg-crowbar#278) when exporting PNGs, which was resolved in v0.6.4.
However, v0.6.4 introduced [an issue with PNG cropping](cy6erskunk/svg-crowbar#75) that has been resolved in 0.6.5.
So in summary, we have now updated to v0.6.5 now that all of these errors have been resolved.

I've also added a small fix for console errors when exporting a null graph. If the graph doesn't exist and graphSize is empty, which can happen if the graph is still loading/rendering, or if all the filters are applied, then NaN console errors will be thrown. This commit prevents these errors.

## QA notes

You will need to update/install the SVG-Crowbar dependency to test this fix.

## Checklist

## Legal notice

### DIFF
--- a/src/components/export-modal/export-graph.js
+++ b/src/components/export-modal/export-graph.js
@@ -21,9 +21,13 @@ const exportGraph = ({ format, theme, graphSize, mockFn }) => {
   clone.classList.add('kedro', `kui-theme--${theme}`, 'pipeline-graph--export');
 
   // Reset zoom/translate
-  let width = graphSize.width + graphSize.marginx * 2;
-  let height = graphSize.height + graphSize.marginy * 2;
-  clone.setAttribute('viewBox', `0 0 ${width} ${height}`);
+  let width, height;
+  const hasGraph = graphSize.width && graphSize.height;
+  if (hasGraph) {
+    width = graphSize.width + graphSize.marginx * 2;
+    height = graphSize.height + graphSize.marginy * 2;
+    clone.setAttribute('viewBox', `0 0 ${width} ${height}`);
+  }
   clone.querySelector('#zoom-wrapper').removeAttribute('transform');
 
   // Impose a maximum size on PNGs because otherwise they break when downloading
@@ -32,8 +36,10 @@ const exportGraph = ({ format, theme, graphSize, mockFn }) => {
     width = Math.min(width, maxWidth);
     height = Math.min(height, maxWidth * (height / width));
   }
-  clone.setAttribute('width', width);
-  clone.setAttribute('height', height);
+  if (hasGraph) {
+    clone.setAttribute('width', width);
+    clone.setAttribute('height', height);
+  }
 
   const style = document.createElement('style');
   if (format === 'svg') {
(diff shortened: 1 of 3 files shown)


studioswong (approved):
Thanks for the quick fix!

richardwestenra:
@studioswong Thanks, but this isn't fixed yet, due to the new bug in SVG-Crowbar ~v0.6.2! We won't be able to update it to merge this PR and properly resolve this issue until that bug is fixed. Hence this is just a draft PR

studioswong:
Thanks for the fix @richardwestenra , and great work for keeping a close eye and enduring all the SVG crowbar saga 👍 

I just checked out the branch and both exports (SVG and PNG) works well 👍 

Tried exporting the graph with all filters applied, while the export works fine ( it just exports an empty image with your theme background), I noticed that the console still throws errors for `NaN` and `Infinity` values ( see below). I guess we would need to apply a check for those x and y values as well to eliminate those errors? 

image

All in all, thanks for the work!

src/components/export-modal/export-graph.js
```diff
@@ -21,9 +21,13 @@ const exportGraph = ({ format, theme, graphSize, mockFn }) => {
   clone.classList.add('kedro', `kui-theme--${theme}`, 'pipeline-graph--export');
 
   // Reset zoom/translate
-  let width = graphSize.width + graphSize.marginx * 2;
-  let height = graphSize.height + graphSize.marginy * 2;
-  clone.setAttribute('viewBox', `0 0 ${width} ${height}`);
+  let width, height;
+  const hasGraph = graphSize.width && graphSize.height;
```
studioswong, line 25: We might need to add the checks for valid values of `width` and `height` to eliminate those console errors?

richardwestenra: good point! fixed 👍

src/components/export-modal/export-graph.js
```diff
@@ -32,8 +36,10 @@ const exportGraph = ({ format, theme, graphSize, mockFn }) => {
     width = Math.min(width, maxWidth);
     height = Math.min(height, maxWidth * (height / width));
   }
-  clone.setAttribute('width', width);
-  clone.setAttribute('height', height);
+  if (hasGraph) {
+    clone.setAttribute('width', width);
+    clone.setAttribute('height', height);
+  }
```
studioswong, line 42: it might be worth to rethink about the `else` scenerio for the empty flowchart - we could perhaps apply a standard `width` and `height` in the meantime for the sake of allowing us to still export and empty graph while we look into the flow when exporting an empty graph?

richardwestenra: I think it works as is without needing an explict height - I don't think we need to think any more about this extreme edge case. I tidied up the console errors for the sake of completeness as I don't like exposing uncaught errors, but I don't think it needs anything else tbh.

studioswong: sure thing, just tested and the errors are gone, so if it works without the need to manually specify a width and height then it's all good.

PR #346, opened 2021-01-07
[KED-2328] Add code panel

bru5:
## Description

Adds a code panel to the metadata sidebar.

code sidebar screenshot

## Development notes

- Adds the code panel component
- Uses highlight.js for code highlighting
- Handles both light and dark highlighting themes
- Improves the responsive approach so that the chart moves and scales to fit
- Adds a feature flag named `code`

## QA notes

Testing requires a live Kedro project running through the viz API.

The `code` feature flag must be enabled before testing.

When a function node is clicked and code is available from the API, the code panel toggle will appear at the top of the metadata panel. Other node types do not show the toggle.

Clicking the toggle will open the panel and show highlighted code.

Test at various browser sizes and with different sidebar configurations, as the chart will responsively scale to fit the space available until a minimum chart size.

The code panel is intentionally a fixed width, if code is wider than the panel it can be scrolled horizontally.

When the clicked node is changed or removed, the code panel will close itself.

Test using both themes.

## Checklist

## Legal notice


src/components/flowchart/index.js
```diff
@@ -213,7 +218,10 @@ export class FlowChart extends Component {
         const margin = 500;
         this.zoomBehaviour.translateExtent([
           [-sidebarWidth / scale - margin, -margin],
-          [width + margin + metaSidebarWidth / scale, height + margin]
+          [
+            width + margin + (metaSidebarWidth + codeSidebarWidth) / scale,
```
studioswong, line 222: With the introduction of the 2 new props to calculate the flowchart size, it is worth to add a test for flowchart to test the recalculation of the scale with different chartsize values?

src/components/metadata/index.js
```diff
@@ -7,22 +7,39 @@ import CopyIcon from '../icons/copy';
...
 import { toggleNodeClicked } from '../../actions/nodes';
+import { toggleCode } from '../../actions';
 import './styles/metadata.css';
 
 /**
  * Shows node meta data
  */
-const MetaData = ({ visible = true, metadata, onToggleNodeSelected }) => {
+const MetaData = ({
+  visible = true,
```
studioswong, line 24: I wonder what is this `visible` value for?

src/components/metadata/index.js
```diff
@@ -7,22 +7,39 @@ import CopyIcon from '../icons/copy';
...
+  // Hide code panel when selected metadata changes
+  useEffect(() => onToggleCode(false), [metadata, onToggleCode]);
+
   const isTaskNode = metadata?.node.type === 'task';
-  const isParametersNode = metadata?.node.type === 'parameters';
   const isDataNode = metadata?.node.type === 'data';
+  const isParametersNode = metadata?.node.type === 'parameters';
+
+  const hasCode = Boolean(metadata?.code);
+  const showCodePanel = codeFlag && visible && visibleCode && hasCode;
```
studioswong, line 41: same as above - what does `visible` check for in this case, given that it is assigned `true` anyways?

src/components/metadata/index.js
```diff
@@ -35,116 +52,147 @@ const MetaData = ({ visible = true, metadata, onToggleNodeSelected }) => {
...
-                    className="pipeline-metadata__copy-button"
-                    icon={CopyIcon}
-                    onClick={onCopyClick}
+    <>
+      <MetaDataCode visible={showCodePanel} value={metadata?.code} />
+      <div className={modifiers('pipeline-metadata', { visible }, 'kedro')}>
+        {metadata && (
+          <>
+            <div className="pipeline-metadata__header-toolbox">
+              {showCodeSwitch && (
```
studioswong, line 62: With the introduction of the codeSwitch there is a need to should add / update the metadata tests to ensure the toggle button only shows up for task type nodes.

studioswong:
Tested with the data from the spaceflight project, nice work @bru5 ! Love the automatic resizing of the code panel for different screen sizes. 

I've added a few comments with regards to adding tests for the newly introduced toggle button and screen resizing - overall it looks good!

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,178 @@
...
+  position: absolute;
+  top: 7.1em;
+  right: $metadata-sidebar-width-open;
+  bottom: 0;
+  left: 0;
+  display: block;
+  padding: 0 0 0 36px;
+  overflow: auto;
+  color: var(--color-metadata-code-text);
+  font-size: 12.5px;
```
richardwestenra, line 90: Can we use em instead of px for font-sizes? Helps keep them all relative to the font-size on the `kedro` class.

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,178 @@
...
+}
+
+.pipeline-metadata-code__code {
+  position: absolute;
+  top: 7.1em;
+  right: $metadata-sidebar-width-open;
+  bottom: 0;
+  left: 0;
+  display: block;
+  padding: 0 0 0 36px;
```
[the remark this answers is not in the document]
richardwestenra: Update: see suggestion below

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,178 @@
...
+  line-height: 1.8;
+  opacity: 1;
+  transition: opacity 0.4s ease 0.4s;
+
+  .pipeline-metadata-code--no-visible & {
+    opacity: 0;
+    transition: opacity 0.3s ease;
+  }
+
+  & * {
```
richardwestenra, line 100: doesn't really matter but you don't need the `&` here: `* {` will suffice.

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,178 @@
...
+    opacity: 1;
+  }
+
+  &--checked:before {
+    background: $color-primary;
+  }
+
+  &--checked:after {
+    background: $color-primary;
+    transform: translate(0.9em, -50%);
```
richardwestenra, line 176: Reckon it might be better to use `%` rather than `em` for horizontal translation?

src/components/metadata/styles/metadata.scss
```diff
@@ -24,9 +24,17 @@
   padding: 6px 0 24px 0;
   overflow-y: auto;
   background: var(--color-metadata-bg);
+
+  // Subtle shadow when overlapping sidebar
+  box-shadow: 0 0 18px rgba(0, 0, 0, 0.04);
```
richardwestenra, line 29: 1. This should be different for light and dark themes.
2. This element already has a themed box-shadow applied (via `@extend %sidebar;`), which this rule is overriding. Is this shadow better than the default one?
  image

richardwestenra: note: just realised that the `%sidebar` rule is broken. will fix it.

src/components/sidebar/sidebar.scss
```diff
@@ -8,16 +8,21 @@
...
   transition: transform ease 0.4s;
+  @media (min-width: $metadata-sidebar-code-breakpoint) {
+    z-index: 3;
+  }
 
   @media (min-width: $sidebar-width-breakpoint) {
     width: $sidebar-width-open;
-    $sidebar-ui-width: $sidebar-width-open - $sidebar-width-closed;
     transform: translateX(-$sidebar-ui-width);
+    $sidebar-ui-width: $sidebar-width-open - $sidebar-width-closed;
```
richardwestenra, line 25: This moves the variable declaration to after where it's used, which causes a Sass error, which means that you can't start the server

src/components/sidebar/sidebar.scss
```diff
@@ -8,16 +8,21 @@
...
+
+  // Ensures sidebar tooltips are above code panel
   z-index: 1;
   display: flex;
   width: 100%;
   transform: translateX(calc(-100% + #{$sidebar-width-closed}));
   transition: transform ease 0.4s;
+  @media (min-width: $metadata-sidebar-code-breakpoint) {
+    z-index: 3;
+  }
```
richardwestenra, line 20: Don't really mind but there should probably be a space above this section.

Also fyi I don't really mind but the tooltips still go below the regular metadata panel at 700-950px widths, but I don't think it matters.

src/config.js
```diff
@@ -30,6 +32,11 @@ export const flags = {
     description: 'Improved sidebar performance',
     default: false,
     icon: '😴'
+  },
+  code: {
+    description: 'Show the code panel',
+    default: false,
+    icon: '{}'
```
richardwestenra, line 39: Could also use #️⃣ if you wanted to stick with emoji but I think the brackets work well

src/selectors/layout.js
```diff
@@ -32,35 +38,48 @@ export const getGraphInput = createSelector(
...
  * and add some useful new ones
  */
 export const getChartSize = createSelector(
-  [getVisibleSidebar, getVisibleMetaSidebar, state => state.chartSize],
-  (visibleSidebar, visibleMetaSidebar, chartSize) => {
+  [
+    getVisibleSidebar,
+    getVisibleMetaSidebar,
+    getVisibleCode,
+    state => state.chartSize
```
richardwestenra, line 53: I didn't notice this before but this should really be `getChartSize` and defined above.

richardwestenra (changes requested):

src/components/metadata/index.js
bru5: Yeah as @richardwestenra says, see also `mapStateToProps` in that function as that's how it gets controlled.

src/components/metadata/index.js
bru5: The `visible` prop controls if the panel is shown, which is decided by something outside of the component.

src/components/metadata/index.js
bru5: Agreed I'll be adding some additional tests around this.

src/components/metadata/metadata-code-toggle.js
```diff
@@ -0,0 +1,31 @@
...
+      disabled={!hasCode}
+      onChange={onChange}
+    />
+    <label
+      className={modifiers('pipeline-metadata__code-toggle-label', {
+        checked: hasCode && showCode
+      })}
+      htmlFor="pipeline-metadata__code-toggle-input">
+      Show Code
+    </label>
```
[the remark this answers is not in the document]
bru5: Agreed I've discussed with @GabrielComymQB and he is currently testing the branch out so we should discuss after.

I'm not keen on the flow with it closing all the time but probably fine for first flagged release, I've suggested moving towards using separate toolbar buttons for each panel going forwards to allow the user more control.

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,178 @@
...
+
+  .hljs-keyword {
+    color: var(--color-metadata-code-text);
+  }
+}
+
+.pipeline-metadata-code pre {
+  width: 100%;
+  margin: 0;
+}
```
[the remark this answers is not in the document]
bru5: Good catch, I've added that.

src/components/metadata/styles/metadata.scss
richardwestenra: update: I fixed this in a separate PR which was merged to main yesterday

src/components/flowchart/index.js
```diff
@@ -66,7 +67,7 @@ export class FlowChart extends Component {
     const { chartZoom } = this.props;
     const changed = (...names) => this.changed(names, prevProps, this.props);
 
-    if (changed('visibleSidebar')) {
+    if (changed('visibleSidebar', 'visibleCode', 'visibleMetaSidebar')) {
       this.updateChartSize();
     }
```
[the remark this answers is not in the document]
bru5: Yeah I've been trying to resolve this, looks like what it really needs to do is respect a minimum zoom level rather than just a width, will get back on this.

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,178 @@
...
+  }
+}
+
+.pipeline-metadata__code-toggle {
+  margin-top: 0.3em;
+  margin-bottom: 2em;
+}
+
+.pipeline-metadata__code-toggle-input {
+  display: none;
```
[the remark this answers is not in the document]
bru5: Thanks I was meaning to ask about how that worked, it's now in then.

src/components/metadata/metadata-code.js
```diff
@@ -0,0 +1,48 @@
...
+  sidebarVisible
+}) => {
+  const codeRef = useRef();
+  const [highlighted, setHighlighted] = useState();
+
+  useEffect(() => {
+    const detected = hljs.highlightAuto(value);
+    const language = detected.language || detected.second_best.language;
+    setHighlighted(hljs.highlight(language, value).value);
+  }, [codeRef, value]);
```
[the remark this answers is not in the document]
bru5: Ah good catch I can remove the effect after since I switched it away from hljs rendering itself.

So it could just be synchronous, but actually I've swapped it out to `useMemo` instead for performance.

src/components/metadata/metadata-code.js
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <div
+      className={modifiers(
+        'pipeline-metadata-code',
+        { visible, sidebarVisible },
+        'kedro'
+      )}>
+      <h2 className="pipeline-metadata-code__title">Code block</h2>
+      <code className="pipeline-metadata-code__code">
+        <pre ref={codeRef} dangerouslySetInnerHTML={{ __html: highlighted }} />
```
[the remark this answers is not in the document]
bru5: I see what you mean but any React wrapper for hljs will still be using this approach internally, either that or letting hljs control its own rendering (which would be worse in a React context).

I'm not sure if there are any pure React highlighters that are up to the same standard all round, but either way I think this approach is fine really so long as it's kept abstracted away like this.

src/components/metadata/metadata-code-toggle.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import modifiers from '../../utils/modifiers';
+import './styles/metadata.css';
+
+/**
+ * Shows a toggle button for code panel
+ */
+const MetaCodeToggle = ({ showCode, hasCode, onChange }) => (
```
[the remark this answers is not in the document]
bru5: For sure I've been meaning to do this, will get back.

src/components/metadata/index.js
```diff
@@ -7,22 +7,39 @@ import CopyIcon from '../icons/copy';
...
+  metadata,
+  codeFlag,
+  visibleCode,
+  onToggleCode,
+  onToggleNodeSelected
+}) => {
   const [showCopied, setShowCopied] = useState(false);
 
+  // Hide code panel when selected metadata changes
+  useEffect(() => onToggleCode(false), [metadata, onToggleCode]);
```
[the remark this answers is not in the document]
bru5: Yeah I addressed this in your first comment, let's keep the discussion there.

src/components/metadata/styles/metadata.scss
bru5: Ah thanks I thought it had been removed, I'll remove this one then.

src/components/sidebar/sidebar.scss
bru5: Weird, not sure why this line was missing (or that variable was defined in the wrong place) maybe a rogue merge or formatter somewhere, I'll keep an eye on it.

Yeah I think for now that case isn't very common (compared to overlapping the code panel), so we can leave it for now?

src/config.js
bru5: Yeah I dug around the emojis but none are quite as nice as the curlys :)

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,178 @@
...
+    width: 0.8em;
+    height: 0.8em;
+    border-radius: 50%;
+    opacity: 0.7;
+    transition: opacity 0.15s ease, transform 0.15s ease;
+  }
+
+  &:hover:after {
+    opacity: 1;
+  }
```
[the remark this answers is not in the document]
richardwestenra: Sorry I didn't do a deep dive on the exact styling, hoped you'd experiment more. But I'd suggest replacing the outline with
```scss
box-shadow: 0 0 0 3px $color-link;
```
which should be a bit better. However the opacity on the pseudo-selectors needs changing to be a rgba background--color opacity for this to work. e.g.
```scss
background: rgba($color-bg-light-1, 0.3);
```

 Are you happy with this? It means changing the transition from opacity to background-opacity but I figure it's small/fast enough that it's barely noticeable.

image

src/components/metadata/metadata-code-toggle.js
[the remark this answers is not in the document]
bru5: Forgot to add, this one is ready.

richardwestenra (approved):
I'd still like [some styling improvements on the toggle switch focus state](https://github.com/quantumblacklabs/kedro-viz/pull/346#discussion_r558417474), but overall you've all my comments, great job, thanks! This is awesome.

src/components/metadata/styles/metadata-code.scss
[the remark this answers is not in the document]
bru5: Sure this is now implemented.

src/components/flowchart/index.js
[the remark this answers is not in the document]
bru5: I've now merged the view refactor work from #363 and implemented the change to account for the code panel and all is working together nicely. 

I've also added some flowchart tests to check the extents are being applied as expected with sidebars off and on.

src/components/metadata/metadata-code-toggle.js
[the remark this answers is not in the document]
richardwestenra: ok cool so we'll look at this later? I'll make a ticket for it.

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,184 @@
...
+
+  @media print {
+    display: none;
+  }
+}
+
+.pipeline-metadata-code__title {
+  flex-grow: 1;
+  margin: 0;
+  margin: 2.1em 36px 1.8em 36px;
```
studioswong, line 73: This should align with using the same unit (`em` ) - @richardwestenra ?

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,184 @@
...
+
+  .hljs-keyword {
+    color: var(--color-metadata-code-text);
+  }
+}
+
+.pipeline-metadata-code pre {
+  display: inline-block;
+  margin: 0;
+  padding: 0 36px 36px;
```
studioswong, line 110: same as above - should be `3.6em`?

src/components/metadata/styles/metadata-code.scss
richardwestenra: Why do you say that? I think this makes sense to use em for the vertical margin (which is based on the relative font-size), but use px for the horizontal margin (which is based on the container's padding, which is based on the overall container width, which is handled in pixels).

src/components/metadata/styles/metadata-code.scss
richardwestenra: As above, I'm not sure that this needs to be in em, because in this case it's not based on font-size but on overall panel width, which is handled in px like most of our layout. There's an argument to be made that literally everything should be done in em, but that's not where we're currently at.

src/components/metadata/styles/metadata-code.scss
bru5: Yeah exactly it's a bit nuanced I know but my approach to spacing usually depends on context. I tend to use `em` for vertical spacing as that's naturally related to typography and vertical rhythm. But really it's not a big deal in practice anymore, but I still find it's still a good way of expressing the intent.

src/components/metadata/styles/metadata-code.scss
bru5: Exactly that thanks for clarifying @richardwestenra

src/components/metadata/styles/metadata-code.scss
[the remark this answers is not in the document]
bru5: Good point I've was able to refactor the positioning approach to flexbox with the same scrolling behaviour, so this should also resolve this potential issue too.

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,184 @@
...
+  margin: 0;
+  margin: 2.1em 36px 1.8em 36px;
+  font-weight: normal;
+  font-size: 1.7em;
+  line-height: 1.6;
+}
+
+.pipeline-metadata-code__code {
+  position: absolute;
+  top: 7.1em;
```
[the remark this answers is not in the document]
bru5: Sure I managed to refactor this to flexbox and retain the scrolling behaviour, so this should now be ready should the title change from being a fixed string.

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,182 @@
...
+  font-weight: normal;
+  font-size: 1.7em;
+  line-height: 1.6;
+}
+
+.pipeline-metadata-code__code {
+  display: block;
+  flex-grow: 1;
+  overflow: auto;
+  color: var(--color-text);
```
richardwestenra, line 85: You've removed the new variable but left the declaration - adding `color: var(--color-text);` seems unnecessary to me because that's already the default color for the application so it just overrides the default value?

src/components/metadata/styles/metadata-code.scss
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  &:hover:after {
+    background-color: $color-bg-light-1;
+  }
+
+  &--checked:before,
+  &--checked:after {
+    background-color: rgba($color-primary, 0.7);
+  }
```
richardwestenra, line 173: The colours here when selected don't look like it did before, and the contrast is a bit low.

Before:
image

After:
image

You might need to alter the rgba opacity separately for the :before and :after elements when checked - maybe something like this?


```suggestion
  &--checked:before {
    background-color: rgba($color-primary, 0.3);
  }
  
  &--checked:after {
    background-color: rgba($color-primary, 0.7);
  }
```

richardwestenra (changes requested):

src/components/metadata/styles/metadata-code.scss
bru5: Ah I misread your earlier comment and thought that was the suggested change, I've removed this line.

src/components/metadata/styles/metadata-code.scss
bru5: Good catch, resolved this thanks.

richardwestenra (approved):
:shipit:🚢:shipit:🚢:shipit:🚢:shipit:🚢

src/components/metadata/styles/metadata-code.scss
studioswong: Thanks for the clarification @richardwestenra - I am quite confused as to our current preference of unit within the codebase - are we using px or em? My understanding is that it is either em or px as a consistent unit within the codebase, and if we are making exceptions for context then it's very easy for us to fall into the trap of individual interpretation?

richardwestenra: That's fair - I've added this to the list of things to add to the styleguide in KED-2413.

- We typically use `em` for font-sizes, and sizing things that might want to be set relative to the font-size - e.g. text margin-top/bottom, line-height, etc. This is partly because it is more accessible, and partly because it allows us to set the font-size in one place (i.e. on the `kedro` class) and then if we want to change later then everything else will update based on that value.
- We typically use `px` and `%` for layout - i.e. things that are more dependant on the size of the viewport than on the size of the font.

studioswong: I see, these 2 types of cases makes sense. In this case can we generally categorize to use `px` / `%`  for all layout-related fields ( i.e fields that would stay constant regardless of font sizing), and `em` for all font related / font-affected values?

PR #347, opened 2021-01-11
[KED-2212] Improve graph layout code quality, performance and docs

bru5:
## Description

Pure refactor to improve graph layout code quality, performance and docs.

## Development notes

Inverts control of `solver.js` to make constraints easier to understand and increase their flexibility.

Splits the graph `layout.js` further into more functions with improved comments for clarity.

There will be some performance improvement from reduced function calls, memory usage and garbage collection (most noticeable in graphs with a high number of edges).

There is still some way to go in fully documenting the graph engine but this will be a gradual process.

## QA notes

There should be no change in results, this is a pure refactoring.

Test all features that trigger graph layout, on various graphs, compare with the previous release.

## Checklist

## Legal notice


src/utils/graph/constraints.js
```diff
@@ -11,96 +11,145 @@ import { distance1d, greaterOrEqual, equalTo, subtract } from './common';
...
-  difference: subtract,
-  distance: distance1d,
-  operator: greaterOrEqual,
-  target: (a, b, co, constants) => constants.spaceY,
-  strength: () => 1,
-  weightA: () => 0,
-  weightB: () => 1,
-  required: true
+
+  solve: (constraint, constants) => {
```
richardwestenra, line 15: I know that the `solve` and `strict` methods are defined in multiple places, but I couldn't find any explanation anywhere about what it is they do. Do you mind adding some (JSDoc?) comments next to them explaining what they're for, and maybe what's different about each one and why?

src/utils/graph/constraints.js
```diff
@@ -11,96 +11,145 @@ import { distance1d, greaterOrEqual, equalTo, subtract } from './common';
...
-  target: (a, b, co, constants) => constants.layerSpace,
-  strength: () => 1,
-  weightA: () => 0,
-  weightB: () => 1,
-  required: false
+
+  solve: (constraint, constants) => {
+    const { a, b } = constraint;
+    const difference = a.y - b.y;
+    const target = constants.layerSpace;
```
richardwestenra, line 48: This `solve` method looks very very similar to the previous one - is it worth making a factory function for it and passing it an argument to create the resulting method? e.g. something like:

```es6
const getSolve = (targetKey) => {
  return (constraint, constants) => {
    const { a, b } = constraint;
    const difference = a.y - b.y;
    const target = constants[targetKey];

    if (difference >= target) {
      return;
    }

    const resolve = difference - target;
    a.y -= 0.5 * resolve;
    b.y += 0.5 * resolve;
  }
};


export const rowConstraint = {
  solve: getSolve('spaceY')
  ...
};


export const layerConstraint = {
  solve: getSolve('layerSpace')
  ...
};
```

If it's too much abstraction, fair enough.

src/utils/graph/constraints.js
```diff
@@ -1,4 +1,4 @@
-import { distance1d, greaterOrEqual, equalTo, subtract } from './common';
+import { Constraint, Operator, Strength } from 'kiwi.js';
 
 /**
  * Constraint base definitions.
```
richardwestenra, line 4: Do you mind expanding on this comment a little to explain what this means?

src/utils/graph/constraints.js
richardwestenra: Ah I found them in solver.js - it just wasn't included in this PR. It might be worth mentioning where it's used in the comments still

richardwestenra (approved):
Nice one! Approved with caveats - it might be worth adding a little more documenation if you don't mind!

src/utils/graph/constraints.js
bru5: Yeah so at the moment that documentation lives in the solver, I'd definitely like to start adding more details here but this API is currently in a bit of flux with more changes coming. 

Once that settles I think then would be best to further document, if that sounds reasonable?

src/utils/graph/constraints.js
bru5: In this case I think I'm aiming to avoid too much abstraction in favour of flexibility and clarity, but once the API solidifies a bit more I'll see if it makes sense to find a natural abstraction for the resulting approach.

src/utils/graph/constraints.js
bru5: There will be some follow up work coming soon which might change this up a bit, but then I think it will be stable enough to get documenting, if that's reasonable?

src/utils/graph/constraints.js
richardwestenra: sure thing

src/utils/graph/constraints.js
richardwestenra: Fair enough, I thought you might say that.

src/utils/graph/constraints.js
richardwestenra: ok, but it might be worth adding a JIRA ticket for these subsequent documentation improvements if they're out of scope for this PR

PR #353, opened 2021-01-18
[KED-2378] Local store of tags

studioswong:
## Description

related ticket: https://jira.quantumblack.com/browse/KED-2378

This ticket enables the app to reload with selected tags by saving the enabled tags in local storage. 

## Development notes

I have added the following:
1. added the new enabled tag field to be saved in localStorage
2. Added a test to specifically ensure the enabled tag field gets stored in localStorage (This test is mainly just to ensure that the object format of tags gets saved in localStorage)
3. Added a test to test the reload of the nodelist to ensure it reloads with the selected tags on page refresh 
4. Ensure the localstorage gets refreshed within the nodelist tests to ensure they are properly tested with the newly added localstorage value

## QA notes

select a few tags, refresh the window, and you should see the selected tags are still selected. 

## Checklist

## Legal notice

### DIFF
--- a/src/store/initial-state.test.js
+++ b/src/store/initial-state.test.js
@@ -29,6 +29,19 @@ describe('mergeLocalStorage', () => {
     window.localStorage.clear();
   });
 
+  it('overrides selected tag values in state with localstorage values if provided', () => {
+    const localStorageValues = {
+      tag: { enabled: 'large' }
+    };
+    saveState(localStorageValues);
+    expect(
+      mergeLocalStorage({
+        tag: { enabled: 'large' }
+      })
+    ).toMatchObject(localStorageValues);
+    window.localStorage.clear();
+  });
+
   it('does not add values if localStorage keys do not match state values', () => {
     const extraValues = {
       additional: 1,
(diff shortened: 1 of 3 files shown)


src/store/initial-state.test.js
```diff
@@ -29,6 +29,19 @@ describe('mergeLocalStorage', () => {
...
+    };
+    saveState(localStorageValues);
+    expect(
+      mergeLocalStorage({
+        tag: { enabled: 'large' }
+      })
+    ).toMatchObject(localStorageValues);
+    window.localStorage.clear();
+  });
+
```
richardwestenra, line 44: I might be wrong but this test seems redundant to me - it's just a modified version of the previous test. Is it necessary to check that this function specifically merges tag values, if we've already established that it already merges other values? The only difference that I can see is that this value is a nested object, in which case it should just be added to the previous test's input and expected result.

richardwestenra (commented):
I'm not sure whether to approve or request changes - overall this is great, but I think the test in initial-state.test.js should be removed.

src/store/initial-state.test.js
studioswong: I agree, I actually had the same thoughts when I added this test, but in the end decided to add it to partly demonstrate / document the type of data we are storing on the localStorage for tags, yet the intention behind this test is the same as the previous test.

In this case I will combine this with the previous one by adding this extra type of data in, and deleting this test.

richardwestenra (approved):

PR #355, opened 2021-01-21
Fix node-list-row keyboard focus states

richardwestenra:
## Description

1. Previously, hidden icons would stay hidden when focused with a keyboard, or would stay semi-transparent when partly faded, which would make their outline difficult to make out. By moving the opacity changes to the SVG children (e.g. path/rect elements), the blue outlines on the parent remain visible at 100% opacity even when the child element has reduced opacity. I've used `> *` on this to select only the immediate children and keep the rule agnostic as to element type.
2. For tag rows, there were two focusable elements that did the same thing on click (a button and a checkbox input). The button was also present on parent rows (e.g. category titles), but did nothing. I've removed the button where it does nothing or is otherwise superfluous, and replaced it with a div.
3. I've refactored a lot of the CSS to make the rules easier to read, mostly by making them indented using Sass, because the long lines of run-on rules were getting difficult to read and I think breaking them up in this way actually makes them easier to parse.

## Development notes

See commit titles + descriptions for a detailed breakdown of changes.

## QA notes

Please check the hover/focus/checked/unchecked/toggle-all styles for the node and tag lists, both with lazy-loading turned on/off (`?lazy=0`). I've tried to keep things working more or less as before, so please compare to the demo branch. The only major difference is that tabbing through the list on a keyboard will skip over unused/superfluous buttons.

## Checklist

## Legal notice

### DIFF
--- a/src/components/node-list/styles/_filter.scss
+++ b/src/components/node-list/styles/_filter.scss
@@ -35,7 +35,7 @@
 }
 
 // Hide all indicators when all unset
-.pipeline-nodelist__group--all-unset .pipeline-row__toggle--kind-filter {
+.pipeline-nodelist__group--all-unset .pipeline-row__toggle--kind-filter > * {
   opacity: 0;
 }
 
@@ -43,9 +43,11 @@
 .pipeline-nodelist__group--all-unset
   .pipeline-nodelist__row:hover
   .pipeline-row__toggle--kind-filter {
-  opacity: 0.55;
+  > * {
+    opacity: 0.55;
+  }
 
-  &:hover {
+  &:hover > * {
     opacity: 0.55;
   }
 }
@@ -60,25 +62,25 @@
 
 .pipeline-row__toggle-icon--kind-filter {
   // Fade indicator when unchecked or unset
-  &.pipeline-row__toggle-icon--unchecked,
-  &.pipeline-row__toggle-icon--unset {
+  &.pipeline-row__toggle-icon--unchecked > *,
+  &.pipeline-row__toggle-icon--unset > * {
     // condition 2: semi-bright indicator when checked but not hovered
     opacity: 0.55;
   }
 
   // brighten up indicator when hovering over toggle all tag button
   &.pipeline-row__toggle-icon--unchecked {
-    .pipeline-row-toggle-icon-visible:hover & {
+    .pipeline-row-toggle-icon-visible:hover & > * {
       opacity: 1;
     }
   }
 
-  &.pipeline-row__toggle-icon--checked {
+  &.pipeline-row__toggle-icon--checked > * {
     // condition 3: Bright indicator when checked but not hovered
     opacity: 0.75;
 
     // condition 4: Brightest indicator when checked and hovered on the entire row
-    .pipeline-nodelist__row--visible:hover & {
+    .pipeline-nodelist__row--visible:hover & > * {
       opacity: 1;
     }
   }
--- a/src/components/node-list/styles/_filter.scss
+++ b/src/components/node-list/styles/_filter.scss
@@ -17,20 +17,20 @@
 }
 
 // Fade row text when unchecked
-.pipeline-nodelist__row--unchecked .pipeline-nodelist__row__text--kind-filter {
+.pipeline-nodelist__row--unchecked .pipeline-nodelist__row__label--kind-filter {
   opacity: 0.55;
 }
 
 // Brighter row text when unchecked and hovered
 .pipeline-nodelist__row--unchecked:hover
-  .pipeline-nodelist__row__text--kind-filter {
+  .pipeline-nodelist__row__label--kind-filter {
   opacity: 0.8;
 }
 
 // Bright row text when all unset
 .pipeline-nodelist__group--all-unset
   .pipeline-nodelist__row--unchecked
-  .pipeline-nodelist__row__text--kind-filter {
+  .pipeline-nodelist__row__label--kind-filter {
   opacity: 1;
 }
 
--- a/src/components/node-list/styles/_filter.scss
+++ b/src/components/node-list/styles/_filter.scss
@@ -3,6 +3,31 @@
 @import '../../../styles/variables';
 @import './variables';
 
+//--- Text label ---//
+
+.pipeline-nodelist__row--unchecked {
+  // Fade row text when unchecked
+  .pipeline-nodelist__row__label--kind-filter {
+    opacity: 0.55;
+  }
+
+  // Brighter row text when unchecked and hovered
+  &:hover {
+    .pipeline-nodelist__row__label--kind-filter {
+      opacity: 0.8;
+    }
+  }
+
+  // Bright row text when all unset
+  .pipeline-nodelist__group--all-unset & {
+    .pipeline-nodelist__row__label--kind-filter {
+      opacity: 1;
+    }
+  }
+}
+
+//--- Toggle icon fills and strokes ---//
+
 .pipeline-row__toggle-icon--kind-filter {
   &.pipeline-row__toggle-icon--checked {
     fill: var(--color-nodelist-filter-indicator-on);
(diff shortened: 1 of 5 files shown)


src/components/node-list/styles/_filter.scss
```diff
@@ -14,72 +39,89 @@
...
-  opacity: 1;
-}
+/*
+ Parent (toggle all tags):
+ | row-hover | icon-hover | checked | opacity | ID |
+ |    🚫     |     🚫      |    🚫   |    0    | 0  |
+ |    ✅     |     🚫      |    🚫   |   0.3   | 1  |
+ |    ✅     |     ✅      |    🚫   |   0.55  | 2  |
+ |    🚫     |     🚫      |    ✅   |   0.55  | 2  |
+ |    ✅     |     ✅      |    ✅   |   0.75  | 3  |
```
studioswong, line 60: Love the clarity on this comment

richardwestenra: It says "Parent (toggle all tags)" as a title and then specifies "opacity" as one of the columns in the comment - what else do you need? I thought it was really obvious.

src/components/node-list/styles/_filter.scss
```diff
@@ -3,6 +3,31 @@
 @import '../../../styles/variables';
 @import './variables';
 
+//--- Text label ---//
+
+.pipeline-nodelist__row--unchecked {
+  // Fade row text when unchecked
+  .pipeline-nodelist__row__label--kind-filter {
+    opacity: 0.55;
```
studioswong, line 11: Perhaps it is worth to also set up a variable system for the text label opacity like what we have below?

richardwestenra: Can do if you think it necessary, but I didn’t set up a similar comment table or variable schema for the visibility (“toggle”) icons, because the CSS there is pretty simple and easy to understand, while on the filter icons it’s complex and hard to follow, so I found the table made it easier to refer to when refactoring the filter icons.

richardwestenra: also: ah I understand now! These rules explicitly _do not_ apply to the visibility icons, that’s why they’re on classes for the tag (“filter”) icons only.

src/components/node-list/styles/_filter.scss
richardwestenra: also: ah I understand now! These rules explicitly do not apply to the visibility icons, that’s why they’re on classes for the tag (“filter”) icons only.

src/components/node-list/styles/_filter.scss
```diff
@@ -14,72 +39,89 @@
...
 
-// Show indicator when all unset and row hovered
-.pipeline-nodelist__group--all-unset
-  .pipeline-nodelist__row:hover
-  .pipeline-row__toggle--kind-filter {
-  opacity: 0.55;
+$icon-opacity-0: 0;
+$icon-opacity-1: 0.3;
+$icon-opacity-2: 0.55;
+$icon-opacity-3: 0.75;
```
studioswong, line 73: Other than just defining these opacity variables just for the icons, given that those values are also used for the text labels,  it might be better to specify it just as an opacity variable (and later on applied to the classes that defines the cases for the toggle button and the label).

For example:

`$opacity-0: 0;`
`$opacity-1: 0.3;`
...
`$opacity-3: 0.75;`

src/components/node-list/styles/_filter.scss
```diff
@@ -3,6 +3,31 @@
...
+.pipeline-nodelist__row--unchecked {
+  // Fade row text when unchecked
+  .pipeline-nodelist__row__label--kind-filter {
+    opacity: 0.55;
+  }
+
+  // Brighter row text when unchecked and hovered
+  &:hover {
+    .pipeline-nodelist__row__label--kind-filter {
+      opacity: 0.8;
```
studioswong, line 17: I wonder whether we should align this opcity value with a value that was defined in the ruleset ( 0.75)? This way it provides more clarity & ailgnment about the opacity rules throughout node-list-row

src/components/node-list/styles/_filter.scss
```diff
@@ -14,72 +39,89 @@
...
+    stroke: $color-primary;
+  }
 }
 
-// Brighter row text when unchecked and hovered
-.pipeline-nodelist__row--unchecked:hover
-  .pipeline-nodelist__row__text--kind-filter {
-  opacity: 0.8;
-}
+//--- Toggle icon opacities ---//
```
studioswong, line 51: quick thought: while we are making the effort to properly set up the style rules, I wonder whether it's worth to separate the style classes for the toggle button out from `_filter.scss` to something like `_toggle.scss`?

src/components/node-list/styles/_filter.scss
```diff
@@ -14,72 +39,89 @@
...
-  .pipeline-row__toggle--kind-filter {
-  opacity: 0.55;
+$icon-opacity-0: 0;
+$icon-opacity-1: 0.3;
+$icon-opacity-2: 0.55;
+$icon-opacity-3: 0.75;
 
-  &:hover {
-    opacity: 0.55;
+.pipeline-row__toggle-icon--kind-filter {
```
studioswong, line 75: From my previous experience of working with these styles without prior context, one of the main things I find hard to grasp amongst our class name declarations is seperating classes for the 'parent' and 'child' toggle button. I believe this button refers to the 'parent' toggle button for tags - perhaps it is worth to name it as `.pipeline-row__toggle-icon--kind-filter-parent` to indicate this is for the parent toggle button?

studioswong:
Thanks for the refactoring and the thoughts behind this PR, this is no easy feat! 👍  One of the main things I like about this PR is its intent to establish a clear logic for the interaction states for the toggle & filter buttons, which having worked with these styles before, this will be very helpful for any further work on the sidebar down the line.

I've added some further suggestions / thoughts on how we can bring this current class declarations to be even more intuitive, especially given that we have now established the important logic of `toggle` vs `filter` buttons and `parent` vs `child` types for node-list , it might be worth to incorporating these terms as part of our class declarations? 

Anyways, great work, and thank you for revisiting / addressing those styling issues!

src/components/node-list/styles/_filter.scss
richardwestenra: The text labels use the same 0.55 value, but the rest is completely different (as it uses 0.55 / 0.8 / 1). So I figured since those are a different scale, it didn't make sense for these variables to be used there, and it would be confusing to use just one of them out of context.

I only used this variable schema here for the filter icon opacities because the classes and opacities in this specific case are complicated and I think this schema makes things easier to understand - we don't need to make every single CSS value a variable everywhere throughout the codebase, as abstractions introduce their own overhead. I just felt it was helpful here.

src/components/node-list/styles/_filter.scss
richardwestenra: I assume that it was set to this value for a reason, according to a design from @GabrielComymQB. 
It might be that the text looks different at a certain opacity than an icon at the same opacity. I'm disinclined to mess with it, that seems out of scope.

src/components/node-list/styles/_filter.scss
richardwestenra: We could do, but some of those ones currently apply to all of the node-list rows, and these override them. There's only a couple of rules that apply there, I'm not sure that it's necessary to have a new file for just the `--kind-toggle` CSS.

_However_, I think there's a decent case to be made for establishing new files for the row-label and row-icon CSS, so that the overrides are all in one place. Could do, but it makes reviewing this PR more difficult as all the styles would be copied to a new file so it's more difficult to compare them, so if we were to do that I'd be inclined to do it in a separate PR.

src/components/node-list/styles/_filter.scss
richardwestenra: I agree, but I'm already doing that here. In the updated CSS in this PR, if it doesn't use the `--parent` modifier then it's not intended to apply to just the parent. The line you're referring to here (line 75) applies to both the parent and child icons. I've got other rules below that use `--parent` and `--child` modifiers, where I wanted to specify one or the other (to be more explicit than it was previously).

richardwestenra: We have four declarations in this set that are intended to apply to both parents and children (so they don't include a specifying modifier), and three that are specifically for just parents or just children.

src/components/node-list/styles/_filter.scss
studioswong: sounds good, the process of migrating those style rules can be done in a seperate PR ( it might be worth to even combine this with the future work in the refactor of the sidebar to allow more generic data as props)

studioswong (approved):
Thanks for the work! Approving this to unblock the merging of this ticket - we can revisit the variables, namings, etc as dicussed here in future tickets for the refactoring of the sidebar ( there has been discussion of refactoring work for the sidebar to allow more generic data as props, which would be a good occasion to revisit any refactoring of the style rules as well.)

PR #358, opened 2021-01-26
[KED-2385] Refactor row toggle/label CSS, and fix row hover icon bug

richardwestenra:
## Description

1. Instead of having just the _row.scss file and then filter overrides in _filter.scss, instead create one file for all the text label scss (both toggle/filter kinds), and one for the toggle-icon scss (both kinds)
2. Fix toggle-icon--kind-filter parent row-hover bug: The toggle-all icon would decrease in opacity on row hover.
3. Rename node-list row toggle `--kind-toggle` to `--kind-element` to make the distinction between the two elements a bit more clear. We can rename them again later if necessary but I think this makes things a lot easier to understand.
4. Refactor node-list group creation selector to improve readability

## Development notes

I discussed the naming of the two node-list row item kinds with @bru5 at length while working on this PR. We want to strike a balance between intuitiveness and extensibility, which is why we've avoided being too specific in the names of the two kinds (e.g. calling filters 'category', which might not work with future usages, or calling them 'blue-dot' and 'eye-icon', which is too dependent on the presentation).

## QA notes

Please double-check the node-list hover/focus/selection states and see that nothing has changed (except for the blue dot checked parent row-hover state, which has been fixed).

## Checklist

## Legal notice


studioswong:
Thanks for the refactor, code looks good to me 👍 

One thing to confirm - I notice that you mentioned that the 'toggle-all' icon would decrease in opacity on hover, however it looks like it is currently set at opacity 0 and only shows up on hover of the toggle button ( see below), which is very different from the existing behaviour of the filter all row on hover. 
toggle 1

This is what we previously have: 
toggle 2

Is this new change something intended, as it seems to deviate from the behaviour of the toggle all icon for the eye ( where the 'toggle-all' eye icon would still show up on toggle of the entire row, and only increase in opacity on hover of the toggle button?) @GabrielComymQB

richardwestenra:
@studioswong Good catch, thanks! 🙌 

Fixed 😅

src/components/node-list/styles/_row-toggle.scss
```diff
@@ -0,0 +1,235 @@
...
+  '--row-selected-transparent',
+  $row-selected-light,
+  $row-selected-dark
+);
+
+.pipeline-row__toggle:before {
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: -100px;
```
studioswong, line 31: optional: I noticed that we use `em` in some places ( for fonts and margins) and `px` in others, do we normally have any preference on which to use?

studioswong (approved):
thanks for fixing - LGTF 👍

src/components/node-list/styles/_row-toggle.scss
richardwestenra: I would typically use `em` where we want it to be related/relative to font-size, and `px` where it should be absolute.

PR #361, opened 2021-01-29
Display warnings on large graphs

studioswong:
## Description

Related ticket: https://jira.quantumblack.com/browse/KED-2280

This ticket is to implement the 'large pipeline warning' user flow, where a warning message will  prompt the user to reselect tags and nodes to allow a smaller pipeline to be displayed.  The user can also choose to proceed with displaying the graph. 

image

Update: There is a new update to the design and the code has been updated accordingly as below:

image

The warning will look like the following under the light theme
image

## Development notes

There are 2 main areas of development for this PR: the set up of the modal, along with the styling; and the set up of a series of state, actions, reducers and selectors that allow the app to determine if the pipeline is a 'large' pipeline before the graph gets rendered. 

These are the main things I have included in this PR:
1. Set up of the modal and its related styling - I have used react-modal for the modal setup, while I have adopted very simple and straight forward styling rules. ( Please do let me know if there are better ways to incorporate inherit styles for the modal)
2. Introducing 4 new fields in the non-pipeline state that governs the display of the modal: ' isLarge', 'DisplayLargeGraph', 'NodesNo' and 'EdgesNo'. Both 'isLarge' is needed to listen to the condition of whether a pipeline is large as a first criteria to trigger the warning, while 'DisplayLargeGraph' is used to listen to the user's preference for whether to display the large pipeline. Both 'NodesNo' and 'EdgesNo' are for quick tally of the no. of Nodes and no. of edges to allow both figures to be displayed without the need for the entire set of nodes and edges to be passed on down to the modal component, thereby allowing the warning to be triggered almost immediately on load. 
3. refactor of actions in `graph.js` to control the display of the warning message without loading the loader and the graph
4. adding in selectors to expose the newly added fields
5. adding in reducers to allow changing the values of the newly added fields
6. Tests for the modal and the newly added actions
7. A field within the config to determine the max combination of nodes and edges to display the warning message

## QA notes

This is best to be tested with a large pipeline, which could be imitated by using the random data genereator. The expected behavior are as follows:
1. On first load, a large pipeline would trigger the warning message. 
2. On selecting different tags, if the combination still results in a large pipeline, the warning message will persist with updated no. of tags and edges in the warning message.
3. On selecting a tag that results in a smaller pipeline, the graph will immediately load.
4. On selecting 'render it anyway', the large pipeline will always render within the lifetime of the session until the app is reloaded. 

## Checklist

## Legal notice


src/components/modal/index.js
```diff
@@ -0,0 +1,76 @@
...
+      <Modal
+        isOpen={true}
+        style={customStyles}
+        overlayClassName="overlay"
+        ariaHideApp={false}>
+        <div className="layout">
+          <img src={goose} alt="goose" />
+          <div>
+            <div className="title">
+              Woahhhhh!
```
stichbury, line 39:
```suggestion

<div className="title">
              Warning!
              <br />
              Your pipeline is too chonky.
            </div>
            <div className="subtitle">
              Your pipeline contains <b>{nodesNo}</b> nodes and <b>{edgesNo}</b>{' '}
              edges.
              <br />
              Use the filters on the left to select tags and elements and display
              a subset of the pipeline, which is less chonky.
            </div>

```

src/components/modal/modal.scss
```diff
@@ -0,0 +1,59 @@
...
+  position: absolute;
+  top: 40px;
+  right: 40px;
+  bottom: 40px;
+  left: 40px;
+  background-color: $color-bg-dark-1;
+}
+
+div {
+  color: white;
```
bru5, line 22: All of these selectors must be scoped to a classname in BEM style, as this project isn't yet using any CSS modules so these styles will affect the whole app (the light theme for example font colour has changed to white on white)

src/actions/actions.test.js
```diff
@@ -207,4 +217,36 @@ describe('actions', () => {
...
+
+  it('should create an action to toggle isChonky', () => {
+    const expectedAction = {
+      type: TOGGLE_IS_CHONKY,
+      isChonky: true
+    };
+    expect(toggleIsChonky(true)).toEqual(expectedAction);
+  });
+
+  it('should create an action to update the no. of nodes for the modal display', () => {
```
bru5, line 237: I think maybe clearer to avoid abbreviations and go with the word `count` or `number`

bru5 (changes requested):

src/components/modal/index.js
studioswong: Thanks @stichbury ! The wording has been updated as per the latest design.

src/actions/actions.test.js
```diff
@@ -33,6 +33,16 @@ import {
   toggleTagActive,
   toggleTagFilter
 } from '../actions/tags';
+import {
+  TOGGLE_DISPLAY_CHONKY_GRAPH,
```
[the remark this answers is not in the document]
studioswong: Thanks @bru5 ! As per our discussion and final decision this morning, I had updated all references of the codebase to refer with the term 'large' instead of 'chonky'.

src/actions/graph.js
```diff
@@ -39,21 +92,42 @@ const chooseLayout = (instance, state) =>
 // Prepare new layout worker
 const layoutWorker = preventWorkerQueues(worker, chooseLayout);
 
+/**
+ * Formula to determine if the pipeline is chonky
```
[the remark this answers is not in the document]
studioswong: Thanks @bru5 , I'll add in some further comments

src/actions/graph.js
```diff
@@ -39,21 +92,42 @@ const chooseLayout = (instance, state) =>
...
+};
+
 /**
  * Async action to calculate graph layout in a web worker
  * whiled displaying a loading spinner
  * @param {Object} graphState A subset of main state
  * @return {function} A promise that resolves when the calcuation is done
  */
-export function calculateGraph(graphState) {
+export function calculateGraph(graphState, customChonkyLimit) {
```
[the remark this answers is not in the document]
studioswong: the word custom is used here mainly to indicate this value being a 'custom defined' value to allow the function to accept a figure that would overwrite the limit defined in the config. The main use case for this is for the test, where we would hope to avoid creating a large pipeline just to test this warning, whereby I can pass in a custom limit ( which would be a much smaller amount to fit in with the animals testing dataset) to allow this to be tested without impacting testing performance.

src/actions/graph.js
```diff
@@ -39,21 +92,42 @@ const chooseLayout = (instance, state) =>
...
   }
   return async function(dispatch) {
-    dispatch(toggleLoading(true));
-    const graph = await layoutWorker(graphState);
-    dispatch(toggleGraph(true));
-    dispatch(toggleLoading(false));
-    return dispatch(updateGraph(graph));
+    const { nodes, edges, displayChonkyGraph } = graphState;
+
+    const chonkyLimit = customChonkyLimit || chonkyNodeAmount;
```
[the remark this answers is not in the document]
studioswong: this refers to the threshold of a large pipeline as calculated by the formula defined in the config.

src/components/modal/index.js
```diff
@@ -0,0 +1,76 @@
...
+    right: 'auto',
+    bottom: 'auto',
+    color: '#FFFFFF',
+    backgroundColor: '#001521',
+    border: 'none'
+  }
+};
+
+export const ChonkyModal = ({
+  nodesNo,
```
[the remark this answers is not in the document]
studioswong: thanks @bru5 , I'll update the naming to reflect that

src/components/wrapper/index.js
```diff
@@ -12,26 +13,32 @@ import './wrapper.css';
...
+    <div
+      className={classnames('kedro-pipeline', {
+        'kui-theme--dark': theme === 'dark',
+        'kui-theme--light': theme === 'light'
+      })}>
+      <h1 className="pipeline-title">Kedro-Viz</h1>
+      <Sidebar />
+      <MetaData />
+      <div className="pipeline-wrapper">
+        {displayWarning === true ? <Modal /> : <FlowChart />}
```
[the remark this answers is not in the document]
studioswong: Thanks for the recommendation!  I completely understand your point about the thoughts on this as an improvement on performance - I'm just not entirely sure whether this will is necessary given the following reasons: 

1. The graphState will still need to be recalculated on every new selection of the tags, which by then the react component will need to be re-rendered anyways. 
2. Rendering the flowchart component in the background at this stage might be an overkill as the user will likely have moved on with a selection within seconds of seeing this warning ( there isn't any point in rendering an empty graph in the background ) 
3.  During our previous discussion, it was suggested that this warning to be displayed in the fastest way on first load, which is best achieved by minimising the amount of elements to be rendered on first load. This is hence my earlier refactor to move the trigger of this warning up in `calculateGraph`, along with the need to set up a series of new actions just for this - to only load this warning, which is a simple div as compared to a complex flowchart component. 
4. I had tried testing the performance by toggling the tag selection back and forth, and for me there has been no huge impact on performance. 

My pragmatic view is that, given that I don't see any huge performance issues, I am sure whether there are much value for this refactor in enhancing the performance ( or whether it's necessary for a simple warning that would not be used 99% of the time. )

src/config.js
```diff
@@ -16,6 +16,9 @@ export const sidebarWidth = {
 
 export const chartMinWidthScale = 0.25;
 
+// this value is used to determine the amount of nodes in pipeline to trigger chonky warning
+export const chonkyNodeAmount = 1000;
```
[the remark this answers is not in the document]
studioswong: thanks @bru5 , I'll use that name!

src/components/modal/modal.scss
studioswong: sounds good, this is the type of comment / suggestion I am hoping to get for the styling - I've changed the styling in my latest commit but will revisit it to comply more with BEM

src/actions/actions.test.js
studioswong: thanks @bru5 , I'll update the wording

src/actions/graph.js
```diff
@@ -27,6 +28,58 @@ export function updateGraph(graph) {
...
+  return {
+    type: TOGGLE_IS_CHONKY,
+    isChonky
+  };
+}
+
+export const UPDATE_NODESNO = 'UPDATE_NODESNO';
+
+/**
+ * updates the nodes no. for warning reference
```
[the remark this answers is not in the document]
studioswong: naming updated in latest commit

src/actions/graph.js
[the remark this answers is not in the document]
studioswong: added in more explanation for this formula in the latest commit

src/components/modal/index.js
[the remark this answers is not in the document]
studioswong: Naming updated in latest commit

src/actions/actions.test.js
studioswong: updated in latest commit

src/config.js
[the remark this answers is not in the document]
studioswong: updated in latest commit

src/components/modal/modal.scss
studioswong: @bru5 updated styling to follow BEM style in latest commit - let me know your thoughts on the new naming of the classes

src/actions/graph.js
```diff
@@ -93,10 +80,12 @@ const chooseLayout = (instance, state) =>
 const layoutWorker = preventWorkerQueues(worker, chooseLayout);
 
 /**
- * Formula to determine if the pipeline is chonky
+ * Formula to determine if the pipeline is large, which is simply the sum of
+ * both the amount of nodes and edges within the graph. Edges are given a
+ * strongger weight of 1.5 with it being more computational heavy to render.
```
bru5, line 85: I was thinking more that this should explain the goal rather than describing the solution? e.g. "Estimate the point at which the graph will take unreasonably long to render in terms of input nodes and edges". Also looks like a spelling issue on `strongger`.

src/actions/graph.js
```diff
@@ -93,10 +80,12 @@ const chooseLayout = (instance, state) =>
...
  */
-const isChonky = (chonkyNodeAmount, nodesNo, edgesNo) => {
-  return nodesNo + 1.5 * edgesNo > chonkyNodeAmount ? true : false;
+const isLarge = (largeNodeCountThreshold, nodeCount, edgeCount) => {
+  return nodeCount + 1.5 * edgeCount > largeNodeCountThreshold ? true : false;
```
bru5, line 88: I didn't notice this earlier, but since this threshold is in terms of both `nodeCount` and `edgeCount` it could be better to go with `largeGraphThreshold` actually?

src/actions/graph.js
```diff
@@ -105,24 +94,23 @@ const isChonky = (chonkyNodeAmount, nodesNo, edgesNo) => {
  * @param {Object} graphState A subset of main state
  * @return {function} A promise that resolves when the calcuation is done
  */
-export function calculateGraph(graphState, customChonkyLimit) {
+export function calculateGraph(graphState, customLargeLimit) {
```
bru5, line 97: Worth considering a default param here to clean this up potentially e.g.

`export function calculateGraph(graphState, displayThreshold=largeGraphThreshold)` that

src/components/largePipelineWarning/index.js
```diff
@@ -0,0 +1,40 @@
...
+import { connect } from 'react-redux';
+import { toggleDisplayLargeGraph } from '../../actions/graph';
+import './largePipelineWarning.css';
+
+export const LargePipelineWarning = ({
+  nodeCount,
+  onToggleDisplayLargeGraph
+}) => {
+  return (
+    <div className="warning">
```
bru5, line 11: Check with @richardwestenra on this, but I think all classnames in this project are prefixed with `pipeline-`

src/components/largePipelineWarning/index.js
```diff
@@ -0,0 +1,40 @@
...
+      <div className="warning__title">Your pipeline is large.</div>
```
bru5, line 12: @stichbury I know this has been discussed, but I think this title itself might be clearer just using the line that follows it ("Your pipeline might take a while") and then adjust the subtitle, so to avoid being unnecessarily prescriptive?

src/components/largePipelineWarning/largePipelineWarning.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  &__subtitle {
+    width: 350px;
+    font-size: 1.48em;
+    line-height: 1.5;
+    opacity: 0.6;
+  }
+
+  &__btn {
```
bru5, line 26: Might be best to be more specific with the name e.g. `render-button`

src/config.js
```diff
@@ -16,8 +16,8 @@ export const sidebarWidth = {
 
 export const chartMinWidthScale = 0.25;
 
-// this value is used to determine the amount of nodes in pipeline to trigger chonky warning
-export const chonkyNodeAmount = 1000;
+// this value is used to determine the amount of nodes in pipeline to trigger large warning
```
bru5, line 19: I think this comment should mention 'nodes and edges'  if it's reflecting the threshold function?

bru5 (changes requested):

src/components/wrapper/index.js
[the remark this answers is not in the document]
bru5: Ah I see your points, but I actually refer more to a general browser best practice worth keeping an eye on.

From a performance perspective it's best to mount components only once and never unmount them where you know they will always be in use, as to avoid DOM thrashing and other initialisation / disposal costs each time (which in turn cause the UI to freeze user input and render jank).

Since we know the flowchart will always be needed and is a core component, it's a good idea to eager mount it in this case and never unmount it, even if it is visibly empty. Maybe the flowchart component needs a little adapting to allow this?

src/actions/graph.js
studioswong: good suggestion, I have assigned the default param later down the code, but I can move it up here 👍

src/components/largePipelineWarning/index.js
studioswong: Thanks for the suggestion @bru5 - updated naming to include the prefix

src/components/largePipelineWarning/largePipelineWarning.scss
studioswong: yea I initially set this up under a similar naming as well, but in the end this still comes down to the class being a general button styling for the warning ( the class name is now `pipeline-warning__btn` with which it is descriptive enough to describe the button styling, hence I think it's still better to keep it as `__btn` .

src/actions/graph.js
studioswong: updated with this new naming in latest commit

src/actions/graph.js
studioswong: updated to reflect this in latest commit

src/components/wrapper/index.js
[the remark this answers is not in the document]
studioswong: yes indeed, the flowchart component do need adapting for this case - I had just tried a quick try / implementation of rendering the flowchart while the warning message is in place ( i.e the toggle graph action along with other actions for graph information calculation has not been dispatched, meaning invalid inputs for the flowChart), and it is returning a series of console errors ( see below). 
image

In the meantime while clicking through the different tags to switch between the warning and the rendering of the flowchart for the selected tags, I don't notice any difference in the speed / performance in the display of the graph ( i.e it still takes a long time to load the newly calculated graph given it still needs to go through `updateGraph` calculations.) 

I completely understand your view on browser DOM best practices, however with all the potential errors thrown ( from the invalid data and further errors thrown each time the user scrolls on the warning message triggered by the flowchart component), furthermore without any particular observed improvements,  it doesn't seem to justify the need to render the flowchart component at the same time. ( of course, if this do get raised by any user in future observations, we can always revisit this again.)

src/config.js
studioswong: updated the comment to reflect the suggested changes

src/actions/graph.js
studioswong: updated with new wording as well as descriptions of params

src/components/wrapper/index.js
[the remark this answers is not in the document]
studioswong: @bru5 I've given this a thought over the weekend, as how you have mentioned before, there might be a better and less intrusive solution to the large pipeline, such as going in the other direction to automatically only render a subset of the nodes and adding a reminder to the user to change their filters on the left if they want to render more nodes, etc. In the meantime, this feature is likely just a temporary solution while we think of implementing a better more intuitive solution to solve this.

Given this, the effort to refactor flowchart ( along with the tests and any potential other tests this breaks) to accomodate this might be an overkill at this stage. However, if you feel strongly about this change, we can file a new ticket to tackle this. Sounds good?

src/components/largePipelineWarning/index.js
stichbury: Sounds good!

src/actions/graph.js
```diff
@@ -27,6 +28,45 @@ export function updateGraph(graph) {
   };
 }
 
+export const TOGGLE_DISPLAY_LARGE_GRAPH = 'TOGGLE_DISPLAY_LARGE_GRAPH';
+
+/**
+ * resets the disaplyLargeGraph setting to enable large graphs to load
```
bru5, line 34: I think best to stick to the standard used throughout here and capitalise the first letter of all comments / end with a full stop (if needed)

src/components/largePipelineWarning/index.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { toggleDisplayLargeGraph } from '../../actions/graph';
+import './largePipelineWarning.css';
+
+export const LargePipelineWarning = ({
```
bru5, line 6: Also best to stick to the same naming style for all component directory names and filenames (kebab case), so here `large-pipeline-warning` throughout

src/components/largePipelineWarning/largePipelineWarning.scss
```diff
@@ -0,0 +1,36 @@
+@import '../../styles/variables';
+
+.pipeline-warning {
+  position: absolute;
+  top: 30%;
+  left: 50%;
+  color: white;
```
bru5, line 7: This will need to take on the standard theming approach, as at the moment this only works for the dark theme, you can refer to how it's done in other components but also check with @richardwestenra on this.

bru5 (changes requested):

src/components/wrapper/index.js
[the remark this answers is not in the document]
bru5: Ah okay I wasn't expecting it would require a lot of changes, I guess since the message is to be only shown once per session at the start then this can probably be revisited later. For anything regular though definitely better to use a CSS based overlay / toggle.

src/actions/graph.js
studioswong: Sure, thanks for picking up on the capitalization, however in looking through the codebase I don't see any comments ending on with the full stop?

src/components/largePipelineWarning/index.js
studioswong: This is a good suggestion, but on revisiting your suggested wording it sounds more like that the pipeline is loading in the background, which is not the case given the intent of this warning ( which is to prevent rendering a large graph by prompting the user to filter by tags/ nodes).

I would say the previous wording in the design ( and as discussed) it more clear.

src/components/largePipelineWarning/index.js
studioswong: hhmm yea, I used pascal case out of second nature - I'll update the file name

src/actions/graph.js
bru5: Ah that's fine on the full stop, I just meant where it makes sense (usually where there are multiple lines in a jsdoc)

src/components/largePipelineWarning/index.js
studioswong: updated the component and file names

src/components/largePipelineWarning/largePipelineWarning.scss
studioswong: Nice suggestion! I have modified the CSS according to the selected theme

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+        className={classnames('pipeline-warning__btn', {
+          'kui-theme--dark': theme === 'dark',
+          'kui-theme--light': theme === 'light'
+        })}
+        onClick={() => onToggleDisplayLargeGraph(true)}>
+        Render it anyway
+      </button>
+    </div>
+  );
+};
```
richardwestenra, line 34: Adding this here because I can't comment on an image file:

The chonky goose image has been removed from this component and yet the image `src/static/images/chonky-goose.jpg` is still present in the images directory. Can you please delete it?

src/actions/graph.js
```diff
@@ -39,21 +79,46 @@ const chooseLayout = (instance, state) =>
...
-    dispatch(toggleLoading(false));
-    return dispatch(updateGraph(graph));
+    const { nodes, edges, displayLargeGraph } = graphState;
+
+    if (
+      isLarge(displayThreshold, nodes.length, edges.length) === true &&
+      displayLargeGraph === false
+    ) {
+      dispatch(toggleIsLarge(true));
+      dispatch(updateNodeCount(nodes.length));
```
richardwestenra, line 114: Why dispatch an action to update the node count? You don't need an action to retrieve a value from the state. Why not use a selector instead? This action doesn't appear necessary.

src/actions/graph.js
```diff
@@ -39,21 +79,46 @@ const chooseLayout = (instance, state) =>
...
-    dispatch(toggleGraph(true));
-    dispatch(toggleLoading(false));
-    return dispatch(updateGraph(graph));
+    const { nodes, edges, displayLargeGraph } = graphState;
+
+    if (
+      isLarge(displayThreshold, nodes.length, edges.length) === true &&
+      displayLargeGraph === false
+    ) {
+      dispatch(toggleIsLarge(true));
```
richardwestenra, line 113: As below, this action could also probably just be a selector instead. You don't need to trigger a change in the state to do this - just `return`.

richardwestenra (changes requested):
Thanks Susanna! I'm not done with this review yet but I need to go have dinner and I've requested quite a lot of changes already, so here is the first phase of my review for you to be getting on with - I'll have another look tomorrow.

package-lock.json
```diff
@@ -26042,6 +26047,22 @@
...
+      "version": "3.12.1",
+      "resolved": "https://registry.npmjs.org/react-modal/-/react-modal-3.12.1.tgz",
+      "integrity": "sha512-WGuXn7Fq31PbFJwtWmOk+jFtGC7E9tJVbFX0lts8ZoS5EPi9+WWylUJWLKKVm3H4GlQ7ZxY7R6tLlbSIBQ5oZA==",
+      "requires": {
+        "exenv": "^1.2.0",
+        "prop-types": "^15.5.10",
+        "react-lifecycles-compat": "^3.0.0",
+        "warning": "^4.0.3"
+      }
+    },
```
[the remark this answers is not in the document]
studioswong: good spot, I'll update the lock file

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+}) => {
+  return (
+    <div
+      className={classnames('pipeline-warning', {
+        'kui-theme--dark': theme === 'dark',
+        'kui-theme--light': theme === 'light'
+      })}>
+      <div className="pipeline-warning__title">Your pipeline is large.</div>
+      <div className="pipeline-warning__subtitle">
+        Your pipeline might take a while to render because it has{' '}
```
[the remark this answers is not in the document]
studioswong: The whitespace was inserted by auto-formatting on my VS coded due to code line-width restriction ( I checked and it applies on text copy) - the white space is added on by the auto-formatting in particular to reflect the white space next to the `<b>` element that follows, which looks like this is the only way to allow the white space to exist while allowing the bold element that follows. 

If this bother you we can change the design and remove the bold element, but not sure whether this design change is  needed?

src/actions/graph.js
```diff
@@ -39,21 +79,46 @@ const chooseLayout = (instance, state) =>
...
+      dispatch(updateNodeCount(nodes.length));
+    } else {
```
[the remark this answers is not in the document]
studioswong: The `else` condition here is to cater for the normal case where we dispatch the actions to load the graph, which is very much needed?

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+  nodeCount,
+  theme,
+  onToggleDisplayLargeGraph
+}) => {
+  return (
+    <div
+      className={classnames('pipeline-warning', {
+        'kui-theme--dark': theme === 'dark',
+        'kui-theme--light': theme === 'light'
+      })}>
```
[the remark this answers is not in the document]
studioswong: great tip, I'll change that!

src/actions/graph.js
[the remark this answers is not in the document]
richardwestenra: What I mean is, you can use a `return` statement instead:
```js
if (!displayLargeGraph) {
  dispatch(toggleIsLarge(true));
  return dispatch(updateNodeCount(nodes.length));
}
dispatch(toggleIsLarge(false));
dispatch(toggleLoading(true));
const graph = await layoutWorker(graphState);
dispatch(toggleGraph(true));
dispatch(toggleLoading(false));
return dispatch(updateGraph(graph));
```

It simplifies the logic, reduces indentation and makes it easier to read.

src/components/large-pipeline-warning/large-pipeline-warning.scss
```diff
@@ -0,0 +1,54 @@
...
+  text-align: center;
+  opacity: 0.85;
+
+  &.kui-theme--light {
+    color: $color-text-light;
+  }
+
+  &.kui-theme--dark {
+    color: $color-text-dark;
+  }
```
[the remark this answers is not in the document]
studioswong: great point you raised - initially I also thought that the component would pick up the value from the parent ( for text, etc), but weirdly it doesn't, hence I had to set up a lot of the properties that were supposed to inherit from the parent here ( colors, font, etc).

src/actions/graph.js
studioswong: Thanks! I have updated the comment with the capitalization!

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+      <div className="pipeline-warning__title">Your pipeline is large.</div>
+      <div className="pipeline-warning__subtitle">
+        Your pipeline might take a while to render because it has{' '}
+        <b>{nodeCount}</b> nodes. Use the sidebar controls to select a smaller
```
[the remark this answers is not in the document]
studioswong: I see, that's a great suggestion, however I am not particularly sure of the need to list out the count of all the elements in the warning as they can be referred to on the sidebar, but rather a total count of elements to be displayed in the graph would be a better indicator of this.

We could perhaps change the wording from `nodes` to `elements`?

src/components/large-pipeline-warning/large-pipeline-warning.scss
```diff
@@ -0,0 +1,54 @@
+@import '../../styles/variables';
+
+.pipeline-warning {
+  position: absolute;
+  top: 30%;
+  left: 50%;
+  font-family: 'Titillium Web', 'sans-serif';
```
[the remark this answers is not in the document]
studioswong: Thanks for the clarification - that was my confusion the whole time too ( I thought those styles would be defined as a globa; style). regardless, I'll add that class.

src/components/large-pipeline-warning/large-pipeline-warning.scss
```diff
@@ -0,0 +1,54 @@
...
+    font-size: 38px;
+    line-height: 1.18;
+    letter-spacing: 0.27px;
+  }
+
+  &__subtitle {
+    width: 350px;
+    font-size: 1.48em;
+    line-height: 1.5;
+    opacity: 0.6;
```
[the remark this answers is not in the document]
studioswong: sounds good, will remove that.

src/components/large-pipeline-warning/large-pipeline-warning.scss
```diff
@@ -0,0 +1,54 @@
+@import '../../styles/variables';
+
+.pipeline-warning {
+  position: absolute;
+  top: 30%;
+  left: 50%;
+  font-family: 'Titillium Web', 'sans-serif';
+  text-align: center;
+  opacity: 0.85;
```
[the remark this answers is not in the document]
studioswong: The opacity setting was specified on the design - I have no comments against taking that out if you feel strongly about it.

src/components/large-pipeline-warning/large-pipeline-warning.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  &__title {
+    padding: 16px 0px;
+    font-size: 38px;
+    line-height: 1.18;
+    letter-spacing: 0.27px;
+  }
+
+  &__subtitle {
+    width: 350px;
```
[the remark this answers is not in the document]
studioswong: yea I would set up a max-width to cater for the full-width scenerio - In terms of using `em`, I am happy to follow, however I do notice that `px` has been used throughout the codebase to define width - is this something that we will align on?

src/actions/graph.js
```diff
@@ -39,21 +79,46 @@ const chooseLayout = (instance, state) =>
...
-    dispatch(toggleLoading(true));
-    const graph = await layoutWorker(graphState);
-    dispatch(toggleGraph(true));
-    dispatch(toggleLoading(false));
-    return dispatch(updateGraph(graph));
+    const { nodes, edges, displayLargeGraph } = graphState;
+
+    if (
+      isLarge(displayThreshold, nodes.length, edges.length) === true &&
+      displayLargeGraph === false
```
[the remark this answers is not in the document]
studioswong: sounds good!

src/components/large-pipeline-warning/index.js
studioswong: just deleted the image, thanks for spotting

src/actions/graph.js
studioswong: good call, not sure why I haven't thought of that 👍 

I've refactored to use the nodes selector and deleted the nodeCount field in the state.

package-lock.json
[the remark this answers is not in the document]
studioswong: updated with latest lock file

src/actions/graph.js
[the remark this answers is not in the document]
studioswong: aargh I see - sounds good!

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+        graph, or click to render.
+      </div>
+      <button
```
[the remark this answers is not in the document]
studioswong: sure thing, I will refactor to use the kedro-ui button

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+  onToggleDisplayLargeGraph
+}) => {
+  return (
+    <div
+      className={classnames('pipeline-warning', {
+        'kui-theme--dark': theme === 'dark',
+        'kui-theme--light': theme === 'light'
+      })}>
+      <div className="pipeline-warning__title">Your pipeline is large.</div>
+      <div className="pipeline-warning__subtitle">
```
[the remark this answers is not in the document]
studioswong: great suggestion, I've just amended that in my latest commit 👍

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+  theme,
+  onToggleDisplayLargeGraph
+}) => {
+  return (
+    <div
+      className={classnames('pipeline-warning', {
+        'kui-theme--dark': theme === 'dark',
+        'kui-theme--light': theme === 'light'
+      })}>
+      <div className="pipeline-warning__title">Your pipeline is large.</div>
```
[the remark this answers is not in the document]
studioswong: great suggestion, just amended in my latest commit 👍

src/components/large-pipeline-warning/index.js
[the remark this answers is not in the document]
studioswong: great suggestion, just amended in my latest commit 👍

src/actions/graph.js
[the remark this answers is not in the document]
studioswong: this is updated in latest commit

src/actions/graph.js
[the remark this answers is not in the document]
studioswong: this is updated in latest commit

src/components/large-pipeline-warning/index.js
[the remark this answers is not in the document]
richardwestenra: [oh my GOD you're right](https://github.com/prettier/prettier/issues/4223), I'd literally never seen Prettier do that before! I wonder how long it's been doing that? ugh. I guess we're stuck with it :(

No worries, best leave it as is I guess.

src/components/large-pipeline-warning/index.js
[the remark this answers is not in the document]
richardwestenra: Yeah I'm okay with changing it from nodes to elements.

richardwestenra (changes requested):
Hey I started writing a few more review comments this morning, submitting them now, will finish the review in a bit!

src/actions/graph.js
```diff
@@ -39,21 +66,46 @@ const chooseLayout = (instance, state) =>
...
-    dispatch(toggleLoading(false));
-    return dispatch(updateGraph(graph));
+    const { nodes, edges, displayLargeGraph } = graphState;
+
+    const largePipeline =
+      isLarge(displayThreshold, nodes.length, edges.length) &&
+      !displayLargeGraph;
+
+    if (largePipeline) {
+      return dispatch(toggleIsLarge(true));
```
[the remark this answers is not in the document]
studioswong: This action is needed to toggle the `isLarge` field in the state that is in turn used to control the display of the warning in the wrapper, in particular to cater for the scenerio where the user clicks between tags to trigger the large warning. 

There are 2 ways to enable the control of the display in the wrapper: we can either import all the nodes and perform the calculation again in the wrapper ( which the calculation is already done in the graph actions), or we can simply set up a variable to log the calculation ( which is `isLarge` in this case). I opted for the second given that it is much simplier without the need to perform the calculation in 2 places.

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,49 @@
...
+      <Button theme={theme} onClick={() => onToggleDisplayLargeGraph(true)}>
+        Render it anyway
+      </Button>
+    </div>
+  );
+};
+
+export const mapStateToProps = state => ({
+  theme: state.theme,
+  visible: state.visible.sidebar,
```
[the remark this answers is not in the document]
studioswong: sure I'll change it to `sidebarVisible`

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,49 @@
...
+}) => {
+  return (
+    <div
+      className={classnames('kedro', 'pipeline-warning', {
+        'pipeline-warning--sidebar-visible': visible
+      })}>
+      <h2 className="pipeline-warning__title">Your pipeline is large.</h2>
+      <p className="pipeline-warning__subtitle">
+        Your pipeline might take a while to render because it has{' '}
+        <b>{nodes.data.length + nodes.parameters.length + nodes.task.length}</b>{' '}
```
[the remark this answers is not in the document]
studioswong: this has been updated

src/components/large-pipeline-warning/index.js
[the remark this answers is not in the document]
studioswong: updated in latest commit

src/actions/graph.js
studioswong: Thanks Richard! I had just pushed the refactor to set this as a selector in the latest commit ( the pairing we did earlier), let me know if you have any further comments! 

Resolving this issue as this is now deleted.

src/actions/graph.js
```diff
@@ -39,21 +66,46 @@ const chooseLayout = (instance, state) =>
...
+/**
+ * Formula to estimate the point at which the graph will take unreasonably
+ * long to render in terms of input nodes and edges.
+ *  @param {Object} displayThreshold The defined threshold for a large graph
+ *  @param {integer} nodeCount The amount of nodes to be displayed in the flowchart
+ *  @param {integer} edgeCount The amount of edges to be displayed in the flowchart
+ */
+const isLarge = (displayThreshold, nodeCount, edgeCount) => {
+  return nodeCount + 1.5 * edgeCount > displayThreshold;
+};
```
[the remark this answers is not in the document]
studioswong: This code is now deleted and moved elsewhere given the recent refactor, hence resolving this matter.

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,50 @@
...
+        Render it anyway
+      </Button>
+    </div>
+  );
+};
+
+export const mapStateToProps = state => ({
+  theme: state.theme,
+  sidebarVisible: state.visible.sidebar,
+  nodes: getGroupedNodes(state)
```
richardwestenra, line 38: As mentioned elsewhere, why not get this value from the redux state in the component itself?
```suggestion
  nodes: getGroupedNodes(state),
  visible: getShouldDisplayLargeWarning(state)
```

src/actions/graph.js
```diff
@@ -50,6 +76,7 @@ export function calculateGraph(graphState) {
     return updateGraph(graphState);
   }
   return async function(dispatch) {
+    dispatch(toggleIsLarge(false));
```
richardwestenra, line 79: There's still a bunch of leftover unnecessary actions & reducers in this - I assume you just haven't gotten around to removing them all yet?

richardwestenra (changes requested):

src/actions/graph.js
studioswong: This has been deleted in latest commit

src/components/wrapper/index.js
```diff
@@ -22,7 +24,7 @@ export const Wrapper = ({ loading, theme }) => (
     <Sidebar />
     <MetaData />
     <div className="pipeline-wrapper">
-      <FlowChart />
+      {showLargeGraphWarning ? <Modal /> : <FlowChart />}
```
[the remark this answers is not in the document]
studioswong: I had initially tried this out, where on toggling between different tags will produce the following error:
image

It seems that we need to further wrap Flowchart in a separate div to control the visibility of the flowchart ( which in this case, would need to implement more styling code), or somehow we need to fire another action to clear off the visible nodes in between toggling different tags. 

@richardwestenra what is your suggestion?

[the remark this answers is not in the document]
studioswong: The requested changes have been resolved and pushed in the latest commit

src/components/wrapper/index.js
```diff
@@ -1,18 +1,20 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
+import Modal from '../large-pipeline-warning';
```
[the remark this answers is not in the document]
studioswong: This has been resolved in the latest commit

src/components/large-pipeline-warning/index.js
studioswong: This has been resolved in the latest commit

src/components/large-pipeline-warning/large-pipeline-warning.scss
```diff
@@ -0,0 +1,46 @@
+@import '../../styles/variables';
+
+.pipeline-warning {
+  position: absolute;
+  top: calc(50% - 14em);
```
bru5, line 5: I think this is fine for now, but later on maybe worth considering more of a flexbox approach for this part if at all possible, just thinking about how this might need to change later, but we can do it then for sure if this works well.

src/components/large-pipeline-warning/large-pipeline-warning.test.js
```diff
@@ -0,0 +1,51 @@
...
+  it('clicking the render anyways button will toggle the graph to display', () => {
+    const mockFn = jest.fn();
+    const props = {
+      onToggleIgnoreLargeWarning: mockFn,
+      nodes: [],
+      visible: true,
+    };
+    const wrapper = setup.mount(<LargePipelineWarning {...props} />);
+
+    wrapper.find('.kui-button__btn').simulate('click');
```
bru5, line 30: Ideally could use a more specific class here like `pipeline-warning__render-button`?

src/components/wrapper/index.js
```diff
@@ -12,7 +13,7 @@ import './wrapper.css';
 /**
  * Main app container. Handles showing/hiding the sidebar nav, and theme classes.
  */
-export const Wrapper = ({ loading, theme }) => (
+export const Wrapper = ({ loading, theme, showLargeGraphWarning }) => (
```
bru5, line 16: Looks like `showLargeGraphWarning` isn't being used here, so think better to remove? Either that or go with something like `visible={showLargeGraphWarning}`.

src/selectors/layout.js
```diff
@@ -20,11 +38,13 @@ export const getGraphInput = createSelector(
     getVisibleLayerIDs,
     getOldgraphFlag,
     getFontLoaded,
+    getTriggerLargeGraphWarning,
   ],
-  (nodes, edges, layers, oldgraph, fontLoaded) => {
-    if (!fontLoaded) {
+  (nodes, edges, layers, oldgraph, fontLoaded, triggerLargeGraphWarning) => {
+    if (!fontLoaded || triggerLargeGraphWarning) {
```
bru5, line 44: I suggest a similar thing here with the naming as earlier, for consistency.

bru5 (changes requested):

src/components/large-pipeline-warning/large-pipeline-warning.scss
richardwestenra: Not a bad point - I considered mentioning that but ultimately decided it probably wasn't worth adding another element for it, and the current approach works well enough. But yeah could look into changing it later if necessary

src/components/large-pipeline-warning/large-pipeline-warning.test.js
richardwestenra: This class is part of Kedro UI so it can't be changed.

src/reducers/index.js
```diff
@@ -68,6 +69,11 @@ const combinedReducer = combineReducers({
   fontLoaded: createReducer(false, UPDATE_FONT_LOADED, 'fontLoaded'),
   textLabels: createReducer(true, TOGGLE_TEXT_LABELS, 'textLabels'),
   theme: createReducer('dark', TOGGLE_THEME, 'theme'),
+  ignoreLargeWarning: createReducer(
+    {},
```
richardwestenra, line 73: Shouldn't this have a Boolean default value here as the first argument?
```suggestion
    false,
```

src/selectors/layout.test.js
```diff
@@ -1,10 +1,19 @@
...
+  getTriggerLargeGraphWarning,
+} from './layout';
 import { updateFontLoaded } from '../actions';
 import reducer from '../reducers';
 import { sidebarWidth } from '../config';
 
 describe('Selectors', () => {
+  describe('getTriggerLargeGraphWarning', () => {
+    expect(getTriggerLargeGraphWarning(mockState.animals)).toEqual(false);
+  });
```
richardwestenra, line 15: Hey this needs more tests - e.g. under what conditions should it return true?

It might be worth using that commit I sent you with the overlarge test dataset creator function.

richardwestenra (changes requested):

src/selectors/layout.test.js
richardwestenra: e.g. something like this:
```js
it('returns true when  the dataset is larger than the threshold', () => {
	  // Modify the animals dataset to add so many nodes that it triggers the warning:
	  const data = { ...animals };
	  let extraNodes = [];
	  const iterations = Math.ceil(largeGraphThreshold / data.nodes.length) + 1;
	  new Array(iterations).fill().forEach((d, i) => {
	    const extraNodeGroup = data.nodes.map(node => ({
	      ...node,
	      id: node.id + i
	    }));
	    extraNodes = extraNodes.concat(extraNodeGroup);
	  });
	  data.nodes = data.nodes.concat(extraNodes);
	  const customMockState = prepareState({ data });
	  expect(getTriggerLargeGraphWarning(customMockState)).toEqual(true);
});
```

src/components/large-pipeline-warning/large-pipeline-warning.test.js
bru5: @richardwestenra you mean that button component doesn't let you add an additional user defined class?

src/components/wrapper/index.js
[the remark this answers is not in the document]
studioswong: This is now resolved with the latest refactor

src/components/large-pipeline-warning/index.js
```diff
@@ -0,0 +1,51 @@
...
+      </Button>
+    </div>
+  ) : null;
+};
+
+export const mapStateToProps = (state) => ({
+  theme: state.theme,
+  sidebarVisible: state.visible.sidebar,
+  nodes: getVisibleNodes(state),
+  visible: getTriggerLargeGraphWarning(state),
```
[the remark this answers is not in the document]
studioswong: I have thought about the naming, but personally I would stand by the naming of `getTriggerLargeGraphWarning` instead as it states exactly what this does, which is to trigger the warning. 

Making the suggested change would entail the refactoring of quite a bit on the selector side, and quite frank I am not entirely sure the effort to refactor for a opinions on naming is quite necessarily worth the effort of going back and forth?

src/components/large-pipeline-warning/large-pipeline-warning.test.js
studioswong: there is no need for an additional user defined class for the button as the design of this button is exactly as is

src/selectors/layout.js
studioswong: As I mentioned before, I don't particularly find the suggested naming to be of an enhancement worth enough for the efforts needed to refactor this?

src/selectors/layout.test.js
studioswong: Thanks @richardwestenra - I had actually set up the test this morning but I had not push the test as it is causing the following errors: 
image

Clearly from the looks of this, using a large dataset would require further set up work to ensure the nodes are in the right format for the setup, which looks like the random data generator would be the only way to truly generate a dataset fit for this purpose. 

Any suggestions on any quick fix I can do to solve this without a huge amount of setup work?

src/components/wrapper/index.js
studioswong: This has been taken out now hence resolving this

src/reducers/index.js
studioswong: This is fixed in the latest commit

src/components/large-pipeline-warning/large-pipeline-warning.test.js
richardwestenra: @bru5 Correct https://github.com/quantumblacklabs/kedro-ui/blob/master/src/components/button/button.jsx

And I agree with Susanna that it's not worth adding a new class just for the test.

bru5: @richardwestenra okay seems like an upstream issue then, seems fine to leave it here

bru5 (approved):

src/components/large-pipeline-warning/index.js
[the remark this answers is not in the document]
bru5: Ah okay fair enough then!

src/selectors/layout.test.js
richardwestenra: Looks like the issue is that it's passing `null` to the layout graphing algorithm which wasn’t prepared for it. It was doing this in `prepareState` in state.mock.js, on line 27. You can fix it by adding a check for falsey graphState input, e.g. like this:
```js
    // Precalculate graph layout:
    (state) => {
      const layout = state.flags.oldgraph ? graphDagre : graphNew;
      const graphState = getGraphInput(state);
      if (!graphState) {
        return state;
      }
      const graph = layout(graphState);
      return updateGraph(graph);
    },
```

src/selectors/layout.test.js
```diff
@@ -1,10 +1,37 @@
...
+  getTriggerLargeGraphWarning,
+} from './layout';
 import { updateFontLoaded } from '../actions';
 import reducer from '../reducers';
-import { sidebarWidth } from '../config';
+import { sidebarWidth, largeGraphThreshold } from '../config';
+import animals from '../utils/data/animals.mock.json';
 
 describe('Selectors', () => {
+  describe('getTriggerLargeGraphWarning', () => {
```
richardwestenra, line 14: (optional) It'd be great to add tests to check that this will show when `!ignoreLargeWarning` or `!graphHasNodes`, but no big deal, we can add this later :)

richardwestenra (approved):
Looks good to me! Thanks so much for your persistence in getting this over the line, I really appreciate it 🌞🙌

PR #356, opened 2021-01-22
[KED-2381] Improve responsive zoom approach

bru5:
## Description

Improves responsive zoom approach to work as follows, in order of precedence

- once, after any graph or sidebar change
  - fit whole graph
  - centre whole graph
  - crop graph when reaching a minimum usable scale
- once, after an element is selected
  - respect the above constraints, and
  - ensure node is clearly visible between panels
  - apply an increased minimum scale
- once, after a browser resize
  - respect all the above constraints
- at any other time
  - allow user pan and zoom freely within extents

## Development notes

This functionality is a usability improvement and is much needed as new sidebars are being added.

A follow up PR will refactor the view system in general now that it is outgrowing its current shape.

## QA notes

Use the above description as a guide for requirements and testing.

Note that some edge cases may not appear to be an 'optimal' view, but none should be broken.

This will include testing in various combinations:

- open / closed sidebar and metadata panel
- resizing the browser to various viewport sizes / aspect ratios
- different graph aspect ratios (square, wide, tall)
- selecting elements on the sidebar and on the chart
- selecting tags and pipelines
- toggling labels

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

## Legal notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #363, opened 2021-02-02
[KED-2381] Refactor zoom and views

bru5:
## Description

Abstracts and improves the viewport implementation. 
This follows up recent view work from #356.

## Development notes

- moves viewport code into a separate `view` utility
- this simplifies and unifies the flowchart and minimap components
- provides a more consistent and convenient API for this usage
- improves test coverage throughout

## QA notes

Behaviour should not change.

All flowchart and minimap functionality involving view effects should be tested.

Refer to notes in #356 for rules on how the view should now act.

This will include testing in various combinations:

- open / closed sidebar and metadata panel
- resizing the browser to various viewport sizes / aspect ratios
- different graph aspect ratios (square, wide, tall)
- selecting elements on the sidebar and on the chart
- selecting tags and pipelines
- toggling labels
- using the minimap

## Checklist

## Legal notice


src/utils/view.test.js
```diff
@@ -0,0 +1,407 @@
...
+        }
+      };
+
+      setViewExtents(view, extents);
+      expect(getViewExtents(view)).toEqual(extents);
+    });
+  });
+
+  describe('viewTransformToFit', () => {
+    it('returns expected transform when example must be scaled down to fit', () => {
```
studioswong, line 321: It's hard to relate to which transforms scenerio these tests are supposed to mimic ( for example, I am guessing this one would likely refer to the opening of the metadata panel, but it is not clear to readers of the code). I understand the tests are structured similar to unit tests here, but perhaps they should be structued in a fashion that is easy for the reader to interpret as actual user scenerios?

src/utils/view.test.js
```diff
@@ -0,0 +1,407 @@
...
+
+      expect(transform).toEqual({
+        k: minScaleFocus,
+        x: -50,
+        y: 60
+      });
+    });
+
+    it('returns expected transform when example must be scaled up to fit', () => {
+      // Example where object is too small, so scale up
```
studioswong, line 344: This might need a bit more clarification to clarify what ratio is viewed as a small object for scaling up

studioswong (commented):
Just tested and played around with the new zoom views and they work well, really great stuff @bru5 ! 👍 

one thing with regards to structuring of the tests - given the code is highly specialised with different formulas / calculations to cater for the different zoom scenerios that was outlined in the QA section, the cases outlined in the tests would be quite crucial in helping the user understand the different functions and how it caters for the different use case. I understand they are structured as unit tests, could you please add some tests cases to outline the different scenerios you listed in the QA ( or if some unit tests covers that, you can also mention that in the test description?)

Thank you!

src/components/flowchart/index.js
```diff
@@ -191,221 +214,148 @@ export class FlowChart extends Component {
   };
 
   /**
-   * Setup D3 zoom behaviour on component mount
+   * On every frame of every view transform change (from reset, pan, zoom etc.)
```
richardwestenra, line 217: Would you mind adding a JSDoc param comment for the `transform` argument?

src/utils/view.js
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * Returns true if the transform meets this definition:
+ *  - any component (x, y or k) is infinite, `NaN` or `undefined`
+ * otherwise returns false.
+ * @param {Object} transform A transform object to test
+ * @returns {Boolean} True if the transform is invalid else false
+ */
+export const isInvalidTransform = ({ x, y, k }) =>
+  !isFinite(x + y + k) || isNaN(x + y + k);
```
richardwestenra, line 78: smort

richardwestenra (approved):

src/utils/view.js
```diff
@@ -0,0 +1,335 @@
...
+ * @returns {Object} The view transform
+ */
+export const getViewTransform = view => {
+  const transform = zoomTransform(view.wrapper.current);
+  return isInvalidTransform(transform) ? origin : negateTransform(transform);
+};
+
+/**
+ * Negates the x and y components, in effect making the direction opposite.
+ * Scale component does not change. Zeroes are always positive.
```
studioswong, line 93: Are we ending comments with full stops within the code base as I noticed that it has not been applied elsewhere?

src/utils/view.js
```diff
@@ -0,0 +1,335 @@
...
+    typeof transform.x !== 'undefined' && typeof transform.y !== 'undefined';
+  let k, x, y;
+
+  if (typeof jest !== 'undefined') {
+    // Transitions not supported in tests
+    duration = 0;
+  }
+
+  if (relative) {
+    // Relative: v = v. + t
```
studioswong, line 195: perhaps need to lose the `.` behind the `v` here?

src/utils/view.js
richardwestenra: I don't think it's something we're picky about. It's not public-facing so I don't think it matters, as long as its meaning is clear to readers

src/utils/view.test.js
bru5: Yeah these are written as unit tests rather than end-to-end style which is why they feel a bit abstract. I know what you mean though I'll look into adding more tests on the flowchart that cover this in a more concrete way.

src/utils/view.js
bru5: Agreed, I usually tend to include them in JSDocs written over multiple sentences, but not usually in comments intended as short notes rather than full sentences, so best to just judge on clarity and brevity.

richardwestenra: Exactly, same.

src/utils/view.js
bru5: I've changed the notation here to make this clearer, so `v'` is used for the view transform after application.

PR #370, opened 2021-02-15
[KED-2437] Fix minimum zoom level

bru5:
## Description

Fixes minimum zoom level back to its original behaviour, such that the whole graph can be made visible.

## Development notes

Implemented the fix, some tidying and updated tests.

## QA notes

Zoom out and ensure graph fits fully in view.

Test using different size graphs of different aspect ratios.

Compare to previous release.

## Checklist

## Legal notice

### DIFF
--- a/src/components/flowchart/flowchart.test.js
+++ b/src/components/flowchart/flowchart.test.js
@@ -75,41 +75,95 @@ describe('FlowChart', () => {
   });
 
   it('applies expected view extents when all sidebars closed', () => {
-    const wrapper = setup.mount(
-      <FlowChart
-        chartSize={mockChartSize({
-          sidebarWidth: 0,
-          metaSidebarWidth: 0,
-          codeSidebarWidth: 0,
-        })}
-      />
-    );
+    // Simulate closed sidebars
+    const chartSize = mockChartSize({
+      sidebarWidth: 0,
+      metaSidebarWidth: 0,
+      codeSidebarWidth: 0,
+    });
 
+    const wrapper = setup.mount(<FlowChart chartSize={chartSize} />);
     const instance = wrapper.find('FlowChart').instance();
     const viewExtents = getViewExtents(instance.view);
 
-    expect(viewExtents.translate).toEqual({ minX: -500, minY: -500, maxX: 1044, maxY: 1553 });
-    expect(viewExtents.scale.minK).toBeCloseTo(0.777);
-    expect(viewExtents.scale.maxK).toBe(2);    
+    const margin = instance.MARGIN;
+    const minScale = instance.MIN_SCALE;
+    const maxScale = instance.MAX_SCALE;
+
+    // Assert expected constants
+    expect(margin).toEqual(500);
+    expect(minScale).toEqual(0.8);
+    expect(maxScale).toEqual(2);
+
+    const { width: chartWidth, height: chartHeight } = chartSize;
+    const { width: graphWidth, height: graphHeight } = instance.props.graphSize;
+
+    // Translate extent should only include margin and graph size
+    expect(viewExtents.translate.minX).toEqual(-margin);
+    expect(viewExtents.translate.minY).toEqual(-margin);
+    expect(viewExtents.translate.maxX).toEqual(graphWidth + margin);
+    expect(viewExtents.translate.maxY).toEqual(graphHeight + margin);
+
+    // The scale at which the full graph in view
+    const fullScale = Math.min(
+      chartWidth / (graphWidth || 1),
+      chartHeight / (graphHeight || 1)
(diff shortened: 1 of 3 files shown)


src/components/flowchart/flowchart.test.js
```diff
@@ -88,10 +88,9 @@ describe('FlowChart', () => {
     const instance = wrapper.find('FlowChart').instance();
     const viewExtents = getViewExtents(instance.view);
 
-    expect(viewExtents).toEqual({
-      translate: { minX: -500, minY: -500, maxX: 1044, maxY: 1553 },
-      scale: { minK: 0.8, maxK: 2 },
-    });
+    expect(viewExtents.translate).toEqual({ minX: -500, minY: -500, maxX: 1044, maxY: 1553 });
+    expect(viewExtents.scale.minK).toBeCloseTo(0.777);
```
studioswong, line 92: What is the value `0.777` represent? This would be more intuitive if this is presented as a calculated value rather than an exact figure.

src/components/flowchart/flowchart.test.js
```diff
@@ -108,10 +107,9 @@ describe('FlowChart', () => {
     const instance = wrapper.find('FlowChart').instance();
     const viewExtents = getViewExtents(instance.view);
 
-    expect(viewExtents).toEqual({
-      translate: { minX: -650, minY: -500, maxX: 1479, maxY: 1553 },
-      scale: { minK: 0.8, maxK: 2 },
-    });
+    expect(viewExtents.translate).toEqual({ minX: -650, minY: -500, maxX: 1479, maxY: 1553 });
+    expect(viewExtents.scale.minK).toBeCloseTo(0.777);
```
studioswong, line 111: same as above

src/components/flowchart/index.js
```diff
@@ -283,26 +281,39 @@ export class FlowChart extends Component {
...
+    const minScale = Math.min(
+      chartWidth / (graphWidth || 1),
+      chartHeight / (graphHeight || 1)
+    );
+
     setViewExtents(this.view, {
       translate: {
         minX: -leftSidebarOffset - margin,
-        maxX: width + margin + rightSidebarOffset,
+        maxX: graphWidth + margin + rightSidebarOffset,
```
studioswong, line 311: We need to add a test to cover the value of `maxX`

src/components/flowchart/index.js
```diff
@@ -283,26 +281,39 @@ export class FlowChart extends Component {
...
         minY: -margin,
-        maxY: height + margin,
+        maxY: graphHeight + margin,
       },
       scale: {
-        minK: this.MIN_SCALE * this.defaultTransform.k,
+        minK: this.MIN_SCALE * minScale,
```
studioswong, line 316: As with above, we need to add a test to test for the value of `minK`

richardwestenra (changes requested):
Thanks Liam, this is great. It works slightly differently than I'd imagined (I'd thought that it might default to always showing the full graph on first load without the metadata panel open etc), but on further reflection I think it makes sense for it not to work that way and I think I see why you've implemented it this way instead.

I think it'd be worth explaining some of the magic numbers in the tests, but those are pretty minor and I'd consider those optional. The only item of pressing concern is that there are a couple of eslint warnings for unused variables that need tidying up. Thanks!

src/components/flowchart/flowchart.test.js
bru5: That's fair I should have mentioned these tests were just an integration check for sanity, partly to avoid testing implementation / re-implementing the tested function in the test itself, but I've managed to wrangle it into something fairly reasonable so please check out the updated version.

src/components/flowchart/index.js
bru5: This should be covered, could you explain further?

src/components/flowchart/index.js
bru5: This should be covered, could you explain further?

src/components/minimap/index.js
```diff
@@ -31,7 +31,6 @@ export class MiniMap extends Component {
     this.isPointerDown = false;
     this.isPointerInside = false;
     this.lastTransitionTime = 0;
-    this.defaultTransform = origin;
```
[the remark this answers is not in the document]
bru5: Yeah so I did investigate this in particular for very large / wide graphs, but things ended up getting very complex to handle all the edge cases smoothly. I think we're able to leave that improvement to a later update if needed, now that this fix resolves the minimum zoom.

See above on the tests, I've improved them and fixed the lint warnings.

richardwestenra (approved):
Nice one, thanks!

PR #374, opened 2021-02-18
[KED-2448] Improve node selected view panning behaviour

bru5:
## Description

Improve selected node view behaviour so panning only occurs when graph is cropped.

## Development notes

- Changes crop detection in view fit function to handle view width and height crops independently
- Allows a smaller zoom level on zoom reset / fit to reduce the need for above in practice

## QA notes

Test by selecting a node at various browser sizes, with various graph sizes, with various open panels.

Check that panning to selected node only occurs when the graph would be cropped by sidebars.
A slight zoom in may occur still when the scale does not meet a minimum.

## Checklist

## Legal notice

### DIFF
--- a/src/utils/view.test.js
+++ b/src/utils/view.test.js
@@ -334,11 +334,11 @@ describe('view', () => {
       });
 
       // Resulting transform should clamp to minimum scale and center
-      expect(transform).toEqual({
-        k: minScaleFocus,
-        x: -50,
-        y: 60,
-      });
+      expect(transform.k).toEqual(minScaleFocus);
+
+      // In this example offset is fractional
+      expect(transform.x).toBeCloseTo(-43.333);
+      expect(transform.y).toBeCloseTo(6.666);
     });
 
     it('returns expected transform when example must be scaled up to fit', () => {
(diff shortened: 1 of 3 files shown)


richardwestenra (approved):
LGTM! Nice one 👍

richardwestenra (approved):
lgtm

src/utils/view.test.js
```diff
@@ -334,11 +334,11 @@ describe('view', () => {
...
-      expect(transform).toEqual({
-        k: minScaleFocus,
-        x: -50,
-        y: 60,
-      });
+      expect(transform.k).toEqual(minScaleFocus);
+
+      // In this example offset is fractional
+      expect(transform.x).toBeCloseTo(-43.333);
+      expect(transform.y).toBeCloseTo(6.666);
```
studioswong, line 341: It is quite interesting to see the y values to change by almost 10 times from the original value - is this directly related to the change in the `minScaleX` value? 

 As with my comment in earlier PRs, it might be more clear if the values are represented and compared against a calculated value rather than a specified number that is subject to change whenever we alter any pre-set value, as well as allowing us to truly test the transform rather than going down the easy path of re-plugging numbers whenever a test fails :) WDYT?

bru5: Yeah these tests are due to be improved, at the moment they are just sanity checks (hence the fixed constants) because the logic is complex - this will be split up later to allow cleaner testing, rather than on the combined result.

PR #376, opened 2021-02-20
[KED-2446] Add new 'sizewarning' flag

richardwestenra:
## Description

1. Add a new 'sizewarning' flag for the large graph warning. This flag is enabled by default, but it's in place so that users can disable it if necessary, if the warning causes them problems. This will also allow us to add a "Don't show this message again" link/button, if we like.
2. Add new tests for the `getTriggerLargeGraphWarning` selector: Add a test for the flag case, but also add new tests to handle the `ignoreLargeWarning` and `graphHasNodes` cases too, which were missed out of #361.
3. Update getTriggerLargeGraphWarning JSDoc comment, to add a bit more detail explaining under what conditions it should return true (i.e. show the warning).

## Development notes

For the `graphHasNodes` test, I needed the state to have a sufficiently large dataset to cross the threshold and trigger the warning, but for the layout to have already been calculated so that the graph object has been updated and populated. In order to achieve this, I prepared the state with a large dataset, then ran it through a reducer with 3 actions: Toggling on a node-type filter to bring it under the threshold, updating the graph state, then toggling the filter back off again to bring the graph back above the threshold.

## QA notes

If you set `?sizewarning=0` in the URL with a very large pipeline, the large graph warning shouldn't display.

## Checklist

## Legal notice

### DIFF
--- a/src/selectors/layout.test.js
+++ b/src/selectors/layout.test.js
@@ -61,6 +61,9 @@ describe('Selectors', () => {
     });
 
     it('returns false if layout has already been calculated', () => {
+      // In cases where a user enables a filter to show the graph, then
+      // disables it again, the graph should still show, so that the user
+      // doesn't keep seeing the warning over and over again.
       const actions = [
         // Filter out all data nodes to reduce node-count below threshold
         () => toggleTypeDisabled('data', true),
--- a/src/selectors/layout.test.js
+++ b/src/selectors/layout.test.js
@@ -69,12 +69,14 @@ describe('Selectors', () => {
           const layout = state.flags.oldgraph ? graphDagre : graphNew;
           return updateGraph(layout(getGraphInput(state)));
         },
+        // Turn the filter back off
+        () => toggleTypeDisabled('data', false),
       ];
       const state = actions.reduce(
         (state, action) => reducer(state, action(state)),
         prepareState({ data: prepareLargeDataset() })
       );
-      expect(getTriggerLargeGraphWarning(state)).toEqual(false);
+      expect(getTriggerLargeGraphWarning(state)).toBe(false);
     });
   });
 
--- a/src/selectors/layout.test.js
+++ b/src/selectors/layout.test.js
@@ -36,12 +36,12 @@ describe('Selectors', () => {
     };
 
     it('returns false for a small dataset', () => {
-      expect(getTriggerLargeGraphWarning(mockState.animals)).toEqual(false);
+      expect(getTriggerLargeGraphWarning(mockState.animals)).toBe(false);
     });
 
     it('returns true for a large dataset', () => {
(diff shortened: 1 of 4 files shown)


src/selectors/layout.test.js
```diff
@@ -27,8 +32,51 @@ describe('Selectors', () => {
...
+
+    it('returns false if ignoreLargeWarning is true', () => {
+      const state = reducer(
+        prepareState({ data: prepareLargeDataset() }),
+        toggleIgnoreLargeWarning(true)
+      );
+      expect(getTriggerLargeGraphWarning(state)).toBe(false);
+    });
+
+    it('returns false if layout has already been calculated', () => {
```
studioswong, line 63: I wonder which scenario this test case refers to? I am not sure whether this case would exist and hence needed given that the graphWarning would be triggered before the layout is calculated ( i.e `updateGraph` is triggered.)

studioswong (commented):
Thanks for setting this PR and adding the tests @richardwestenra, looks great! I just have one confusion about the last test you added in, which I don't think is a scenerio that would exist and having that test in would be rather confusing on the expected behavior for this feature. 

Regardless, great stuff!

src/selectors/layout.test.js
richardwestenra: This is the case where, instead of clicking 'render anyway', the user enables a filter, then disables it again. This use-case is exactly why we added the `graphHasNodes` argument - so that if the user has already dismissed the warning via a filter, it won't appear again. This is because we only want the warning to appear once on load, and not for subsequent filter changes.

If necessary I can add more detail to explain this in the JSDoc comment for the layout selector?

studioswong: aargh I see - thanks for the clarification! 

It might be clearer to state the case as `'returns false once the user has selected a filter that triggered a layout calculation'` ( or something along those lines) that will indicate the expected behaviour of the warning only appearing once during the lifetime of the session. 
( This could also be further elaborated with this comment: ` // This is the ensure that the warning will only appear once during the lifetime of the session`)

studioswong (approved):
Thanks for the clarification - all clear about the test case.

I've left a suggested change for the wording of the test case to make the expected behaviour more apparent.

src/selectors/layout.test.js
richardwestenra: Thanks, no worries!

I'm unsure about this part: `‘returns false once the user has selected a filter that triggered a layout calculation’`.
On the one hand, this is more specific so it gives more of an indication of the typical flow, but it’s not the only situation that might cause this behaviour, so I think it's inaccurate. We want the test name to only refer to the actual circumstances that cause the test to pass. The motivation for the feature can be handled by the comment.

I'll add the other comment. Thanks!

PR #377, opened 2021-02-20
[KED-2453] Update large pipeline warning copy

richardwestenra:
## Description

1. Let's avoid saying "your pipeline" - we don't want to shame our users for their big-ass pipelines or make them look bad in front of a client. It's not a bad thing to be little on the large side, there's just more to love 😌
2. Let's inject a little levity and humanity into this message - [studies show that even in corporate environments, when people use humour in the workplace, they are 23% more respected and are seen as more competent and more confident](https://www.npr.org/2021/02/08/965572857/the-power-of-humor). Even executives have a sense of humour, they just don’t want us to imply that their work is of poor quality.
3. Let's use 'will' instead of 'might' - as @stichbury said, "I never like the 'might' word, because it implies doubt, and certainty is a nice thing for a reader."
4. Let's avoid needless repetition. There's no need to say "click to render", just show the button which says this anyway. Also, we prefer to avoid saying "click" because this implies that the user is navigating with a mouse, when they might be using a keyboard, touchpad or touchscreen.

image

## Development notes

I've changed the max-width of the paragraph text to `30em`, as this is the ideal width to split the two sentences onto two lines, and it is dependent on the font-size so the line-split will remain constant if the text size changes.

## QA notes

Updated copy. The only thing worth checking is how this renders across screen widths and when the sidebar is open/closed.

## Checklist

## Legal notice

### DIFF
--- a/src/components/large-pipeline-warning/index.js
+++ b/src/components/large-pipeline-warning/index.js
@@ -19,11 +19,13 @@ export const LargePipelineWarning = ({
       className={classnames('kedro', 'pipeline-warning', {
         'pipeline-warning--sidebar-visible': sidebarVisible,
       })}>
-      <h2 className="pipeline-warning__title">Your pipeline is large.</h2>
+      <h2 className="pipeline-warning__title">
+        Whoa, that’s a chonky pipeline!
+      </h2>
       <p className="pipeline-warning__subtitle">
-        Your pipeline might take a while to render because it has{' '}
-        <b>{nodes.length}</b> elements. Use the sidebar controls to select a
-        smaller graph, or click to render.
+        This graph contains <b>{nodes.length}</b> elements, which will take a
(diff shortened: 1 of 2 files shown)


studioswong (commented):
Thanks @richardwestenra ! I have fine with both the previous copy and this one and will leave it up to @stichbury and @yetudada . I recall Joel mentioned about the tone of voice for the warning given we have to be mindful for client presentations, so might be worth checking in with him as well?

datajoely (approved):
I like it - let's ship it. Can we track how many times it's fired and the associated node count?

richardwestenra:
@datajoely Thanks! We don't have tracking yet, but I'd love to track node/dataset/edge counts once we do.

datajoely:
> @datajoely Thanks! We don't have tracking yet, but I'd love to track node/dataset/edge counts once we do.

Render time would be great too

PR #379, opened 2021-02-23
Add Container component to simplify app/lib entrypoint

richardwestenra:
## Description

1. Create new `Container` component, and move a lot of the logic for the app entrypoint to it, in order to keep src/index.js as simple as possible. This should help make it easier to understand the different entrypoints for the app/lib use-cases.
2. Move the 'what-input' import to the App component - it should have been there to begin with.
3. Move the global app styles from src/styles.index.scss to Container component, as they are most related to this component.
4. Change import order in App component - cosmetic change to ensure that absolute imports come first.
5. Add better explanation in App/Container JSDoc comments

## Development notes

The only functionality change is that focus highlights should work in app imports now - I don't think they would have before.

I wasn't sure about the 'Container' component name, but I wasn't sure what else to call it. 'KedroViz'? 'AppContainer' and 'LibContainer'? I think it works for now with the comments I've added. Besides, this is all fairly temporary, no need to spend too much time scrutinising it.

## QA notes

Check both lib/app usages.

## Checklist

## Legal notice


bru5 (approved):

src/components/app/index.js
```diff
@@ -9,12 +12,15 @@ import getInitialState, {
   preparePipelineState,
 } from '../../store/initial-state';
 import { getFlagsMessage } from '../../utils/flags';
-import '@quantumblack/kedro-ui/lib/styles/app-no-webfont.css';
-import LoadWebFont from '@quantumblack/kedro-ui/lib/utils/webfont.js';
 import './app.css';
 
 /**
- * Main wrapper component. Intialises the Redux store
+ * Entry-point component for the use-case where Kedro-Viz is imported as a
```
studioswong, line 18: The newer component is good as it explains the use case for this component 👍

studioswong (approved):

src/components/container/index.js
```diff
@@ -0,0 +1,18 @@
...
+import App from '../app';
+import EasterEgg from '../easter-egg';
+import getPipelineData from '../../utils/data-source';
+import './container.css';
+
+/**
+ * Top-level component for the use-case where Kedro-Viz is run as a standalone
+ * app rather than imported as a library/package into a larger application.
+ */
+const Container = () => (
```
studioswong, line 11: Optional: I might've preferred to call `AppContainer` just to allow this to be more specific.

bru5: Maybe even `KedroViz` as that's the alias being used? Makes sense based on intended usage, since that's the public component.

richardwestenra: Yeah that's a good shout! I'm inclined to leave it for now because we'll be revisiting/refactoring this when we deprecate the library/package use-case, but can revisit if you feel strongly about it :)

PR #381, opened 2021-02-23
[KED-2471] Add "Don't show this again" message to large-pipeline-warning

richardwestenra:
## Description

Allow users to turn off the large-pipeline-warning permanently via the UI. They can reenable it using the 'sizewarning' flag.

image

## Development notes

I also did some slight refactoring of the CSS:

- Rename some component props to simplify them.
- Changed nested BEM child modifiers to be standalone - I'll document it in the style-guide, but we try to avoid this with children (as opposed to modifiers). It's very much a stylistic thing, it doesn't actually matter.
- I changed the vertical centring from position:absolute to flexbox.

## QA notes

Test on a large pipeline. Clicking the new button (which looks like a link) should hide the warning and turn off the flag.

## Checklist

## Legal notice

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -44,7 +44,7 @@ export const flags = {
   sizewarning: {
     description: 'Show a warning before rendering very large graphs',
     default: true,
-    icon: '⚠️',
+    icon: '🐳',
   },
 };
 
--- a/src/components/large-pipeline-warning/index.js
+++ b/src/components/large-pipeline-warning/index.js
@@ -32,7 +32,9 @@ export const LargePipelineWarning = ({
         Render it anyway
       </Button>
       <div className="pipeline-warning__disable">
-        <button onClick={onDisable}>Don't show this again</button>
+        <button className="pipeline-warning__disable__btn" onClick={onDisable}>
+          Don't show this again
+        </button>
       </div>
     </div>
   ) : null;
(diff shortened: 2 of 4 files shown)


bru5 (approved):
Looks good.

Should the underline on "Don't show this again" button remain on hover rather than disappear though? Maybe make the text a little darker as a hover effect instead.

studioswong (commented):
One idea to add on: the wording "Don't show this again" seems to convey to the user that it won't be shown ever again ( which in this case, it would once the user refreshes the browser ) - shall we also save this flag to localStorage to save the user's preference?

src/config.js
```diff
@@ -44,7 +44,7 @@ export const flags = {
   sizewarning: {
     description: 'Show a warning before rendering very large graphs',
     default: true,
-    icon: '⚠️',
+    icon: '🐳',
```
studioswong, line 47: 👍  chonky whale

richardwestenra:
@studioswong It won't be shown again: That's the point! :)

It saves the user's preference to localStorage by updating the `sizewarning` flag value. Users can revert this change later if needed but only by updating the flag value via the URL.

richardwestenra:
@bru5 I had another go at the button hover effect and then it occurred to me that I could just use the Kedro-UI Button with `size="small"` and `mode="secondary"` props. Works really well and lets me delete my other styling! Hope this works for you:

image

studioswong (approved):
the new styling on the 'Don't show again' button looks great!

PR #394, opened 2021-03-12
[KED-1951] Backend to send modular pipelines to kedro viz

MerelTheisenQB:
## Description

As part of the work on showing modular pipelines in viz.

## Development notes

Added modular pipelines as top level key as well as to individual nodes (both "node"-nodes and parameter/dataset nodes). 

Merging it into feature branch `feature/modular-pipelines-complete`, where the FE changes will be merged as well. Then we'll do a full check and make sure it all works before creating a final PR to main.

## QA notes

Update test setup and `animals.mock.json` to contain modular pipeline data.

FYI: JS tests will fail now I've updated the test data. @studioswong will address these tests once this PR has been merged to the main feature branch `feature/modular-pipelines-complete`

## Todo:
- [ ] Handle modular pipelines for parameter nodes. Will be done in a separate PR.
 
## Checklist

## Legal notice


package/kedro_viz/server.py
```diff
@@ -317,9 +318,22 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
     # keep track of node_id -> set(child_node_ids) for layers sorting
     node_dependencies = defaultdict(set)
     tags = set()
+    # keep track of modular pipelines
+    modular_pipelines = set()
 
     for pipeline_key, pipeline in pipelines.items():
         pipelines_list.append({"id": pipeline_key, "name": _pretty_name(pipeline_key)})
+        modular_pipelines.update(
```
idanov, line 326: This will include the modular pipelines from all pipelines, but we actually need only the ones from the currently selected pipeline.

package/kedro_viz/server.py
```diff
@@ -342,13 +356,19 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
         else pipelines_list[0]["id"]
     )
 
+    sorted_modular_pipelines = [
+        {"id": modular_pipeline, "name": _pretty_name(modular_pipeline.split(".")[-1])}
+        for modular_pipeline in sorted(modular_pipelines)
+    ]
```
idanov, line 362: 👍

package/kedro_viz/server.py
```diff
@@ -317,9 +318,22 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
...
         pipelines_list.append({"id": pipeline_key, "name": _pretty_name(pipeline_key)})
+        modular_pipelines.update(
+            set(
+                chain.from_iterable(
+                    [
+                        _expand_namespaces(x.namespace)
+                        for x in pipeline.nodes
+                        if x.namespace
+                    ]
+                )
```
idanov, line 334: I think this can be made even more clear and concise:
```suggestion
modular_pipelines = list({y for y in _expand_namespaces(x.namespace)
                        for x in pipeline.nodes})
```

idanov (changes requested):
Currently the code combines all modular pipelines from all registered pipelines. However for the proper functioning of the frontend, we need the modular pipelines only from the currently selected registered pipeline.

package/kedro_viz/server.py
MerelTheisenQB: When viz first fetches data about the pipelines and nodes etc it will call this `format_pipelines_data`. When this call is made, the default pipeline will be the selected pipeline. As far as I understand we don't do any filtering at this point, so the pipeline data returned here contains everything. Subsequent calls will then go to the `/api/pipelines/<string:pipeline_id>` endpoint, where indeed the data will be filtered to only return details for the selected data.

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
+    for chunk in namespace_chunks:
+        if prefix:
+            prefix = f"{prefix}.{chunk}"
+        else:
+            prefix = chunk
+        namespace_list.append(prefix)
+    return namespace_list
+
+
+def _add_parameter_data_to_node(dataset_namespace, task_id):
```
MerelTheisenQB, line 513: Extracted this to a method as linting was failing for number of statements in `format_pipeline_data`

package/kedro_viz/server.py
```diff
@@ -501,11 +561,14 @@ def pipeline_data(pipeline_id):
 
     pipeline_node_ids = set()
     pipeline_nodes = []
+    modular_pipelines = set()
 
     for node in _DATA["nodes"]:
         if pipeline_id in node["pipelines"]:
             pipeline_node_ids.add(node["id"])
             pipeline_nodes.append(node)
+            if node["type"] == "task" and node.get("modular_pipelines"):
```
MerelTheisenQB, line 570: Nodes that aren't tasks could be part of multiple modular pipelines (same dataset used in several modular pipelines), so the modular pipeline will only be part of the selected pipeline if it comes from a task node.

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
         else:
             nodes[task_id]["pipelines"].append(pipeline_key)
 
         for data_set in node.inputs:
-            namespace = data_set.split("@")[0]
-            namespace_to_layer[namespace] = dataset_to_layer.get(data_set)
-            namespace_id = _hash(namespace)
+            dataset_namespace = data_set.split("@")[0]
+            if node.namespace:
+                namespace_to_modular_pipelines[dataset_namespace] = _expand_namespaces(
```
limdauto, line 426: Maybe extract `_expand_namespaces(node.namespace)` into a variable so you can re-use it in 3 places?

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
-                if namespace == "parameters":
-                    _JSON_NODES[task_id]["parameters"] = _get_dataset_data_params(
-                        namespace
-                    ).load()
-                else:
-                    parameter_name = namespace.replace("params:", "")
-                    parameter_value = _get_dataset_data_params(namespace).load()
-                    _JSON_NODES[task_id]["parameters"][parameter_name] = parameter_value
+            if _is_namespace_param(dataset_namespace):
+                _add_parameter_data_to_node(dataset_namespace, task_id)
```
limdauto, line 441: 👍

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
-
-                if namespace == "parameters":
-                    _JSON_NODES[task_id]["parameters"] = _get_dataset_data_params(
-                        namespace
-                    ).load()
-                else:
-                    parameter_name = namespace.replace("params:", "")
-                    parameter_value = _get_dataset_data_params(namespace).load()
-                    _JSON_NODES[task_id]["parameters"][parameter_name] = parameter_value
+            if _is_namespace_param(dataset_namespace):
```
limdauto, line 440: Can we rename this to `_is_dataset_param`?

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
+                "layer": dataset_namespace_to_layer[dataset_namespace],
                 "pipelines": [pipeline_key],
+                "modular_pipelines": namespace_to_modular_pipelines.get(
+                    dataset_namespace, []
+                ),
             }
             nodes_list.append(nodes[node_id])
         else:
             nodes[node_id]["pipelines"].append(pipeline_key)
+            mod_pipelines = namespace_to_modular_pipelines.get(dataset_namespace, [])
```
limdauto, line 490:
```suggestion
            modular_pipelines = namespace_to_modular_pipelines.get(dataset_namespace, [])
```

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
+            check = nodes[node_id]["modular_pipelines"] and all(
```
limdauto, line 491: Sorry, what is this checking?

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
+                item in mod_pipelines for item in nodes[node_id]["modular_pipelines"]
+            )
+            if not check:
+                nodes[node_id]["modular_pipelines"] += mod_pipelines
+
+
+def _expand_namespaces(namespace):
```
limdauto, line 498: Could we put a docstring here with an example input and output of this function?

package/tests/test_server.py
```diff
@@ -106,6 +119,7 @@ def de_pipeline():
                 outputs=["pig", "giraffe"],
                 name="shark",
                 tags=["medium", "large"],
+                namespace="pipeline1.data_engineering",
```
limdauto, line 122: Could we add another pipeline called `pipeline2.data_engineering`? I'd like to see how it behaves when there is conflict in names.

limdauto (commented):
A few comments. I think we are nearly there. Could we add an unit tests for `_expand_namespace` to document its behaviour please? Also if we make the logic to prettify a modular pipeline name into a function to unit test it as well, that'd be helpful.

package/kedro_viz/server.py
limdauto: Let's add this explanation as a comment in the code.

package/tests/test_server.py
```diff
@@ -92,7 +96,16 @@ def get_pipeline(name: str = None):
 
 def ds_pipeline():
     ds_pipeline = Pipeline(
-        [node(trout, inputs=["pig", "sheep"], outputs=["whale"], name="trout")]
+        [
+            node(trout, inputs=["pig", "sheep"], outputs=["whale"], name="trout"),
```
idanov, line 100: 🐟 (🐷 , 🐑 ) = 🐳

package/tests/test_server.py
```diff
@@ -337,6 +351,8 @@ def test_pipelines_endpoint(cli_runner, client):
     # make sure only edges in the selected pipelines are returned
 
     assert data["edges"] == [
+        {"source": "6525f2e6", "target": "689862f0"},
```
idanov, line 354: These random hex numbers confuse me a lot... not sure why the frontend wants them like this....

src/utils/data/animals.mock.json
```diff
@@ -1,89 +1,67 @@
 {
   "edges": [
+    { "source": "9d989e8d", "target": "f42dab68" },
```
idanov, line 3: This formatting is so much better and sanity preserving! Thanks @MerelTheisenQB and/or `black` (I presume) 👏

src/utils/data/animals.mock.json
```diff
@@ -1,89 +1,67 @@
...
-    {
-      "source": "f42dab68",
-      "target": "fac8f1a3"
+      "id": "pipeline1",
+      "name": "Pipeline1"
     },
     {
-      "source": "2cd4ba93",
-      "target": "e27376a9"
+      "id": "pipeline1.data_engineering",
```
idanov, line 28: Hm, I think we might've not foreseen this in the ticket description... Having `pipeline1` as well here is probably quite useful for the UX in the frontend  (despite not directly assigned to any node). I suppose we can reiterate on that once we have this one merged and just keep a note when we can see things in the frontend.

package/kedro_viz/server.py
idanov: Is that true though? The membership of a pipeline is entirely determined from the name of the dataset and cannot be shared between two hierarchies. Isn't it more correct to say that a node (task) can consume and produce datasets from/to different modular pipelines than its own?

package/kedro_viz/server.py
```diff
@@ -376,20 +386,26 @@ def format_pipeline_data(
         node_dependencies: Dictionary of id and node dependencies.
         edges_list: List of all edges.
         nodes_list: List of all nodes.
+        modular_pipelines: Set of modular pipelines for all nodes.
 
     """
     # keep_track of {data_set_namespace -> set(tags)}
-    namespace_tags = defaultdict(set)
+    dataset_namespace_tags = defaultdict(set)
```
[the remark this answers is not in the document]
MerelTheisenQB: Okay cool, I wanted to rename it to that at first but then hesitated because it seemed like maybe the namespace part did have a deeper meaning.

package/kedro_viz/server.py
```diff
@@ -501,11 +561,14 @@ def pipeline_data(pipeline_id):
...
     pipeline_node_ids = set()
     pipeline_nodes = []
+    modular_pipelines = set()
 
     for node in _DATA["nodes"]:
         if pipeline_id in node["pipelines"]:
             pipeline_node_ids.add(node["id"])
             pipeline_nodes.append(node)
+            if node["type"] == "task" and node.get("modular_pipelines"):
+                modular_pipelines.update(node.get("modular_pipelines"))
```
idanov, line 571:
```suggestion
                modular_pipelines.update(node["modular_pipelines"])
```

package/kedro_viz/server.py
```diff
@@ -342,13 +345,19 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
         else pipelines_list[0]["id"]
     )
 
+    sorted_modular_pipelines = [
+        {"id": modular_pipeline, "name": _pretty_name(modular_pipeline.split(".")[-1])}
```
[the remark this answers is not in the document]
MerelTheisenQB: On the ticket it said that the name for `pipeline1.pipeline2.pipeline3` would be `Pipeline3`. I think your point is very valid, but just wondering what the initial reasoning was on the ticket.

package/kedro_viz/server.py
MerelTheisenQB: These are lists so the check is verifying whether there are any modular pipelines and if so that the ones on the dataset don't already exist in the list.

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
             }
             nodes_list.append(nodes[task_id])
         else:
             nodes[task_id]["pipelines"].append(pipeline_key)
 
         for data_set in node.inputs:
-            namespace = data_set.split("@")[0]
-            namespace_to_layer[namespace] = dataset_to_layer.get(data_set)
-            namespace_id = _hash(namespace)
+            dataset_namespace = data_set.split("@")[0]
```
idanov, line 424:
```suggestion
            dataset_full_name = data_set.split("@")[0]
```

package/kedro_viz/server.py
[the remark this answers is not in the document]
idanov: Yep, here we have a clash of terminology (it's something we no longer refer to as namespace, but that's how it was introduced initially). Dataset names have the following format:

```
(na.me.sp.ace.)name(@format)
```
The namespace part and the format part are optional, the name is not. Maybe worth having a helper function to extract the ones we care about - `namespace` and `name`.

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
+            if _is_namespace_param(dataset_namespace):
+                _add_parameter_data_to_node(dataset_namespace, task_id)
 
         for data_set in node.outputs:
-            namespace = data_set.split("@")[0]
-            namespace_to_layer[namespace] = dataset_to_layer.get(data_set)
-            namespace_id = _hash(namespace)
+            dataset_namespace = data_set.split("@")[0]
+            if node.namespace:
+                namespace_to_modular_pipelines[dataset_namespace] = _expand_namespaces(
```
idanov, line 446: The belonging to modular pipelines of a dataset is not determined by the nodes using it, but only by it's name. This one assigns more modular pipelines for a dataset and will cause a lot of confusion when expanding / collapsing pipelines. E.g. what will happen if I collapse one of the modular pipelines this belongs to?

The hierarchical relationship is exclusive unlike tags, i.e. a node / dataset can belong to only one hierarchy of namespaces.

package/kedro_viz/server.py
```diff
@@ -398,70 +414,114 @@ def format_pipeline_data(
...
             }
             nodes_list.append(nodes[node_id])
         else:
             nodes[node_id]["pipelines"].append(pipeline_key)
+            mod_pipelines = namespace_to_modular_pipelines.get(dataset_namespace, [])
+            check = nodes[node_id]["modular_pipelines"] and all(
+                item in mod_pipelines for item in nodes[node_id]["modular_pipelines"]
+            )
+            if not check:
+                nodes[node_id]["modular_pipelines"] += mod_pipelines
```
idanov, line 495: If we are using sets here, all of those checks for membership, combination without duplication, etc... can be handled super easily...
```
nodes[node_id]["modular_pipelines"] |= mod_pipelines
```

We won't even need the variable here... E.g. if we made `namespace_to_modular_pipelines = defaultdict(set)`, then we just need to do this:
```
nodes[node_id]["modular_pipelines"] |= namespace_to_modular_pipelines[dataset_name]
```

package/kedro_viz/server.py
idanov: 👏 Super nice you started a slow refactoring of the code here! I think the code for the Viz backend can definitely benefit from a bigger refactoring...

idanov (commented):
Glad to see we're getting there and almost ready! There's two main things to address here:

- In the current implementation, the membership of a dataset is determined by the nodes using it, which is incorrect. Modular pipeline membership is determined **entirely from the name of the dataset**, otherwise it will produce incorrect result.
- Currently the modular pipelines for the whole pipeline do not include all expanded pipelines. Not sure if that is added to the requirement in the ticket and we can probably get away for now without adding it, but the most correct definition of the top level `modular_pipelines` field should be a superset of all `modular_pipelines` fields of all the nodes in the pipeline. That's actually a very easy way to implement it as well, so maybe worth looking into doing it this way. Could be in a separate PR though, since it's not critical :)

Well done 👏  on actually prompting a refactoring for the code here and highlighting how confusing it is already.

package/kedro_viz/server.py
MerelTheisenQB: Ah yes your phrasing is much better. The thing I was mainly struggling with here is the list of modular pipelines that should be shown for the selected pipeline and I ended up only showing those which task nodes are part of and not the input/output of those nodes. I'll add a comment explaining this.

MerelTheisenQB:
> Glad to see we're getting there and almost ready! There's two main things to address here:
> 
> * In the current implementation, the membership of a dataset is determined by the nodes using it, which is incorrect. Modular pipeline membership is determined **entirely from the name of the dataset**, otherwise it will produce incorrect result.

I'm **definitely** missing something here. How do you get the modular pipeline from the name of the dataset? How does a user add that? 🤯

> * Currently the modular pipelines for the whole pipeline do not include all expanded pipelines. Not sure if that is added to the requirement in the ticket and we can probably get away for now without adding it, but the most correct definition of the top level `modular_pipelines` field should be a superset of all `modular_pipelines` fields of all the nodes in the pipeline. That's actually a very easy way to implement it as well, so maybe worth looking into doing it this way. Could be in a separate PR though, since it's not critical :)


I might be misunderstanding something, but the modular pipelines added for the full pipeline does include all extended pipelines. If you look at the `animals.mock.json` file you'll see that spread around the nodes we have `["pipeline1", "pipeline1.data_engineering"]` and `["pipeline2", "pipeline2.data_science"]` and for the full pipeline we have:
```
"modular_pipelines": [
    {
      "id": "pipeline1",
      "name": "Pipeline1"
    },
    {
      "id": "pipeline1.data_engineering",
      "name": "Data Engineering"
    },
    {
      "id": "pipeline2",
      "name": "Pipeline2"
    },
    {
      "id": "pipeline2.data_science",
      "name": "Data Science"
    }
  ],
```

src/utils/data/animals.mock.json
MerelTheisenQB: It's just above (looks a bit odd through the way git is showing the diff)

package/kedro_viz/server.py
MerelTheisenQB: The problem is that sets aren't JSON serialisable and so `nodes[node_id]["modular_pipelines"]` eventually has to be a list.

src/utils/data/animals.mock.json
```diff
@@ -1,89 +1,71 @@
...
-      "source": "6525f2e6",
-      "target": "e27376a9"
+      "id": "pipeline2",
+      "name": "Pipeline2"
     },
     {
-      "source": "e27376a9",
-      "target": "1769e230"
+      "id": "pipeline2.data_science",
+      "name": "Data Science"
```
MerelTheisenQB, line 41: @limdauto Added `pipeline2.data_science` and `pipeline1.data_science` to show the "nameclash" case.

package/tests/test_server.py
```diff
@@ -71,10 +78,15 @@ def trout(pig, sheep):
...
+def tuna(sheep):
+    return sheep
+
+
 def get_pipelines():
     return {
         "de": de_pipeline(),
-        "ds": ds_pipeline(),
+        "pre_ds": pipeline(pre_ds_pipeline(), namespace="pipeline1.data_science"),
+        "ds": pipeline(ds_pipeline(), namespace="pipeline2.data_science"),
```
MerelTheisenQB, line 89: Changed the setup to include modular pipelines properly.

package/kedro_viz/server.py
```diff
@@ -398,72 +425,108 @@ def format_pipeline_data(
...
                 "id": node_id,
-                "name": _pretty_name(namespace),
-                "full_name": namespace,
+                "name": _pretty_name(dataset_full_name),
+                "full_name": dataset_full_name,
                 "tags": sorted(tag_names),
-                "layer": namespace_to_layer[namespace],
+                "layer": dataset_name_to_layer[dataset_full_name],
                 "pipelines": [pipeline_key],
+                "modular_pipelines": dataset_modular_pipelines,
```
limdauto, line 489: I think we discussed that people can technically name their dataset `foo.bar` with `foo`  & `bar` not being modular pipeline, so we need to check whether they are in the valid modular pipelines list (which comes from the nodes)

package/kedro_viz/server.py
```diff
@@ -398,72 +425,108 @@ def format_pipeline_data(
...
             nodes_list.append(nodes[node_id])
         else:
             nodes[node_id]["pipelines"].append(pipeline_key)
 
 
+def _expand_namespaces(namespace):
+    """
+    Expand a node's namespace to the modular pipelines this node belongs to.
+    For example, if the node's namespace is: "pipeline1.data_science"
+    it should be expanded to: ["pipeline1", "pipeline1.data_science"]
```
limdauto, line 500: Thank you 👍

package/kedro_viz/server.py
```diff
@@ -398,72 +425,108 @@ def format_pipeline_data(
...
-
-                if namespace == "parameters":
-                    _JSON_NODES[task_id]["parameters"] = _get_dataset_data_params(
-                        namespace
-                    ).load()
-                else:
-                    parameter_name = namespace.replace("params:", "")
-                    parameter_value = _get_dataset_data_params(namespace).load()
-                    _JSON_NODES[task_id]["parameters"][parameter_name] = parameter_value
+            if _is_dataset_param(dataset_full_name):
```
idanov, line 445: 👏

package/kedro_viz/server.py
```diff
@@ -296,7 +296,15 @@ def _pretty_name(name: str) -> str:
     return " ".join(parts)
 
 
-def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
+def _pretty_modular_pipeline_name(modular_pipeline: str) -> str:
```
idanov, line 299: 👏

package/kedro_viz/server.py
```diff
@@ -342,18 +353,27 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
         else pipelines_list[0]["id"]
     )
 
+    sorted_modular_pipelines = [
+        {
+            "id": modular_pipeline,
+            "name": _pretty_modular_pipeline_name(modular_pipeline),
+        }
+        for modular_pipeline in sorted(modular_pipelines)
```
idanov, line 361: 👏

package/kedro_viz/server.py
```diff
@@ -342,18 +353,27 @@ def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, list]:
...
         "layers": sorted_layers,
         "pipelines": pipelines_list,
         "selected_pipeline": selected_pipeline,
+        "modular_pipelines": sorted_modular_pipelines,
     }
 
 
-def _is_namespace_param(namespace: str) -> bool:
-    """Returns whether a dataset namespace is a parameter"""
+def _is_dataset_param(namespace: str) -> bool:
```
idanov, line 375: 👏

package/kedro_viz/server.py
```diff
@@ -398,72 +425,108 @@ def format_pipeline_data(
...
         else:
             nodes[task_id]["pipelines"].append(pipeline_key)
 
         for data_set in node.inputs:
-            namespace = data_set.split("@")[0]
-            namespace_to_layer[namespace] = dataset_to_layer.get(data_set)
-            namespace_id = _hash(namespace)
+            dataset_full_name = data_set.split("@")[0]
+            dataset_name_to_layer[dataset_full_name] = dataset_to_layer.get(data_set)
+            namespace_id = _hash(dataset_full_name)
```
idanov, line 437: Maybe call it `dataset_id` or `graph_node_id`?

package/kedro_viz/server.py
```diff
@@ -398,72 +425,108 @@ def format_pipeline_data(
...
-                if namespace == "parameters":
-                    _JSON_NODES[task_id]["parameters"] = _get_dataset_data_params(
-                        namespace
-                    ).load()
-                else:
-                    parameter_name = namespace.replace("params:", "")
-                    parameter_value = _get_dataset_data_params(namespace).load()
-                    _JSON_NODES[task_id]["parameters"][parameter_name] = parameter_value
+            if _is_dataset_param(dataset_full_name):
+                _add_parameter_data_to_node(dataset_full_name, task_id)
```
idanov, line 446: 👏

limdauto (commented):
Apart from one last comment around dataset namespace, this is it :D Thank you so much for pushing through!

package/kedro_viz/server.py
MerelTheisenQB: Ah okay, thanks for flagging that! I had that in mind for parameters, but will check for it for datasets as well 👍

idanov (approved):
Well done 🎉 Super curious to see this served to Viz. I think we can start experimenting even before we have the `parameters` work done.

package/kedro_viz/server.py
MerelTheisenQB: Done in https://github.com/quantumblacklabs/kedro-viz/pull/394/commits/50ca16c73d0e363ac69ae870e51368470fa915b1

PR #398, opened 2021-03-16
[KED-2365] Simplify layout algorithm, improve layout quality and performance

bru5:
## Description

Simplify layout algorithm, improve layout quality and performance.

## Development notes

Performance and algorithmic improvements:

- simplified and reduced overall implementation
- reduced average complexity / run time
- reduced memory usage and garbage collection

Example performance change for 868 nodes + 1555 edges test case
- 56% time reduction (2.27x speedup)
- 91% less object allocation

Output changes and quality improvements:

- consistently compact layouts
- improved node alignment, regularity, symmetry, linearity
- increased spacing around nodes with high number of edges
- improved layout spacing and routing when node labels enabled
- added more specific constraint tests

## QA notes

Check various pipelines, selections and compare results with main branch.

Load time should be significantly reduced and larger pipelines should show some visual improvement described above. 

Results may look somewhat different than previous renderings but in general should be as good or better than before.

## Checklist

## Legal notice

### DIFF
--- a/src/utils/graph/graph.js
+++ b/src/utils/graph/graph.js
@@ -4,12 +4,12 @@ import { routing } from './routing';
 
 const defaultOptions = {
   layout: {
-    spaceX: 16,
+    spaceX: 14,
     spaceY: 110,
     layerSpaceY: 55,
-    basisX: 1500,
+    spreadX: 2.2,
     padding: 100,
-    iterations: 20,
+    iterations: 25,
   },
   routing: {
     spaceX: 26,
--- a/src/utils/graph/layout.js
+++ b/src/utils/graph/layout.js
@@ -1,5 +1,5 @@
 import { halfPI, snap, angle, compare, groupByRow } from './common';
-import { solve } from './solver';
+import { solveLoose, solveStrict } from './solver';
 import {
   rowConstraint,
   layerConstraint,
@@ -52,7 +52,7 @@ export const layout = ({
   const layerConstraints = createLayerConstraints(nodes, layers);
 
   // Find the node positions given these constraints
-  solve([...rowConstraints, ...layerConstraints], constants, 1, true);
+  solveStrict([...rowConstraints, ...layerConstraints], constants, 1);
 
   // Find the solved rows using the node positions after solving
   const rows = groupByRow(nodes);
@@ -63,15 +63,15 @@ export const layout = ({
 
   // Solve these constraints iteratively
   for (let i = 0; i < iterations; i += 1) {
-    solve(crossingConstraints, constants, 1);
-    solve(parallelConstraints, constants, 50);
+    solveLoose(crossingConstraints, 1, constants);
+    solveLoose(parallelConstraints, 50, constants);
   }
 
   // Constraints to maintain a minimum horizontal node spacing
   const separationConstraints = createSeparationConstraints(rows, constants);
 
   // Find the final node positions given these strict constraints
-  solve([...separationConstraints, ...parallelConstraints], constants, 1, true);
+  solveStrict([...separationConstraints, ...parallelConstraints], constants, 1);
 
   // Adjust vertical spacing between rows for legibility
   expandDenseRows(edges, rows, spaceY);
--- a/src/utils/graph/layout.js
+++ b/src/utils/graph/layout.js
@@ -16,9 +16,9 @@ import {
  * @param {array} params.nodes The input nodes
  * @param {array} params.edges The input edges
  * @param {object=} params.layers The node layers if specified
- * @param {number} params.spreadX The amount to adapt spacing in X
  * @param {number} params.spaceX The minimum gap between nodes in X
  * @param {number} params.spaceY The minimum gap between nodes in Y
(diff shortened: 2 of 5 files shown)


src/utils/graph/graph.js
```diff
@@ -4,12 +4,12 @@ import { routing } from './routing';
 
 const defaultOptions = {
   layout: {
-    spaceX: 16,
+    spaceX: 14,
     spaceY: 110,
     layerSpaceY: 55,
-    basisX: 1500,
+    spreadX: 2.2,
```
studioswong, line 10: What does both the `basisX` and `spreadX` values refer to, and in this case, what does this change effectively means and do? It would help if we have some comments for the `defaultOptions` section for context.

src/utils/graph/layout.js
```diff
@@ -16,7 +16,7 @@ import {
  * @param {array} params.nodes The input nodes
  * @param {array} params.edges The input edges
  * @param {object=} params.layers The node layers if specified
- * @param {number} params.basisX The basis relating diagram width in X
+ * @param {number} params.spreadX The amount to adapt spacing in X
```
studioswong, line 19: I wonder what is the main motivation behind adapting spacing instead of the relation to the width of the diagram, and how does it improve the layout engine? It would be worth mentioning which goal this change relates to in the PR decription.

src/utils/graph/layout.js
```diff
@@ -39,56 +39,36 @@ export const layout = ({
...
-
-    // Update and solve separation constraints given the updated positions
-    updateSeparationConstraints(separationConstraints, rows, spaceX);
-    solve(separationConstraints, constants, iterations * 0.5);
+    solve(parallelConstraints, constants, 50);
   }
 
-  // Update separation constraints but ensure spacing is exact
-  updateSeparationConstraints(separationConstraints, rows, spaceX, true);
+  // Constraints to maintain a minimum horizontal node spacing
```
studioswong, line 70: This change is very interesting! Similar to my comment above, it would be great to document the reasoning / motivation behind this change in the PR description.

studioswong (commented):
Thanks for introducing these improvements @bru5 , I haven't extensively tested the new layout yet, I'm sure it's good stuff!

Just some initial comments: 

- I notice in the PR decription on the list of improvements, but seems to be missing the 'how' - I notice the various changes that was introduced in this PR, such as change in constraints, etc, and it would be great if you could include the explanation of the relating changes for each improvement so whoever reviewing this understand and relate the changes with the motivations behind. Currently it's a bit of a guessing game in understanding and relating the how to the list of improvements. 
- There are a lot of calculations for the constraints - these needs unit tests. 

I will have a more thorough read and further poke around and let you know if I see anything!

limdauto:
+1 to @studioswong's suggestion. I would love to know how these improve the current engine for future reference.

studioswong (commented):
I've just tested out the graph, the new layout and improvements looks good! I notice that these new constraints and improvement only apply to the new graph - I persume these improvements does not apply to the old Dagre graph layout? 

Following my previous comment, the constraints itself now contains quite a bit of calculations  - tests would need to set up for these calculations within the constraints. I notice that there is currently no test file set up for `constraints.js` - perhaps it's time we set one up to ensure the constraints are correctly tested? 

Thank you!

richardwestenra (approved):
💯

src/utils/graph/layout.js
bru5: Ah I've updated the comment here to make it a bit clearer what that parameter does, though it's not really expected that this value will change from the included default constant which is pretty implementation relative and was found by trial and error.

If you're really curious though it's worth following the parameter through the code and playing with the value while watching in the browser.

src/utils/graph/layout.js
bru5: Ah this part was just a refactoring and simplification so I've not much to add on that.

src/utils/graph/layout.js
```diff
@@ -337,7 +275,7 @@ const expandDenseRows = (edges, rows, spaceY) => {
     const density = densities[i];
 
     // Snap to improve vertical rhythm
-    const offsetY = snap(density * spaceY, Math.round(spaceY * 0.25));
+    const offsetY = snap(density * 1.25 * spaceY, Math.round(spaceY * 0.25));
```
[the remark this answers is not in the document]
bru5: Yeah I know what you mean. As much as I'd like to often the problem is these kind of constants tend to be really abstract and highly implementation specific, so I tend to avoid making them appear like they are easily 'configurable' unless I know they are very well defined. In this case I gave it a shot though, so hopefully it helps.

bru5:
Thanks for the comments all, I've now added more specific tests for each constraint. 

I realise there is still some way to go in explaining the algorithm at a higher level than the implementation. I'll work on a higher level supporting document for that, since this is not particularly straightforward and is very abstract. 

Often I do find that code comments are not the best format for discussing things at a high level, for now they are describing the specific implementation and the API but I can definitely add cross references once complete.

PR #400, opened 2021-03-18
Only look at target nodes when finding nearest node with layers

limdauto:
Since layers dependency flows one way, we want to make sure that when finding nearest node with layers, we must also follow the one-way direction from source to target.

## Description

Imagine this pipeline shape:

Screenshot 2021-03-18 at 15 03 13

There is a dataset in the `raw` layer that serves as the input to the `models` layer. Therefore, the current behaviour is that nodes in the `models` layer will think the closest node with layer to them is in `raw` and create a direct constraint between `raw` and `models`, which is incorrect. When looking for nearest nodes with layer, we should only look ahead along the layers dependency flow, not behind. 

## QA notes

I have manually tested this.

## Checklist

## Legal notice

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,7 +15,7 @@ Please follow the established format:
 
 ## Bug fixes and other changes
 
-<!-- Add release notes for the upcoming release here -->
+- Fix layers visualisation to only look for target nodes when building layers dependency.
 
 # Release 3.10.1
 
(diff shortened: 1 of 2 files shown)


bru5 (changes requested):
As discussed looks like there are some issues with this approach, so hold off on this for now until we can take a closer look.

RELEASE.md
```diff
@@ -15,7 +15,7 @@ Please follow the established format:
 
 ## Bug fixes and other changes
 
-<!-- Add release notes for the upcoming release here -->
+- Fix layers visualisation to only look for target nodes when building layers dependency.
```
richardwestenra, line 18: Do you mind leaving the comment (`<!-- Add release notes for the upcoming release here -->`) here, so that it gets carried over and is visible for subsequent contributors?

PR #405, opened 2021-03-24
Use full namespace as name for modular pipeline

MerelTheisenQB:
## Description

Only sending the last part of the modular pipeline as the name can lead to having the same name show up on the viz side multiple times.  

## Development notes

Prettify full modular pipeline namespace and send it to the frontend. 

## QA notes

Updated test data to match this. 

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -296,12 +296,11 @@ def _pretty_name(name: str) -> str:
     return " ".join(parts)
 
 
-def _pretty_modular_pipeline_name(modular_pipeline: str) -> str:
-    """Takes the namespace of a modular pipeline and prettifies the
-    last part to show as the modular pipeline name."""
-    chunks = modular_pipeline.split(".")
-    last_chunk = chunks[-1]
-    return _pretty_name(last_chunk)
+def _pretty_modular_pipeline_name(modular_pipeline_name: str) -> str:
+    """Takes the namespace of a modular pipeline and prettifies it
+     to show as the human readable modular pipeline name."""
+    modular_pipeline_name = modular_pipeline_name.replace(".", " ")
+    return _pretty_name(modular_pipeline_name)
 
 
 def format_pipelines_data(pipelines: Dict[str, "Pipeline"]) -> Dict[str, Any]:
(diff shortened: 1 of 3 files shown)


src/utils/data/animals.mock.json
```diff
@@ -26,19 +26,19 @@
     },
     {
       "id": "pipeline1.data_engineering",
-      "name": "Data Engineering"
+      "name": "Pipeline1 Data Engineering"
```
MerelTheisenQB, line 29: @studioswong Does it make sense to send the name like this? Or should it still contains some sort of "level" separation. E.g. `Pipeline1 - Data Engineering`?

datajoely: If we're doing some sort of nesting we should use `.` or `/` syntax as this is analogous to a directory structure

package/kedro_viz/server.py
```diff
@@ -296,12 +296,11 @@ def _pretty_name(name: str) -> str:
     return " ".join(parts)
 
 
-def _pretty_modular_pipeline_name(modular_pipeline: str) -> str:
-    """Takes the namespace of a modular pipeline and prettifies the
-    last part to show as the modular pipeline name."""
-    chunks = modular_pipeline.split(".")
-    last_chunk = chunks[-1]
-    return _pretty_name(last_chunk)
+def _pretty_modular_pipeline_name(modular_pipeline_name: str) -> str:
```
datajoely, line 299: what's the rationale of taking it out of the pythonic way of thinking?

MerelTheisenQB: I think if we want to keep the dots, there's not really a point in sending both the `id` which is basically just the full namespace as well as a sanitised `name`. WDYT?

datajoely: I'm not sure where I should comment on implementation. But using the analogy of a file path, I think may be helpful provide  the `absolute` path so that we can do stuff later without the backend as well the equivalent of  `pathlib.Path.name` i.e. the last part of the path.

limdauto (commented):
The implementation looks good to me but I agree we might as well just take the `id` and display it. If we want to prettify it, we should keep some separator IMO.

MerelTheisenQB:
This is not needed at this time.

PR #407, opened 2021-03-26
[KED-2528] Improve layer solving approach when layers partially defined

bru5:
## Description

This PR improves the approach for inferring and rendering nodes in pipelines with only partially defined layers.

## Development notes

The algorithm for nearest layer search was changed to strictly follow rank order, avoiding edge cases where layer constraints could not be solved with the previous algorithm.

The strict constraint solver was also changed to allow partial solving whilst throwing a console warning, though this was no longer observed to happen in practice.

## QA notes

Tested using a large pipeline that has only partially defined layers (i.e. not the 'demo').
The 'demo' dataset, which has fully defined layers, was also tested.

Using both cases above, open the console to observe any errors or warnings, then test by enabling layers and toggling various combinations of nodes as well as toggling top level pipelines in various combinations. 

Each time check that the resulting render is in a sensible order, does not appear broken and there are no console errors or warnings. Since the order is inferred using only limited information, it may not always be exactly as desired but in this case the idea is that the user should add more layer information to nodes.

## Checklist

## Legal notice


src/utils/graph/graph.js
```diff
@@ -77,39 +77,41 @@ export const addEdgeLinks = (nodes, edges) => {
...
+    for (const layer of layers) {
+      validLayers[layer] = true;
+    }
+
+    const hasValidLayer = (node) => Boolean(node?.layer in validLayers);
+    const lastLayer = layers[layers.length - 1];
 
+    // For each node
     for (const node of nodes) {
+      // Find first descendant node that has a valid layer following rank order (including itself)
```
datajoely, line 97: What are they ranked by?

src/utils/graph/solver.js
```diff
@@ -75,15 +75,25 @@ const solveStrict = (constraints, constants) => {
...
+          variables[variableId(co.b, co.base.property)]
+        )
+      );
+    } catch (err) {
+      unsolvableCount += 1;
+    }
+  }
+
+  if (unsolvableCount > 0) {
+    console.warn(`Skipped ${unsolvableCount} unsolvable constraints`);
```
datajoely, line 96: Is there no more information worth spitting out to the console?

src/utils/graph/graph.js
limdauto: Nodes are ranked by their batch-topological order

src/utils/graph/solver.js
limdauto: Yea, would be nice to improve this error message some how

limdauto (approved):
Thank you 🙏  this does make sense.

src/utils/graph/solver.js
bru5: So I did experiment with this, but essentially the problem is that the constraint systems involved are not very human interpretable given their size and complexity, such that the individual constraints here themselves don't really directly relate to the cause of the user's problem in a meaningful way.

I think for now the notification will be enough to keep an eye on given we're not expecting to see it after this fix, but if there are further complications still then we can invest more into somehow making this more interpretable.

datajoely: Could we at least put the notification on screen rather than in the console?

PR #410, opened 2021-03-31
Modular pipeline tags concept

studioswong:
## Description

related PR: https://jira.quantumblack.com/browse/KED-2520

This PR is the main development work for the initial MVP for the modular pipeline feature,  in allowing the user to select the display of nodes on the flowchart by modular pipelines using a similar interaction as the tag filters on the current sidebar. (see below)

image

## Development notes

This PR contains the following main pieces of work:

1. Initial set up of the modular pipeline selectors to format the data to be consumed for the sidebar 
2. Addition of a series of actions related to toggle the selection of modular pipeline items ( I have set them up in a consistent fashion with the tag actions for code consistency), along with the setup of tests
3. Set up a new 'modularPipeline' type within the items array used in node-list to allow refactor work of all components related to the sidebar ( e.g node-list, node-list-group, etc)
4. Refactoring of a series of functions related to the tag filtering logic to cater for the new 'modularPipeline' item type. This includes adding a series of functions to cater for the toggling of modular pipeline items, as well as refactoring the main function that determines the filtering logic, `onTagItemChange`, to `onCategoryItemChange` to allow for more category item types beyond tags.
5. Adding a new display logic for the name of modular pipelines under `node-list-row-list` to represent the nested relationship of pipelines by indentation
6. Incorporating modularPipeline type items within the search results so users can search by modular pipelines from the search bar
7. Addition of a series of modular pipeline selectors used within the node-list component under `node-list-items.js` 
8. Refactoring of tests to cater for the new moularPipeline item types to test the modular pipeline filter functionality ( all related tests  for the filtering by modular pipelines sits under `node-list-items.test.js`) 
9. Addition of the 'modularpipeline' flag, where the modular pipeline field will only be exposed via enabling this flag
10. I have also refactored the `sections` selector to a simple function that takes in the value of the flag to return different object in determining the fields displayed on the sidebar. 
11. Update the `demo` test dataset and the `random-data` generator to include the new modular pipeline related fields

## QA notes

This feature needs to be tested alongside the new changes to the Kedro-viz BE under the `feature/modular-pipelines-complete` branch. I also recommend testing this with the new Optimus AI reference project. 

If you are testing with the Optimus AI project, please note that currently there are some errors with the display of layers on the flowchart until [this ticket](https://jira.quantumblack.com/browse/KED-2528) is merged. Please test with the layers turned off for the flowchart. 

## Checklist

## Legal notice


studioswong:
it looks like the pylint tests are currently failing on the CI - will resolve this with the core team in the meantime
image

src/components/node-list/node-list-items.js
```diff
@@ -176,14 +223,16 @@ export const getFilteredNodeItems = createSelector(
 
 /**
  * Get formatted list of sections
+ * @param {boolean} flag value of modularpipeline flag
  * @return {array} List of sections
  */
-export const getSections = createSelector(() =>
-  Object.keys(sidebar).map((name) => ({
+export const getSections = (flag) => {
```
bru5, line 229: Might be clearer to name the `flag` as something like `modularPipelineFlag`?

src/components/node-list/node-list-items.js
```diff
@@ -176,14 +223,16 @@ export const getFilteredNodeItems = createSelector(
...
+  const sidebarObject = sidebar(flag);
```
bru5, line 230: I think `sidebarObject` is probably the `sidebarConfig` here?

src/components/node-list/node-list-row-list.js
```diff
@@ -4,6 +4,19 @@ import { connect } from 'react-redux';
 import NodeListRow, { nodeListRowHeight } from './node-list-row';
 import LazyList from '../lazy-list';
 
+// modify display of labels for modular pipelines to show nested relationship
+const getItemLabel = (item) => {
+  const layer = '・';
+  const whiteSpace = '&nbsp;&nbsp;&nbsp;&nbsp;';
+  if (item.type === 'modularPipeline') {
+    const levels = item.id.match(/\./g) ? item.id.match(/\./g).length : 0;
```
bru5, line 12: I think it might be worth commenting on what the levels are based on here e.g. something around 'parse module depth from namespace'. Could even break it out into a specific function like `indentedLabel` for that part?

src/config.js
```diff
@@ -41,15 +41,33 @@ export const flags = {
...
-    Datasets: 'data',
-    Parameters: 'parameters',
+  modularpipeline: {
+    description: 'enables modular pipeline features',
+    default: false,
+    icon: '⛓️',
   },
 };
+
+export const sidebar = (flag) =>
```
bru5, line 51: Similar to earlier I think the flag should be specifically named here.

src/selectors/modular-pipelines.js
```diff
@@ -0,0 +1,54 @@
...
+const getModularPipelineName = (state) => state.modularPipeline.name;
+const getModularPipelineEnabled = (state) => state.modularPipeline.enabled;
+
+/**
+ * Retrieve the formatted list of modular pipeline filters
+ */
+export const getModularPipelineData = createSelector(
+  [getModularPipelineIDs, getModularPipelineName, getModularPipelineEnabled],
+  (modularPipelineIDs, modularPipelineName, modularPipelineEnabled) =>
+    modularPipelineIDs.sort().map((id) => ({
```
bru5, line 15: Might be a good idea to clone the array first, since sort is in-place on the original array. Usually I do this as a pattern just to be safe when the array comes from outside: `modularPipelineIDs.slice(0).sort()`

src/selectors/modular-pipelines.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * returns an array of modular pipelines with the corresponding
+ * nodes for each modular pipeline
+ */
+export const getModularPipelineNodes = createSelector(
+  [getNodesModularPipelines, getModularPipelineIDs],
+  (allNodes, modularPipelines) => {
+    return modularPipelines.map((modularPipeline) => {
+      let nodes = [];
+      Object.keys(allNodes).map((key) =>
```
bru5, line 43: I think since the return result of `map` isn't used here and the loop uses side effects it might be best to convert this into a `for` loop?

bru5 (commented):
Awesome! This works really nicely.

Looks very good overall, I've still some more testing to do on it but for now I've left a few suggestions.

src/actions/modular-pipelines.js
```diff
@@ -0,0 +1,33 @@
...
+      ? modularPipelineIDs
+      : [modularPipelineIDs],
+    active,
+  };
+}
+
+export const TOGGLE_MODULAR_PIPELINE_FILTER = 'TOGGLE_MODULAR_PIPELINE_FILTER';
+
+/**
+ * Toggle a modular pipeline's filtering on/off (or array of tags)
```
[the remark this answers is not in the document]
studioswong: good spot! Let me change that

src/components/node-list/index.js
```diff
@@ -87,41 +102,64 @@ const NodeListProvider = ({
   const onItemMouseLeave = (item) => {
     if (isTagType(item.type)) {
       onToggleTagActive(item.id, false);
+    } else if (isModularPipelineType(item.type)) {
+      onToggleModularPipelineActive(item.id, true);
```
richardwestenra, line 106: ⚠️

I think this should be `onToggleModularPipelineActive(item.id, false)` - this appears to be the cause of a bug that results in modular pipelines staying active (highlighted) once hovered over, even after mouseleave. Try hovering on/off and see if it behaves as you'd expect!

src/components/node-list/node-list-row-list.js
```diff
@@ -4,6 +4,19 @@ import { connect } from 'react-redux';
...
+    return `${whiteSpace
+      .repeat(levels)
+      .concat(layer.repeat(levels).concat(item.highlightedLabel))}`;
```
richardwestenra, line 15: Why are we wrapping this in a multiline string? It doesn't seem necessary to me.

```suggestion
    return whiteSpace
      .repeat(levels)
      .concat(layer.repeat(levels))
      .concat(item.highlightedLabel);
```

richardwestenra, line 15: Are more deeply nested pipelines supposed to have more than one dot? I thought they would just have a single dot, and that the number of spaces should determine the nesting level?

image

src/components/node-list/node-list-row-list.js
```diff
@@ -4,6 +4,19 @@ import { connect } from 'react-redux';
 import NodeListRow, { nodeListRowHeight } from './node-list-row';
 import LazyList from '../lazy-list';
 
+// modify display of labels for modular pipelines to show nested relationship
+const getItemLabel = (item) => {
```
richardwestenra, line 8: The rest of the label formatting is handled in selectors in `node-list-items.js` to improve performance. Do you think it would make more sense to handle this there?

src/selectors/modular-pipelines.js
```diff
@@ -0,0 +1,54 @@
...
+  })
+);
+
+/**
+ * returns an array of modular pipelines with the corresponding
+ * nodes for each modular pipeline
+ */
+export const getModularPipelineNodes = createSelector(
+  [getNodesModularPipelines, getModularPipelineIDs],
+  (allNodes, modularPipelines) => {
```
richardwestenra, line 40: I'm finding the variable name `allNodes` a bit confusing here - this is the state prop for the modular pipelines that correspond to certain nodes, right?

Is it possible that there might be a node that has a modular pipeline but which isn't included in this list? If so then using this prop in this way could lead to potential bugs.

src/selectors/modular-pipelines.js
```diff
@@ -0,0 +1,54 @@
...
+    return modularPipelines.map((modularPipeline) => {
+      let nodes = [];
+      Object.keys(allNodes).map((key) =>
+        allNodes[key]
+          ? allNodes[key].includes(modularPipeline) && nodes.push(key)
+          : null
+      );
```
richardwestenra, line 47: (optional)

I'd suggest using `forEach` here for clarity, and also preferring an `if` statement over the nested ternary, again to make it clearer to read. The current code implies that something is being returned, when it isn't.
```suggestion
      Object.keys(allNodes).forEach((key) => {
        if (allNodes[key] && allNodes[key].includes(modularPipeline)) {
          nodes.push(key);
        }
      });
```

src/utils/random-data.js
```diff
@@ -59,6 +62,24 @@ class Pipeline {
     return pipelines.filter(unique);
   }
 
+  /**
+   * Create the modualr pipelines aray
```
richardwestenra, line 66:
```suggestion
   * Create the modular pipelines array
```

richardwestenra (changes requested):
Hey! Overall this is great work! I've included a lot of optional suggestions, feel free to include or ignore. The most important things I've marked with a ⚠️  emoji. Thanks! :)

src/components/node-list/index.js
```diff
@@ -40,21 +45,30 @@ const NodeListProvider = ({
   onToggleNodeActive,
   onToggleTagActive,
   onToggleTagFilter,
+  onToggleModularPipelineActive,
+  onToggleModularPipelineFilter,
   onToggleTypeDisabled,
+  modularPipelines,
+  modularPipelinesEnabled,
+  flag,
```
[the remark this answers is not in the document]
studioswong: This is referring to the modular pipeline flag - sure I can rename it to the modular pipeline flag

src/components/node-list/node-list-items.js
studioswong: I have already elaborated that this refers to the modular pipeline flag in the comments, not sure of the need to repeat this in the naming and opt for a much longer variable name?

src/components/node-list/node-list-items.js
studioswong: sure however it's just a difference of preference in naming?

src/components/node-list/node-list-row-list.js
studioswong: sure I'll add on more comments about the indentation level in the comments

src/selectors/modular-pipelines.js
studioswong: Thanks for the comment, though I am not entirely sure of the advantages of converting this to a `for` loop given that in our codebase ( and generally for most JS development nowadays) we often favor using `map`?  This is a pttern that is used in our other selectors as well. 

personally I also think using a `map` allows the code to be much cleaner.

richardwestenra: I commented on this one too below and suggested using `forEach`. `map` is typically used to assign the result of a loop to a variable (e.g. `const result = array.map(d => d.foo)`), which this bit of code doesn't do. Hence, using `forEach` or `for` is clearer to a reader, as it doesn't imply that you intended to return the iterated array.

src/actions/actions.test.js
```diff
@@ -155,6 +161,62 @@ describe('actions', () => {
...
+    const active = false;
+    const expectedAction = {
+      type: TOGGLE_MODULAR_PIPELINE_ACTIVE,
+      modularPipelineIDs,
+      active,
+    };
+    expect(toggleModularPipelineActive(modularPipelineIDs, active)).toEqual(
+      expectedAction
+    );
+  });
```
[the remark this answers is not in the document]
studioswong: yes technically they are - you're right! I will take the first test out then

src/actions/actions.test.js
```diff
@@ -155,6 +161,62 @@ describe('actions', () => {
...
+    const enabled = false;
+    const expectedAction = {
+      type: TOGGLE_MODULAR_PIPELINE_FILTER,
+      modularPipelineIDs,
+      enabled,
+    };
+    expect(toggleModularPipelineFilter(modularPipelineIDs, enabled)).toEqual(
+      expectedAction
+    );
+  });
```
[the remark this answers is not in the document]
studioswong: sure thing!

src/components/node-list/node-list-row-list.js
studioswong: Yea I had a thought about this too, but given that this is more like a 'hack' rather than a permanent solution ( this indentation is subjected to further design changes) I believe this sits better here than in a selector ( so when the next design iteration happens this is more noticeable to spot as a temporary solution for modular pipelines). 

How about I add a comment to mention that this is a temporary formatting subject to delete / change on further design changes?

src/components/node-list/node-list-row-list.js
studioswong: great spot, I think I added this on following the comment at some point on the slack thread and we decided not to go for the 2 dots. I'll take out the multiple dot in the formatting

src/components/node-list/node-list-row-list.js
richardwestenra: Okay that's fine. The main reason I suggested this is as a performance improvement but it's not a big deal.

src/components/node-list/index.js
```diff
@@ -65,20 +79,21 @@ const NodeListProvider = ({
   };
 
   const onItemChange = (item, checked) => {
-    if (isTagType(item.type)) {
-      onTagItemChange(item, checked);
+    if (isTagType(item.type) || isModularPipelineType(item.type)) {
+      onCategoryItemChange(item, item.checked);
```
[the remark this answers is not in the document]
studioswong: yea indeed checked is passed into this function - I've just checked and using either value would work. I'll use `checked` in this case then!

src/actions/actions.test.js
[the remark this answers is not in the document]
studioswong: speaking of, @richardwestenra , the tests for tags in this same test file are also set up in the same fashion (i.e having the above same 2 tests ) - there's no need for them and hence I should delete them as well?

src/actions/actions.test.js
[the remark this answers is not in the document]
studioswong: see my comment above - following this logic I should delete the same repeating tests for tags as well?

src/components/node-list/index.js
studioswong: great spot - thank you!

src/components/node-list/node-list-items.js
```diff
@@ -31,10 +31,12 @@ export const highlightMatch = (nodes, searchValue) => {
...
-  const addLabelsToNodes = (newNodes, type) => ({
-    ...newNodes,
-    [type]: nodes[type].map(addHighlightedLabel),
-  });
+  const addLabelsToNodes = (newNodes, type) => {
+    return {
+      ...newNodes,
+      [type]: nodes[type].map(addHighlightedLabel),
+    };
+  };
```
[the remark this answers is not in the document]
studioswong: I've reverted the change

src/components/node-list/node-list-items.js
```diff
@@ -176,14 +223,16 @@ export const getFilteredNodeItems = createSelector(
...
+export const getSections = (flag) => {
+  const sidebarObject = sidebar(flag);
+  return Object.keys(sidebarObject).map((name) => ({
     name,
-    types: Object.values(sidebar[name]),
-  }))
-);
+    types: Object.values(sidebarObject[name]),
+  }));
+};
```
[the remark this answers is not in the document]
studioswong: I had set this up as a simple function given that it does not technically read anything from the state it is merely a function to return an object, but after our discussion yesterday I appreciate the potential benefits of using a selector ( which, however, it does raises the question on which should technically be be arranged and what shouldn't in our codebase, given that we do set up functions for calculations all over the codebase, with many that still do rely on the state but still structured as functions). 

regardless i can set it up as a selector.

Also speaking of, I've always wanted to raise this question - I wonder the reason behind setting up the selectors for the sidebar ( `node-list-items`) under the components folder rather than under the `selectors` folder>

src/components/node-list/index.js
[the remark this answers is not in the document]
bru5: Yeah I think that is for the best to show the intent at least, if I remember it correctly the callback provides the new `checked` value from the event and `item.checked` in theory would (or maybe could sometimes) be the previous value.

src/components/node-list/node-list-items.js
bru5: For sure it's good that it's documented, but yeah I also think in this case a more verbose name is a good idea, since it's a specific flag the function is looking for.

src/components/node-list/node-list-items.js
bru5: Yeah in this case the name can be more specific since we know it's a config object, right?

src/selectors/modular-pipelines.js
bru5: Yeah `map` and friends return an array themselves already. You're right it's a sound idea in this case to convert this to a functional style though instead, but what you need here looks like `filter` rather than `map`  e.g. something like:

```js
  const nodes = Object.keys(allNodes)
        .filter(key => allNodes[key] && allNodes[key].includes(modularPipeline));
```

src/components/node-list/node-list-row-list.js
studioswong: I've updated the label formatting and it now looks like this: 
image

src/config.js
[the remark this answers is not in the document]
studioswong: Thanks for the suggestion Richard! I definitely had intended and would've set it up in your suggested way but unfortunately it would not work with the existing sidebar setup ( having a field as `undefined` would cause errors in the sidebar). 

If course we could refactor the sidebar to change how we accept the sidebar config object, but given that hiding this behind a flag is a temporary measure ( and will subject to iteration in the next round) I had decided to keep it as is.

src/actions/actions.test.js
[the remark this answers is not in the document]
bru5: @richardwestenra These are actually two different tests that are very similar, as the function accepts both a singular id or an array of ids both are tested - so should be kept I think.

src/config.js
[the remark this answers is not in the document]
bru5: Yeah a static config is definitely preferred generally, so would be good to add support for this case at the source if practical. 

Flags do make this tricky though so if that's not straightforward then maybe this pattern gets closer to the above suggestion e.g.

```js
export const sidebar = (flags) => ({
  Categories: {
    Tags: 'tag',
    ...(flags.modularPipeline && { 
      ModularPipelines: 'modularPipeline' 
    }),
  },
  Elements: {
    Nodes: 'task',
    Datasets: 'data',
    Parameters: 'parameters',
  },
})
```

src/actions/actions.test.js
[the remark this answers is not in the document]
bru5: Added a comment to the previous one on this as well.

src/actions/modular-pipelines.js
```diff
@@ -0,0 +1,33 @@
+export const TOGGLE_MODULAR_PIPELINE_ACTIVE = 'TOGGLE_MODULAR_PIPELINE_ACTIVE';
+
+/**
+ * Toggle a modular pipeline item's highlighting on/off (or array of modular pipelines)
+ * @param {string|Array} modularPipelineIDs Modular pipeline id(s)
+ * @param {Boolean} active True if modualr pipeline(s) active
+ */
+export function toggleModularPipelineActive(modularPipelineIDs, active) {
```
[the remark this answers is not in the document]
bru5: It looks like this was following consistency from `toggleTagActive` - that one was originally singular but was later upgraded to accept batch changes but kept the name to reduce the CR.

That said plural is better ideally, but if we do so then we should change both I think.

src/components/node-list/node-list-items.js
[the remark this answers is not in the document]
bru5: > Also speaking of, I've always wanted to raise this question - I wonder the reason behind setting up the selectors for the sidebar ( node-list-items) under the components folder rather than under the selectors folder>

I think the main reason is that they're highly specific to that component and not intended to be reusable across the whole app, where as those that live in `src/selectors` are designed to be generic and reusable. 

There's definitely other ways to get the performance benefits of selectors in pure react these days, but I don't think these were around at the time this was implemented.

src/components/node-list/node-list-row-list.js
studioswong: @bru5 please see additional comments I have added in my latest commits

src/actions/modular-pipelines.js
```diff
@@ -0,0 +1,33 @@
...
+  return {
+    type: TOGGLE_MODULAR_PIPELINE_ACTIVE,
+    modularPipelineIDs: Array.isArray(modularPipelineIDs)
+      ? modularPipelineIDs
+      : [modularPipelineIDs],
```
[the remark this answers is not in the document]
richardwestenra: Note: We're gonna discuss this separately and maybe bump it to a future PR.

src/components/node-list/node-list-row-list.js
```diff
@@ -4,6 +4,20 @@ import { connect } from 'react-redux';
...
+  const whiteSpace = '&nbsp;&nbsp;&nbsp;&nbsp;';
+  if (item.type === 'modularPipeline') {
+    // parse depth of modular pipeline from namespace(i.e id)
+    const levels = item.id.match(/\./g) ? item.id.match(/\./g).length : 0;
+
+    return whiteSpace.repeat(levels) + layer + item.highlightedLabel;
+  }
+  return item.highlightedLabel;
+};
+
```
richardwestenra, line 20: The dots appear for all levels now, not just nested ones:
image

Suggested fix:
```suggestion
// Modify display of labels for modular pipelines to show nested relationship.
// Note: This label indentation could be subject to further design changes.
const getItemLabel = (item) => {
  if (item.type === 'modularPipeline') {
    // Parse depth of modular pipeline from namespace(i.e id)
    const levels = item.id.match(/\./g) ? item.id.match(/\./g).length : 0;
    const whiteSpace = '&nbsp;&nbsp;&nbsp;&nbsp;';
    const layer = levels ? '・' : '';

    return whiteSpace.repeat(levels) + layer + item.highlightedLabel;
  }
  return item.highlightedLabel;
};

```

richardwestenra (changes requested):

richardwestenra (approved):
LGTM! Great work :)

src/selectors/modular-pipelines.js
```diff
@@ -0,0 +1,59 @@
...
+        )
+          nodes.push(key);
+      });
+      return {
+        modularPipeline,
+        nodes,
+      };
+    });
+  }
+);
```
richardwestenra, line 59: @studioswong Hey I merged this branch into the one I'm working on, and realised that this selector isn't actually being used anywhere? Anyway I've deleted it on my branch.

PR #416, opened 2021-04-13
[KED-2480] Turn off parameters by default

rashidakanchwala:
## Description

Parameters have now been turned off by default. Also, parameters are no longer saved in local storage hence if user refreshes the page after turning parameters on, it will always go back to default behavior (turn-off) 

## Development notes

We turn parameters off by first checking if the parameters flag is set to true. If it is, then on initial load, we set the state to nodeType.disabled.parameters=True. So on every page refresh, the parameters will be turned off by default. 

Other changes 
- for multiple tests we were getting getNodeID to get nodes on the flowchart instead of getVisibleNodeID. This issue has been fixed in this PR
- @bru5 has also refactored the test setup in this PR

## QA notes

npm t run's successfully

## Checklist

## Legal notice


studioswong (changes requested):
Thanks for the work Rashida!

On initially checking the requirements  for this ticket, we would need to set a flag to enable the users to turn on the parameters by default, which is currently missing in this PR?

rashidakanchwala:
> Thanks for the work Rashida!
> 
> On initially checking the requirements for this ticket, we would need to set a flag to enable the users to turn on the parameters by default, which is currently missing in this PR?

Thanks I missed that bit. I will work on it today!

src/components/metadata/metadata.test.js
```diff
@@ -1,25 +1,24 @@
...
+import { setup, prepareState } from '../../utils/state.mock';
+import animals from '../../utils/data/animals.mock.json';
 
 const salmonTaskNodeId = '443cf06a';
 const catDatasetNodeId = '9d989e8d';
 const rabbitParamsNodeId = 'c38d4c6a';
 
 describe('MetaData', () => {
-  // Add edge links, can be removed when new graph is default
-  addEdgeLinks(mockState.animals.graph.nodes, mockState.animals.graph.edges);
```
studioswong, line 13: I think the new graph is currently loaded by default?

In case for some reason it's not loaded in the tests, instead of adding this pending removal for the oldgraph flag, this should ideally not needed by specifying `{flag: {oldgraph: false}}` within the `prepareState` method on line 35 here?

rashidakanchwala: @bru5 has done this bit of changes and prepareState. I am not sure of the above -- can we get on a call to discuss?

bru5: @studioswong That used to be the case but as the line in question is being removed here I don't think it's an issue?

studioswong (commented):
Thanks for adding in the flag @rashidakanchwala!

With the new changes with the display of parameters, we would need to add the following 2 types of tests:
- ensuring the parameters are turned off by default of the app
- test the display of parameters on the flowchart when the parameter flag is turned on

The relevant tests could be added on existing test suites we have set up. For example, this test could be added by adding a test for the `getVisibleEdgesByNode` function under `linked-nodes`. 

Ideally we can also do some pairing on setting this test tomorrow too 👍

rashidakanchwala:
> Thanks for adding in the flag @rashidakanchwala!
> 
> With the new changes with the display of parameters, we would need to add the following 2 types of tests:
> 
> * ensuring the parameters are turned off by default of the app
> * test the display of parameters on the flowchart when the parameter flag is turned on
> 
> The relevant tests could be added on existing test suites we have set up. For example, this test could be added by adding a test for the `getVisibleEdgesByNode` function under `linked-nodes`.
> 
> Ideally we can also do some pairing on setting this test tomorrow too 👍

Thanks, 

I missed writing new tests bit. QQ - parameters are turned off default in the app and turned on default for the test. So how can I test this bit ? Also looping in @bru5

studioswong:
As discussed with Rashida in our pairing yesterday, I took a look at the other approach to fixing the tests ( i.e taking out the `isTest` param added in normalize-data.js), and can conclude that they can be fixed with some refactor. (FYI, @bru5 , as per our discussion yesterday, the current method of temporarily taking out the assignment of parameters under the disabled value in a test environment is imo not the right way to set up the tests as it set up our test environment completely different from how the app is run in reality, hence I do think it is problematic in taking this shortcut. I know there were some problems with the tests under edges.test.js, which should be fixed by adding a new function to filter out the parameter nodes within the tests. )

However, I noticed that currently the newly added `parameter` flag does not enable the parameter nodes when assigned to true - @rashidakanchwala this will need to be fixed before I can proceed with the remaining failing tests ( which I will enable the parameter flag to true within the test itself as part of the fix, which currently does not work now.) 

I had pushed the current tests that I was able to fix ( it is still a WIP) before the above problem with the parameter flag under the branch `fix/fix-failing-tests` - please let me know once the problem with the parameter flag are fixed and we can find some time to pair to fix the test ( or I can fix it too, up to you :) )

Thank you!

src/store/normalize-data.js
```diff
@@ -38,7 +40,9 @@ export const createInitialPipelineState = () => ({
       task: 'Nodes',
       parameters: 'Parameters',
     },
-    disabled: {},
+    disabled: {
+      parameters: isTest ? false : true,
```
studioswong, line 44: I am against this set up mainly because this essentially causes great seperation of the app that was loaded in the test environment and the actual set up ( with parameters disabled by default) and in turn the tests that was subsequently tested in all tests suites cannot be used to identify other breaking points in the app - @bru5 Rashida mentioned this was proposed as a quick fix to fix the existing tests, but I do advise strongly against this. 

I had a look at the current failling tests and they should be able to be fixed with a bit of refactor, which should be the right way to approach this given that the refactor is needed anyways as parameters off will be the default going onwards.

src/config.js
```diff
@@ -36,6 +36,11 @@ export const flags = {
     default: false,
     icon: '😴',
   },
+  parameters: {
+    description: 'Turns on parameters',
```
studioswong, line 40: perhaps change this to 'Turns on parameters by default' for more clarity?

studioswong:
@rashidakanchwala I've added in the fix for the problem with the parameters flag I mentioned above ( setting parameters flag to true does not enable the parameters to be turned on by default when the data is loaded from an endpoint.) Reading the value of the `parameters` flag to assign the value of the disabled node type during data normalization will ensure the flag value gets reflected within the app regardless of how the data is loaded.

src/store/normalize-data.js
```diff
@@ -201,6 +204,12 @@ const normalizeData = (data) => {
     return state;
   }
 
+  // check the current parameters flag to determine if parameters are disabled
+  const flagState = getFlagsFromUrl();
+  if (flagState.parameters) {
+    state.nodeType.disabled.parameters = false;
+  }
```
richardwestenra, line 211: @studioswong This only gets flags from URL - it doesn't know about localStorage values or the initial default values. So if a flag is set to true by default, or true in localStorage, it will be ignored here.

richardwestenra (changes requested):

src/store/normalize-data.js
studioswong: Thanks @richardwestenra ! That solution was done given an misunderstanding that the parameters flag was not stored in localstorage, hence I opted for a more straight forward approach! I see that you had implemented a new solution now, so I believe it was been resolved :)

src/utils/state.mock.js
```diff
@@ -9,18 +9,30 @@ import reducer from '../reducers';
...
+  beforeLayoutActions = [],
+  afterLayoutActions = [],
+  ...props
+}) => {
   const initialState = getInitialState(props);
   const actions = [
     // Set fontLoaded = true:
     () => updateFontLoaded(true),
+    // Enable parameters as they are disabled by default:
+    () => toggleTypeDisabled('parameters', false),
```
studioswong, line 33: Please add within the comments that this is a temporary setup and will need to be taken out once the failing tests due to the new parameter setup are revisited :)

src/store/initial-state.test.js
```diff
@@ -157,4 +156,15 @@ describe('getInitialState', () => {
...
+
+  it('hides parameters when parameter flag is true', () => {
+    const state = getInitialState({
+      ...props,
+      flags: {
+        parameters: true,
+      },
+    });
+    const parametersDisabled = state.nodeType.disabled.parameters;
+    expect(parametersDisabled).toBe(true);
```
studioswong, line 168: Thanks for adding this test in! 

One suggestion: this test is based on the assumption that assigning `nodeType.disabled.parameters` will automatically ensure parameters are not displayed within the app. Once we have removed the 'hack' in mockstate and go further into fixing the tests, we definitely need to update the tests under `actions/nodes.test.js` to ensure parameters nodes are not returned within the actions and flowchart.

studioswong (approved):
Approving this to bring this forward to the next PR :D  Thanks for fixing the above comments Rashida!

2 things to follow up:
1. As I mentioned above, the current set up of enabling parameters in the mockstate to fix the breaking tests should only be a temporary approach. As Liam mentioned, we should file a ticket for fixing the failing tests caused by the change in the default behaviour of the parameters - @rashidakanchwala can you please follow up in filing a ticket for it and fixing the tests?  This is definitely a great chance to learn about the current test setup within our app.

2. Given that the parameters flag is now enabled by default, we should update the readme on the new default behaviour of the parameter nodes ( moreover to provide clarity on the confusion that the user will still need to toggle the parameters on the sidebar for them to be visible after changing the flag due to the localstorage setting saved from the previous load) - something like the following? 
`On first load, the visibility for the parameter nodes will now be disabled on the flowchart. If you hope to revert back to the previous behaviour, please disable the 'newparams' flag. Note that on disabling the flag, you will still need to toggle the parameters on in order for them to load by default.`. 

Thank you!

bru5 (approved):

rashidakanchwala:
Closing this PR in favour of #424, as these changes are included in that PR

richardwestenra:
Note: This PR has been closed in favour of #424, as these changes are included in that PR.

PR #429, opened 2021-04-21
Refactor node list selectors and modular pipeline config

richardwestenra:
## Description

1. Remove unused `modularPipelineFlag` prop, which doesn't appear to be used for anything - idk why it didn't throw a linter error.
2. Simplify modular pipeline flag config, and Move utility functions and flag manipulation to a selector in the component so that the config file contains pure config objects/properties.
3. Simplify node-list filter selectors: The separate 'enabled' arguments don't seem to be necessary here, as  these values are already passed down in the main list data argument.
4. Fix a bug in the filter behaviour, where if you toggle on all nodes via the toggle-all button, then manually disabling each one individually, it would not previously revert to the default state, and now it does. This behaviour change was kind of accidental, but apparently this is how it was supposed to work all along anyway, so I don't think it's a problem.

## QA notes

Please compare the behaviour in the [demo](https://quantumblacklabs.github.io/kedro-viz/?data=animals&modularpipeline=1) to ensure that nothing is broken.

## Checklist

## Legal notice

This prop doesn't appear to be used for anything - idk why it doesn't throw a linter error

### DIFF
--- a/src/components/node-list/node-list-items.js
+++ b/src/components/node-list/node-list-items.js
@@ -101,12 +101,11 @@ export const getFilteredTags = createSelector(
 /**
  * Return filtered/highlighted tag list items
  * @param {object} filteredTags List of filtered tags
- * @param {object} tagsEnabled Map of enabled tags
  * @return {array} Node list items
  */
 export const getFilteredTagItems = createSelector(
-  [getFilteredTags, (state) => state.tagsEnabled],
-  (filteredTags, tagsEnabled) => ({
+  getFilteredTags,
+  (filteredTags) => ({
     tag: filteredTags.tag.map((tag) => ({
       ...tag,
       type: 'tag',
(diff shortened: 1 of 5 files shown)


src/components/node-list/node-list-items.js
```diff
@@ -219,19 +216,32 @@ export const getFilteredNodeItems = createSelector(
   }
 );
 
+/**
+ * Get sidebar node-list sections
```
bru5, line 220: Description needs updating I think?

src/components/node-list/node-list-items.js
```diff
@@ -117,8 +116,8 @@ export const getFilteredTagItems = createSelector(
       faded: false,
       visible: true,
       disabled: false,
-      unset: typeof tagsEnabled[tag.id] === 'undefined',
-      checked: tagsEnabled[tag.id] === true,
+      unset: !tag.enabled,
```
bru5, line 119: Originally the `unset` state was specifically different from `true` and `false` for the UI pattern to work, but maybe things have changed since then.

Looks to work from my testing, so if you're confident this is no longer needed then I'd say consider removing the `unset` flag here and just use `!checked` where it was used before - though if that's a lot of work no problem.

(if it turns out that doesn't work, then maybe it still needs to be distinguished?)

richardwestenra: Yeah, I have a suspicion that that should be removed. If you don't mind, I might make a new ticket and leave that for a separate PR? It seems tricky.

bru5 (approved):

PR #430, opened 2021-04-22
Expand/collapse modular pipelines on the graph - Part 1

richardwestenra:
## Description

Part 1 of a WIP feature. I'm making this PR to a feature branch in order to allow review while it's still _only_ ~500 additional lines 😖 .

The aim of this feature is to allow the user to replace a group of graph nodes with a new modular pipeline node, when this pipeline is toggled to contracted.

Would you mind please having a look over this PR and giving me your thoughts, before I move on to fixing tests etc? Ideally I'd do that in subsequent PRs.

image image

## Development notes

- This was based on a WIP feature branch that has since been squashed, hence the messy Git history.
- The selectors aren't necessarily always in the file I would have preferred to have them in, and that's mostly due to requirements about the import order, in order to avoid circular imports.
- I've used some abbreviations in selector variable names (e.g. `modPip`) because they were getting out of control in length. Happy to discuss best practice here over Zoom, but of the alternatives I thought they were the best of several bad options (e.g. I preferred `modPipID` when used repeatedly in long lines, over `id` or `modularPipelineID`).

## TODO (things that I'm planning to add in subsequent PRs):

- Unit tests
- Proper icons that aren't emoji, and other sidebar design tweaks. This is pending design consultation.
- Handle modular pipelines that are both parents AND children to other nodes. This causes errors at present - e.g. see `?data=random` for examples.
- Selector performance improvements?
- Support for other datasets, e.g. demo data
- Figuring out what exactly is supposed to be going on with layers???
- Fix disabled state fadeout on the sidebar when an MP's parent is contracted

## QA notes

You'll need to visit http://localhost:4141/?data=animals&modularpipeline=true

This is gonna be a tricky one because it's still a WIP. Try toggling the MPs on and off. Talk to me if you spot something funny.

## Checklist

## Legal notice


src/selectors/disabled.js
```diff
@@ -84,28 +144,100 @@ export const getNodeDisabled = createSelector(
...
+    nodeDisabledModularPipelineFilter,
+    nodeDisabledPipeline,
+    nodeType,
+    typeDisabled,
+    modularPipelineChildren
+  ) =>
+    arrayToObject(modularPipelineIDs, (modPipID) => {
+      const children = Object.keys(modularPipelineChildren[modPipID]);
+      const allChildrenAreDisabled = children.every(
+        (nodeID) =>
```
bru5, line 190: Is this function the same as used in `getNodeDisabledExclModPip`? Might be a good idea to move this out to a function if so.

src/selectors/modular-pipelines.js
```diff
@@ -1,34 +1,220 @@
...
+
+    const addNewEdge = (source, target) => {
+      const id = [source, target].join('|');
+      edges.ids[id] = true;
+      edges.sources[id] = source;
+      edges.targets[id] = target;
+    };
+    modularPipelineIDs.forEach((modPipID) => {
+      const modPipNodes = modularPipelineChildren[modPipID];
+      edgeIDs.forEach((edgeID) => {
```
bru5, line 123: Wondering if this could get slow with lots of edges with the nested loops, maybe it could be made linear looking at it another way if so?

src/selectors/nodes.js
```diff
@@ -61,13 +67,18 @@ export const getNodeActive = createSelector(
...
-      return Boolean(activeViaTag) || Boolean(activeViaModularPipeline);
+      const tags = nodeTags[nodeID] || [];
+      const activeViaTag = tags.some((tag) => tagActive[tag]);
+      const modularPipelines = nodeModularPipelines[nodeID] || [];
+      const activeViaModularPipeline = modularPipelines.some(
+        (modularPipeline) => modularPipelineActive[modularPipeline]
+      );
+      const isActiveModularPipeline = modularPipelineActive[nodeID];
+      return (
+        Boolean(activeViaTag) ||
```
bru5, line 78: Probably don't need the `Boolean` for each of these as it's pretty clear and not used to cast (if it was used to cast though I'd say move the `Boolean` to the original assignment)

src/selectors/disabled.js
richardwestenra: ah good catch! I think I added `getNodeDisabledExclModPip` after adding this selector. I can just import the result of that selector instead, which will simplify this one tremendously.

src/selectors/modular-pipelines.js
```diff
@@ -1,34 +1,220 @@
...
+);
+
+/**
+ * List all the parent modular pipelines for a given modular pipeline
+ */
+export const getModularPipelineParents = createSelector(
+  [getModularPipelineIDs],
+  (modularPipelineIDs) =>
+    arrayToObject(modularPipelineIDs, (modPipID) =>
+      modPipID.split('.').reduce((parents, part, i, parts) => {
```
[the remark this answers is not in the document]
richardwestenra: Yes, this is way better! Thanks, I wasn't happy with my first attempt. This one needs a `.slice(1)` on the end but other than that it's perfect 👍

src/selectors/modular-pipelines.js
```diff
@@ -1,34 +1,220 @@
...
+      edges.ids[id] = true;
+      edges.sources[id] = source;
+      edges.targets[id] = target;
+    };
+    modularPipelineIDs.forEach((modPipID) => {
+      const modPipNodes = modularPipelineChildren[modPipID];
+      edgeIDs.forEach((edgeID) => {
+        const source = edgeSources[edgeID];
+        const target = edgeTargets[edgeID];
+        if (modPipNodes[target] && !modPipNodes[source]) {
```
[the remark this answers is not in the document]
richardwestenra: The comment below each on was the comment explaining this (I never know where to put the comment with `if` statements), but it evidently wasn't clear so I've made these lines into named variables so that comments are unnecessary.

src/selectors/modular-pipelines.js
richardwestenra: Yeah this is a concern of mine also - but I'm honestly not sure how to do it differently? I'll look into it. The good news is that this will only run on first loading the data, and won't re-run when toggling nodes'/MPs' disabled state.

bru5 (approved):

PR #439, opened 2021-05-08
Refactor initial-state setup to reduce complexity

richardwestenra:
## Description

1. Remove the pipeline state initialisation (`createInitialPipelineState`) from `normalize-data.js` and combine it all with the main createInitialState. The data normalizer should be kept relatively pure.
2. Get rid of the separation between `prepareNonPipelineState` and `preparePipelineState`. Instead, when we want to reset the pipeline data (e.g. when loading a new pipeline), use a dedicated reset action that manually deletes the required properties from the overall state.

## Development notes

1. Based on a suggestion from @studioswong, the app no longer deletes existing pipeline data when loading new data from another pipeline endpoint. Instead, it relies on the pipeline selectors, and just adds the new data to the store so that it's ready immediately. We could capitalize on this further by checking whether a pipeline is loaded before making a new redundant request.
2. We'll need to update the Architecture docs & particularly the diagram, as it's now changed significantly. @studioswong I'll need your help with this!

## QA notes

These changes significantly affect the data loading and state initialisation, so we'll need to thoroughly QA how it interacts with localStorage application, switching pipelines, async/sync data loading, and resetting the state when running inside another application.

## Checklist

## Legal notice

Avoid array.includes for performance reasons

### DIFF
--- a/src/actions/pipelines.js
+++ b/src/actions/pipelines.js
@@ -1,6 +1,6 @@
 import { getUrl } from '../utils';
 import loadJsonData from '../store/load-data';
-import { preparePipelineState } from '../store/initial-state';
+import normalizeData from '../store/normalize-data';
 import { resetData, toggleGraph } from './index';
 
 /**
@@ -95,12 +95,14 @@ export function loadInitialPipelineData() {
     // whether the active pipeline (from localStorage) exists in the data.
     const url = getUrl('main');
     let newState = await loadJsonData(url).then((data) =>
-      preparePipelineState(data, true)
+      normalizeData(state, data)
     );
     // If the active pipeline isn't 'main' then request data from new URL
     if (requiresSecondRequest(newState.pipeline)) {
       const url = getPipelineUrl(newState.pipeline);
-      newState = await loadJsonData(url).then(preparePipelineState);
+      newState = await loadJsonData(url).then((data) =>
+        normalizeData(state, data)
+      );
     }
     dispatch(resetData(newState));
     dispatch(toggleLoading(false));
@@ -114,19 +116,21 @@ export function loadInitialPipelineData() {
  */
 export function loadPipelineData(pipelineID) {
   return async function (dispatch, getState) {
-    const { dataSource, pipeline } = getState();
-    if (pipelineID && pipelineID === pipeline.active) {
+    const state = getState();
+    if (pipelineID && pipelineID === state.pipeline.active) {
       return;
     }
-    if (dataSource === 'json') {
+    if (state.dataSource === 'json') {
       dispatch(toggleLoading(true));
       // Remove the previous graph to show that a new pipeline is being loaded
       dispatch(toggleGraph(false));
       const url = getPipelineUrl({
-        main: pipeline.main,
+        main: state.pipeline.main,
         active: pipelineID,
       });
-      const newState = await loadJsonData(url).then(preparePipelineState);
+      const newState = await loadJsonData(url).then((data) =>
+        normalizeData(state, data)
+      );
       // Set active pipeline here rather than dispatching two separate actions,
       // to improve performance by only requiring one state recalculation
       newState.pipeline.active = pipelineID;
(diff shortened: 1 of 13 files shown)


studioswong (changes requested):
Thanks for the refactor @richardwestenra ! great stuff!!

On initially testing this branch with the kedro-factory project that Joel had set up, I notice that the app does not update with the new pipeline data on selecting different registered pipelines from the dropdown. I am not entirely sure which change causes this problem, but I suspect it could be caused by the new `resetPipelineState` function? ( I could be completely wrong, this needs to be further investigated). 

Thank you!

src/actions/pipelines.js
```diff
@@ -114,19 +116,21 @@ export function loadInitialPipelineData() {
...
       // Remove the previous graph to show that a new pipeline is being loaded
       dispatch(toggleGraph(false));
       const url = getPipelineUrl({
-        main: pipeline.main,
+        main: state.pipeline.main,
         active: pipelineID,
       });
-      const newState = await loadJsonData(url).then(preparePipelineState);
+      const newState = await loadJsonData(url).then((data) =>
+        normalizeData(state, data)
```
bru5, line 132: Could there be any problems here when loading is slow and the `state` becomes stale? Should it call `getState()` again?

src/components/app/index.js
```diff
@@ -72,7 +72,10 @@ class App extends React.Component {
    * Dispatch an action to update the store with new pipeline data
    */
   updatePipelineData() {
-    const newState = preparePipelineState(this.props.data, true);
+    const newState = normalizeData(
```
bru5, line 75: Seems a lot is happening in these few lines, could you explain in a comment a bit more about what is involved in this update?

src/store/helpers.js
```diff
@@ -44,6 +45,22 @@ export const saveState = (state) => {
...
 
+/**
+ * Load values from localStorage and combine with existing state,
+ * but filter out any unused properties
+ * @param {object} state Initial/extant state
+ * @return {object} Combined state from localStorage
+ */
+export const mergeLocalStorage = (state) => {
+  const localStorageState = loadState();
+  Object.keys(localStorageState).forEach((key) => {
```
bru5, line 56: Cleaner to use a for-of loop here I think

src/store/initial-state.js
```diff
@@ -1,25 +1,73 @@
...
 import { getFlagsFromUrl, Flags } from '../utils/flags';
 import { sidebarWidth } from '../config';
 
-/**
- * Create new default state instance for properties that aren't overridden
- * when the pipeline is reset with new data via the App component's data prop
- * @return {object} state
- */
-export const createInitialState = () => ({
+export const initialState = {
```
bru5, line 6: Great to see all of the state in one place here. Optional but I think I'd prefer to keep this wrapped in a factory function though for protection rather than risk a single mutable reference.

richardwestenra:
@studioswong @bru5 @rashidakanchwala @limdauto Hey, I no longer have write access to this repo so I can't make the requested changes to this PR! If you want to get it merged (and I suggest doing so sooner rather than later in order to avoid more conflicts piling up), then one of you might need to take it over to get it over the line. Thanks! :)

studioswong:
Hello @richardwestenra ! great to see you on here buddy!
Yea indeed we do need to revisit this PR - things are going full steam ahead with the tickets in the current sprint, however once modular pipeline is cross the line we should definitely revisit and pick back up on this work. ( I see that lodash clone deep is introduced here which is great, as I had introduced this in my latest PR for some of our modular pipeline work as well :D )

Luckily the conflicts are mainly on the additional of fields introduced by plotly, and given that we are not introducing any new data fields in our upcoming work, it should be fine for the time being :)

limdauto:
Thanks again for this @richardwestenra Unfortunately we didn't have a chance to get this in before the codebase diverges from this implementation. We would scope out a project to re-build the state management and data loading mechanism and would surely take ideas from this PR onboard when we get there. Gonna close this for now.

PR #454, opened 2021-05-24
[KED-2642] Filter panel updates and refactoring

bru5:
## Description

Screenshot 2021-07-05 at 10 52 37

This PR includes the following feature and behaviour changes:

- implements filter panel layout changes per design
- adds element types filters for Nodes, Datasets, Parameters
  - resets all these filters on load to avoid potential confusing saved filter states
  - based on existing implementation used by the old sidebar
- adds node counts to all filters representing the fixed total number of nodes in each set excluding any filters
- changes group headings to be 'sticky' during scroll

To support and implement the above in a reasonable and understandable way, as well as handle the merge with the recent tree list changes, I've applied some refactoring and implementation changes to avoid introducing further complication:

- refactors the sidebar config to be more convenient and simpler to use
- refactors parts of `node-list-items.js` to simplify and make them easier to follow
- improves some naming and comments for clarity
- removes now redundant sidebar 'sections'
- removes now redundant original modular pipeline filters from the filter panel
- moves the hidden parameters highlight state to be on the 'Parameters' element filter hover
- moves most handling of 'unset' states into reducers and removes it elsewhere
- changes some actions to support batch changes with individual values to ensure state consistency
- improves label sizing in `node-list-row.js` to handle usage in both filters panel and the tree panel
- improves tab active state for group filter toggle (blue button)

## QA notes

For some changes the diff is not accurate due to some things moving or being removed, so it's worth looking at the changes as a whole file first for a better sense.

See the description of changes above as a guide for QA testing.

## Checklist

## Legal notice


rashidakanchwala (approved):
Great stuff! Works well for me.

studioswong (changes requested):
Thanks for setting up the filters panel @bru5 !

I noticed some major discrepancies with the design for the filter panel as outlined on zeplin [here](https://app.zeplin.io/project/5d5ea7a05efca76a74f7d0ea/screen/603cf6f8e7e0167c72815d9f) and [here](https://app.zeplin.io/project/5d5ea7a05efca76a74f7d0ea/screen/6098e18f3cf12211b4d9ea4b) - see below latest design. @GabrielComymQB also appreciate your design eyes and further confirmation as well. 
image

1. The 'elements types' filter section is missing from the filters panel.  ( see below)

image


2. The modular pipeline filter exist under the elements filters section, and should be all condensed as one filter to toggle all modular pipelines on or off, however it is still arranged as a dropdown that lists all modular pipeline items for the main pipeline. ( see below screenshot)

image

2. Other elements filters (nodes, datasets, parameters) are missing on the filters panel - they should now exists as a filter for each element type to toggle all elements on and off. ( see below screenshot)

image

Currently this setup of the filters panel would not work with the tree-list work as I worked on it based on the design, with the assumption that the modular pipeline filter would not be an expanded list but instead condensed into 1 filter, moreover all other elements ( nodes, parameters, datasets) would not be arranged as the list that we have in the old sidebar, hence it is crucial for this to be set up according to the design in order for the tree-list to work.

However, please let me know in case there is another PR coming up for the filters panel to complete the filters panel work - if this is the complete filters panel work, the above design discrepancy would definitely need to be addressed in this PR. 

On another note, I do not think that this should be merged directly into main given that implementing the filters panel on its own will be very confusing  to the user and disruptive for their journey, particularly given that we will be releasing next week which we should definitely avoid releasing this as part of the next release.  (cc @yetudada )

This should be merged into a feature branch that combines both the tree-list and the filter panel together for the new sidebar. I had set up a feature branch `feature/sidebar-2.0` to consolidate both of our filter-panel and tree list work. While we might want to remain agile, given that the new sidebar needs both the tree-list and the filters panel in order to complete the journey, it is still best to consolidate both into 1 feature for release. I simply do not see any benefit in merging this directly into main.

src/components/node-list/styles/_group.scss
```diff
@@ -46,22 +42,44 @@ $placeholder-fade: 120px;
...
 
-.pipeline-nodelist__placeholder-lower:after {
-  top: -$placeholder-fade;
-  background: linear-gradient(
-    0deg,
-    var(--color-bg-3) 0%,
-    var(--nodelist-bg-transparent) 100%
-  );
+.pipeline-nodelist__filter-panel {
+  .pipeline-nodelist__placeholder-upper:after {
```
studioswong, line 66: is this a class that is supposed to be deleted and replaced once the tree items are merged in?

src/components/node-list/node-list-row.js
```diff
@@ -76,29 +77,32 @@ const NodeListRow = memo(
...
+              }
+            )}
+            type={type}
+          />
+        )}
         <TextButton
-          className="pipeline-nodelist__row__text"
+          className={classnames(
+            'pipeline-nodelist__row__text',
+            `pipeline-nodelist__row__text--kind-${kind}`
```
studioswong, line 99: for readability purposes I would format this as `{'pipeline-nodelist__row__text--kind-filter': kind === 'filter'}` , given that there aren't any classes applied other than for the `filter` type rows.

studioswong (commented):
@bru5 generally the code set up works well, I have some more minor comments but nothing major. 

A few things to follow up:
1. I observe while testing this branch: the 'blue dot' for the highlighted tag actually reduces its opacity on hover ( see below), which is contrary to the previous sidebar interaction setup. This might be accidentally introduced from one of the style changes in this ticket? 
new tag interation

2. Since this ticket is mainly just for exposing the numbering next to the tags, it would be good to mention in the description that this is part of the filter panel work in exposing the numbering of tags, with the addition of the  combined filters for modular pipelines, nodes, datasets and parameters to follow in seperate PRs ( and would even be better if we could set up the relevant tickets for those tasks on Jira so this work don't get lost in the sprint) . This would help with the communication within the team, especially the modular pipeline filter would be deleted on later merging the tree-list work as we discussed. 

Thank you!

bru5:
Sorry about any confusion here, I'll try to be clearer in future on the scope of my PRs. Thanks for the reviews, I've now finished implementing those additional features in this PR as suggested. 

I've done some (much due) refactoring and general clearing of now unused implementation details. While this was in aid of the new features to avoid adding extra complication, it should also make future sidebar changes a little easier for all. Check out the updated PR description above for specific details on what's changed.

I realise this may have become a bit of a difficult diff to review, so let me know if anything is unclear or if the PR description doesn't explain. As some things have moved around a bit, the diff is not always accurate, so worth looking at the updated files as a whole first.

src/components/node-list/styles/_group.scss
bru5: Check the latest changes, but this is still around as the filters panel uses this as part of the lazy list loading effect.

src/components/node-list/node-list-row.js
bru5: I think we've discussed but this is just a pattern to allow styling of multiple kinds, I think it's worth keeping as a pattern as the sidebar evolves there may be more.

src/components/node-list/styles/_panels.scss
```diff
@@ -2,6 +2,23 @@
 @import '../../../styles/mixins';
 @import '../../../styles/variables';
 
+.pipeline-nodelist__filter-panel {
+  z-index: 1;
```
limdauto, line 6: Don't we have a z-index scale as scss variables?

src/components/node-list/styles/_row.scss
```diff
@@ -11,17 +16,24 @@
   display: flex;
   align-items: center;
 
-  // force gpu layers to avoid drawing lag on scroll
+  // Fixed height required for lazy list (see note above).
+  height: 37px;
```
limdauto, line 20: We might want to add a comment that it needs to match a hard-coded variable in `node-list-row.js` as well.

limdauto (approved):
LGTM! A few minor comments but really well done. Thank you 💯

src/reducers/node-type.js
```diff
@@ -1,13 +1,55 @@
...
+const allNodeTypesUnset = {
+  parameters: NODE_TYPE_DISABLED_UNSET,
+  task: NODE_TYPE_DISABLED_UNSET,
+  data: NODE_TYPE_DISABLED_UNSET,
+};
+
+const isNodeTypeUnset = (nodeTypeValue) =>
+  nodeTypeValue === NODE_TYPE_DISABLED_UNSET;
+
+const isNodeTypeEnabled = (nodeTypeValue) =>
```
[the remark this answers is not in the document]
limdauto: Actually realised that it was removed in your latest commit.

src/actions/node-type.js
```diff
@@ -1,14 +1,22 @@
 export const TOGGLE_TYPE_DISABLED = 'TOGGLE_TYPE_DISABLED';
 
 /**
- * Toggle a node-type's visibility on/off
- * @param {string} typeID Type id
- * @param {Boolean} disabled True if type is disabled
+ * The default enabled / disabled value for all types is 'unset', meaning not explicitly set by user.
+ * In practice 'unset' acts like enabled but is a distinct state for UI purposes.
```
studioswong, line 5: I am slightly confused about the wording here, in particular the  `UI purpose` for `unset` - is this action type particularly used to cater for unsetting all element type filters? It would be nice to further elaborate on what the `distinct state for UI purposes` actually us.

src/components/node-list/index.js
```diff
@@ -34,33 +36,35 @@ const NodeListProvider = ({
...
     searchValue,
   });
 
-  const groups = getGroups({ types, items });
+  const groups = getGroups({ items });
 
   const onItemClick = (item) => {
-    if (isTagType(item.type) || isModularPipelineType(item.type)) {
-      onCategoryItemChange(item, item.checked);
+    if (isGroupType(item.type)) {
```
studioswong, line 66: What does `isGroupType` checks for? I've cross referenced this with `node-list-items` and there is no `isGroupType` function declared. Is this check still needed?

src/components/node-list/index.js
```diff
@@ -2,29 +2,31 @@ import React, { useEffect, useState } from 'react';
...
-  toggleModularPipelineActive,
-  toggleModularPipelineFilter,
-} from '../../actions/modular-pipelines';
-import { toggleTypeDisabled } from '../../actions/node-type';
+  getFilteredItems,
+  getGroups,
+  isTagType,
+  isModularPipelineType,
+  isElementType,
+  isGroupType,
```
studioswong, line 11: I've crossed check these function imports with `node-list-items.js` and they are not declared in the file - are these functions and their relevant usage in this file still needed?

src/components/node-list/index.js
```diff
@@ -71,8 +75,8 @@ const NodeListProvider = ({
   };
 
   const onItemChange = (item, checked) => {
-    if (isTagType(item.type) || isModularPipelineType(item.type)) {
-      onCategoryItemChange(item, checked);
+    if (isGroupType(item.type)) {
+      onGroupItemChange(item, checked);
```
studioswong, line 79: same comment as above - is this check still relevant?

src/components/node-list/index.js
```diff
@@ -86,6 +90,9 @@ const NodeListProvider = ({
       onToggleTagActive(item.id, true);
     } else if (isModularPipelineType(item.type)) {
       onToggleModularPipelineActive(item.id, true);
+    } else if (isElementType(item.type) && item.id === 'parameters') {
```
studioswong, line 93: as mentioned above, `isElementType` was not declared in `node-list-items.js` - the `isElementType` check does not seem to be needed?

src/components/node-list/index.js
```diff
@@ -96,73 +103,42 @@ const NodeListProvider = ({
       onToggleTagActive(item.id, false);
     } else if (isModularPipelineType(item.type)) {
       onToggleModularPipelineActive(item.id, false);
+    } else if (isElementType(item.type) && item.id === 'parameters') {
```
studioswong, line 106: same as above comment

src/components/node-list/index.js
```diff
@@ -96,73 +103,42 @@ const NodeListProvider = ({
...
+
+    if (isTagType(groupType)) {
+      onToggleTagFilter(
+        groupItems.map((item) => item.id),
+        groupItemsDisabled
+      );
+    } else if (isElementType(groupType)) {
+      onToggleTypeDisabled(
+        groupItems.reduce(
+          (o, item) => ({ ...o, [item.id]: !groupItemsDisabled }),
```
studioswong, line 129: Please use a more intuitive variable name ( such as `obj`) rather than `o` here to match the naming pattern of the codebase

src/components/node-list/index.js
```diff
@@ -219,9 +192,6 @@ export const mapDispatchToProps = (dispatch) => ({
   onToggleModularPipelineActive: (modularPipelineIDs, active) => {
     dispatch(toggleModularPipelineActive(modularPipelineIDs, active));
   },
-  onToggleModularPipelineFilter: (modularPipelineIDs, enabled) => {
-    dispatch(toggleModularPipelineFilter(modularPipelineIDs, enabled));
-  },
```
studioswong, line 224: This action would be needed for focus mode implementation, please retain this - you can add a comment to reference this if needed

rashidakanchwala (approved):
LGTM. Hi, I have tested this and it works well based on what we discussed on Monday.

src/components/node-list/node-list-items.js
```diff
@@ -10,80 +10,86 @@ import InvisibleIcon from '../icons/invisible';
 import { arrayToObject } from '../../utils';
 const { escapeRegExp, getHighlightedText } = utils;
 
+export const isTagType = (type) => type === 'tag';
+export const isParameterType = (type) => type === 'parameters';
```
studioswong, line 14: on cross-referencing the sidebar code this does not seem to be used anymore - looks like we can delete that.

src/components/node-list/node-list-items.js
```diff
@@ -159,13 +165,63 @@ export const getFilteredModularPipelineItems = createSelector(
...
+/**
+ * Return filtered/highlighted element type items
+ * @param {object} filteredTags List of filtered element types
+ * @param {array} nodeTypes List of node types
+ * @return {object} Element type items
+ */
+export const getFilteredElementTypeItems = createSelector(
+  [getFilteredElementTypes, (state) => state.nodeTypes],
+  (filteredElementTypes, nodeTypes) => ({
+    elementType: filteredElementTypes.elementType.map((elementType) => {
```
studioswong, line 205: I wonder what is the reasoning behind wrapping the returned results from `getFilteredElementTypes` under the `elementType` field in the first place, as the name of this function suggests it is only returning results based on element types?

src/components/node-list/node-list-items.js
```diff
@@ -188,7 +244,7 @@ const compareEnabledThenAlpha = (itemA, itemB) => {
  */
 const compareEnabledThenType = (itemA, itemB) => {
   const byEnabledTag = Number(itemA.disabled_tag) - Number(itemB.disabled_tag);
-  const nodeTypeIDs = ['task', 'data', 'parameters'];
+  const nodeTypeIDs = Object.keys(sidebarElementTypes);
```
studioswong, line 247: there isn't any need to declare or reference nodeTypeIDs `sidebarElementTypes` from `config.js` here as `nodeTypes` are being passed into the state of the selectors via the `nodeTypes` value in `node-list/index.js` - we can just simply pass that on as a variable from the selector that references it.

src/components/node-list/node-list-items.js
```diff
@@ -159,13 +165,63 @@ export const getFilteredModularPipelineItems = createSelector(
...
+    )
+);
+
+/**
+ * Return filtered/highlighted element type items
+ * @param {object} filteredTags List of filtered element types
+ * @param {array} nodeTypes List of node types
+ * @return {object} Element type items
+ */
+export const getFilteredElementTypeItems = createSelector(
```
studioswong, line 202: There aren't any related tests set up for this new selector - please add the relevant tests for this under `node-list-items.test.js`

src/reducers/reducers.test.js
```diff
@@ -157,12 +157,14 @@ describe('Reducer', () => {
     it('should toggle whether a type is disabled', () => {
       const newState = reducer(mockState.animals, {
         type: TOGGLE_TYPE_DISABLED,
-        typeID: '123',
+        typeIDs: { '123': true },
```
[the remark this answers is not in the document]
studioswong: I second Lim's comment - given the new condition with the 2 different cases for this reducer, please add in the relevant tests for those 2 cases listed in the reducer ( ' if no types will be enabled' and 'At least one type is enabled, set any unset types to disabled'

src/selectors/tags.js
```diff
@@ -29,3 +30,21 @@ export const getTagCount = createSelector(
     enabled: tagIDs.filter((id) => tagEnabled[id]).length,
   })
 );
+
+/**
+ * Get the total number of nodes for each tag
+ */
+ export const getTagNodeCounts = createSelector(
```
studioswong, line 37: Please add the relevant test for `getTagNodeCounts` in `tags.test.js`

src/components/node-list/node-list-row.js
```diff
@@ -116,42 +117,49 @@ const NodeListRow = memo(
...
-                  'pipeline-row__toggle-icon--all-unset': allUnset,
-                }
-              )}
-            />
-          </label>
+          <>
+            {typeof count === 'number' ? (
+              <span
+                onClick={onClick}
+                className={'pipeline-nodelist__row__count'}>
```
studioswong, line 125: I don't see any tests relating to the display of the count / existance of this css class for filter type rows - please add a test for this.

src/components/node-list/styles/_panels.scss
bru5: Not currently but probably a good idea for a future refactor?

src/components/node-list/node-list-items.js
bru5: I thought the ordering used here is fixed and depends on the sidebar design though? If that's the case then the config seems like a good place to declare that intention?

src/components/node-list/node-list-items.js
bru5: This is just to keep consistency with the same existing pattern used for all the other filters in this file (filter inputs -> create sidebar items). I agree though it would be nice to refactor this but at this point I think the priority is to merge this first?

src/components/node-list/styles/_row.scss
bru5: Ah that was noted above this comment, either way I've now moved it directly to this line instead.

src/components/node-list/node-list-items.js
studioswong: I see - yea I agreem we can refactor this away in a future PR.

src/components/node-list/node-list-items.js
studioswong: aargh I see your confusion - currently the order of the elements in the tree follows the order of the array declared under `nodes.type` in `normalize-data.js` ( i.e the value of `nodeTypes` in the state passed onto the reducers), and I had set up the order here as an array more so for clarity purposes to reiterate that order ( which in retrospective it might be better if I set the function to take in a `nodeTypeIDs` array instead and pass in the `nodeTypes`)

In looking at the elements declared in config, I do see that it is more so for the purpose of structuring the groups for the sidebar filter, which does make sense to declare it separately instead of reusing `nodeTypes` from state, as the grouping on the filter panel depends on the design. 

In terms of the `nodeTypeIDs` value used here, given that the function to sort by nodetype  is mainly utilised to sort elements for the tree, I would follow the setup by the other tree selectors in using `nodeTypes` from the state as a reference for the order, in case we do add any new node types that will be declared in `normalize-data.js` in the future. ( It might be even better to bring this forward that, instead of using a hardcoded object in config to declare the element types, we can also instead reference the node types declared in the state to set up the `sidebarElementTypes`. )

I understand this might be a bit of refactor work, I am also happy to do it in the focus mode PR if you wish to avoid blocking this PR :smiley

bru5:
Thanks for the reviews, I've now responded or pushed changes for your comments. Sorry about the delay pushing the tests, they're now up here too in keeping with the existing approach.

src/components/node-list/node-list-items.js
bru5: Ah ok then happy to leave it to you, but I think probably worth adding a comment explaining the order and maybe pulling it from a named constant somewhere, especially if it requires being in sync with something else.

limdauto (approved):
Awesome 🏆

studioswong (approved):
LGTM - thanks for the work and adding the tests 👍

PR #478, opened 2021-06-09
[KED-2681] Implement settings panel for flags

rashidakanchwala:
## Description

This PR is about implementing a settings panel for flags. 

Currently a lot of new features that are introduced on Kedro are hidden behind a flag. The user can enable/disabled a flag by using query string.

This ticket was created to enhance the user experience around enabling/disabling flags by creating a settings panel that has list of toggles that can be used to turn on/off a flag. 

## Development notes

As an MVP, we have created a modal that has toggles which allows users to enable/disable flags. This will change in future. 
The following design has been implemented - https://zpl.io/2vj0AGY 

The toggle component has been refactored because it is used in multiple places as a part of this PR. 

Note - During code review; we realised that there were missing tests around metadata toggle from before, so we have added those as well as a part of this PR> 

## QA notes

npm t run successfully

## Checklist

## Legal notice


src/components/primary-toolbar/index.js
```diff
@@ -85,6 +88,15 @@ export const PrimaryToolbar = ({
         disabled={disableLayerBtn}
         visible={visible.layerBtn}
       />
+      <IconButton
+        ariaLabel={`Change the settings flags`}
+        className={'pipeline-menu-button--settings'}
+        onClick={() => onToggleSettingsModal(true)}
+        icon={SettingsIcon}
+        labelText={`Settings`}
```
studioswong, line 96: There is no use for a backtick here - please change this to a `'`

src/components/primary-toolbar/primary-toolbar.test.js
```diff
@@ -37,6 +38,7 @@ describe('PrimaryToolbar', () => {
     ['.pipeline-menu-button--labels', 'onToggleTextLabels'],
     ['.pipeline-menu-button--export', 'onToggleExportModal'],
     ['.pipeline-menu-button--layers', 'onToggleLayers'],
+    ['.pipeline-menu-button--settings', 'onToggleSettingsModal'],
```
studioswong, line 41: While this function call is added, there has been no relevant test to use and test this in `mapDispatchToProps` - please add that test

studioswong (changes requested):
Please refer to comments for the relevant requested changes, moreover to add the test

studioswong:
I had communicated this to @rashidakanchwala , but adding this in for transparency:

- all PRs should be raised as draft if it is not ready to be reviewed
- @rashidakanchwala mentioned that there is a note mentioning the intention to add tests later, however I did not see the note this morning during the review, and the note was added afterwards. For future reference, if a note is to be added afterwards in the description, please mentioned this as an edit.

rashidakanchwala:
> I had communicated this to @rashidakanchwala , but adding this in for transparency:
> 
> * all PRs should be raised as draft if it is not ready to be reviewed
> * @rashidakanchwala mentioned that there is a note mentioning the intention to add tests later, however I did not see the note this morning during the review, and the note was added afterwards. For future reference, if a note is to be added afterwards in the description, please mentioned this as an edit.


**I added the note regarding testing to my description yesterday**. And i didn't add you as a reviewer yet because my PR was not ready. I have changed it draft PR

studioswong:
Thanks for your clarification Rashida - unfortunately github does not update the time of edit on the description ( as evident from the example that, the latest checkbox for 'draft' was added within the last minutes but it still shows up as last commented at 17 hours ago), hence if this is a misunderstanding then I will of course believe you, and thanks again for clarifying. The point I raised is more of a reminder, please do not take it the wrong way.

src/components/metadata/metadata-code.test.js
```diff
@@ -49,57 +48,3 @@ describe('MetaDataCode', () => {
     ).toBe(true);
   });
 });
-
-describe('MetaCodeToggle', () => {
```
studioswong, line 53: I wonder why are these tests deleted? The should not be deleted given the change to use the toggle component, but rather to update the `wrapper` selector below to the new css class for the toggle component

src/components/metadata/metadata-code.test.js
```diff
@@ -49,57 +48,3 @@ describe('MetaDataCode', () => {
...
-  const input = (wrapper) =>
-    wrapper.find('.pipeline-metadata__code-toggle-input');
```
studioswong, line 55: As I mentioned in the comment above, the css class should be updated to reflect the new css rather than the entire test suite being deleted.

src/components/settings-modal/index.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Modal from '@quantumblack/kedro-ui/lib/components/modal';
+import { changeFlag, toggleSettingsModal } from '../../actions';
+import { getFlagsData } from '../../utils/flags';
+import './settings-modal.css';
+import Toggle from '../toggle';
+
+/**
+ * Kedro-UI modal to allow users to choose between SVG/PNG export formats
```
studioswong, line 10: This is a copied comment from the export modal - please delete

src/components/metadata/metadata-code.test.js
rashidakanchwala: Yes, i was a bit confused regarding this bit. So now, there is a toggle component with its own unit tests which are very similar to the MetaCodeToggle tests. Since this is unit testing and the toggle tests are handled by the core component (Toggle.js). I thought maybe it should not be repeated here...

src/components/settings-modal/settings-modal.scss
```diff
@@ -0,0 +1,88 @@
...
+  --color-modal-icon: #{$color-text-light};
+  --color-modal-header-text: #{rgba($color-dark, 0.55)};
+  --color-toggle--on: #{$color-primary-2};
+  --color-toggle-bar--on: #{rgba($color-primary, 0.3)};
+  --color-toggle--off: #{$color-bg-dark-9};
+  --color-toggle-bar--off: #{rgba($color-bg-dark-9, 0.3)};
+}
+
+.kui-theme--dark {
+  --color-modal-bg: #{$color-bg-dark-8};
```
studioswong, line 17: This is more likely a question for @GabrielComymQB - I noticed that the settings modal is using a different background color and font size from the export modal - I understand that this is the color adopted from the design, however I noticed that visually the 2 modals do deviate quite a bit which perhaps isn't the best alignment visually given both are modals ( the large pipeline warning also uses a similar design language to the export modal). 

I wonder if we would need to either align with the background color of the old modal, or update the background color of the other modals for consistency? 

image

image

src/components/toggle/index.js
```diff
@@ -0,0 +1,39 @@
...
+const Toggle = ({
+  id = '',
+  checked,
+  enabled = true,
+  title,
+  onChange,
+  className = '',
+}) => (
+  <div
+    className={modifiers(classnames('pipeline-toggle', className), {
```
studioswong, line 18: Not sure why this css class is named as `pipeline-toggle` given that this toggle is a general component

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.kui-theme--dark {
+  --color-toggle-on: #{$color-primary};
+  --color-toggle-on-bar: #{rgba($color-primary, 0.3)};
+  --color-toggle-off: #{$color-light};
+  --color-toggle-off-bar: #{rgba($color-light, 0.3)};
+}
+
+.pipeline-toggle {
```
studioswong, line 17: I see that this naming was adopted from the classes from `metadata-code` ( which I agree also shouldn't be named as `pipeline-(something)` - as per my other comment, given that this is a general react component, please rename this to a classname that is scoped within the toggle button. Naming this `pipeline-` will make future debugging harder and more confusing.

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,86 @@
...
+  transition: opacity 0.2s ease;
+}
+
+.pipeline-toggle--no-hasCode {
```
studioswong, line 21: same as above

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,86 @@
...
+  opacity: 0.35;
+  pointer-events: none;
+}
+
+.pipeline-toggle {
```
studioswong, line 26: same as above

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,86 @@
...
+  margin-top: 0.3em;
+  margin-bottom: 2em;
+}
+
+.pipeline-toggle-input {
```
studioswong, line 31: same as above

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,86 @@
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  overflow: hidden;
+  white-space: nowrap;
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+}
+
+.pipeline-toggle-label {
```
studioswong, line 41: same as above

src/components/toggle/toggle.scss
```diff
@@ -0,0 +1,86 @@
...
+  position: relative;
+  padding: 0 0 0 2.8em;
+  cursor: pointer;
+  user-select: none;
+}
+
+.pipeline-toggle-label:before,
```
studioswong, line 48: same as above

src/components/toggle/toggle.test.js
```diff
@@ -0,0 +1,30 @@
...
+  const input = (wrapper) => wrapper.find('.pipeline-toggle-input');
+  const label = (wrapper) => wrapper.find('.pipeline-toggle-label');
+
+  it('is checked when checked is true', () => {
+    const wrapper = setup.mount(
+      <Toggle checked={true} onChange={jest.fn()}>
+        {' '}
+      </Toggle>
+    );
+    expect(input(wrapper).prop('checked')).toBe(true);
```
studioswong, line 15: I'm not entirely sure about the effectiveness of this test given this is technically testing the functionality of react in passing props which isn't needed - this could be deleted.

src/components/toggle/toggle.test.js
```diff
@@ -0,0 +1,30 @@
...
+    expect(label(wrapper).hasClass('pipeline-toggle-label--checked')).toBe(
+      true
+    );
+  });
+
+  it('is not checked when checked is false', () => {
+    const wrapper = setup.mount(
+      <Toggle checked={false} onChange={jest.fn()}></Toggle>
+    );
+    expect(input(wrapper).prop('checked')).toBe(false);
```
studioswong, line 25: same for above

src/components/toggle/toggle.scss
rashidakanchwala: heyy thanks I will fix this! 

Just a request - if you have already highlighted the problem, then you can avoid 'same as above' comment .. it's just that it creates too much noise when reading the reviews :D

studioswong (commented):
good work @rashidakanchwala  - see my comments below. It is worth looking into how we align this settings modal design with the existing design.

src/components/toggle/toggle.scss
studioswong: totally understand - I added these comments mainly to aid my review as it helps to highlights the areas to address and helps me to track the further updates on github

src/components/toggle/index.js
bru5: I think `pipeline-` is the general prefix for everything in Viz - we've talked before about changing it to `kv` to avoid this confusion but for now it should be there.

src/components/metadata/metadata-code.test.js
studioswong: I see - I had a look at the tests at `toggle.test.js`, and the tests that are set up there are testing different things - those tests are scoped to just within the toggle component,  while all the tests here are focused on the performance of the code toggle and all relevant actions for the code panel, which is essential to test the entire code panel feature. These should not be deleted - please put them back in. Thank you :)

src/components/metadata/metadata-code.test.js
studioswong: @rashidakanchwala following up on my comment, the quickest way to tackle this is to update the css class here to reflect the new classes that were set up under `toggle.js` :) Thank you

src/components/primary-toolbar/primary-toolbar.test.js
studioswong: I saw that a test has been added, hence resolving this.

src/components/settings-modal/index.js
studioswong: @rashidakanchwala please see above comment

src/components/settings-modal/settings-modal.scss
studioswong: I see that Kedro-UI has been updated and the relevant classes deleted, hence resolving this.

src/components/toggle/toggle.scss
studioswong: As per @bru5 's comment, I understand about the `pipeline` prefix - this is definitely problematic, but I see that this is out of scope for this work and hence resolving this.

src/components/toggle/index.js
studioswong: I see, thanks for clarifying @bru5 . I don't think this has been communicated within the team before, good to know about this, and on checking the code I do see this as a pattern.

I wonder what was the reason behind using `pipeline-` as the general prefix? This is definitely confusing and worth a future ticket and discussion about how to go forward with this.

src/components/toggle/toggle.test.js
rashidakanchwala: @studioswong  -- i thought it was doing that. I am confused now, can we have a quick 1:1 where you can explain to me this further. thanks

src/components/metadata/metadata-code.test.js
rashidakanchwala: @studioswong , i added the MetaDataToggle tests back and i renamed the component to Toggle and updated the classnames and prop names. Now it looks exactly like the test functions in toggle . So I am not sure -- whenever you are free, can we do a quick 1:1 to address it. thanks. I don't want to push it as I am still not convinced !

studioswong: I had a further detailed look at the tests that were deleted and see that it was set up in a similar fashion to the tests in `toggle.test.js`, and definitely understand where you are coming from. 

For me the main value I am looking for in setting up toggle related tests in `meta-data-code` is to test the `onChange` callback that was passed into the toggle, with which this needs to be set up in the context with the `meta-data-code` file given that the callback was setup and passed on from here ( refer to the `'onChange callback fires when input changed'` test that was deleted here) - the tests set up in `Toggle.js` was scoped to testing the dummy component itself, while setting up the test here would allow us to test it end to end. ( the situation is similar to the difference between tests for individual selectors, while stlll setting up tests within the component that consumes that selector to test the end to end funcionality of that feature). 

I would say the simpliest way to do this is to set up a test similar to the `'onChange callback fires when input changed'` test with the new toggle, while the other tests that only tests the passed prop for the toggle component can stay within `toggle.test.js`. 

Other than the current setup, I would even extend the tests within `toggle.test.js` to ensure that the right css classes gets shown for the different prop ( current it just tests whether the passed prop is right which is pretty much useless, as it's testing an assumed react capability rather than testing our app setup). I know the previous testing setup was really not good and your work was built on from there, so definitely this is no fault on you - let's use this as an oppourtunity to do it well ;) we can pair on this tomorrow in our 1:1.

src/components/metadata/metadata-code.test.js
studioswong: resolving this as per our discussion in our 1:1. Thanks @rashidakanchwala

src/components/toggle/toggle.test.js
studioswong: resolving this as per our discussion in our 1:1. Thanks @rashidakanchwala

src/reducers/reducers.test.js
```diff
@@ -301,6 +302,17 @@ describe('Reducer', () => {
     });
   });
 
+  describe('TOGGLE_CODE', () => {
+    it('should toggle the value of hoveredParameters', () => {
```
studioswong, line 306: The description here is irrelevant to the test - it should be something along the lines as `should toggle / update the value of `visible.code` in the state`

src/reducers/reducers.test.js
```diff
@@ -301,6 +302,17 @@ describe('Reducer', () => {
...
+      const newState = reducer(mockState.animals, {
+        type: TOGGLE_CODE,
+        visible: true,
+      });
+      expect(mockState.animals.visible.code).toBe(false);
```
studioswong, line 311: This test is not needed / relevant to what we are testing - please delete

rashidakanchwala: @studioswong  -- in our 1:1 the other day, you highlighted that this unrelated test was missing and it should added  as well. I personally think it should be in a new PR because it is unrelated but the consensus last time was to do everything int hsi PR.

studioswong: Hi Rashida, there seems to be some misunderstand - I am not referring for you to delete this entire test, but only this line (`expect(mockState.animals.visible.code).toBe(false);`) as it is not testing anything relevant to this test.

rashidakanchwala: thanks Susanna for the clarification. It wasn't clear in the initial comment. I will fix this.

src/utils/flags.js
```diff
@@ -83,3 +83,14 @@ export const getFlagsMessage = (flagsEnabled) => {
     return info;
   }
 };
+
+export const getFlagsData = () => {
```
bru5, line 87: Might be worth considering this as `getFlagsState`?

bru5 (commented):
Overall looks great, added a few optional suggestions. 

As discussed I think once the font sizing and responsive changes are finished think this should be ready?

studioswong (approved):
LGTM 👍 

Approving this on my side, however looks like Liam has raised some suggestions with regards to font sizing and responsiveness - would be great to tackle them and get his approval before merging this in. 

Thank you :)

bru5 (approved):

bru5:
Awesome work 🙌

PR #481, opened 2021-06-14
Integrate with kedro-telemetry

limdauto:
## Description

With the release of [kedro-telemetry](https://github.com/quantumblack/kedro-telemetry), once user has opted into sending telemetry to us, we want to enable automatic tracking of features usage on Kedro-Viz as well. This is done by including [Heap SDK](https://heap.io/) if and only if user's consent has been previously acquired by kedro-telemetry. 

## QA notes

If you want to test this:

* Make sure you have the environment variable `HEAP_APPID_DEV` set to our Heap dev account.
* Check out this branch and run `make install`
* When you open Heap, you should be able to see your events sent to Heap successfully both in the Live view and the Analyze view.

Screenshot 2021-06-14 at 18 15 01

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/integrations/kedro/telemetry.py
+++ b/package/kedro_viz/integrations/kedro/telemetry.py
@@ -1,0 +1,59 @@
+# Copyright 2021 QuantumBlack Visual Analytics Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+# OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, AND
+# NONINFRINGEMENT. IN NO EVENT WILL THE LICENSOR OR OTHER CONTRIBUTORS
+# BE LIABLE FOR ANY CLAIM, DAMAGES, OR OTHER LIABILITY, WHETHER IN AN
+# ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF, OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+# The QuantumBlack Visual Analytics Limited ("QuantumBlack") name and logo
+# (either separately or in combination, "QuantumBlack Trademarks") are
+# trademarks of QuantumBlack. The License does not grant you any right or
+# license to the QuantumBlack Trademarks. You may not use the QuantumBlack
+# Trademarks or any confusingly similar mark as a trademark for your product,
+# or use the QuantumBlack Trademarks in any other manner that might cause
+# confusion in the marketplace, including but not limited to in advertising,
+# on websites, or on software.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""`kedro_viz.integrations.kedro.telemetry` helps integrate Kedro Viz with Kedro-Telemetry
+"""
+import hashlib
+import logging
+import socket
+from pathlib import Path
+from typing import Optional
+
+import yaml
+from kedro_telemetry.plugin import _get_heap_app_id, _is_valid_syntax
+
+logger = logging.getLogger(__name__)
+
+
+def get_heap_app_id(project_path: Path) -> Optional[str]:
+    """Return the Heap App ID used for Kedro telemetry if user has given consent."""
+    telemetry_file_path = project_path / ".telemetry"
+    if not telemetry_file_path.exists():
+        return None
+    with open(telemetry_file_path) as telemetry_file:
+        telemetry = yaml.safe_load(telemetry_file)
+        if _is_valid_syntax(telemetry) and telemetry["consent"]:
+            return _get_heap_app_id()
+    return None
+
+
+def get_heap_identity() -> Optional[str]:  # pragma: no cover
+    """Return the user ID in heap identical to the id used by kedro-telemetry plugin."""
+    try:
+        return hashlib.sha512(bytes(socket.gethostname(), encoding="utf8")).hexdigest()
+    except socket.timeout:
+        return None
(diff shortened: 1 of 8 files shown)


package/kedro_viz/integrations/kedro/telemetry.py
```diff
@@ -0,0 +1,59 @@
...
+    if not telemetry_file_path.exists():
+        return None
+    with open(telemetry_file_path) as telemetry_file:
+        telemetry = yaml.safe_load(telemetry_file)
+        if _is_valid_syntax(telemetry) and telemetry["consent"]:
+            return _get_heap_app_id()
+    return None
+
+
+def get_heap_identity() -> Optional[str]:  # pragma: no cover
```
limdauto, line 54: Ideally there should be a function like this in `kedro-telemetry`. I can mock out hashlib and test it's being called but it's not testing anything useful. Adding `no cover` instead and will PR on the telemetry side to expose this logic as a function.

package/kedro_viz/integrations/kedro/telemetry.py
```diff
@@ -0,0 +1,59 @@
...
+from pathlib import Path
+from typing import Optional
+
+import yaml
+from kedro_telemetry.plugin import _get_heap_app_id, _is_valid_syntax
+
+logger = logging.getLogger(__name__)
+
+
+def get_heap_app_id(project_path: Path) -> Optional[str]:
```
limdauto, line 42: Need to wrap `kedro-telemetry`'s `_get_heap_app_id` and add some consent checking logic because in `kedro-telemetry`, using `_check_for_telemetry_consent` will prompt user to confirm consent. I will also extract a boolean-checking logic only on the telemetry side.

public/index.html
```diff
@@ -9,6 +9,13 @@
     <meta name="theme-color" content="#000000" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
+    {% if should_add_telemetry %}
+    <script type="text/javascript">   
+      window.heap=window.heap||[],heap.load=function(e,t){window.heap.appid=e,window.heap.config=t=t||{};var r=document.createElement("script");r.type="text/javascript",r.async=!0,r.src="https://cdn.heapanalytics.com/js/heap-"+e+".js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(r,a);for(var n=function(e){return function(){heap.push([e].concat(Array.prototype.slice.call(arguments,0)))}},p=["addEventProperties","addUserProperties","clearEventProperties","identify","resetIdentity","removeEventProperty","setEventProperties","track","unsetEventProperty"],o=0;o<p.length;o++)heap[p[o]]=n(p[o])};
+      heap.load("{{ heap_app_id }}"); 
+      heap.identify("{{ heap_user_identity }}");
```
lorenabalan, line 16: Hmm we don't have `heap.identify` on kedro docs for example - why do you need this?

package/tests/test_api/test_apps.py
```diff
@@ -192,6 +192,19 @@ class TestIndexEndpoint:
     def test_index(self, client):
         response = client.get("/")
         assert response.status_code == 200
+        assert "heap" not in response.text
+
+    @mock.patch("kedro_viz.integrations.kedro.telemetry.get_heap_app_id")
+    @mock.patch("kedro_viz.integrations.kedro.telemetry.get_heap_identity")
+    def test_heap_enabled(
+        self, mock_get_heap_identity, mock_get_heap_app_id, client, tmpdir
+    ):
```
lorenabalan, line 201: We normally do the patching inside the test function, but not sure if there's a different style guide on viz.

package/requirements.txt
```diff
@@ -1,5 +1,6 @@
 semver~=2.10 # Needs to be at least 2.10.0 to get VersionInfo.match
 kedro>=0.16.0
+kedro-telemetry>=0.1.0
```
lorenabalan, line 3: I would've been a little more restrictive but up to you.
```suggestion
kedro-telemetry~=0.1.0
```

lorenabalan (approved):
LGTM! 👍

public/index.html
```diff
@@ -9,6 +9,13 @@
...
+      </script>
+    {% endif %}
```
AntonyMilneQB, line 18: Curious... what is this and where does it come from?! Why does it look so weird?

package/requirements.txt
limdauto: Yea my passive-agressive way of saying if you break my software, I will come find you :D

public/index.html
AntonyMilneQB: I see that you are passing the jinja variables in from `templates.TemplateResponse` but everything else here is a complete mystery to me.

public/index.html
limdauto: In viz, we need to match the user on the web with the user on the command line. The CLI user is identified by the host machine so I pass the same hash to the client. In the doc, we don't have the command line so we just let Heap work out who it is based on the session.

AntonyMilneQB (approved):
This looks good as far as I understand it 😅 

One thing to check: what happens if I do `kedro viz` before I've run any other `kedro` command. Do I get the question about whether I want to enable telemetry?

package/tests/test_api/test_apps.py
AntonyMilneQB: I guess this is a difference between `unittest.mock` and pytest's mocker?

public/index.html
limdauto: It's copied and pasted from here: https://developers.heap.io/docs/web :D

public/index.html
lorenabalan: Makes sense, thank you!

limdauto:
> This looks good as far as I understand it 😅
> 
> One thing to check: what happens if I do `kedro viz` before I've run any other `kedro` command. Do I get the question about whether I want to enable telemetry?

@AntonyMilneQB Yes! It would follow the same process as another other command.

package/tests/test_api/test_apps.py
limdauto: Yea I prefer this style maybe just out of old habits.

package/tests/test_integrations/test_telemetry.py
```diff
@@ -0,0 +1,55 @@
...
+    assert kedro_telemetry.get_heap_app_id(tmpdir) is None
+
+
+def test_get_heap_app_id_no_consent(tmpdir):
+    telemetry_file = tmpdir / ".telemetry"
+    telemetry_file.write_text("consent: false", encoding="utf-8")
+    assert kedro_telemetry.get_heap_app_id(tmpdir) is None
+
+
+@mock.patch("kedro_viz.integrations.kedro.telemetry._get_heap_app_id")
```
lorenabalan, line 50: There's a mix of `unittest` and `pytest` in the `tests` folder. 😂  Assuming this still runs when calling `pytest` right?

PR #494, opened 2021-06-29
[KED-2711] Fix parameters display on metadata panel

rashidakanchwala:
## Description

Parameters are passed as nested objects from the backend and on the front end they were simply being  equated to strings. For this reason the on the metadata panel, the parameters were rendered like below. 

Screenshot 2021-06-29 at 11 18 18

In this ticket we used react-json-view to display parameters in a collapsable tree view. The react-json-view also offers functionality like copy to clipboard to the parameters view

Screenshot 2021-07-01 at 11 47 46

Please note - the count in the title has been removed as part of the design (requested by Joel & Gabriel) 

## Development notes

**Prior to this ticket parameters were rendered as a metadata list and the limit of the list was set to 10. So it wouldn't display more than 10 parameters. Parameters are no longer metadata list but are now part of metadata object and do not have the limit. Moreover, the count in the title has been removed; for this reason the test which was checking for more than 10 parameters and expanding if there were more is not relevant and has been been replaced by checking limit on a generic metadata list**

## QA notes

All npm t run successfully. 

## Checklist

## Legal notice


limdauto:
Could you take a couple more screenshots how this looks for other type of parameters (e.g. primitives, array) please?

rashidakanchwala:
> Could you take a couple more screenshots how this looks for other type of parameters (e.g. primitives, array) please?

updated the description to show the edge case (gif) and one more simple example.

src/components/metadata/index.js
```diff
@@ -106,8 +107,10 @@ const MetaData = ({
                   value={metadata.filepath}
                 />
                 <MetaDataRow
-                  label={`Parameters (${metadata.parameters?.length || '-'}):`}
+                  label={'Parameters:'}
```
limdauto, line 110:
```suggestion
                  label="Parameters:"
```
No need for this syntax with a raw string

src/selectors/metadata.js
```diff
@@ -24,6 +23,14 @@ const runCommandTemplates = {
   task: (name) => `kedro run --to-nodes ${name}`,
 };
 
+const renderParameters = (params) => {
+  if (!params || !params[Object.keys(params)[0]]) {
```
limdauto, line 27: Sorry what does `Object.keys(params)[0]]` mean? Also if we are going to reuse it, can we extract it into a variable and give it a descriptive name?

src/components/metadata/styles/metadata.scss
```diff
@@ -299,3 +300,8 @@ $list-inline-spacing: 0.2em;
   color: var(--color-button-text);
   font-size: 1.6em;
 }
+
+.react-json-view {
+  font-family: Consolas, Monaco, 'Courier New', Courier, monospace !important;
```
limdauto, line 305: Do we want to force `!important` here or should we prefix it with a more specific selector, e.g. `.pipeline-metadata__params-json-view`

limdauto (commented):
I have a few minor questions. It looks good to me generally. I will test it out as well ⭐

src/components/metadata/styles/metadata.scss
rashidakanchwala: i am going to give this another go. I tried it before and it didn't override the font. so basically without important it applies the font of the react-json-view package. and i am overriding that class here. let me refine it on Monday again.

src/selectors/metadata.js
rashidakanchwala: i have edited this code so it's more clear. changed renderParameters to unwrapParameters. it's actually because of the way we get the json from backend. it looks like this 
{ 
   params_name: {
       params_key:1 { 
       
      }
} --> there's this extra wrapper around the json, and I need to unwrap it and access the first object to get all the parameters output.

src/components/metadata/styles/metadata.scss
rashidakanchwala: i am going to give this another go. I tried it before and it didn't override the font. so basically without important it applies the font of the react-json-view package. and i am overriding that class here. let me refine it on Monday again.

studioswong:
Thanks for the work Rashida - very glad and happy to see the json tree is part of this work, well done :) 

I will review this after getting the filter panel off the line - thank you!

src/components/metadata/metadata-row.js
```diff
@@ -19,7 +21,7 @@ const MetaDataRow = ({
   children,
 }) => {
   const showList = Array.isArray(value);
-
+  const showObject = typeof value === 'object' && !showList;
```
studioswong, line 24: I am not sure of the need for the `!showList` check as showList would not be positive if `value` is an object. 
I would, however, add an additional check to ensure that the object is not an object ( i.e `const showObject = typeof value === 'object' && Object.keys(value).length > 0`) before displaying it.

studioswong:
On testing the app across multiple datasets, I have been getting the following errors on the console on having clicked a pameter node:
image
image

It looks like the error concerns `metadata-value.js` - please look into the issue. ( fixing this issue would also fix the current failing JS tests on the CI as they are failing for the same error)

src/components/metadata/metadata.test.js
```diff
@@ -30,42 +31,41 @@ describe('MetaData', () => {
...
           afterLayoutActions: [() => toggleNodeClicked(salmonTaskNodeId)],
         })
       );
-      // Add extra mock parameters
-      metadata.parameters = Array.from({ length: 20 }, (_, i) => `Test: ${i}`);
-
+      metadata.inputs = Array.from({ length: 20 }, (_, i) => `Test: ${i}`);
       const wrapper = setup.mount(
-        <MetaData visible={true} metadata={metadata} />
+        <MetaDataList
```
studioswong, line 49: Given that this test focuses on testing the entire metadata panel and the integration of all its individual component, it would be better to follow the previous set up in mounting and test with the entire metadata component instead of the individual `MetaDataList` component - what is the reason behind changing the mount component?

rashidakanchwala: hi Susanna, please read the development notes I have mentioned it. Ideally the test should check if expand collapse function happens on metadata list if there's a limit and the new test does that. Currently, no metadata property uses this feature --parameters was using it previously but it no longer does.

rashidakanchwala:
> On testing the app across multiple datasets, I have been getting the following errors on the console on having clicked a pameter node:
> image
> image
> 
> It looks like the error concerns `metadata-value.js` - please look into the issue. ( fixing this issue would also fix the current failing JS tests on the CI as they are failing for the same error)

Thanks for flagging this. I have fixed this issue!!!

src/components/metadata/metadata.test.js
studioswong: Hi Rashida, yes I have read the development notes you mentioned, and understand the situation where the metadata does not have a `parameter` field and is replaced b the `input` field. However, this does not affect the setup of this test - we are still importing `metadata` that includes the `inputs` field into the `metadata` component and the test should pick up the same CSS classes as we previously does. 

Setting up the test to only test the individual component essentially changes the nature of this test from an integration test to simply a unit test of the component.

src/components/metadata/metadata-row.js
rashidakanchwala: @studioswong  -- it's weird but if you do type of array in javascript it's an object, hence showList is also true for objects. however it's not true for object with key value pairs -- hence I think this works the same way as your suggested way of implementation - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

src/selectors/metadata.js
```diff
@@ -24,6 +23,18 @@ const runCommandTemplates = {
   task: (name) => `kedro run --to-nodes ${name}`,
 };
 
+/**
+ * Unwrap parameters by extracting the first object
```
studioswong, line 27: I am not familiar with the object that is returned for parameters, but I wonder why are we only extracting the object of the first key of the parameter object? Is this based on any assumption of the data returned by the server? It would be good to elaborate on this assumption here in this comment. 

Also, using the term 'unwrapping' sounds a bit confusing - I would rephrase it as `returns a destructured parameter object by extracting the first object of the returned parameter metadata ( based on the assumption of the returned data from the server( or XX - insert assumption) `

src/selectors/metadata.js
```diff
@@ -24,6 +23,18 @@ const runCommandTemplates = {
...
+ *  @param {?object} parameters The parameters
+ *  @returns {object} The unwrapped parameters
+ */
+const unwrapParameters = (parameters) => {
```
studioswong, line 31: As per above, I would rename this function as `destructuredParameters`, or anything more clear other than unwrap

src/selectors/metadata.js
```diff
@@ -24,6 +23,18 @@ const runCommandTemplates = {
   task: (name) => `kedro run --to-nodes ${name}`,
 };
 
+/**
+ * Unwrap parameters by extracting the first object
+ *  @param {?object} parameters The parameters
```
studioswong, line 28: I assumed this refers to the raw parameter object returned from the server?

studioswong:
> > On testing the app across multiple datasets, I have been getting the following errors on the console on having clicked a pameter node:
> > image
> > image
> > It looks like the error concerns `metadata-value.js` - please look into the issue. ( fixing this issue would also fix the current failing JS tests on the CI as they are failing for the same error)
> 
> Thanks for flagging this. I have fixed this issue!!!

I've tested this and it works well now, thanks for pushing this fix!

Happy to approve this once the above comments are addressed - thank you and good job.

src/selectors/metadata.js
rashidakanchwala: yes

src/selectors/metadata.js
rashidakanchwala: thanks @studioswong  -- i will leave the function name as unWrap because i think that sounds correct as well. 
I will mention 'because of returned data from server' for more clarity. thanks

studioswong: thanks for the follow up Rashida - I am not convinced that `unwrap` is the right word for it, this is not a term uncommonly used in codebases. What you are doing is essentially destructing the object and it would be much more intuitive / clear that way.

limdauto (approved):
Great stuff. I think we could do with a test, either at the `<MetaDataObject />` level or the `<MetaData />` level, to check that it shows up properly for a non-empty parameters object.

LGTM otherwise!

src/components/metadata/metadata-row.js
studioswong: aargh I see - good point! yea in that case let's keep this

studioswong (approved):
Nice to have: Currently there is a visual jump between '0 Items' and 'X items' when the parameter metadata is fetched from the endpoint. Ideally this would be solved if we only display the JSON object viewer when the object is not an empty object. 

But otherwise it looks good! We can also update this via a later PR in order not to block the release.

PR #496, opened 2021-07-05
[KED-2736] Fix metadata inputs outputs

rashidakanchwala:
## Description

Previously Inputs and Outputs field on the metadata panel used to reflect only the graph nodes. If the graph nodes were hidden, the Inputs/Output field wouldn't display those nodes. This was not the correct representation of the kedro pipeline.

In this PR, the metadata info reads the kedro pipeline data and displays all the inputs and outputs associated with the node. It also highlights all the input/output nodes that are visible on the graph and fades all input/output nodes that are missing from the graph

## Development notes

I have added test cases for the back-end and front-end. 

**Metadata-test.js in the front-end has been updated to reflect how it would run in a live environment. i.e. when the node is clicked, addNodeMetadata would load metadata content from api.** 

## Checklist

## Legal notice


limdauto (commented):
The backend code looks good to me but I wonder why the frontend tests unrelated to inputs and outputs have changed?

bru5 (approved):
Awesome, looks good!

src/components/metadata/metadata.test.js
```diff
@@ -93,70 +97,59 @@ describe('MetaData', () => {
...
       expect(textOf(rowObject(row))[0]).toEqual(
         expect.stringContaining('3 items')
       );
     });
 
-    it('shows the node inputs when parameters are disabled (default)', () => {
-      const wrapper = mount({ nodeId: salmonTaskNodeId });
-      const row = rowByLabel(wrapper, 'Inputs:');
-      expect(textOf(rowValue(row))).toEqual(['Cat', 'Dog']);
-    });
```
studioswong, line 145: I am confused as to why this test is deleted? Looking at the test it should still be valid amongst the changes and should still pass.

studioswong (changes requested):
Thanks for the work Rashida!

In looking at the execution, it looks like that we are now expecting an `input` and `output` field from the returned response from the metadata endpoint for nodes to populate the input/output fields. However, in checking the current response data returned from the node metadata endpoints, it doesn't not seem to include that field ( see below screenshot of the response of a task node). This is problematic as on checking out this branch and playing around with the app, none of the task nodes now displays data for the input/output ( given that the field is not returned in the response.)
image

Are we currently expecting work to expose the `input` and `output` field for the BE? If so this ticket should only be raised after that work is done. Otherwise, we can still calculate the true state of the input/output from the edges data, which would be a completely different execution from that raised in this PR.

src/components/metadata/metadata.test.js
```diff
@@ -20,6 +22,8 @@ describe('MetaData', () => {
       afterLayoutActions: [
         // Click the expected node
         () => toggleNodeClicked(props.nodeId),
+        //simulating loadNodeData in node.js
```
studioswong, line 25: rewording this to `simulating loadNodeData action in 'action/node.js'` would be more explicit

src/components/metadata/metadata.test.js
```diff
@@ -20,6 +22,8 @@ describe('MetaData', () => {
...
+        () => addNodeMetadata({ id: props.nodeId, data: props.mockMetadata }),
```
studioswong, line 26: It looks like this is added to simulate the `input` and `output` data field that is now returned from the endpoint -  rather than adding this to the mount function that affects all tests, it is simplier to simply adopt the same set up as the test for `code toggle`, where the `afterLayoutAction` method of simulating the fetching of node data is done only on the mock metadata rather than on the actual mount function itself.), hence reducing the need to add the `mockMetadata` for tests that does not need the mock metadata. 

Introducing the `afterLayoutAction` here would effectively introduce a lot of redundant refactor of this test.

src/components/metadata/metadata.test.js
```diff
@@ -93,70 +97,59 @@ describe('MetaData', () => {
     });
 
     it('shows the node type as an icon', () => {
-      const wrapper = mount({ nodeId: salmonTaskNodeId });
+      const wrapper = mount({
+        nodeId: salmonTaskNodeId,
+        mockMetadata: node_task,
```
studioswong, line 102: see my above comment - this change would not be needed if we take out the mock action in the mount function, as here we are testing something that does not need mock data and hence this proposed change is redundant

src/components/metadata/metadata.test.js
```diff
@@ -93,70 +97,59 @@ describe('MetaData', () => {
...
-            },
-          ],
-        });
-      };
-      const wrapper = mount({ nodeId: salmonTaskNodeId });
+    it('shows the node inputs', () => {
+      const wrapper = mount({
+        nodeId: salmonTaskNodeId,
+        mockMetadata: node_task,
+      });
```
studioswong, line 152: To test the node inputs, I would now reformat the test as `shows the node inputs with metadata fetched from the endpoint`, and then reference the setup of the `toggle code` test to format this test. ( this is the same setup given the code toggle is also relying on returned metadata to format the test.)

```describe('shows the node inputs with metadata fetched from the endpoint', () => {
      // Get metadata for a sample node
      const metadata = getClickedNodeMetaData(
        prepareState({
          data: animals,
          afterLayoutActions: [() => toggleNodeClicked(salmonTaskNodeId)],
        })
      );

      const wrapper = setup.mount(
        <MetaData visible={true} metadata={metadata} />
      );

      lett row = rowByLabel(wrapper, 'Inputs:');
      expect(textOf(rowValue(row))).toEqual([
        'Cat',
        'Dog',
        'Parameters',
        'Params:rabbit',
      ]);
    });

      row = rowByLabel(wrapper, 'Outputs:');
      expect(textOf(rowValue(row))).toEqual(['Horse', 'Sheep']);

    });```

File: src/components/metadata/metadata.test.js [hunk 6]
```diff
@@ -323,51 +353,75 @@ describe('MetaData', () => {
...
           afterLayoutActions: [
-            () => {
-              // Click the expected node
-              return toggleNodeClicked(props.nodeId);
-            },
+            // Click the expected node
+            () => toggleNodeClicked(props.nodeId),
+            //simulating loadNodeData in node.js
+            () =>
+              addNodeMetadata({ id: props.nodeId, data: props.mockMetadata }),
```
studioswong, line 360: again, this setup is not needed - please refer to my above comment

src/components/metadata/metadata.test.js
```diff
@@ -376,54 +430,64 @@ describe('MetaData', () => {
...
-      metadata.plot = node_plot.plot;
-
-      const wrapper = setup.mount(
-        <MetaData visible={true} metadata={metadata} />
-      );
-
+      const wrapper = mount({
+        nodeId: bullPlotNodeID,
+        mockMetadata: node_plot,
+      });
```
studioswong, line 490: with the above proposed setup, I would instead keep the previous setup, as it is much more readable to see that this test is testing the display of a field returned from the metadata.

studioswong (commented):
the overall setup works, however on looking further into the test it would be much easier to set up the mock metadata on individual test rather than forcing all tests to load the mock metadata, even for tests that does not require the mock metadata in order to tests its functionality. 

This allows the tests to be more readable ( as only the tests that truly needs the mock metadata will load it, thereby allowing more clairty as to which field relies on the metadata response), as well as reducing the redundant code changes introduced in all the tests. We can pair on this on Monday in case you have any questions about the setup.

rashidakanchwala:
> Thanks for the work Rashida!
> 
> ~In looking at the execution, it looks like that we are now expecting an `input` and `output` field from the returned response from the metadata endpoint for nodes to populate the input/output fields. However, in checking the current response data returned from the node metadata endpoints, it doesn't not seem to include that field ( see below screenshot of the response of a task node). This is problematic as on checking out this branch and playing around with the app, none of the task nodes now displays data for the input/output ( given that the field is not returned in the response.)~
> image
> 
> ~Are we currently expecting work to expose the `input` and `output` field for the BE? If so this ticket should only be raised after that work is done. Otherwise, we can still calculate the true state of the input/output from the edges data, which would be a completely different execution from that raised in this PR.~
> 
> I've spoken to Lim and looks like this PR includes the BE work to expose the input/output field for the `node` endpoint response - sorry for my false alarm 😂 ( for future reference the 'development notes' would be a great place to elaborate these related BE changes 😉)
> 
> I've reinstalled and reran the server and the returned data works well 👍
> 
> Will go through the code in more detail and let you know if I have any further comments

Thanks Susanna. Yes, i had a good realisation today when I saw your PR and how you had listed all the work you did in points. I think it's very helpful in general and I will do that as well.

rashidakanchwala:
> I've did some testing and the overall setup works well, thank you Rashida!
> 
> My main concern / comment on this PR is on the series of introduced changes to the metadata tests - On looking further into the test it would be much easier to set up the mock metadata on individual test rather than forcing all tests to load the mock metadata, even for tests that does not require the mock metadata in order to tests its functionality.
> 
> This allows the tests to be more readable ( as only the tests that truly needs the mock metadata will load it, thereby allowing more clairty as to which field relies on the metadata response), as well as reducing the redundant code changes introduced in all the tests. We can pair on this on Monday in case you have any questions about the setup.

Sounds good. Yes, initially I was writing mockdata test for individual test cases that required mockdata then I realised that a lot of them did, so just wrote it such that it worked for all.

I had a talk with Liam on this as well, and he said to try and make the test as close as possible to how the app works in production and basically all nodes when they are clicked will call the metadata api hence I then decided to do for all.

src/components/metadata/metadata.test.js
rashidakanchwala: So previously input/output were dependent on what nodes are hidden in the flowchart. So if parameters were disabled, the input/output would not show parameters and if they were enabled the input/output would list them. This has been changed in this ticket. We no longer read from the flowchart -- hence if parameters are disabled/enabled, the output would always be the same unlike before. This is the reason I removed the test.

studioswong:
Thanks for the follow up @rashidakanchwala , I had a look at the test as well and I understand where you are coming from. 

So in short there is 2 approach to our tests:
the old way: the wrapper function would not initiate any mock calls to get metadata, and if we do want to test the display functionality of mock data, we set up a function to mock the metadata object and pass in that object as a prop. ( see test setup for test `'limits parameters to 10 values and expands when button clicked'`, lines 45-76, and `Code Toggle`, lines 79 - 107)

The new proposed way in this PR: set up the mock call to fetch metadata in wrapper function so now all tests needs to include mock metadata. 

The older approach is more explicit in the test intention ( passing in a certain bit of information to test the display), and simplier in the test setup. To enable the testing of input/output data we only need to refactor the relevant test to the setup similar to `Code toggle`.

The changes in the setup proposed by this PR is pretty much a rewrite of the tests, which I appreciate and is ok with this approach as well. However, if we were to go for this completely new set up for the tests, we essentially need to rewrite the existing tests that setup the mock call in the metadata ( the `'limits parameters to 10 values and expands when button clicked'`, `Code Toggle` tests, and any tests that adopts a similar setup) to instead return the mock data in the wrapper function ( which I see you had kind of started on that with the bull plot tests.) Currently having both setup in this test is redundant and confusing. We would also need to revisit some of the wording of the existing tests as the situation with this mock function is now different. 

In summary: either we keep existing approach and only refactor the input/output test, or we need to do a fundamental revisit of even the setup of all tests ( instead of having only done bits of it here) for this new setup. 

I am happy with either approach, so it's up to you to decide which direction you want to go, would appreciate for you to make the relevant further changes if you do decide to go the full refactor route.

src/components/metadata/metadata.test.js
```diff
@@ -323,51 +353,75 @@ describe('MetaData', () => {
...
-        const wrapper = mount({ nodeId: rabbitParamsNodeId });
+        const wrapper = mount({
+          nodeId: rabbitParamsNodeId,
+          mockMetadata: node_parameters,
+        });
         const row = rowByLabel(wrapper, 'File Path:');
         expect(textOf(rowValue(row))).toEqual(['-']);
       });
 
       it('does not display the node parameter when it is an empty object', () => {
```
studioswong, line 400: if we were to go for the new approach, this wording will now need to be revisited

rashidakanchwala:
> Thanks for the follow up @rashidakanchwala , I had a look at the test as well and I understand where you are coming from.
> 
> So in short there is 2 approach to our tests:
> the old way: the wrapper function would not initiate any mock calls to get metadata, and if we do want to test the display functionality of mock data, we set up a function to mock the metadata object and pass in that object as a prop. ( see test setup for test `'limits parameters to 10 values and expands when button clicked'`, lines 45-76, and `Code Toggle`, lines 79 - 107)
> 
> The new proposed way in this PR: set up the mock call to fetch metadata in wrapper function so now all tests needs to include mock metadata.
> 
> The older approach is more explicit in the test intention ( passing in a certain bit of information to test the display), and simplier in the test setup. To enable the testing of input/output data we only need to refactor the relevant test to the setup similar to `Code toggle`.
> 
> The changes in the setup proposed by this PR is pretty much a rewrite of the tests, which I appreciate and is ok with this approach as well. However, if we were to go for this completely new set up for the tests, we essentially need to rewrite the existing tests that setup the mock call in the metadata ( the `'limits parameters to 10 values and expands when button clicked'`, `Code Toggle` tests, and any tests that adopts a similar setup) to instead return the mock data in the wrapper function ( which I see you had kind of started on that with the bull plot tests.) Currently having both setup in this test is redundant and confusing. We would also need to revisit some of the wording of the existing tests as the situation with this mock function is now different.
> 
> In summary: either we keep existing approach and only refactor the input/output test, or we need to do a fundamental revisit of even the setup of all tests ( instead of having only done bits of it here) for this new setup.
> 
> I am happy with either approach, so it's up to you to decide which direction you want to go, would appreciate for you to make the relevant further changes if you do decide to go the full refactor route.

Thanks @studioswong for flagging that. I did further refactoring of the metadata-test so that all the tests are consistent

studioswong (approved):
Thanks for making the suggested changes to the tests :)

PR #500, opened 2021-07-13
Fix object check when displaying metadata panel

limdauto:
## Description

Currently the object check will pass a `null` value as object because `typeof null === "object"`, so it will try to render the `null` value as an object.

This fix will make sure `null` doesn't pass the object check and will get display as an empty primitive, i.e. `-`, as before. 

## Development notes

I think we should extract the type checking out into utility functions so we can unit tests but not doing it to avoid blowing the scope of change for the release.

That, or we use lodash utility functions which are well tested.

## QA notes

I amended the mock data to add a realistic condition where the backend would return `null` for `dataset_type`. Without this fix, the test will break.

## Checklist

## Legal notice

### DIFF
--- a/src/components/metadata/metadata-row.js
+++ b/src/components/metadata/metadata-row.js
@@ -21,7 +21,7 @@ const MetaDataRow = ({
   children,
 }) => {
   const showList = Array.isArray(value);
-  const showObject = typeof value === 'object' && !showList;
+  const showObject = typeof value === 'object' && value !== null && !showList;
   return (
     visible && (
       <>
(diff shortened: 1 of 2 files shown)


src/components/metadata/metadata-row.js
```diff
@@ -21,7 +21,7 @@ const MetaDataRow = ({
   children,
 }) => {
   const showList = Array.isArray(value);
-  const showObject = typeof value === 'object' && !showList;
+  const showObject = typeof value === 'object' && value !== null && !showList;
```
studioswong, line 24: Thanks for this implementation!! I might put the null value check to the check in the return statement ( line 26 of the same file) so that way the label would not get rendered as well for null values

limdauto: Actually this is for backward-compatibility. Current behaviour in production is that if it's an empty value, e.g. `null`, it gets displayed as `-`. If we want to change that behaviour, we need to get Gab to okay first. So let's keep it backward compat for now.

studioswong (approved):
LGTM! thanks for the fix

PR #512, opened 2021-07-14
[KED-2679] Hover interaction for parameters indicator

rashidakanchwala:
## Description

Currently, when you hover over the parameter indicator attached to task nodes; the tooltip shows the name of the task node which was not very helpful to the users. 
In this ticket, we implemented a method for the tooltip to show the 
- name of the hidden parameter node if there's one parameter linked to the task node
- the count of the parameters nodes if there's more than one parameter linked to the task node. 

## Development notes

Below are the things implemented in this PR :- 

1. This ticket required refactoring some of the old code related to parameters indicator. Previously we got information about which nodes have parameters as source using a selector called getNodeswithInputParams -- now the backend provides this information and hence it is directly read from state.node.parameters. 
3. Refactored showTooltip function so instead of taking a node as the prop  it takes the name of the node. This way the function can be reused to show parameter names when parameter indicator is hovered
4. Create a mouseOver event called handleParamsIndicatorMouseOver which calls showTooltip with node parameters passed to it. If there's more than one parameter it displays the count otherwise it displays the name of the parameter. 
5. Object.keys(nodeParameters[node.id]).length> 0 is added in multiple places so that the parameters indicator is not applied to task nodes with 0 parameters that are reflected in the backend as **parameters : {}**
6.  NOTE - Flowchart tests haven't been covered as there are no tests for other flowchart event handlers. 

## QA notes

**To test for node with multiple parameter please test on node - load regressor**

## Checklist

## Legal notice


src/components/flowchart/draw.js
```diff
@@ -121,11 +121,10 @@ export const drawNodes = function (changed) {
     nodeActive,
     nodeSelected,
     hoveredParameters,
-    nodesWithInputParams,
+    nodeParameters,
```
studioswong, line 124: What is the cause behind this name change? The old name, `nodesWithInputParams`, is more intuitive than `nodeParameters`.

src/components/flowchart/index.js
```diff
@@ -418,7 +414,21 @@ export class FlowChart extends Component {
...
+
+  handleParamsIndicatorMouseOver = (event, node) => {
+    const nodeParameterNames = Object.keys(
+      this.props.nodeParameters[node.id] || {}
+    );
+    if (nodeParameterNames.length) {
+      const label =
+        nodeParameterNames.length > 1
+          ? `Parameters:${nodeParameterNames.length}`
+          : `Params:${nodeParameterNames[0]}`;
```
studioswong, line 428: do we need `Params` before the parameter name?  perhaps displaying only the parameter name for the single parameter is enough?  (cc. @GabrielComymQB )

studioswong (commented):
The above are my initial comments on skimming through the changes - will be doing a more thorough QA and will leave any further comments if I spot anything.

studioswong:
I noticed a discrepancy between the displayed name of a parameter node on the flowchart and the name of the hovered parameter ( see below). On cross-checking with the response from the BE, the BE does return the name `uplift.pdf` for the parameter name.

It looks like there might be some discrepancy between the displayed names on flowchart nodes with the response? 

image
image

Another case of the discrepancy of the displayed node name with the hover parameter name: 

image
image

Given now that we rely on the BE to pass on the parameters for the parameter hover information, we need to revisit on aligning the displayed name of the parameter nodes on the FE. 

One further thing I noticed: we have applied pretty name on the display of the node ( i.e no underscrore for spaces), and this should be aligned in the display of the parameter name on hover as well. 

image
image

studioswong (changes requested):
As with the discrepancy of the names as spotted above, we need to ensure the naming of the parameters are aligned before shipping this feature.

bru5 (approved):

studioswong (commented):
I've initially tested the new changes, looks good!

one observation: the current hover note still exists when the newparams flag is off / yellow dot is non-existant - I know we are going to delete the newparams flag in this coming release hence this will not be a problem, but given this case we would need to be mindful to only merge this in after the newparams flag is deleted ( which would likely be very soon)
image

Will let you know of any further comments on the code.

rashidakanchwala:
> I've initially tested the new changes, looks good!
> 
> one observation: the current hover note still exists when the newparams flag is off / yellow dot is non-existant - I know we are going to delete the newparams flag in this coming release hence this will not be a problem, but given this case we would need to be mindful to only merge this in after the newparams flag is deleted ( which would likely be very soon)
> image
> 
> Will let you know of any further comments on the code.

Hey good spot!! Thanks for flagging -- makes sense I will first push newparams to main and then this! Thanks

src/components/flowchart/draw.js
```diff
@@ -174,6 +174,7 @@ export const drawNodes = function (changed) {
     enterNodes
       .append('rect')
       .attr('class', 'pipeline-node__parameter-indicator')
+      .on('mouseover', this.handleParamsIndicatorMouseOver)
```
studioswong, line 177: Optional: in the event that the newparams flag still exists, we would need to cater for the `newparams` flag here as well

src/components/flowchart/index.js
```diff
@@ -418,7 +418,19 @@ export class FlowChart extends Component {
...
+  };
+
+  handleParamsIndicatorMouseOver = (event, node) => {
+    const parameterNames = this.props.nodesWithInputParams[node.id];
+    if (parameterNames) {
+      const label =
+        parameterNames.length > 1
+          ? `Parameters:${parameterNames.length}`
+          : parameterNames[0];
+      this.showTooltip(event, label);
```
studioswong, line 431: please set tests to test the different display options for the parameter nodes under `flowchart.test.js`.

rashidakanchwala: @studioswong -- thanks for flagging this. New tests have been added -- the function logic has been extracted to another function.

studioswong (approved):
LGTM 👍

PR #514, opened 2021-07-16
[KED-2700] Focus Mode MVP

studioswong:
## Description

related ticket: https://jira.quantumblack.com/browse/KED-2700

This ticket is to implement the first version of focus mode feature for modular pipelines. 

focus mode demo

## Development notes

I have implemented the following:
- addition of new 'focus-mode' icon for focus mode toggle button
- addition of handlers for handling focus mode state for the flowchart for `onChange` handlers in node-list
- new checks to enable / disable modular pipelines according to focus mode state
- handlers for handling focus mode isolated state in node-list-tree
- tests to test focus-mode isolated state

## Checklist

## Legal notice


src/actions/actions.test.js
```diff
@@ -200,7 +200,7 @@ describe('actions', () => {
     );
   });
 
-  it('should create an action to toggle an array of tags on/off', () => {
+  it('should create an action to toggle an array of modular pipelines on/off', () => {
```
studioswong, line 203: This change is irrelevant to the scope of my ticket but getting this change in to fix a wording issue I spot while working on this PR

src/components/node-list/node-list-row.test.js
```diff
@@ -19,7 +19,6 @@ describe('NodeListRow', () => {
       onMouseEnter: jest.fn(),
       onMouseLeave: jest.fn(),
       onChange: jest.fn(),
-      type: node.type,
```
studioswong, line 22: this field is not used in any of the other tests hence deleting this

src/components/node-list/node-list-row.js
```diff
@@ -57,11 +58,72 @@ const NodeListRow = memo(
...
+        return disabled;
+      }
+      return (
+        parentDisabled !== false &&
+        disabled === true &&
+        applyFocusMode === false
+      );
+    };
+
+    const [applyFocusMode, setApplyFocusMode] = useState(
```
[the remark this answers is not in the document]
limdauto: Also I don't think you need 2 levels of indirections with both `useState` and `useEffect` here. I think this use case could work with just `useCallback`:

```
const determineFocusMode = useCallback(
  () =>
    focusMode !== null &&
    type === 'modularPipeline' &&
    id === focusMode?.id,
    [focusMode, type, id]
);
const isInFocusMode = determineFocusMode();

// then in the component
...
'pipeline-row__toggle-icon--focus-checked': isInFocusMode,
```

Re-render will be triggered when either one of `focusMode, type, id` changes. No need to keep a local state in this case IMO.

Similar idea applies to the label disabled state.

src/components/node-list/node-list-row.test.js
```diff
@@ -85,6 +84,37 @@ describe('NodeListRow', () => {
         false
       );
     });
+
+    describe('focus mode', () => {
+      it('sets the visibility toggle to the checked mode when the row is selected for focusmode', () => {
+        const { props } = setupProps();
+        const wrapper = setup.mount(
+          <NodeListRow {...props} focusMode={node} type="modularPipeline" />
```
limdauto, line 92: Nit: As a public props API, I do find `focusMode={node}` a bit hard to parse. The prior art of `nodeSelected={node}` makes more sense to me. A suggestion is `nodeFocused={node}`.

src/components/node-list/node-list-row.test.js
```diff
@@ -85,6 +84,37 @@ describe('NodeListRow', () => {
         false
       );
     });
+
+    describe('focus mode', () => {
+      it('sets the visibility toggle to the checked mode when the row is selected for focusmode', () => {
```
limdauto, line 89:
```suggestion
      it('sets the visibility toggle to the checked mode when the row is selected for focus mode', () => {
```

src/components/node-list/node-list-row.test.js
```diff
@@ -85,6 +84,37 @@ describe('NodeListRow', () => {
...
+      it('sets the label to match the selected parent when the row belongs to a modular pipeline selected by focus mode', () => {
+        const selectedItem = { id: 'abc' };
+        const { props } = setupProps();
+        const wrapper = setup.mount(
+          <NodeListRow
+            {...props}
+            focusMode={selectedItem}
+            parentDisabled={true}
+            parentPipeline={'abc'}
+            disabled={true}
```
limdauto, line 109: Should we test a couple more combo of this condition? It's a complex condition in the code.

src/components/node-list/node-list-tree.js
```diff
@@ -111,16 +115,25 @@ const TreeListProvider = ({
     }
   };
 
-  const renderModularPipelines = (treeData) =>
-    treeData.children.map((node) =>
+  const renderModularPipelines = (treeData, parentStatus) => {
+    // this value is needed to determine whether the children modular pipeline belongs to a parent under focusMode
+    const status =
+      parentStatus === false && treeData.id !== 'main'
```
limdauto, line 121: Is this `main` a convention on the frontend?

src/components/node-list/node-list-row.test.js
limdauto: Do you think it's worth integration testing at the `node-list-tree.js` level as well?

src/components/node-list/node-list-tree.js
```diff
@@ -111,16 +115,25 @@ const TreeListProvider = ({
     }
   };
 
-  const renderModularPipelines = (treeData) =>
-    treeData.children.map((node) =>
+  const renderModularPipelines = (treeData, parentStatus) => {
```
limdauto, line 118: Is this status the focused status of a parent? Or the disabled status? What are the possible values?

src/components/node-list/node-list-row.js
```diff
@@ -57,11 +58,72 @@ const NodeListRow = memo(
...
+        : checked
+        ? visibleIcon
+        : invisibleIcon;
     const isButton = onClick && kind !== 'filter';
     const TextButton = isButton ? 'button' : 'div';
 
+    const determineFocusMode = (focusMode, type, id) =>
+      focusMode !== null && type === 'modularPipeline' && id === focusMode?.id;
+
+    const determineDisabledLabel = (
```
limdauto, line 77: Sorry I'd need some help understanding this bit:

Am I correct in reading this condition as a node row label is enabled if and only if either the application isn't in focused mode or its parent or itself is the currently the focused modular pipeline?

limdauto (commented):
Thanks @studioswong I have tested this out and the behaviour looks good to me. A bit odd that we have to explicitly get out of focus mode before we can focus on another modular pipeline.

I have a few other comments, mostly questions to understand this better and some suggestions around styles and `useCallback`. Happy to approve once we are aligned on those.

src/components/node-list/node-list-tree.js
studioswong: yes, it is used to represent the top level representation of the tree ( as designated in the `node-list-items` selectors)

src/components/node-list/index.js
```diff
@@ -47,10 +47,11 @@ const NodeListProvider = ({
   onToggleTagFilter,
   onToggleModularPipelineActive,
   onToggleTypeDisabled,
+  onToggleModularPipelineFilter,
   modularPipelines,
 }) => {
   const [searchValue, updateSearchValue] = useState('');
-
+  const [focusMode, setFocusMode] = useState(null);
```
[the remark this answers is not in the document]
studioswong: In terms of `runFocusMode`, unfortunately I am not sure of the context and have not heard of that feature - is that for experimental tracking? if so I can keep in mind in my next task when I move the focusMode variable to the global state level.

In terms of the difference between the 3, from what I know: 

- `Select` is not a concept widely used within our codebase for variable naming; 

- `Filter` is literally what it is, and is used for most actions / variable naming that relates to the functionality performed under the filter panel. 

- 'Focus' is a relatively new concept which is only introduced with focus mode in this PR - it seems like there is potential additional usage of 'focus' with experimental tracking, which I personally am not aware of as of the details yet when setting this PR up. 

You have brought up a great point in terms of what's pontentially coming up next for focusmode - perhaps it would be great if we can catch up to align on what's in store for focuMode so I can plan my variable naming better for this and my upcoming PR?

src/components/node-list/node-list-items.js
```diff
@@ -162,8 +162,8 @@ export const getFilteredModularPipelineItems = createSelector(
         selected: false,
         faded: false,
         visible: true,
-        disabled: false,
-        checked: true,
+        disabled: focusMode !== null && focusMode?.id !== modularPipeline.id,
```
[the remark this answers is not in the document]
studioswong: we need to check for `null` value given that this variable is set as `null` by default, and we need to enforce the first check to ensure that there is a selected focusMode before conducting the second check - i.e without the `focusMode !== null` check then all modular pipelines would be disabled by default when not in focusMode, which is not what we want. ( You can also verify this by taking this test out and you will see what I mean 😀 )

Thanks for raising this nonetheless, this would've been a valid point otherwise

src/components/node-list/node-list-row.test.js
studioswong: Thanks for raising this - indeed that's a great point! 

In terms of testing on the `node-list-tree`, it's a great point that you raised this - I had looked into setting tests for that level as well, however there isn't anything to test on that level given that we are only passing in the `focusMode` data through that level to `node-list-row`, hence I didn't add any tests to that. 

To elaborate on the further tests needed to set up, there are 2 few key things in this PR that we need to set up tests for to ensure its functionality:

1. ensuring all actions, reducers and selectors relating to the selection of an individual modular pipeline works - those tests have all been previously set up when we set up the modular pipeline as tags concept. 
2. the behaviour of the individual row on the tree - the main addition of a condition where the row is `disabled` while it needing to appear in a 'non-disabled' way when it is part of a selected modularpipeline under focus mode - those are the tests set up here. 

Given that most of the tests have been set up previously, hence this is why the tests set up here does seem quite minimal 😁 However, I can  perhaps look into adding 1-2 more cases for node-list-row to cross test the non-selected focus mode scenerio as a cross reference to the current tests. 

Heads up: in my next PR I will set up a lot more selectors and actions for focusmode to include the new graph nodes selection for the related dataset nodes, which by then I do have a bunch of new updated tests coming up in the next PR once we add in a series of new actions to call the related database notes for the flowchart 😁 By then you would be able to see how the entire feature would be tested end to end.

src/components/node-list/node-list-tree.js
studioswong: yes it is the focused Status of the parent pipeline, which is used to determine the display state of the pipeline - the possible values are only `true` or `false`

- if any higher parent is selected for focusMode, the parentStatus will remain true for all child modular pipelines down all levels. 
- If this level is true, then the label of that modular pipeline row will be rendered in a 'non-disabled' way while still having disabled status

src/components/node-list/node-list-row.js
studioswong: `determineFocusMode` is the variable used to determine if the modular pipeline row _IS_ the selected modular pipeline; `determineDisabledLabel` is used to determine if the modular pipeline row belongs to part of a selected pipeline but is not the selected pipeline itself. 

and yes, your reading of this condition is correct 💯

src/components/node-list/node-list-row.js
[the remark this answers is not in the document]
studioswong: good suggestion, happy to explore and try this out instead of the `useState` and `useEffect` pattern and let you know if it works

src/components/node-list/node-list-row.js
```diff
@@ -57,11 +58,72 @@ const NodeListRow = memo(
...
+      parentPipeline,
+      parentDisabled,
+      applyFocusMode,
+      disabled
+    ) => {
+      if (parentPipeline === 'main') {
+        return disabled;
+      }
+      return (
+        parentDisabled !== false &&
```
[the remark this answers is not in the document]
studioswong: yes it is - we are checking that `parentDisabled` is not `true` or `undefined` ( which happens in the case of rows representing nodes instead of modular pipelines). The parentDisabled value is assigned `false` by default for modular pipelines hence using this check allows us to cater for both the `true` or `undefined` cases as well.

src/components/node-list/node-list-items.js
```diff
@@ -149,8 +149,8 @@ export const getFilteredModularPipelines = createSelector(
  * @return {array} Node list items
  */
 export const getFilteredModularPipelineItems = createSelector(
```
rashidakanchwala, line 151: Hey, I just realised that node-list-items contains a lot of selectors. Shouldn't selectors be in the selectors folder of the project and imported to components as required. Why is it not this case here :S

src/components/node-list/index.js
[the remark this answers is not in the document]
studioswong: P.S @limdauto , this local state will actually be removed in my next iteration as I move this state to a global level. I am now aware of the situation with the potential use case of focusmode beyond modular pipelines, which I will name the global state accordingly. In the meantime, to save time in editing back and forth, I will keep the naming as is ( which this variable will not exist in my next iteration.)

src/components/node-list/node-list-row.js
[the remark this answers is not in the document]
studioswong: I've updated to use `useCallback` instead and it works - thanks for the suggestion!

src/components/node-list/node-list-row.test.js
studioswong: @limdauto I've added a bunch of tests to illustrate all the possible scenerios for node-list-row under this new setup in my latest commit

src/components/node-list/node-list-items.js
studioswong: yes, this is indeed a problem that had been raised multiple times - this has been set up historically. `node-list-items` are the files that is used to contain all selectors related to the sidebar. 

The main reason behind this seperation, I believe, is that this set of selectors:
a) are uniquely arranged just for the use of the node-list component rather than pertaining the selection of a more 'general' data type ( tags, modular-pipeline, etc)
b) the selectors feed in from a 'state' that is an object passed in from node-list rather than from the global state object

I also had the same confusion as to why this is arranged here, however on having developed extensively on the sidebar I do find that this arrangement makes sense. You can treat this as a series of util functions of the sidebar that makes uses of the selector setup to allow constant updates for all of its related selectors. 

FYI - In terms of this question itself,  this is out of scope of this PR and would be better raised in a seperate occasion.

rashidakanchwala: Thanks for the clarification Susanna. 
I think we should set norms on raising questions in and out or PR because I always get asked questions that are outside the scope of the project in my PR's as well. Maybe next retro.

src/components/node-list/index.js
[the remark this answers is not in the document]
limdauto: Thank you for explaining. Yes, happy to keep naming the same in this PR. I will raise this with design as a potential design debt.

limdauto (approved):
Thank you for helping me understand the details. This looks good to me now. 💯

PR #530, opened 2021-08-05
[KED-2770] - Expose related dataset nodes for focus mode flowchart

studioswong:
## Description

related ticket: https://jira.quantumblack.com/browse/KED-2770

This is the work to expose the related input and output nodes and display them with a dotted outline and edges. ( see below). Those nodes are displayed for indicative purposes only, where all typical click interations for nodes on the flowchart and sidebar will remain disabled for those input and output nodes. 

image

image

image

## Development notes

The main changes to enable this feature includes:
- updating the `getNodeDisabledModularPipeline` selector so that it will check for input and output nodes related to the selected modular pipeline and enable that node
- constructing selectors to return the list of input / output nodes and edges for reference
- refactoring the `focusMode` state from a local state to a global state value stored under the `visible` prop
- refactoring node-list to take in the new global state for focusMode and the related selectors for the input / output nodes and edges
- added in new actions to allow the update of the focusMode global state
- adding in the new css classes for the new display of those input / output nodes
- adding in new changes for the flowchart draw function to assign the newly added css classes for those input / output nodes on the flowchart
- Adding in checks for the sidebar node-list to check for these input / output nodes in focus mode and ensure the nodes remains disabled on the sidebar
- added all relevant tests for the newly added selectors, actions, and checks throughout the codebase

I have also added in some minor refactor of the focus mode icon. 

Please note that, due to the latest decision of removing the newparams flag, I have removed all considerations of styling for when `newparams` flag is off ( i.e old version of parameters). 

## QA notes

Please play around with the focus mode and toggle between filters to test the display of those input / output nodes. 

## Checklist

## Legal notice


src/selectors/nodes.test.js
```diff
@@ -370,4 +376,17 @@ describe('Selectors', () => {
       expect(Object.keys(nodesWithInputParams)).toEqual(parameterNodesID);
     });
   });
+
+  describe('getInputOutputDataNodes', () => {
+    it('includes input output ndoes related to a modular pipeline in the returned object', () => {
```
rashidakanchwala, line 381: typo - nodes

limdauto: You can use github suggestion to suggest changes for these small issues like this:

```suggestion
    it('includes input output nodes related to a modular pipeline in the returned object', () => {
```

src/components/flowchart/draw.js
```diff
@@ -301,12 +320,17 @@ export const drawEdges = function (changed) {
   const exitEdges = this.el.edges.exit();
   const allEdges = this.el.edges.merge(enterEdges).merge(exitEdges);
 
-  if (changed('edges')) {
-    enterEdges
-      .append('path')
+  if (changed('edges', 'newParamsFlag', 'focusMode', 'inputOutputDataNodes')) {
```
limdauto, line 323: qq: do you think we should remove `mewParamsFlag` from this PR since its removal has already been merged? Or should that be a follow up PR?

src/components/flowchart/draw.js
```diff
@@ -301,12 +320,17 @@ export const drawEdges = function (changed) {
...
+    enterEdges.append('path');
+    allEdges
+      .select('path')
       .attr('marker-end', (edge) =>
         edge.sourceNode.type === 'parameters'
-          ? `url(#pipeline-arrowhead--accent)`
+          ? focusMode !== null && inputOutputDataEdges[edge.id]
```
limdauto, line 329: nit: I think giving `focusMode !== null && inputOutputDataEdges[edge.id]` a name might help with this block as it's repeated twice in this nested tenary operator. You can also reuse that variable further down as well.

src/components/flowchart/draw.js
```diff
@@ -343,11 +367,33 @@ export const drawEdges = function (changed) {
...
+      'inputOutputDataEdges'
+    )
+  ) {
     allEdges
       .classed(
         'pipeline-edge--parameters',
-        (edge) => edge.sourceNode.type === 'parameters'
+        (edge) =>
+          edge.sourceNode.type === 'parameters' &&
+          !inputOutputDataEdges[edge.id]
```
limdauto, line 385: Sorry I don't understand why this condition `!inputOutputDataEdges[edge.id]` is needed here. Is it to fix the problem you identified the other day?

src/components/flowchart/flowchart.test.js
```diff
@@ -281,6 +281,60 @@ describe('FlowChart', () => {
...
+          '09f5edeb': { id: '09f5edeb', name: 'Bear' },
+        }}
+      />
+    );
+    expect(wrapper.render().find('.pipeline-node--dataset-input').length).toBe(
+      2
+    );
+  });
+
+  it('applies pipeline-edge--data--input class to input dataset edges under focus mode', () => {
```
limdauto, line 307: Nit: shouldn't this modifier be named as `pipeline-edge--dataset--input` to be consistent with `pipeline-node--dataset-input` (or the other way around)? Same comment about `pipeline-edge--parameters--input` and `pipeline-node--parameter-input` (although I have a feeling this naming is intentional because pipeline-edge might have more than 1 parameters, right?)

src/selectors/nodes.test.js
studioswong: Thanks both - great spot! I've updated the typo in my latest commit

src/components/flowchart/flowchart.test.js
```diff
@@ -361,6 +415,9 @@ describe('FlowChart', () => {
       visibleSidebar: expect.any(Boolean),
       visibleCode: expect.any(Boolean),
       visibleMetaSidebar: expect.any(Boolean),
+      inputOutputDataNodes: expect.any(Object),
+      inputOutputDataEdges: expect.any(Object),
+      focusMode: expect.any(Object),
```
limdauto, line 420: This set of tests is awesome. Very clear!!! 🏆

src/selectors/disabled.js
```diff
@@ -37,33 +37,92 @@ export const getNodeDisabledTag = createSelector(
...
+          if ((isInput || isOutput) && isMPEdge === false) {
+            isMPEdge = true;
+          }
+        });
+        return !isMPEdge;
       }
-      return true;
+
+      // Hide nodes that don't have at least one modular pipeline filter enabled
+      return !isNodeOfActiveModularPipeline(
```
limdauto, line 121: This seems duplicated with the earlier `if` check. Should we short circuit by moving it up instead?

```js
if (!isNodeOfActiveModularPipeline(
      nodeModularPipelines,
      nodeID,
      modularPipelineEnabled
    ) || (nodeType[nodeID] === 'task') {
   return false;
}
```

limdauto (approved):
This PR is 🔥 Everything seems very clear to me. Amazing work @studioswong Just a few minor nitpicks. 

I will do a bit more testing but the code looks good to me.

rashidakanchwala (approved):
Hi Susanna, Great work! it work's well for me!

src/components/flowchart/draw.js
studioswong: I can remove this in this PR :) thanks for the reminder!

src/components/flowchart/draw.js
studioswong: sounds good, I'll extract that out.

bru5 (commented):
Works really nicely! Great stuff 👏

Still a bit more for me to check out, but for now I've left some suggestions if you're interested, mostly around simplifying a few parts, if it makes sense.

src/components/flowchart/draw.js
studioswong: this condition is needed to prevent applying this class to parameter nodes that are also input/output nodes - if this check is not applied this will create potential conflicts with the css class that will be applied to parameter nodes that are part of input/output. ( i.e taking this check out will result in some input/output parameter nodes not adopting the `pipeline-edge--parameters--input` class. )

src/components/flowchart/flowchart.test.js
studioswong: sure, I can rename this as `pipeline-edge--dataset--input` to align this with the naming for the input nodes css class

src/selectors/disabled.js
studioswong: Thanks for the suggestions - I had a look into this, and while the arrangement of the returned statement would stand as is (i.e we still need to return the result of the check and this cannot be combined with the check for the task node), I had extracted the check for `isNodeOfActiveModularPipeline` out to a variable so the check will not be duplicated and allows more readability. I had also added further comments to explain the exceptions that the `if` statements filters out before returning the result of this check as the final returned value.

src/components/node-list/index.js
```diff
@@ -84,7 +100,11 @@ const NodeListProvider = ({
     if (isGroupType(item.type) || isModularPipelineType(item.type)) {
```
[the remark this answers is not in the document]
studioswong: I had a look at this, and given that both conditions calls `onGroupItemChange(item, checked);`, I think it is still better to combine that into a single if statement. The nesting is still readable, and I would rather avoid duplicate code.

src/components/node-list/node-list-items.js
```diff
@@ -269,14 +274,16 @@ export const getFilteredNodeItems = createSelector(
           const disabled =
             node.disabled_tag ||
             node.disabled_type ||
-            node.disabled_modularPipeline;
+            node.disabled_modularPipeline ||
+            (focusMode !== null && !!inputOutputDataNodes[node.id]);
```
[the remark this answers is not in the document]
studioswong: I am glad you raised this question - I am using `!!inputOutputDataNodes[node.id]` to force `disabled` to return still be assigned under a boolean value under this check. If we use `inputOutputDataNodes[node.id]` the check will return an error. ( you can also verify this by trying this out).

src/selectors/disabled.js
```diff
@@ -37,33 +37,92 @@ export const getNodeDisabledTag = createSelector(
...
-        // Hide task nodes that don't have at least one modular pipeline filter enabled
-        return !nodeModularPipelines[nodeID].some(
-          (modularPipeline) => modularPipelineEnabled[modularPipeline]
+      if (
+        !isNodeOfActiveModularPipeline(
+          nodeModularPipelines,
+          nodeID,
+          modularPipelineEnabled
+        ) &&
+        (nodeType[nodeID] === 'parameters' || nodeType[nodeID] === 'data')
```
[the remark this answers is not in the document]
studioswong: I've refactored this check into a separate constant variable - please see my latest commit

src/selectors/edges.js
```diff
@@ -48,10 +66,17 @@ export const getTransitiveEdges = createSelector(
           return;
         }
         const target = edgeTargets[edgeID];
+
+        // Further filter out connections between indicative input / output nodes under focus mode
+        const isNotInputEdge =
```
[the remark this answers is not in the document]
studioswong: It is a performance thing as checking `isNotInputEdge` is more straight forward and has less cases rather than the positive situation ( this is apparent if you read through the condition - checking for isInputEdge has much more possibility rather than checking for `isNotInputEdge`. ) 

Out of curiosity, I wonder what is the upside of the refactor work for flipping this to positive?

studioswong:
Thanks all for the comments! I've gone through all your comments and made the suggested changes.

studioswong (commented):
- Please add a set of tests for the newly introduced selectors under  `contracted.js` , as well as the other modified selectors, such as `getVisibleLayerIDs` which is now under `contracted.js` instead of `disabled.js`

PR #532, opened 2021-08-10
[KED-2208] Rules for JS linter

rashidakanchwala:
## Description

In this ticket, we have implement a few new js lint rules and applied code changes to ensure it meets the new rules. 

## Checklist

## Legal notice


src/utils/graph/solver.js
```diff
@@ -20,8 +20,8 @@ import { Solver, Variable } from 'kiwi.js';
  */
 export const solveLoose = (constraints, iterations, constants) => {
   for (let i = 0; i < iterations; i += 1) {
-    for (const co of constraints) {
-      co.base.solve(co, constants);
+    for (const constrain of constraints) {
```
limdauto, line 23:
```suggestion
    for (const constraint of constraints) {
```
Need to change the one below too

limdauto (approved):
LGTM. Couple of minor comments!

src/utils/random-utils.js
```diff
@@ -48,9 +48,9 @@ export const getSeedFromURL = () => {
 
 /**
  * Get an array of numbers
- * @param {number} n Length of the array
+ * @param {number} num Length of the array
```
bru5, line 51: I think since the description is 'Length of the array' then maybe better name for this one is 'length'?

src/utils/random-utils.js
```diff
@@ -67,15 +67,15 @@ const randomUtils = () => {
 
   /**
    * Get a random number between 0 to n-1, inclusive
-   * @param {number} n Max number
+   * @param {number} num Max number
```
bru5, line 70: This one could be 'max'?

src/utils/random-utils.js
```diff
@@ -95,8 +95,8 @@ const randomUtils = () => {
    * @param {number} n Number of words in the name
    * @param {string} join The character(s) used to join each word
    */
-  const getRandomName = (n, join = '_') =>
-    getNumberArray(n)
+  const getRandomName = (num, join = '_') =>
```
bru5, line 98: Maybe 'length'?

src/utils/random-utils.js
```diff
@@ -106,16 +106,16 @@ const randomUtils = () => {
    * @param {array} arr List from which to choose
    * @param {number} n Number of items to select
    */
-  const getRandomSelection = (arr, n) => {
-    const result = new Array(n);
+  const getRandomSelection = (arr, num) => {
```
bru5, line 109: Also maybe 'length' here too?

src/utils/graph/common.js
```diff
@@ -1,6 +1,8 @@
 /**
  * PI / 2 constant
  */
+
+/* eslint-disable id-length */
 export const halfPI = Math.PI * 0.5;
```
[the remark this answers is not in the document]
bru5: Yeah fair point I think that's the standard elsewhere in the project and now style guide, so this can be `HALF_PI` in that case

bru5 (changes requested):
This is a great step forward in standardising things as well as for smoother future reviews too 👏

Great stuff. I think before the merge though, I left a few suggestions where I think the name could be a little more specific but other alternatives can also work I'm sure.

src/utils/random-utils.js
rashidakanchwala: thanks for highlighting these. makes sense... i changed this one to numWords

src/utils/random-utils.js
rashidakanchwala: i changed it to numItems

src/utils/graph/common.js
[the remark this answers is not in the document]
rashidakanchwala: nope, the lint rule is kept here because there are a lot of two-letter words such as ax, bx in this file.

.eslintrc.json
```diff
@@ -1,6 +1,12 @@
 {
   "extends": "react-app",
   "rules": {
-    "curly": ["error"]
+    "curly": ["error"],
+    "valid-typeof": ["error"],
+    "camelcase": "error",
+    "id-length": ["error", { "min": 3, "exceptions": ["_","$","a","b","d","e","i","j","k","x","y","id","el","pi","PI"] }],
```
studioswong, line 7: I don't recall this rule being mentioned in our previous discussion - what is the premise or discussion behind adding this rule?

src/store/normalize-data.test.js
```diff
@@ -46,7 +46,7 @@ describe('normalizeData', () => {
     });
     expect(normalizeData(data).pipeline.active).toBe(undefined);
   });
-
+  /* eslint-disable id-length,camelcase */
```
studioswong, line 49: it seems like we only need to disable the `camelcase` rule here - I wonder what is the reason behind disabling `id-length` here as well?

rashidakanchwala: good spot, not required

.eslintrc.json
rashidakanchwala: hi Susanna, the rule in style-guide is not to use single-letter variable names but more meaningful names and hence we put up this rule. However, we specified the minimum length to be 3 

 But there are a lot of exception to this rule as double letter words have been used a lot in the graphing algorithm (ax,bx) so in certain places I have put up the eslint ignore rule and for some I have added them to exceptions.

src/utils/graph/common.js
```diff
@@ -1,6 +1,8 @@
 /**
  * PI / 2 constant
  */
+
+/* eslint-disable id-length */
```
studioswong, line 5: If this is applied to the entire file I recommend putting this at the top / first line of this file for clarity.

src/utils/plot-templates/dark.js
```diff
@@ -3,6 +3,8 @@ Plotly templates are defined to override user-specified styles with Kedro-viz st
 More details can be found here - https://plotly.com/python/templates/
 */
 
+/* eslint-disable id-length,camelcase */
+
```
studioswong, line 7: I notice that there is no extra line when specifying eslint disabled rules in the other files - personally I am up for an extra line after the rule given that this will impact the entire file, and should be specified differently from how we specify block comments.

If we are adopting this practice of adding an extra line after specifying the disabled eslint rules, I suggest we should add this for the other files that has also specified disabled rulesets -  either way is fine, the key is just to be consistent. :)

src/utils/random-data.js
```diff
@@ -182,6 +182,7 @@ class Pipeline {
    * @param {number} initialRank Rank index
    * @returns {object} Node object
    */
+  /* eslint-disable camelcase */
```
studioswong, line 185: see my above comment - I would suggest adding an extra line here to remain consistent with the other files.

Furthermore, is this disabled rule only intended to be applied to `createNode`, or the rest of the file from line 186 onwards? If it's the former, we will also need to add in `/* eslint-disable camelcase */` at the end of the function to ensure it does not affect the rest of the file.

studioswong (commented):
Good work Rashida!! Thanks for adding in the rules and adding in all the relevant changes to adhere to the ruleset. 

Overall this looks good, my main comment is to check the format of specifying the disabled eslint statements - I notice some of them are only declared in the middle of the file, which I am guessing some of them might only be intended to be applied to a certain function - we will need to ensure to add in the closing eslint disabled statement at the end of the function to ensure it does not affect the rest of the file.

.eslintrc.json
studioswong: yes I do see the rule is for restricting variable naming, my question relates more to the context as to why this rule was added - is this something added on a further discussion of the JS lint rules that was advised to be added in via a separate discussion after our team discussion on the JS lint rules last Monday?

Furthermore, with regards to the exceptions, I wonder what is the significance to include single letters such as `"a","b","d","e","i","j","k"` in the rule? While I understand the exception of  "x", "y", "id", etc, specifying those other letters seems to defeat the idea behind setting this rule, which is for us to create meaningful and readable variable names. 

For ESlint rules, from how I see it, should relate to general best practice that can be applied universally. If there is so many possible exceptions that can apply to this rule, then it might be worth revisiting whether this is something that we should specify as a linter rule.

rashidakanchwala: I agree. Even I am slightly confused as to what to include in the exception. I can provide reasoning behind the above but I believe we can make this rules better. Let's set up sometime with Lim/Liam to go again through it and align and agree together. I will send a MR

bru5: I think this is pretty well covered by this section of style guide here with the general reasoning behind it:

https://github.com/quantumblacklabs/kedro-viz/blob/main/STYLE_GUIDE.md#variable-naming

This lint rule of avoiding 1 or 2 letter names is to try to encourage clearer naming. 

There are fair exceptions for the cases where the naming is well established by standard practices - so `i` and `j` for loops are standard in JS, `x` and `y` for position are standard in mathematics (ideally these would be written into the lint config, but don't think comments are possible there)

bru5 (approved):

PR #538, opened 2021-08-17
[KED-2517] Contracting modular pipelines

bru5:
## Description

Implements modular pipelines nodes on the flowchart component, that represent multiple elements of a modular pipeline in a single node when collapsed using the sidebar tree controls.

Screenshot 2021-08-25 at 12 18 14

## Development notes

The majority of this work was originally implemented by Richard, but required some further work to complete that I have implemented here.

### Caveats
This work is specifically front-end based, since the user's filters can change the graph dynamically at any time, including forming graphs that are unnatural by Kedro standards, so implementation has to be more general and robust than might be expected.

### Algorithm
This PR uses an [edge contraction](https://en.wikipedia.org/wiki/Edge_contraction) based algorithm to merge adjacent nodes together resulting in a simplified final graph.

### Handling disconnected graph components
If a modular pipeline has disconnected components, then this will naturally result in two disconnected modular pipeline nodes.

### Handling external graph components
In the case that a modular pipeline links to an element external to the modular pipeline, such as a dataset from a different namespace, these elements can not merge to a single node. This naturally forms a split of the modular pipeline over two or more modular pipeline nodes, with externals inbetween.

### Handling graph cycles
In particular it's possible that an external component will form a cycle if it links back into the same modular pipeline, a practical example of which can be seen in the shuttles demo dataset, but can be specifically seen in this new minimal [test dataset](http://localhost:4141/?data=contracting).

To handle this there is a special rule ([code](https://github.com/quantumblacklabs/kedro-viz/pull/538/files#diff-7488e78efcbb12d967ddb953ce65d912fec7cc5dd0d7ac4bd3763d7f9782b8c2R131-R154)) introduced to avoid introducing cycles when contracting these cases which would otherwise not form a valid DAG as required, which could break other features such as the layout engine. 

The approach to handle them was discussed with @idanov  and @limdauto and concluded that choosing a preferred edge direction and dropping the cyclic edge was an acceptable solution when following Kedro logic.

## QA notes

A new minimal [test dataset](http://localhost:4141/?data=contracting) was added that includes a potential cycle when the modular pipeline is contracted.

**Note:** I've updated some of the existing unit tests but there are still some more to do as it stands. I've also added some tests to cover the new feature but more might be needed (especially around the sidebar UI interaction).

To manually test:

- open the browser console to check for any errors or warnings
- open a datasets `animals` and `demo` (that have modular pipeline nodes)
- on the sidebar toggle various modular pipelines open / closed
- also try toggling sub modular pipelines at various depths
- check that the output graph has contracted the closed modular pipelines in a reasonable way
- try in combination with various filters on the sidebar
- try in combination with layers enabled

## Checklist

## Legal notice


src/components/node-list/node-list-tree.js
```diff
@@ -100,6 +102,13 @@ const TreeListProvider = ({
     focusMode,
   ]);
 
+  useEffect(() => {
+    // Initial state contracts all modular pipelines
+    for (const id of modularPipelineIds) {
+      onToggleContracted(id, true);
+    }
```
studioswong, line 109: rather than looping through all the modular pipelines to contract one by one, it would be easier if we just set this in the default state when initialising the state>

src/components/node-list/node-list-tree.js
```diff
@@ -118,6 +127,23 @@ const TreeListProvider = ({
...
+  const onItemExpandToggle = (event, expandedItemIds) => {
+    // Reset clicked node
+    onToggleNodeClicked(null);
+
+    // Contract all modular pipelines that are not expanded on sidebar
+    for (const id of modularPipelineIds) {
+      if (!expandedItemIds.includes(id)) {
+        onToggleContracted(id, true);
+      }
+    }
```
studioswong, line 139: As per my comment above this would likely not be needed if all modularPipelines are set as contracted by default in the state?

src/components/node-list/node-list-tree.js
```diff
@@ -118,6 +127,23 @@ const TreeListProvider = ({
     }
   };
 
+  const onItemExpandToggle = (event, expandedItemIds) => {
```
studioswong, line 130: the `event` is not used here and can be taken out.

src/components/node-list/node-list-tree.js
```diff
@@ -194,6 +220,7 @@ const TreeListProvider = ({
       className={classes.root}
       defaultCollapseIcon={<ExpandMoreIcon />}
       defaultExpandIcon={<ChevronRightIcon />}
+      onNodeToggle={onItemExpandToggle}
```
studioswong, line 223: naming it as `onNodeToggle` could be quite confusing - this might be better to be renamed as `onPipelineToggle`?

studioswong (commented):
Looks good @bru5 , thanks for picking this up and resolving the edge case!

I've played around with this branch and overall looks good, some things I spot below that needs further fixing:

- the input/output edges for both the dataset and parameter nodes under focus mode does not pick up the right class, hence the edges does not show up as dotted edges ( see below left) - it should be styled with dotted lines as shown in the latter image. 

image

image

- Please add a set of tests for the newly introduced selectors under  `contracted.js` , as well as the other modified selectors, such as `getVisibleLayerIDs` which is now under `contracted.js` instead of `disabled.js`

src/components/node-list/node-list-tree.js
bru5: With it being the first argument I'm not sure if it can be? Can you show how you mean?

src/components/node-list/node-list-tree.js
bru5: Even with the default state set as you mentioned, I think this is still needed since the material-ui `TreeView` component holds its own internal state. I looked to see if it was possible to drive the `TreeView` state through props but it doesn't seem to expose that.

src/components/node-list/node-list-tree.js
bru5: This is the name of the prop on material-ui `StyledTreeView` component so think we'll need to keep it?

bru5:
@studioswong thanks for the review, I've got back on some of your points. As for the focus mode issue, I think that one should now be fixed (it turns out a selector it relied on was moved, so I reverted and updated both).

src/components/node-list/node-list-tree.js
studioswong: indeed - thanks for clarifying! Funny it does sound exactly like how we would name our callback prop :grin:

src/components/node-list/node-list-tree.js
limdauto: This is also needed because once user collapses a node on the sidebar, we need to dispatch an action to collapse it on the main graph as well.

src/selectors/disabled.js
```diff
@@ -169,23 +166,6 @@ export const getVisibleNodeIDs = createSelector(
   (nodeIDs, nodeDisabled) => nodeIDs.filter((id) => !nodeDisabled[id])
 );
 
-/**
- * Get a list of just the IDs for the remaining visible layers
- */
-export const getVisibleLayerIDs = createSelector(
```
limdauto, line 175: This is now moved to `src/selectors/contracted.js`

src/selectors/edges.js
```diff
@@ -101,50 +99,21 @@ export const getTransitiveEdges = createSelector(
...
-        source: edgeSources[id] || transitiveEdges.sources[id],
-        target: edgeTargets[id] || transitiveEdges.targets[id],
-      }))
-);
-
-/**
- * Obtain all the edges that belongs to input and output data
- * nodes when under focus mode.
- */
-export const getInputOutputDataEdges = createSelector(
```
limdauto, line 129: Also move to contracted selectors

limdauto:
This is superseded by #571

PR #542, opened 2021-08-23
[KED-2753] Turn on/off pretty naming

rashidakanchwala:
## Description

Pretty names is a feature that capitalises each word in Kedro-Viz so that things look "prettier" on the navigation sidebar and pipeline graph
This ticket makes it possible for users to turn off this feature  using the Settings panel so that it reflects the actual name

## Checklist

## Legal notice


src/config.js
```diff
@@ -34,6 +34,14 @@ export const flags = {
   },
 };
 
+export const settings = {
+  prettyName: {
+    name: 'Pretty Name',
+    description: 'Prettify the kedro node name',
```
studioswong, line 40: 'Prettify' sounds ambiguous - I would reword this as something along the lines of ' display a more read-able version of the node name'

studioswong (changes requested):
I've tested this feature and it works well, good work!

There are a few tests that needs to be added to supplement this feature:
- tests on the sidebar (the node-list tests) to ensure it displays the pretty and non-pretty name according to the settings. This can be set up by mocking the state of the prettyName to be passed in as a prop in the tests.
- tests for the metadata panel to ensure the name reflects according to the state. The setup will be similar to the setup above.

It looks like there is currently a failing JS test on the CI so worth double checking on this as well.

src/components/node-list/node-list.test.js
```diff
@@ -142,6 +143,64 @@ describe('NodeList', () => {
...
+        ['cat', true],
+        ['dog', true],
+        ['elephant', true],
+        ['giraffe', true],
+        ['horse', true],
+        ['pig', true],
+        ['sheep', true],
+        ['parameters', false],
+        ['params:pipeline100.data_science.plankton', false],
+        ['params:rabbit', false],
```
studioswong, line 176: This set of tests, while does test the displayed title of the wrapper, is more structured to test the `enabled` feature as set out in the other tests - we won't need to test whether they are disabled but only need to test the names, so you can take out the check for the `disabled` class here. This will also apply to the test below from lines 184 - 199 :)

rashidakanchwala: Now, i get it . i was wondering why there's true and false when i copy/pasted it from another tests. thanks I have removed it :D

src/components/metadata/metadata.test.js
```diff
@@ -66,6 +67,40 @@ describe('MetaData', () => {
       // Should show all 20 values
       expect(parametersRow().find('.pipeline-metadata__value').length).toBe(20);
     });
+
+    it('when pretty name is turned off the metadata title displays the full node name', () => {
+      const props = {
+        nodeId: salmonTaskNodeId,
```
studioswong, line 73: It would be better to set up a node that has a more distinctive difference between their pretty name and non-pretty name other than `salmon` - the `params:pipeline100.data_science.plankton` node would work much better.

studioswong (approved):
Thanks for adding the tests! I highly recommend switching to use the parameter node other than the salmon node for the metadata tests in testing the pretty name, as that node name includes the difference of capitalisation as well as the underscore, which would be much more effective in testing the pretty name feature.

limdauto (approved):
LGTM!

PR #553, opened 2021-09-03
[KED-2831] Fix focus mode error when switching pipelines

studioswong:
## Description

related ticket: https://jira.quantumblack.com/secure/RapidBoard.jspa?rapidView=525&projectKey=KED&view=detail&selectedIssue=KED-2831

This is to fix the error thrown when switching between pipelines while in focus mode. 

## Development notes

The bug is caused by the persistence of the focus mode state while switching between pipelines - the problem will be solved by resetting the value of the selected focus mode modular pipeline when switching between pipelines. 

I have added the dispatch of the relevant action, as well as the relevant test to ensure this action to reset the focus mode state gets dispatched everytime we switch between pipelines. 

## QA notes

select a modular pipeline to be under focus mode, and try switching between pipelines. The error should not happen again. 

## Checklist

## Legal notice


limdauto (approved):
This LGTM as a fix but I wonder if we can do something at a higher level, e.g. parameterise the state store based on the selected registered pipelines?

```
pipeline_a: <state>
pipeline_b: <state>
__default__: <state>
```

studioswong:
> This LGTM as a fix but I wonder if we can do something at a higher level, e.g. parameterise the state store based on the selected registered pipelines?
> 
> ```
> pipeline_a: <state>
> pipeline_b: <state>
> __default__: <state>
> ```

Heya that is an interesting suggestion! yea I see what you mean - Saving all the loaded pipeline data on a higher level in the state could even help us preserve the individual state of each pipeline ( e.g selected focus mode, etc), great idea. 

I think we haven't set it up this way previously as there isn't any particular need to preserve the app state of each pipeline, but now that we are implementing more features ( i.e experiment tracking) that suggests more complicated individual states for each pipelines that would need to be preserved, it makes sense for us to revisit the data structure of our store. This would definitely require quite a bit of rewrite, but I can factor that in for the FE architecture spike and make a new ticket for implementing this ( if needed)?

limdauto:
@studioswong 👍  agreed. Let's track this. I really want to move away from a single state store at some point.

PR #554, opened 2021-09-06
[KED-2808/2809] Display metrics as json and plot time series

rashidakanchwala:
## Checklist

## Legal notice


package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +117,17 @@ def load_data(
 
     context = load_context(project_path=project_path, env=env)
     return context.catalog, context.pipelines
+
+def load_all_versions(filepath: str) -> Dict[str, any]:
+    version_list = [x for x in Path(filepath).iterdir() if x.is_dir()]
```
MerelTheisenQB, line 122: I would recommend using clearer variable names instead of `x`, `i` etc.. In core Kedro we actually have a lint check for that. But in general it helps to understand the code more easily if proper names are given. So in this case I'd probably do `path for path in Path(filepath).iterdir() ...`

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +117,17 @@ def load_data(
...
+    versions = {}
+    for v in version_list:
```
MerelTheisenQB, line 124: Same comment here about variable names.

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +117,17 @@ def load_data(
...
+        timestamp = datetime.strptime(v.name,VERSION_FORMAT)
+        path = v / Path(filepath).name
+        with open(path) as fs_file:
+            versions.update({timestamp: json.load(fs_file)})
```
MerelTheisenQB, line 128: We're filling an empty dictionary here, so instead of using `update` I'd just do: `versions[timestamp] = json.load(fs_file)`. 

`update` is less efficient and commonly only used when updating a dictionary that already has data in it.

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +117,17 @@ def load_data(
 
     context = load_context(project_path=project_path, env=env)
     return context.catalog, context.pipelines
+
+def load_all_versions(filepath: str) -> Dict[str, any]:
```
MerelTheisenQB, line 121: Bit nitpicky, but when reading the function name and looking at the return type, I wasn't sure what kind of data I would get back from the method. I'd personally call it `load_data_for_all_versions` to make it clear we actually load the data itself and not just the versions.

package/kedro_viz/models/graph.py
```diff
@@ -33,16 +33,23 @@
...
+import pandas as pd
+from pandas.core.frame import DataFrame
+import plotly.express as px
+import plotly.io as pio 
 
 from kedro.io import AbstractDataSet
 from kedro.io.core import get_filepath_str
 from kedro.pipeline.node import Node as KedroNode
 from kedro.pipeline.pipeline import TRANSCODING_SEPARATOR, _strip_transcoding
+from kedro_viz.integrations.kedro import data_loader as kedro_data_loader
```
MerelTheisenQB, line 52: Why rename the import? Is there another type of data loader we're using? I think since we're in Kedro it speaks for itself that the data loader loads Kedro data 🙂

package/kedro_viz/models/graph.py
```diff
@@ -455,6 +468,11 @@ def __post_init__(self):
         )
 
 
+def create_metrics_plot(df: DataFrame) -> Dict[str, any]:
+            df = df.reset_index().rename(columns={'index': 'version'})
+            df = pd.melt(df, id_vars='version', var_name='metrics').sort_values(by='version')
+            return json.loads(pio.to_json(px.line(df, x="version", y='value', color='metrics')))
```
MerelTheisenQB, line 474: I don't know much about plots, so this mostly looks like magic to me 😁 
But one thing I would suggest is rather than mutating your input `df` is to create new variables. It's best practice to not mutate your input, because it can cause nasty bugs that are hard to trace and explain. So here you could do something like:
```
renamed_columns_df = ...
melted_and_sorted_df = ...
return
```

package/kedro_viz/models/graph.py
```diff
@@ -497,6 +515,20 @@ def __post_init__(self, data_node: DataNode):
             with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
                 self.plot = json.load(fs_file)
 
+        if data_node.is_metric_node():
+            if not dataset._exists():
+                return
+
```
MerelTheisenQB, line 521: Too many blank lines - I guess we might have to update the linters for viz, because we check this stuff on kedro core 😉

package/kedro_viz/models/graph.py
```diff
@@ -497,6 +515,20 @@ def __post_init__(self, data_node: DataNode):
...
+            
+
+            load_path = get_filepath_str(dataset._get_load_path(), dataset._protocol)
+            with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
+                self.metrics = json.load(fs_file)    
+            self.plot = create_metrics_plot(
+                pd.DataFrame.from_dict(
+                kedro_data_loader.load_all_versions(self.filepath),
+                orient='index'))
+
```
MerelTheisenQB, line 530: Again, remove a line here.

package/kedro_viz/models/graph.py
```diff
@@ -497,6 +515,20 @@ def __post_init__(self, data_node: DataNode):
...
+                return
+            
+
+            load_path = get_filepath_str(dataset._get_load_path(), dataset._protocol)
+            with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
+                self.metrics = json.load(fs_file)    
+            self.plot = create_metrics_plot(
+                pd.DataFrame.from_dict(
+                kedro_data_loader.load_all_versions(self.filepath),
+                orient='index'))
```
MerelTheisenQB, line 529: This is personal preference, so feel free to discard, but because of the nesting of so many methods it's a bit hard to read this line. I'd separate the call to `load_all_versions`, so call that first, assign it to a variable and put that into the `dataframe.from_dict`. So you would get:

```
metrics_data = kedro_data_loader.load_all_versions(self.filepath)
self.plot = create_metrics_plot(pd.DataFrame.from_dict(metrics_data, orient='index'))
```

MerelTheisenQB (commented):
First of all I want to say well done @rashidakanchwala! Nice and simple implementation 👏  

Most of my comments are minor and suggestions for stylistic improvements. A bigger question I have is whether we want to add a limit to the load method. I don't know if this was discussed on the ticket or not, but @limdauto had some thoughts on this when we talked about implementing this earlier on.

limdauto (approved):
This is very nice!!! Well done 👏 I think just tests are missing and we are good. Pre-approve to indicate that the approach would be my approach as well.

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +117,20 @@ def load_data(
...
+    """Load data for all versions of the dataset
+    Args:
+        filepath: the path whether the dataset is located.
+    Returns:
+        A dictionary containing the version and the json data inside each version
+    """
+    version_list = [path for path in Path(filepath).iterdir() if path.is_dir()]
+    versions = {}
+    for version in version_list:
+        timestamp = datetime.strptime(version.name, VERSION_FORMAT)
```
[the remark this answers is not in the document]
rashidakanchwala: Done. Would it be possible for kedro-core to pass the user-specified version so it doesn't break on viz?

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +120,36 @@ def load_data(
...
+    versions = {}
+    for index, version in enumerate(version_list):
+        if index == limit:
+            break
+
+        try:
+            timestamp = datetime.strptime(version.name, VERSION_FORMAT)
+        except ValueError:
+            logger.warning(
+                "Version %s is not a timestamp. Skip when loading metrics.",
```
MerelTheisenQB, line 147: This warning could be confusing, because it might be that the user has provided a timestamp which is just different from the one we're expecting. I'd suggest saying something like: "Expected timestamp of format ..."

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +120,36 @@ def load_data(
...
+        filepath: the path where the dataset is located.
+        limit: the maximum number of past versions we want to load.
+    Returns:
+        A dictionary containing the version and the json data inside each version
+    """
+    version_list = [
+        path for path in sorted(Path(filepath).iterdir(), reverse=True) if path.is_dir()
+    ]
+    versions = {}
+    for index, version in enumerate(version_list):
```
MerelTheisenQB, line 139: You can simplify this a bit by doing:
```suggestion
    for version in version_list[:limit]
```

And then you can remove the `if index == limit` below.

package/tests/test_integrations/test_data_loader.py
```diff
@@ -0,0 +1,48 @@
...
+            "projected_optimization": 0.5799845430760487,
+        },
+        datetime.datetime(2021, 9, 10, 9, 3, 23, 733000): {
+            "recommendations": 0.200383330721228,
+            "recommended_controls": 0.250479163401535,
+            "projected_optimization": 0.30057499608184196,
+        },
+    }
+    filepath = str(Path(__file__).parent.parent / "example_metrics/metrics.json")
+    assert data_loader.load_data_for_all_versions(filepath) == mock_metrics_json
```
MerelTheisenQB, line 48: It would be good if we also had a test that checked that the `limit` we set on the data loader is actually working.

package/requirements.txt
```diff
@@ -6,3 +6,5 @@ fastapi>=0.63.0, <0.67.0
 aiofiles==0.6.0
 uvicorn~=0.13.4
 watchgod~=0.7
+plotly>=4.14.3
```
MerelTheisenQB, line 9: We could make this a bit less strict and do:
```suggestion
plotly~=4.14
```
that way it would be pinned the same way as in Kedro core.

package/requirements.txt
```diff
@@ -6,3 +6,5 @@ fastapi>=0.63.0, <0.67.0
...
+pandas>=1.3.0
```
MerelTheisenQB, line 10:
```suggestion
pandas~=1.3
```

package/test_requirements.txt
```diff
@@ -13,6 +13,8 @@ pytest~=6.2.0
 pytest-cov~=2.11.1
 pytest-mock~=3.6.1
 trufflehog~=2.1.0
+plotly>=4.14.3
+pandas>=1.3.0
```
MerelTheisenQB, line 17: The same comments here for the requirements 🙂

package/tests/test_api/test_apps.py
```diff
@@ -530,3 +530,11 @@ def test_api_app_from_json_file_index(self):
         client = TestClient(api_app)
         response = client.get("/")
         assert response.status_code == 200
+
+    def test_api_app_from_fe_json_file_main_api(self):
+        print(str(Path))
```
MerelTheisenQB, line 535: Left over print statement.

package/tests/test_integrations/test_data_loader.py
```diff
@@ -0,0 +1,48 @@
...
+            "recommended_controls": 0.48332045256337397,
+            "projected_optimization": 0.5799845430760487,
+        },
+        datetime.datetime(2021, 9, 10, 9, 3, 23, 733000): {
+            "recommendations": 0.200383330721228,
+            "recommended_controls": 0.250479163401535,
+            "projected_optimization": 0.30057499608184196,
+        },
+    }
+    filepath = str(Path(__file__).parent.parent / "example_metrics/metrics.json")
```
MerelTheisenQB, line 47: It might be better if we create a temporary file that contains that data rather than adding it as a test file. This is an example of how we do that in Kedro core: https://github.com/quantumblacklabs/kedro/blob/be3022ecf9ea18f6a448ef9d2c57bd8ad089c769/tests/config/test_config.py#L84

MerelTheisenQB (commented):
Nice work on adding those tests! 🥇  A couple more comments from me, but we're super close to getting this merged!

package/tests/test_integrations/test_data_loader.py
rashidakanchwala: Thanks Merel. I am still a bit confused, happy to connect tomm if you have time to understand this better. :)

package/kedro_viz/models/graph.py
```diff
@@ -497,10 +507,52 @@ def __post_init__(self, data_node: DataNode):
             with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
                 self.plot = json.load(fs_file)
 
+        if data_node.is_metric_node():
+            if not dataset._exists() or self.filepath is None:
+                return
+
+            import pandas as pd  # pylint: disable=import-outside-toplevel
```
MerelTheisenQB, line 514: Minor: you should be able to move this disable to the top of the function and then it should apply to everything inside the function. That way you won't have to add it for multiple lines.

package/kedro_viz/models/graph.py
```diff
@@ -497,10 +507,52 @@ def __post_init__(self, data_node: DataNode):
...
 
+    @staticmethod
+    def create_metrics_plot(data_frame: DataFrame) -> Dict[str, Any]:
+        """
+        Args:
+            data_frame: dataframe with the metrics data from all versions
+        Returns:
+            a plotly line chart object with metrics data
+        """
+        import pandas as pd  # pylint: disable=import-outside-toplevel
```
MerelTheisenQB, line 542: Same here. Just move it to the top of the method.

package/tests/test_integrations/test_data_loader.py
```diff
@@ -0,0 +1,89 @@
...
+import json
+from pathlib import Path
+
+import pytest
+
+from kedro_viz.integrations.kedro import data_loader
+
+
+@pytest.fixture
+def metrics_filepath(tmpdir):
```
MerelTheisenQB, line 38: exactly what I meant with using tmpdir 😄 👏 👏

MerelTheisenQB (approved):
Awesome work @rashidakanchwala ! This is good to go from the Python side 🐍 👍

package/kedro_viz/models/graph.py
```diff
@@ -497,10 +508,57 @@ def __post_init__(self, data_node: DataNode):
...
+                return
+
+            import pandas as pd
+            from kedro.extras.datasets.tracking.metrics_dataset import MetricsDataSet
+
+            dataset = cast(MetricsDataSet, dataset)
+            load_path = get_filepath_str(dataset._get_load_path(), dataset._protocol)
+            with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
+                self.metrics = json.load(fs_file)
+            metrics_data = data_loader.load_data_for_all_versions(self.filepath)
```
limdauto, line 522: I'm a bit concerned that we call the API `load_data_for_all_versions` but it has a hidden implicit limit of 10. I'm also concerned that the underlying implementation is limited to just JSON but the interface is generic. IMO we can improve the interface this way:

```python
data_loader.load_data_for_multiple_versions(dataset=dataset, num_versions=10)
```

Then all of the path wranging will happen inside `load_data_for_multiple_versions` as well as the load operation will use the native `dataset.load`. For each version, you can instantiate a new dataset object and load it this way:

```python
dataset_class = type(dataset)
for version in versions:
    dataset_instance = dataset_class(filepath=filepath, version=version)
    data = dataset_instance.load()
```

Something like this. Let me know if you want to pair on this.

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -114,3 +120,33 @@ def load_data(
...
+        except ValueError:
+            logger.warning(
+                """Expected timestamp of format YYYY-MM-DDTHH:MM:SS.ffffff.
+                 Skip when loading metrics."""
+            )
+            continue
+        else:
+            path = version / Path(filepath).name
+            with open(path) as fs_file:
+                versions[timestamp] = json.load(fs_file)
```
limdauto, line 151: We will need to catch an exception here as well if the data can't be loaded with `json`.

package/test_requirements.txt
```diff
@@ -13,6 +13,8 @@ pytest~=6.2.0
 pytest-cov~=2.11.1
 pytest-mock~=3.6.1
 trufflehog~=2.1.0
+plotly~=4.14.3
+pandas~=1.3.0
```
limdauto, line 17: You don't need it here. `test_requirements` automatically uses `requirements.txt`.

package/tests/test_api/test_apps.py
```diff
@@ -530,3 +530,10 @@ def test_api_app_from_json_file_index(self):
         client = TestClient(api_app)
         response = client.get("/")
         assert response.status_code == 200
+
+    def test_api_app_from_fe_json_file_main_api(self):
```
limdauto, line 534: Sorry what are we testing with this one?

package/kedro_viz/models/graph.py
```diff
@@ -497,10 +508,57 @@ def __post_init__(self, data_node: DataNode):
...
+            if not dataset._exists() or self.filepath is None:
+                return
+
+            import pandas as pd
+            from kedro.extras.datasets.tracking.metrics_dataset import MetricsDataSet
+
+            dataset = cast(MetricsDataSet, dataset)
+            load_path = get_filepath_str(dataset._get_load_path(), dataset._protocol)
+            with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
+                self.metrics = json.load(fs_file)
```
limdauto, line 521: I think if you replace these lines with just

```python
self.metrics = dataset.load()
```

it will work the same way, i.e. always load the latest version of the dataset, and it's more reliable because it's native Kedro API. We couldn't use it for plotly because plotly was asymmetrical.

limdauto (commented):
This generally looks good to me. I have some comments around the python API. Let me know if you want to pair to resolve them.

package/tests/test_api/test_apps.py
rashidakanchwala: good spot. Sorry this was me trying something for another ticket, I will remove it.

package/kedro_viz/models/graph.py
rashidakanchwala: I get the error -  Loading not supported for `MetricsDataSet`

MerelTheisenQB: @limdauto after our discussions we decided not to implement the `load` on metrics dataset.
https://github.com/quantumblacklabs/private-kedro/blob/master/kedro/extras/datasets/tracking/metrics_dataset.py#L107-L108

rashidakanchwala: Yeah, I got that error.

src/selectors/metadata.js
```diff
@@ -56,6 +58,12 @@ export const getClickedNodeMetaData = createSelector(
     if (!node) {
       return null;
     }
+    //rounding of metrics data
+    nodeMetrics[node.id] &&
+      Object.entries(nodeMetrics[node.id]).forEach(([key, value]) => {
+        nodeMetrics[node.id][key] = Math.round(value * 100) / 100;
```
studioswong, line 64: I see that we are rounding this up to 2 decimal values - for displaying rounded up numbers in our codebase we usually use `toFixed`  rather than using `Math.round`  ( which in our case, would be `+value.toFixed(2)` to eliminate the unnecessary zeros).

In terms of accuracy, the results of both are very similar ( both `toFixed` and `Math.round` would still throw oddities for certain edge cases), however `toFixed(2)` is a much more readable notation.

studioswong (commented):
Great work @rashidakanchwala ! I've tested it and it works perfectly.

My main comment relates to how we structure the tests for the dataset nodes under `metadata.test.js`, now that we have a extended set of different dataset nodes. As per my comment below, this is a good chance for us to rearrange the test suite so that it presents a clearer structure to arrange the tests for those 4 types of dataset nodes. 

The structure would be something like this:

``` 
    describe('different types of dataset nodes', () => {
        describe('dataset nodes', () => {
           // tests for testing normal dataset node functions, such as display dataset node names, icons, etc
        },
         describe('plot dataset node', () => {
           // only add the  addition tests for the plot
        },
          describe('transcoded dataset node', () => {
           // only add the tests for testing addditional functionality for transcoded dataset node 
        },
          describe('metrics data', () => {
           // only add the metrics data and plot test
        },
  })
```


@limdauto given the improvements this could bring in maintaining the metadata test files ( it currently has too many redundant tests making it very hard to maintain), I would highly suggest getting this fix in as part of completing this feature before our release.

src/selectors/metadata.js
rashidakanchwala: @studioswong  - toFixed() returns a string and it messes up the way the data is represented on the json tree and can confuse the user hence it was decided to use Math.round() instead.

src/components/metadata/metadata.test.js
```diff
@@ -449,6 +450,88 @@ describe('MetaData', () => {
     });
   });
 
+  describe('Metrics dataset nodes', () => {
```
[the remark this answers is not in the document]
rashidakanchwala: yes, that makes sense. I will create another ticket to refactor metadata.tests and we can work on it in the future.

PR #573, opened 2021-09-27
[KED-2870] Fix : Metrics Tree doesn't pick up the latest kedro run metrics data

rashidakanchwala:
## Description

This issue occurs when you run kedro viz and kedro run simultaneously on two terminals. After your last kedro run, if you refresh the browser running kedro viz you will notice that while the plot gets updated with the latest. The json tree with metrics data does not. 

## Development notes

Initially, we would get metrics data using get_load_path function from kedro-core which would load the most recent version -- but this does not happen in the above settings. 

So in this ticket we change that, now we iterate through the metrics json directory and get the latest one and load the json from it. 

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/models/graph.py
+++ b/package/kedro_viz/models/graph.py
@@ -514,9 +514,7 @@ class DataNodeMetadata(GraphNodeMetadata):
             dataset = cast(MetricsDataSet, dataset)
             if not dataset._exists() or self.filepath is None:
                 return
-            load_path = get_filepath_str(dataset._get_load_path(), dataset._protocol)
-            with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
-                self.metrics = json.load(fs_file)
+            self.metrics = self.load_latest_metrics_data(self.filepath)
             metrics_data = self.load_metrics_versioned_data(self.filepath)
             if not metrics_data:
                 return
@@ -527,6 +525,19 @@ class DataNodeMetadata(GraphNodeMetadata):
         # Run command is only available if a node is an output, i.e. not a free input
         if not data_node.is_free_input:
             self.run_command = f'kedro run --to-outputs="{data_node.full_name}"'
+
+    @staticmethod
+    def load_latest_metrics_data(filepath: str) -> Dict:
+        """Load data for latest versions of the metrics dataset
+        Args:
+            filepath: the path whether the dataset is located.
+        Returns:
+            A dictionary containing json data for the latest version
+        """
+        latest_version = sorted(Path(filepath).iterdir(), reverse=True)[0]
+        path = latest_version / Path(filepath).name
+        with open(path) as fs_file:
+            return json.load(fs_file)
 
     @staticmethod
     def load_metrics_versioned_data(


package/kedro_viz/models/graph.py
```diff
@@ -27,6 +27,7 @@
 # limitations under the License.
 """`kedro_viz.models.graph` defines data models to represent Kedro entities in a viz graph."""
 # pylint: disable=protected-access
+# pylint: disable=logging-fstring-interpolation
```
limdauto, line 30: This is actually a good linting error. We shouldn't disable it.

rashidakanchwala: Yes, I will remove this as I am also no longer using this.

package/kedro_viz/models/graph.py
```diff
@@ -528,6 +526,22 @@ def __post_init__(self, data_node: DataNode):
         if not data_node.is_free_input:
             self.run_command = f'kedro run --to-outputs="{data_node.full_name}"'
 
+    @staticmethod
+    def load_latest_metrics_data(dataset: Any) -> Optional[Dict[Any, Any]]:
+        """Load data for latest versions of the metrics dataset
+        Args:
+            dataset: the latest version of the metrics dataset
```
MerelTheisenQB, line 533: I'm a little bit confused about what the input is here. Is it the actual dataset or just a version? And could we make the type more specific, it should only ever be `MetricsDataSet` that gets processed here right? For me personally, that already makes it clearer what this method does.

rashidakanchwala: You are right!! Initially I had it as MetricsDataSet but it kept failing circleCI then I tried doing forward reference "Metrics Dataset" but that also didn't work so I change it to Any as a last resort just so it could pass circleCI. 
I can share with you the fail circleCI error if u have time to help investigate

package/kedro_viz/models/graph.py
```diff
@@ -528,6 +526,22 @@ def __post_init__(self, data_node: DataNode):
...
 
+    @staticmethod
+    def load_latest_metrics_data(dataset: Any) -> Optional[Dict[Any, Any]]:
+        """Load data for latest versions of the metrics dataset
+        Args:
+            dataset: the latest version of the metrics dataset
+        Returns:
+            A dictionary containing json data for the latest version
+        """
+        pattern = str(dataset._get_versioned_path("*"))
```
limdauto, line 545: Can you put a comment here linking to the place on kedro core where the cache happens and explains that we have to do this, not relying on core's native API because it has a cache?

package/kedro_viz/models/graph.py
limdauto: Fixed

package/tests/test_models/test_graph/test_graph_nodes.py
```diff
@@ -551,6 +553,15 @@ def test_load_metrics_versioned_data(self, metrics_filepath):
...
         )
 
+    def test_load_latest_metrics(self, metrics_filepath):
+        dataset = MetricsDataSet(filepath=f"{metrics_filepath}")
+        mock_metrics_json = {
+            "recommendations": 0.200383330721228,
+            "recommended_controls": 0.250479163401535,
+            "projected_optimization": 0.30057499608184196,
+        }
+        assert DataNodeMetadata.load_latest_metrics_data(dataset) == mock_metrics_json
```
limdauto, line 563: would it be possible to write a test to check that if we write another version of this dataset and load it again, it picks up the change?

limdauto (commented):
2 minor comments other wise lgtm 🌮

PR #576, opened 2021-09-29
[KED-2872] Create global toolbar UI

tynandebold:
## Description

This PR adds a new, left-aligned, vertical global toolbar, which is the first step in what will eventually become the full Experiment Tracking view and feature. 

## Development notes

image

I've added the global toolbar, moved the relevant icon buttons to it, added the Kedro logo and the two icon buttons for the respective pages that will be created soon, and created new tests and updated the existing ones.

I'm not thrilled with the naming for some of my class names, e.g. `pipeline-global-control-toolbar`. Since we're now introducing a new paradigm into the application with Experiment Tracking, perhaps we no longer need to prefix all class names with `pipeline-`, because not everything is a pipeline anymore. And that's what I did in an earlier commit, with something like `global-control-toolbar` but then reverted back to the old pattern. Curious to hear other thoughts!

Lastly, I haven't added anything to the release markdown file because I don't think we've aligned on when we should add in those notes.

## QA notes

The changes implemented in this PR can be [viewed here](http://kedro-viz-fe.s3-website.eu-west-2.amazonaws.com/feature/create-global-toolbar-ui%23ked-2872/?data=animals).

@hamzaoza I wasn't sure if the icons in the global toolbar need to be enlarged. In Zeplin, they look slightly larger than what's in the current primary toolbar. For now, I've left them the same size as what they were in their previous location. Let me know if that's incorrect. 

## Checklist

## Legal notice


studioswong:
Thanks for the great work @tynandebold , I had a skim through and a play around with the deployed link and it looks absolutely cracking 💯 

Re: `pipeline-` prefix, indeed this is definitely something in the pipeline for us to get rid of ( the `pipeline-` prefix has existed historically and has been raised multiple times to be gotten rid of over the past months - thanks for raising this concern as well 👍 ! ) This should ideally be dealt with in [this ticket](https://jira.quantumblack.com/browse/KED-2137).  

I'll take a deeper dive into the code and let you know of comments, but so far it looks awesome!

hamzaoza:
Great to see this starting to come to life 🚀 and already looking great 😄 

You're right, the icons in the global navigation do look small. They should be 24x24px both on the global navigation and utility bar so probably need to be bumped up a little.

Two minor questions:
- Can we remove the tooltip from the top two icons on hover (pipeline and experiments tracking)? It looks a bit strange and would prefer the full square background hover interactions that we have in the designs.
- Is the plan to change the `hide menu` icon and the order of icons in the utility bar in this PR or a separate PR?

tynandebold:
> Great to see this starting to come to life 🚀 and already looking great 😄
> 
> You're right, the icons in the global navigation do look small. They should be 24x24px both on the global navigation and utility bar so probably need to be bumped up a little.
> 
> Two minor questions:
> 
> * Can we remove the tooltip from the top two icons on hover (pipeline and experiments tracking)? It looks a bit strange and would prefer the full square background hover interactions that we have in the designs.
> * Is the plan to change the `hide menu` icon and the order of icons in the utility bar in this PR or a separate PR?

Cool, so 24px everywhere then. I'll update.

Yes, can remove the tooltip and add the hover interaction. I'll also update the `hide menu` icon and change the order. Didn't see that those hand changed. Thanks for pointing that out.

tynandebold:
> Thanks for the great work @tynandebold , I had a skim through and a play around with the deployed link and it looks absolutely cracking 💯
> 
> Re: `pipeline-` prefix, indeed this is definitely something in the pipeline for us to get rid of ( the `pipeline-` prefix has existed historically and has been raised multiple times to be gotten rid of over the past months - thanks for raising this concern as well 👍 ! ) This should ideally be dealt with in [this ticket](https://jira.quantumblack.com/browse/KED-2137).
> 
> I'll take a deeper dive into the code and let you know of comments, but so far it looks awesome!

Oh nice! Didn't realize there was another ticket for that. That's great.

Also, there's a huge amount of changes in my `package-lock`. Not sure why. All I've done in clone the project and `npm install`. Maybe a fresh install hasn't been done in a while and then committed though. Let me know what should be done there.

src/components/global-toolbar/global-toolbar.test.js
```diff
@@ -0,0 +1,90 @@
...
+    expect(wrapper.find('.pipeline-icon-toolbar__button').length).toBe(3);
+  });
+
+  it('hides one button when visible prop is false for one of them', () => {
+    const visible = {
+      themeBtn: false,
+    };
+    const wrapper = setup.mount(<ConnectedGlobalToolbar />, { visible });
+    expect(wrapper.find('.pipeline-icon-toolbar__button').length).toBe(4);
+  });
```
studioswong, line 30: Looking at these tests, it does remind me that there is likely no need for setting the `visible` prop for the theme and settings panel button - there simply is no cases at all within our codebase for `visible.themeBtn` or `visible.settingsBtn` to be false within our codebase ( or in general anywhere in the app.) 

This just reminds me that we would need to revisit the current set of state items that existed - for now I would say that the `visible` prop for the settings panel and theme button are not needed given the lack of a use case for it, and we can delete these tests to help this component and set of tests to be lighter weight.  

( You definitely did the right thing here as I completely understand that you had followed how we had previously set up the theme button and settings panel, and definitely nothing wrong with these tests here. )

src/components/global-toolbar/global-toolbar.test.js
```diff
@@ -0,0 +1,90 @@
...
+      expect(mockFn.mock.calls.length).toBe(1);
+    }
+  );
+
+  it('maps state to props', () => {
+    const expectedResult = {
+      theme: expect.stringMatching(/light|dark/),
+      visible: expect.objectContaining({
+        exportBtn: expect.any(Boolean),
+        settingsBtn: expect.any(Boolean),
```
studioswong, line 58: As per my previous comment, I don't think we need to import the `settingsBtn` and `themeBtn` here as there isn't any use case for them to be false

src/components/global-toolbar/index.js
```diff
@@ -0,0 +1,94 @@
...
+            ariaLabel={`Change to ${
+              theme === 'light' ? 'dark' : 'light'
+            } theme`}
+            className={
+              'pipeline-menu-button--theme pipeline-menu-button--large'
+            }
+            onClick={() => onToggleTheme(theme === 'light' ? 'dark' : 'light')}
+            icon={ThemeIcon}
+            labelText="Toggle theme"
+            visible={visible.themeBtn}
```
studioswong, line 61: Same as my previous comment, I don't think we need the `visible` prop here given that we don't have any use case for the themeBtn to be false 😆

src/components/global-toolbar/index.js
```diff
@@ -0,0 +1,94 @@
...
+          <IconButton
+            ariaLabel={'Change the settings flags'}
+            className={
+              'pipeline-menu-button--settings pipeline-menu-button--large'
+            }
+            onClick={() => onToggleSettingsModal(true)}
+            icon={SettingsIcon}
+            disabled={false}
+            labelText={'Settings'}
+            visible={visible.settingsBtn}
```
studioswong, line 72: same here 😄

src/components/primary-toolbar/primary-toolbar.test.js
```diff
@@ -62,7 +57,6 @@ describe('PrimaryToolbar', () => {
     const expectedResult = {
       disableLayerBtn: expect.any(Boolean),
       textLabels: expect.any(Boolean),
-      theme: expect.stringMatching(/light|dark/),
       visible: expect.objectContaining({
         exportBtn: expect.any(Boolean),
         settingsBtn: expect.any(Boolean),
```
studioswong, line 62: This can be deleted - same for `themeBtn` for line 68 as well

src/components/global-toolbar/global-toolbar.test.js
tynandebold: Removed!

src/components/primary-toolbar/primary-toolbar.test.js
tynandebold: It's gone now.

hamzaoza:
Thanks for this @tynandebold this is 🔥 🔥 🔥 and it's looking great! I'm basing my comments on [this](http://kedro-viz-fe.s3-website.eu-west-2.amazonaws.com/feature/create-global-toolbar-ui%23ked-2872/?data=animals) so let me know if this is an old build and I need to look at a new version and my comment's no longer apply.

Two minor points:

1.  The tab of the selected page i.e. tree view or experiments tracking should have the active colour on it. Right now the highlight shows on hover.
2. The global bar should slide out from view when you click the `hide menu` icon. Have a play with the `hide menu` icon on this [example](https://mckinsey.invisionapp.com/share/4F15YJ56CJ#/164221455_Experiments_-_Comparison_-_Triple)

and thanks for cleaning up some of the inconsistencies in the toolbar, it has made a huge difference already 😄

tynandebold:
> Thanks for this @tynandebold this is 🔥 🔥 🔥 and it's looking great! I'm basing my comments on [this](http://kedro-viz-fe.s3-website.eu-west-2.amazonaws.com/feature/create-global-toolbar-ui%23ked-2872/?data=animals) so let me know if this is an old build and I need to look at a new version and my comment's no longer apply.
> 
> Two minor points:
> 
> 1. The tab of the selected page i.e. tree view or experiments tracking should have the active colour on it. Right now the highlight shows on hover.
> 2. The global bar should slide out from view when you click the `hide menu` icon. Have a play with the `hide menu` icon on this [example](https://mckinsey.invisionapp.com/share/4F15YJ56CJ#/164221455_Experiments_-_Comparison_-_Triple)
> 
> and thanks for cleaning up some of the inconsistencies in the toolbar, it has made a huge difference already 😄

My pleasure @hamzaoza! That's the correct link to look at. Responding to your points:

1. We'll wire that up when we're actually able to switch between two pages.
2. Oh good catch. I missed this. If this is the intended behavior though I think it defeats the purpose of having a global toolbar. Wouldn't we always want the user to be able to navigate to the other page or control the theme/settings no matter where they are? If the global toolbar collapses with the sidebar it'll take an extra click to be able to perform any of these actions.

FYI: we may elect to get this merged into our `experiment-tracking` base branch now and continue to tweak behavior as we go along.

studioswong (approved):
great stuff - thanks for all the changes and congrats on a great first PR on Kedro-viz!!! 💯 

@hamzaoza I share the same views as Tynan on the global toolbar for it to be in place on hide menu, mainly given the easier navigation between the experiment and flowchart page for those users who normally keeps their sidebar hidden ( which our localstate memorizes this pattern and hence reloads the collapsed sidebar everytime the page reloads.) 

As Tynan mentioned, this is merged into a feature branch so can definitely be revisited before this feature is released.

PR #602, opened 2021-10-20
[KED-2902] `RunTrackingData` resolver

merelcht:
## Description

https://jira.quantumblack.com/browse/KED-2902

## Development notes

* Added a resolver to fetch run details. Run tracking data is any data stored in tracking `MetricsDataSet` or `JSONDataSet` instances during a Kedro run. 
* Run tracking data is not stored in the session store for the time being. In a later iteration of this work, this data will be added to the session store and it will be indexed so it can be searched efficiently.
* The run tracking data is matched to a run through the session ID. The session ID is saved with the run and the session ID is the version of the datasets. 

## QA notes

* Added a new test to check that only data from tracking `MetricsDataSet` or `JSONDataSet` is added in the response.

## Checklist

## Legal notice

- [X] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

### DIFF
--- a/package/tests/test_api/test_graphql.py
+++ b/package/tests/test_api/test_graphql.py
@@ -1,32 +1,3 @@
-# Copyright 2021 QuantumBlack Visual Analytics Limited
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
-# OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, AND
-# NONINFRINGEMENT. IN NO EVENT WILL THE LICENSOR OR OTHER CONTRIBUTORS
-# BE LIABLE FOR ANY CLAIM, DAMAGES, OR OTHER LIABILITY, WHETHER IN AN
-# ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF, OR IN
-# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-#
-# The QuantumBlack Visual Analytics Limited ("QuantumBlack") name and logo
-# (either separately or in combination, "QuantumBlack Trademarks") are
-# trademarks of QuantumBlack. The License does not grant you any right or
-# license to the QuantumBlack Trademarks. You may not use the QuantumBlack
-# Trademarks or any confusingly similar mark as a trademark for your product,
-# or use the QuantumBlack Trademarks in any other manner that might cause
-# confusion in the marketplace, including but not limited to in advertising,
-# on websites, or on software.
-#
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-
 import shutil
 from unittest import mock
 from unittest.mock import PropertyMock, call, patch
@@ -81,6 +52,10 @@ def example_tracking_catalog(save_version):
 
     yield catalog
 
+    shutil.rmtree("test.json", ignore_errors=True)
+    shutil.rmtree("metrics.json", ignore_errors=True)
+    shutil.rmtree("tracking.json", ignore_errors=True)
+
 
 @pytest.fixture
 def example_tracking_output(save_version):
@@ -113,18 +88,6 @@ def example_tracking_output(save_version):
             },
         ),
     ]
-
-
-@pytest.fixture(scope="class", autouse=True)
-def cleanup(request):
-    """Cleanup a testing directory once we are finished."""
-
-    def remove_test_dir():
-        shutil.rmtree("test.json")
-        shutil.rmtree("metrics.json")
-        shutil.rmtree("tracking.json")
-
-    request.addfinalizer(remove_test_dir)
 
 
 class TestTrackingData:


package/kedro_viz/api/graphql.py
```diff
@@ -126,6 +120,92 @@ def get_all_runs() -> List[Run]:
...
+            tracking_keys.add(nested_keys)
+    runs_tracking_data = {
+        key: [
+            {"runId": run_id, "value": tracking_data[run_id][key]}
+            for run_id in tracking_data
+            if key in tracking_data[run_id]
+        ]
+        for key in sorted(tracking_keys)
+    }
+    return json.loads(json.dumps(runs_tracking_data))
```
limdauto, line 131: Why do we have to do this? Can't we just return `runs_tracking_data`?

package/kedro_viz/api/graphql.py
```diff
@@ -126,6 +120,92 @@ def get_all_runs() -> List[Run]:
...
+            { runId: 'My Favorite Sprint', value: 0.8 },
+            { runId: 'Another favorite sprint', value: 0.5 },
+            { runId: 'Slick test this one', value: 1 },
+            ],
+            classWeight: [
+            { runId: 'My Favorite Sprint', value: 23 },
+            { runId: 'Another favorite sprint', value: 21 },
+            { runId: 'Slick test this one', value: 21 },
+            ]
+    }]
```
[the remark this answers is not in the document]
limdauto: Could you also add an example for `tracking_data` as well?

package/kedro_viz/api/graphql.py
```diff
@@ -99,6 +93,92 @@ def get_all_runs() -> List[Run]:
...
+            if key in tracking_data[run_id]
+        ]
+        for key in sorted(tracking_keys)
+    }
+    return json.loads(json.dumps(runs_tracking_data))
+
+
+def get_run_tracking_data(run_ids: List[ID]) -> List[TrackingDataSet]:
+    # pylint: disable=protected-access,import-outside-toplevel
+    """Get all details for a specific run. Run details contains the data from the
```
limdauto, line 136:
```suggestion
    """Get all details for a list of runs. Run tracking data contain the data from the
```

package/kedro_viz/api/graphql.py
```diff
@@ -99,6 +93,92 @@ def get_all_runs() -> List[Run]:
...
+    tracking MetricsDataSet and JSONDataSet instances that have been logged
+    during that specific `kedro run`.
+    Args:
+        run_ids:  List of IDs of runs to fetch the details for.
```
limdauto, line 140:
```suggestion
        run_ids:  List of IDs of runs to fetch the tracking data for.
```

package/kedro_viz/api/graphql.py
```diff
@@ -99,6 +93,92 @@ def get_all_runs() -> List[Run]:
...
+    all_datasets = []
+    catalog = data_access_manager.catalog.get_catalog()
+    experiment_datasets = [
+        (ds_name, ds_value)
+        for ds_name, ds_value in catalog._data_sets.items()
+        if (isinstance(ds_value, (MetricsDataSet, JSONDataSet)))
+    ]
+
+    if not experiment_datasets:
+        logger.warning("No tracking datasets found in catalog")
```
limdauto, line 157: I don't think we should warn here. Instead, we should return an empty list.

package/kedro_viz/api/graphql.py
```diff
@@ -99,6 +93,92 @@ def get_all_runs() -> List[Run]:
...
+
+    Returns:
+        List of TrackingDataSets
+
+    """
+    from kedro.extras.datasets.tracking import JSONDataSet, MetricsDataSet  # noqa: F811
+
+    all_datasets = []
+    catalog = data_access_manager.catalog.get_catalog()
+    experiment_datasets = [
```
limdauto, line 150: Can we name this as `tracking_datasets` to faithfully reflect what they are?

package/kedro_viz/api/graphql.py
```diff
@@ -99,6 +93,92 @@ def get_all_runs() -> List[Run]:
...
+        all_runs = {}
+        for run_id in run_ids:
+            runid = ID(run_id)
+            file_path = dataset._get_versioned_path(str(runid))
+            if Path(file_path).is_file():
+                with dataset._fs.open(
+                    file_path, **dataset._fs_open_args_load
+                ) as fs_file:
+                    json_data = json.load(fs_file)
+                    all_runs[runid] = json_data
```
limdauto, line 169: Could you add a comment here explaining why we need to rely on low-level I/O methods and not being able to utilise `catalog.load`?

Also I thought we agree to abstract this bit out so we could re-use it?

package/kedro_viz/api/graphql.py
```diff
@@ -127,6 +207,12 @@ class TrackingDataSet:
 class Query:
     """Query endpoint to get data from the session store"""
 
+    @strawberry.field
+    def run_tracking_data(self, run_ids: List[ID]) -> List[TrackingDataSet]:
+        """Query to get data for specific runs from the session store"""
+        runs = get_run_tracking_data(run_ids)
+        return runs
```
limdauto, line 214:
```suggestion
        return get_run_tracking_data(run_ids)
```

Why do we need to have a method called `get_run_tracking_data`? Can't the logic be written inside this resolver instead? Is it to make it easier to test?

package/tests/test_api/test_graphql.py
```diff
@@ -1,5 +1,186 @@
...
+            },
+        ),
+    ]
+
+
+class TestTrackingData:
+    def test_graphql_run_tracking_data_query(
+        self, tmp_path, data_access_manager: DataAccessManager, example_tracking_output
+    ):
+        save_version = "2021-11-02T18.24.24.379Z"
```
limdauto, line 85: I feel like this string is repeated, so it could be a fixture. This fixture could be used in `example_tracking_output` as well.

package/tests/test_api/test_graphql.py
```diff
@@ -1,5 +1,186 @@
...
+        with mock.patch(
+            "kedro_viz.api.graphql.data_access_manager", new=data_access_manager
+        ):
+            metrics_dataset = MetricsDataSet(
+                filepath=Path(tmp_path / "test.json").as_posix(),
```
limdauto, line 90: `as_posix()` isn't cross-platform. You can do `str()` instead: https://docs.python.org/3/library/pathlib.html#operators

package/tests/test_api/test_graphql.py
```diff
@@ -1,5 +1,186 @@
...
+            catalog = DataCatalog(
+                data_sets={
+                    "csv": dataset,
+                }
+            )
+            data_access_manager.add_catalog(catalog)
+
+            assert get_run_tracking_data([ID(save_version)]) == []
+
+            patched_warning.assert_has_calls(
```
limdauto, line 141: I think this warning might not be necessary. Once we have the design, this is a legit path that we need to return some response to the client. I'd ignore it for now.

package/tests/test_api/test_graphql.py
```diff
@@ -61,3 +242,72 @@ def test_graphql_runs_metadata_endpoint_no_dbsession(self, client):
...
+
+    @patch("kedro_viz.api.graphql.get_run_tracking_data")
+    def test_graphql_runs_tracking_data_endpoint(
+        self, patched_tracking, client, example_tracking_output
+    ):
+        patched_tracking.return_value = example_tracking_output
+        response = client.post(
+            "/graphql",
+            json={
+                "query": """{runTrackingData(runIds: [1534326,1234,1234])
```
limdauto, line 254: This test is confusing. We request these 3 run ids here and in the response, each metric only has 1 run in it. And that run doesn't match any of the three given run ids.

package/kedro_viz/api/graphql.py
rashidakanchwala: it threw me pylint errors for some reason. but i reckon it should work now that u changed JSONOBject to take a dict. I will give it a go

rashidakanchwala: it's still giving me this :( 
 error: Incompatible return value type (got "Dict[Any, List[Dict[str, Any]]]", expected "JSONObject")

package/tests/test_api/test_graphql.py
```diff
@@ -1,5 +1,184 @@
...
+
+
+@pytest.fixture(scope="class", autouse=True)
+def cleanup(request):
+    """Cleanup a testing directory once we are finished."""
+
+    def remove_test_dir():
+        shutil.rmtree("test.json")
+        shutil.rmtree("metrics.json")
+        shutil.rmtree("tracking.json")
```
limdauto, line 125: Nit: I'd suggest adding `ignore_errors=True` in case they don't exist for whatever reason.

Also you don't need to have a dedicated cleanup fixture. You can do all of this after `yield catalog` in `example_tracking_catalog` and it would be called after the test. This way is fine too of course.

limdauto (approved):
Thank you. In the future, I want to have test to cover more than just 1 `save_version`s but this is probably enough for now.

package/tests/test_api/test_graphql.py
```diff
@@ -1,5 +1,184 @@
...
+# license to the QuantumBlack Trademarks. You may not use the QuantumBlack
+# Trademarks or any confusingly similar mark as a trademark for your product,
+# or use the QuantumBlack Trademarks in any other manner that might cause
+# confusion in the marketplace, including but not limited to in advertising,
+# on websites, or on software.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
```
limdauto, line 29: Can you remove this as well? This header no longer exists in `main`

PR #623, opened 2021-11-11
[KED-2953] Update RunTrackingData schema to show difference

rashidakanchwala:
## Description

Jira ticket here - https://jira.quantumblack.com/browse/KED-2953

## Development notes

Earlier, runTrackingData for multiple runs returned only all the available run tracking data. Now for showDiff:true; it returns only runTrackingData that is common for multiple runs and for false, it returns all the available run tracking data
  

The code logic has changed and needs to be reviewed. 

## Checklist

## Legal notice

### DIFF
--- a/package/kedro_viz/api/graphql.py
+++ b/package/kedro_viz/api/graphql.py
@@ -143,15 +143,6 @@ def format_run_tracking_data(tracking_data: Dict, show_diff: bool) -> JSONObject
 
     if show_diff:
         for key in sorted(set(tracking_keys)):
-            if tracking_keys.count(key) == len(tracking_data):
-                for run_id in tracking_data:
-                    if key in tracking_data[run_id]:
-                        formatted_tracking_data[key].append(
-                            {"runId": run_id, "value": tracking_data[run_id][key]}
-                        )
-
-    else:
-        for key in sorted(set(tracking_keys)):
             for run_id in tracking_data:
                 if key in tracking_data[run_id]:
                     formatted_tracking_data[key].append(
@@ -161,7 +152,15 @@ def format_run_tracking_data(tracking_data: Dict, show_diff: bool) -> JSONObject
                     formatted_tracking_data[key].append(
                         {"runId": run_id, "value": None}
                     )
-
+    else:
+        for key in sorted(set(tracking_keys)):
+            if tracking_keys.count(key) == len(tracking_data):
+                for run_id in tracking_data:
+                    if key in tracking_data[run_id]:
+                        formatted_tracking_data[key].append(
+                            {"runId": run_id, "value": tracking_data[run_id][key]}
+                        )
+ 
     return JSONObject(formatted_tracking_data)
 
 
--- a/package/kedro_viz/api/graphql.py
+++ b/package/kedro_viz/api/graphql.py
@@ -94,7 +94,7 @@ def get_all_runs() -> List[Run]:
     return runs
 
 
-def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
+def format_run_tracking_data(tracking_data: Dict, showDiff: bool) -> JSONObject:
     """Convert tracking data in the front-end format.
 
     Args:
@@ -118,7 +118,7 @@ def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
         >>>         'classWeight":21
         >>>     },
         >>> }
-        >>> format_run_tracking_data(tracking_data)
+        >>> format_run_tracking_data(tracking_data,false)
         {
             bootstrap: [
                 { runId: 'My Favorite Run', value: 0.8 },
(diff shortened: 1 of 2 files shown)


package/kedro_viz/api/graphql.py
```diff
@@ -94,11 +94,12 @@ def get_all_runs() -> List[Run]:
     return runs
 
 
-def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
+def format_run_tracking_data(tracking_data: Dict, show_diff: bool) -> JSONObject:
     """Convert tracking data in the front-end format.
 
     Args:
         tracking_data: JSON blob of tracking data for selected runs
+        show_diff : user specified boolean for comparison view
```
AntonyMilneQB, line 102: I think this needs a bit more explanation about what it actually does, like you gave in the PR. Maybe:
```suggestion
        show_diff: whether to include metrics that are not defined for some of the selected runs in comparison view
```

package/kedro_viz/api/graphql.py
```diff
@@ -118,7 +119,7 @@ def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
         >>>         'classWeight":21
         >>>     },
         >>> }
-        >>> format_run_tracking_data(tracking_data)
+        >>> format_run_tracking_data(tracking_data,false)
```
AntonyMilneQB, line 122:
```suggestion
        >>> format_run_tracking_data(tracking_data, False)
```
Note `true/false` doesn't work in Python - it needs to be `True/False`. If we give this a default value of `False` then this argument wouldn't be needed here at all though.

package/kedro_viz/api/graphql.py
```diff
@@ -94,11 +94,12 @@ def get_all_runs() -> List[Run]:
     return runs
 
 
-def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
+def format_run_tracking_data(tracking_data: Dict, show_diff: bool) -> JSONObject:
```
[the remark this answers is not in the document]
AntonyMilneQB: Oh, I see it gets its default value in `run_tracking_data`, which then gets passed through `get_run_tracking_data`  and then to `format_run_tracking_data` - did I understand that right?

package/kedro_viz/api/graphql.py
```diff
@@ -135,22 +136,42 @@ def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
     """
     formatted_tracking_data = defaultdict(list)
 
-    for run_id, run_tracking_data in tracking_data.items():
-        for tracking_name, data in run_tracking_data.items():
-            formatted_tracking_data[tracking_name].append(
-                {"runId": run_id, "value": data}
-            )
+    tracking_keys = []
```
AntonyMilneQB, line 139: I would suggest doing this a bit differently:
* leave the stuff that just generates `formatted_tracking_data` as it was before
* this should include all the metrics (both common across runs and not common across runs)
* _then_, if  `show_diff = False`, go through `formatted_tracking_data` and remove the ones where it's not common across all runs
* return whatever is left in `formatted_tracking_data`

It might look something like this (completely untested...):
 ```suggestion
 # Same as before:
formatted_tracking_data = defaultdict(list)
for tracking_name, data in run_tracking_data.items():
    formatted_tracking_data[tracking_name].append({"runId": run_id, "value": data})
    
# New stuff:
if show_diff:
     for tracking_name, all_run_data in list(formatted_tracking_data.items()):
         if any(one_run_data["value"] is None for one_run_data in all_run_data):
              del formatted_tracking_data[tracking_name]
```

File: package/kedro_viz/api/graphql.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (rashidakanchwala) [reply package/kedro_viz/api/graphql.py:97]: correct, I add the default = False in the run_tracking_data query.

File: package/kedro_viz/api/graphql.py (hunk 4 above)
Comment (rashidakanchwala) [reply package/kedro_viz/api/graphql.py:139]: Thanks, this is smarter. But this doesn't handle the case where show_diff = false and we need to explicitly add 'null' to the keys which have no value right? Also Merel's question is should we do that in the backend?

Comment (AntonyMilneQB) [reply package/kedro_viz/api/graphql.py:139]: Ohh, I thought the original `formatted_tracking_data` already contained those nulls and then what you were trying to do is remove them if `show_diff=True`. But you're saying the original `formatted_tracking_data` didn't contain the nulls and if `show_diff=False` you need to add them?

Comment (AntonyMilneQB) [reply package/kedro_viz/api/graphql.py:139]: Also, just wondering... this kind of dictionary reshaping thing might be much easier to do in a pandas DataFrame. I see pandas is already a dependency of kedro-viz but don't know if you use it for data manipulation anywhere or it's preferred to do all this sort of thing with dictionaries?

File: package/kedro_viz/api/graphql.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (AntonyMilneQB) [reply package/kedro_viz/api/graphql.py:97]: Do you think `run_tracking_data` or `get_run_tracking_data` will ever get called outside `format_run_tracking_data`? If so then I think you should give them default value `show_diff = False` too. If not then all good how it is 🙂

**Review by MerelTheisenQB — APPROVED** [review]
Just one minor comment on splitting the test, but otherwise this looks good 🙂

File: package/kedro_viz/api/graphql.py [hunk 5]
```diff
@@ -94,11 +94,13 @@ def get_all_runs() -> List[Run]:
...
 
 
-def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
+def format_run_tracking_data(tracking_data: Dict, show_diff: bool = False) -> JSONObject:
     """Convert tracking data in the front-end format.
 
     Args:
         tracking_data: JSON blob of tracking data for selected runs
+        show_diff : if false, it will show runs with only common tracking
+        data else it will show all available tracking data
```
AntonyMilneQB, line 103:
```suggestion
        show_diff: If false, show runs with only common tracking
            data; else show all available tracking data
```

package/kedro_viz/api/graphql.py
```diff
@@ -140,17 +142,23 @@ def format_run_tracking_data(tracking_data: Dict) -> JSONObject:
...
-def get_run_tracking_data(run_ids: List[ID]) -> List[TrackingDataSet]:
+def get_run_tracking_data(run_ids: List[ID], show_diff: bool = False) -> List[TrackingDataSet]:
     # pylint: disable=protected-access,import-outside-toplevel
     """Get all tracking data for a list of runs. Tracking data contains the data from the
     tracking MetricsDataSet and JSONDataSet instances that have been logged
     during that specific `kedro run`.
     Args:
         run_ids:  List of IDs of runs to fetch the tracking data for.
+        show_diff : if false, it will show runs with only common tracking
+        data else it will show all available tracking data
```
AntonyMilneQB, line 161:
```suggestion
        show_diff: If false, show runs with only common tracking
            data; else show all available tracking data
```

package/tests/test_api/test_graphql.py
```diff
@@ -57,6 +63,35 @@ def example_tracking_catalog(save_version):
...
+    # Note - filepath is assigned without using tmp_path as it fails on windows build.
+    # This is a temp soln and will be cleaned up in the future.
+    new_metrics_dataset = MetricsDataSet(
+        filepath="test.json",
+        version=Version(None, save_version),
+    )
+    new_metrics_dataset.save({"col1": 1, "col3": 3})
+
+    time.sleep(1)
+
```
AntonyMilneQB, line 77:
```suggestion
    new_metrics_dataset.save({"col1": 1, "col3": 3})
```

AntonyMilneQB (approved):

package/tests/test_api/test_graphql.py
```diff
@@ -57,6 +63,35 @@ def example_tracking_catalog(save_version):
...
+    new_metrics_dataset = MetricsDataSet(
```
AntonyMilneQB, line 78:
```suggestion
    new_metrics_dataset.save({"col1": 1, "col3": 3})
    new_metrics_dataset = MetricsDataSet(
```

package/tests/test_api/test_graphql.py
```diff
@@ -1,4 +1,5 @@
 import shutil
+import time
```
AntonyMilneQB, line 2:
```suggestion
```

MerelTheisenQB (approved):
LGTM 👍

PR #624, opened 2021-11-11
update schema and queries with new showDiff variable

studioswong:
## Description

related ticket: https://jira.quantumblack.com/browse/KED-2914

This ticket is to update the schema to reflect the latest design of query and data type to allow the 'showDifference' feature

## Development notes

I had mainly made changes to the schema that will be referenced for the BE, as well as added an additional query to showcase the usage of this variable to obtain the showDiff data

## QA notes

N/A

## Checklist

## Legal notice

### DIFF
--- a/src/apollo/queries.js
+++ b/src/apollo/queries.js
@@ -38,3 +38,14 @@ export const GET_RUN_TRACKING_DATA = gql`
     }
   }
 `;
+
+/** query for collapsable run details component */
+export const GET_RUN_TRACKING_DATA_WITH_DIFF = gql`
+  query getRunTrackingData($runs: [ID]!, $showDiff: Boolean) {
+    runTrackingData(runIDs: $runs, showDiff: $showDiff) {
+      datasetName
+      datasetType
+      data
+    }
+  }
+`;
--- a/src/apollo/queries.js
+++ b/src/apollo/queries.js
@@ -30,17 +30,6 @@ export const GET_RUN_METADATA = gql`
 
 /** query for collapsable run details component */
 export const GET_RUN_TRACKING_DATA = gql`
-  query getRunTrackingData($runs: [ID]!) {
-    runTrackingData(runIDs: $runs) {
-      datasetName
-      datasetType
-      data
-    }
-  }
-`;
-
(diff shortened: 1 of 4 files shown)


src/apollo/queries.js
```diff
@@ -38,3 +38,14 @@ export const GET_RUN_TRACKING_DATA = gql`
...
+/** query for collapsable run details component */
+export const GET_RUN_TRACKING_DATA_WITH_DIFF = gql`
+  query getRunTrackingData($runs: [ID]!, $showDiff: Boolean) {
+    runTrackingData(runIDs: $runs, showDiff: $showDiff) {
+      datasetName
+      datasetType
+      data
+    }
+  }
+`;
```
tynandebold, line 51: I don't think you need this. We should just be able to add the `showDiff` argument to the `GET_RUN_TRACKING_DATA` query.

src/apollo/schema.graphql
```diff
@@ -7,7 +7,7 @@ schema {
 type Query {
   runsList: [Run]!
   runMetadata(runIDs: [ID]!): [Run!]!
-  runTrackingData(runIDs: [ID]!): [TrackingDataset]!
+  runTrackingData(runIDs: [ID]!, showDiff: Boolean): [TrackingDataset]! # showDiff as true will return all tracked fields amongst compared runs - default setting is false
```
tynandebold, line 10:
```suggestion
  runTrackingData(runIDs: [ID]!, showDiff: Boolean = false): [TrackingDataset]! 
```
It would be good to set an explicit default value here for the `showDiff` argument. That way we can safely leave that argument out of this query when we need to.

tynandebold (changes requested):
Great stuff! 

I don't think we need to create a new query export though, as we should be able to get both the `showDiff` data and the shared-values data from the same `GET_RUN_TRACKING_DATA` query.

src/apollo/schema.graphql
studioswong: great reminder! Haven't thought of this as I had this as an optional variable in mind but this set up is safer!

tynandebold (approved):

PR #625, opened 2021-11-12
[KED-2954] Format experiment tracking timestamp

tynandebold:
## Description

We'd like to show a human-readable timestamp for experiment tracking runs. This PR implements that.

## Development notes

I've used the `.fromNow()` method of the [Day.js](https://day.js.org/docs/en/display/from-now) library, which is [nice and small](https://bundlephobia.com/package/dayjs@1.10.7) and does exactly what we need.

## QA notes

image

Check out the branch and take a look at the dates.

## Checklist

## Legal notice

### DIFF
--- a/src/utils/date-utils.js
+++ b/src/utils/date-utils.js
@@ -1,6 +1,6 @@
-import * as dayjs from 'dayjs';
-import * as relativeTime from 'dayjs/plugin/relativeTime';
-dayjs.extend(relativeTime);
+import dayjs from 'dayjs';
+import relativeTime from 'dayjs/plugin/relativeTime';
+const _dayJs = dayjs.extend(relativeTime);
 
 /**
  * Take a timestamp and return a meaningful length of time, e.g. 5 months ago
@@ -8,5 +8,5 @@ dayjs.extend(relativeTime);
  * @returns A human-readable from-now date
  */
 export const toHumanReadableTime = (timestamp) => {
-  return dayjs(timestamp).fromNow();
+  return _dayJs(timestamp).fromNow();
 };
--- a/src/utils/date-utils.js
+++ b/src/utils/date-utils.js
@@ -1,0 +1,12 @@
+import * as dayjs from 'dayjs';
+import * as relativeTime from 'dayjs/plugin/relativeTime';
+dayjs.extend(relativeTime);
+
+/**
+ * Take a timestamp and return a meaningful length of time, e.g. 5 months ago
+ * @param {ISO 8601 string} timestamp The timestamp to be converted
+ * @returns A human-readable from-now date
+ */
+export const toHumanReadableTime = (timestamp) => {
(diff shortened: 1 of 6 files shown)


studioswong (commented):
code looks good - only thing that's left is to test and modify this to work with the timestamp format returned from the actual endpoint ( which on my initial testing with the actual data it doesn't seem to be recognised by `dayjs` ). 

Let's this this in the afternoon with the actual experiment data setup after merging [this PR](https://github.com/quantumblacklabs/kedro-viz/pull/630) in 
😄

studioswong (changes requested):
I've given this a thought - while we are waiting to align with the BE on the format of the timestamp, it would be worth for us to set up a small util function to format the timestamp from the BE to return a JS date object as this might happen again somewhere down the line.

I've just merged the latest feature branch into this branch locally to test this with my local experiment data - pasting it here for your reference. 
image

tynandebold:
> I've given this a thought - while we are waiting to align with the BE on the format of the timestamp, it would be worth for us to set up a small util function to format the timestamp from the BE to return a JS date object as this might happen again somewhere down the line.
> 
> I've just merged the latest feature branch into this branch locally to test this with my local experiment data - pasting it here for your reference. image

Whoops. I already merged this in after speaking with Antony about the date format. I was tired of it still being in PR.

If you pull this branch down or also check out our main feature branch it should be working now. Let me know if it isn't.

src/utils/date-utils.js
```diff
@@ -4,9 +4,30 @@ const _dayJs = dayjs.extend(relativeTime);
...
+    if (i < 2) {
+      formattedTimestamp += `${splitTimestamp[i]}:`;
+    } else if (i === 2) {
+      formattedTimestamp += `${splitTimestamp[i]}.`;
+    } else {
+      formattedTimestamp += `${splitTimestamp[i]}`;
+    }
+
+    i++;
+  }
```
studioswong, line 30: can we simplify this with the `.replace` method with Regex ( obviously we need to find a Regex pattern than matches and replace the first 2 occurrences of the `.` instances in the string, but from what I recall that is doable with Regex, though this might require more digging in with Regex patterns.)

studioswong (commented):
Arrgh I didn't know about the new pushes - you should've told me to review the new code pushes before merging it in 😄 

I've tested this and it seems to work now with the data, have left a further comment about the new code that you pushed. It would be great if we can also add a unit test to test the string manipulation as well.

src/utils/date-utils.js
tynandebold: I updated it once more after that push. [See here](https://github.com/quantumblacklabs/kedro-viz/commit/77f04d68f38badfff62e0bb2a889301228cd40c9). It's a one-liner again.

studioswong: the new changes looks great - thank you for putting that in 😄

PR #626, opened 2021-11-12
Add extra -a click command for autoreload

datajoely:
The `kedro viz --autoreload` function is super useful for developing modular pipelines live and getting the namespace stuff just right.

I've found myself writing `--auto-reload` by accident so many times, I think this shortcut would be appreciated!


limdauto (approved):
LGTM!

PR #636, opened 2021-11-18
[KED-2969] - Fix the issue with showDiff when comparing with run without tracking data

rashidakanchwala:
## Description

There was a bug in showDiff wherein if one run had dataset that had no tracking data and another run had dataset which had some tracking data, it wouldn't show the common data between the two runs i.e. no data because there's nothing common. 
This issue has been fixed in this PR. 

## Checklist

## Legal notice


studioswong (changes requested):
I've just tested this and the `showDiff` variable and it works well for both `true` and `false` conditions, thanks for pushing this fix!

Can you please add the related tests for both `showDiff = true` and `showDiff = false` so the expected behaviour for this variable will be clear in the codebase ( to prevent confusion down the line). Thank you.

package/tests/test_api/test_graphql.py
```diff
@@ -87,6 +87,52 @@ def example_multiple_run_tracking_catalog(save_version, save_new_version):
...
+    # Note - filepath is assigned without using tmp_path as it fails on windows build.
+    # This is a temp soln and will be cleaned up in the future.
+    new_metrics_dataset = MetricsDataSet(
+        filepath="test.json",
+        version=Version(None, save_version),
+    )
+    new_metrics_dataset = MetricsDataSet(
+        filepath="test.json",
+        version=Version(None, save_new_version),
+    )
```
limdauto, line 125: Why are there 2 objects with the same name here?

package/tests/test_api/test_graphql.py
```diff
@@ -202,6 +248,104 @@ def test_graphql_run_tracking_data_query_show_diff_false(
                     ),
                 )
             ]
+    def test_graphql_run_tracking_data_show_diff_true_one_empty_run(
```
limdauto, line 251: You can use https://docs.pytest.org/en/6.2.x/parametrize.html to make these tests cleaner. They are essentially 4 variations of the same test, not 4 separate tests IMHO.

limdauto (commented):
I think the test could be a bit cleaner (added a couple of comments) but logic looks good to me.

package/tests/test_api/test_graphql.py
rashidakanchwala: I am trying to create two runs for the same metrics -- using two versions save_version, and save_new_version.

package/tests/test_api/test_graphql.py
rashidakanchwala: Thanks, this looks good. Will parametrize the below!!

limdauto (approved):
LGTM!

studioswong (approved):
Thanks for adding the tests!

PR #645, opened 2021-11-24
[KED-2980] Rewrite Kedro UI Modal 

rashidakanchwala:
## Description

This PR was created based on the ticket - https://jira.quantumblack.com/browse/KED-2980

Demo link - http://kedro-viz-fe.s3-website.eu-west-2.amazonaws.com/KED2980/kedro-ui-migration-modal/

## Development notes

Removed references to kedro-ui modal. Export modal still uses kedro-ui buttons which will be removed in another ticket. 

I had a question regarding tests for modal component. I have added only one test to check if modal renders. The other test to close the modal are already covered in settings-modal and export-modal so I was wondering if I should write another one -- because then I will have to explicitly create an onClose() method for the test. 

Update - Have also rewritten the escape key event in 2 places (modal, and nodelist) as these previously used kedro-ui utils

## QA notes

npm t runs successfully. 

## Checklist


studioswong (changes requested):
Thanks for setting this up!

Please add more tests to test the basic functionality of the modal, but otherwise looks good.

src/components/modal/index.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import classnames from 'classnames';
+import CloseIcon from '../icons/close';
+import IconButton from '../../components/icon-button';
+import './modal.css';
+
+/**
+ * Shows a toggle button for code panel
+ */
```
tynandebold, line 9: This comment doesn't seem relevant for this file.

src/components/modal/modal.scss
```diff
@@ -0,0 +1,113 @@
...
+  --color-modal-description: #{rgba($color-dark, 0.55)};
+}
+
+.kui-theme--dark {
+  --color-modal-bg: #{rgba($color-bg-dark-1, 0.9)};
+  --color-modal-content: #{$color-bg-dark-4};
+  --color-modal-title: #{rgba($color-light, 0.85)};
+  --color-modal-description: #{rgba($color-light, 0.55)};
+}
+/** Implementation **/
```
tynandebold, line 23: Add a space between these two lines.

tynandebold (changes requested):
Looking good overall! It's nearly there.

In the Kedro-UI modal, we were able to close it by pressing the [escape key](https://github.com/quantumblacklabs/kedro-ui/blob/master/src/components/modal/modal.jsx#L45-L51). I think you should aim to recreate that behavior here.

I also noticed there are two empty files in the PR that should be removed: 
- src/components/button/button.scss
- src/components/button/index.js

Lastly, the `Settings` button is always clickable now, since it's in the global toolbar. When the modal is open and that button is clicked, nothing happens. Perhaps clicking that button should close the modal when the modal is already open?

src/components/modal/index.js
```diff
@@ -0,0 +1,57 @@
...
+import utils from '@quantumblack/kedro-ui/lib/utils';
+
+/**
+ * Generic Kedro Modal
+ */
+const Modal = ({ title, onClose, visible, message, children }) => {
+  const handleKeyDown = (event) => {
+    utils.handleKeyEvent(event.keyCode, {
+      escape: onClose,
+    });
```
tynandebold, line 15: Appreciate you using the utils here from `kedro-ui` though perhaps we want to move away from that since we're also moving away from that library as a whole? What do you think?

You could get the same behavior using this:

```suggestion
if (event.keyCode === 27) {
  onClose();
}
```

tynandebold (approved):
Great job! I'm going to approve now with some additional feedback and notes that I hope you can take a look at and change before merging 😊

src/components/modal/index.js
rashidakanchwala: Sure, i will do this. There is another ticket to handle utils but since it's being used here I will address it here itself :) thanks!

src/components/global-toolbar/index.js
```diff
@@ -70,7 +71,7 @@ export const GlobalToolbar = ({
             className={
               'pipeline-menu-button--settings pipeline-menu-button--large'
             }
-            onClick={() => onToggleSettingsModal(true)}
+            onClick={() => onToggleSettingsModal(!visible.settingsModal)}
```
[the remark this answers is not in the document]
rashidakanchwala: Thanks, so the general rule is that modal should be on top of everything! Because one of the things I realised was plotly modal overlapped the global sidebar but others didn't.

src/components/modal/modal.scss
```diff
@@ -0,0 +1,114 @@
...
+
+.modal {
+  position: fixed;
+
+  top: 0;
+  left: 0;
+
+  width: 100%;
+  height: 100%;
+  z-index: 4;
```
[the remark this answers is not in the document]
rashidakanchwala: thanks, i changed it to 5 for some reason global toolbar is accessible at 4.

src/components/global-toolbar/index.js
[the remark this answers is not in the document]
tynandebold: Yes I think the modal should sit on top of all content :)

src/components/modal/index.js
```diff
@@ -0,0 +1,57 @@
...
+import CloseIcon from '../icons/close';
+import IconButton from '../../components/icon-button';
+import './modal.css';
+
+/**
+ * Generic Kedro Modal
+ */
+const Modal = ({ title, onClose, visible, message, children }) => {
+  const handleKeyDown = (event) => {
+    if (event.keyCode === 27) {
```
studioswong, line 12: great one for adding this in ( good spot @tynandebold ) . With the introduction of the keycode events into our codebase, we should set up a `handleKeyEvents` function in utils to handle all the key events that are used in our app, especially now that we are migrating kedro-UI into our codebase - the set up of the function can reference the structure as set out in Kedro-UI [here](https://github.com/quantumblacklabs/kedro-ui/blob/master/src/utils/key-events.js). 

Setting this up will also make the future migration of other components from Kedro-ui easier. 😄

src/components/node-list/index.js
```diff
@@ -175,9 +174,9 @@ const NodeListProvider = ({
 
   // Deselect node on Escape key
   const handleKeyDown = (event) => {
-    utils.handleKeyEvent(event.keyCode, {
-      escape: () => onToggleNodeSelected(null),
-    });
+    if (event.keyCode === 27) {
+      onToggleNodeSelected(null);
+    }
```
studioswong, line 179: Nice spot on refactoring this - as suggested, once the `handleKeyEvents` function is set up in utils, we can follow a similar structure as set out before.

studioswong (commented):
Thanks for all the fixes as per our previous comments - great stuff. 

I left a further comment about setting up a `handleKeyEvent` util function to handle the `escape` key, otherwise looks good!

src/components/modal/index.js
rashidakanchwala: hey @studioswong  -- there's a future ticket in the backlog to do that https://jira.quantumblack.com/browse/KED-2985 
I was thinking of doing the migration to utils folder as part of that ticket.

studioswong: I see the utility migration being in the pipeline as a later ticket - that said, I would rather we take the migration of utility functions as part of the end-to-end migration of the component as a complete feature rather than as a separate ticket, particularly given that we only want to migrate the parts within the utility function that we use into the codebase, and doing it seperately will easily follow the trap of migrating unnecessary code ( while also extremely easy to forget to come back to refactor other part of the codebase, especially if the utility migration ticket is picked up by someone else.)

in other words, my recommendation is:
1) set up a `handleKeyEvents` utility function that only tackles the handleKeyDown event as needed here as part of the complete work for this ticket
2) No need to migrate the `handleKeyFunction` later on as it is set up already.

Hope this is clear?

studioswong (approved):
Approving this to move this forward, with the remarks to add in a comment to refactor the `handleKeyDown` event to use the `handleKeyEvent` that will be set up as part of the future migration work.

That said, on revisiting the ticket for the future migration work, I would suggest that ticket is not needed, and instead the migration / rewrite of the utility function should be set up as part of a end-2-end of a component migration to reflect the full functionality of the component. Given this is the first PR for component migration from Kedro-UI, this is definitely a great chance for us to also learn on how to tackle this for future. 

Otherwise, good job on setting up the first component migration from Kedro-UI 👍

src/utils/key-events.js
```diff
@@ -0,0 +1,48 @@
+// Keyboard character codes
+const KEYS = {
+  13: 'enter',
+  27: 'escape',
+  32: 'space',
+  37: 'left',
+  38: 'up',
+  39: 'right',
+  40: 'down',
```
studioswong, line 9: I would delete all of the unneeded key codes.

PR #646, opened 2021-11-24
[KED-2896] Continuously validate graphql structure against the main schema

rashidakanchwala:
## Description

Jira ticket - https://jira.quantumblack.com/browse/KED-2896

## Development notes

We are using this package to compare differences between schema - https://github.com/Ambro17/graphql-schema-diff 

It also has a very good cli view when schema don't match and what needs to be added/removed to make sure the schemas are aligned. 

NOTE -  i had to make some changes to the Apollo schema like runIDs to runIds because on the BE run_IDs is not correct python style. Also had to add some '!' to highlight the need of the fields as they are not optional on BE. 
## Checklist


package/tests/test_api/test_graphql.py
```diff
@@ -513,3 +520,16 @@ def test_graphql_runs_tracking_data_endpoint(
...
         assert response.json() == expected_response
+
+
+class TestGraphQLSchema:
+    def testApolloSchema(self):
+        here = path.abspath(
+            path.join(path.dirname(__file__), "..", "..", "..", "src", "apollo")
+        )
+        changes = []
+        with open(path.join(here, "schema.graphql")) as data:
```
limdauto, line 531: This could be improved with pathlib. Something like

```python
schema_file = Path(__file__).parents[3] / "src" / "apollo" / "schema.graphql")

with schema_file.open() as f:
    ... 
```

package/tests/test_api/test_graphql.py
```diff
@@ -513,3 +520,16 @@ def test_graphql_runs_tracking_data_endpoint(
...
+            apollo_schema = data.read()
+            changes = diff(apollo_schema, str(schema))
+            print_diff(changes)
+        assert changes == []
```
limdauto, line 535: This coul dbe done with 
```suggestion
            changes = diff(apollo_schema, str(schema))
        assert not changes, print_diff(changes)
```
This only prints if the assert fails.

limdauto (commented):
Logic LGTM! Let me sort out the license.

src/apollo/schema.graphql
```diff
@@ -1,17 +1,18 @@
...
   subscription: Subscription
 }
 
 type Query {
-  runsList: [Run]!
-  runMetadata(runIDs: [ID]!): [Run!]!
-  runTrackingData(runIDs: [ID]!, showDiff: Boolean = false): [TrackingDataset]! # showDiff as true will return all tracked fields amongst compared runs
+  runsList: [Run!]!
+  runMetadata(runIds: [ID!]!): [Run!]!
+  runTrackingData(runIds: [ID!]!, showDiff: Boolean! = false): [TrackingDataset!]! # showDiff as true will return all tracked fields amongst compared runs
```
tynandebold, line 11: I'm not sure `showDiff` needs to be required here since we're setting a default value for it.

In your PR description I see you wrote, "Also had to add some '!' to highlight the need of the fields as they are not optional on BE." Maybe they should be changed on the BE then?

tynandebold (approved):
The JS part looks mostly fine to me. I'm going to tentatively approve now with some comments about the required fields.

studioswong (changes requested):
The changes on the `schema.graphql` file looks good to me.

one thing to note:  the schema outlined in `schema.graphql` file should always be consistent with that outlined in `schema.js` for the dev server ( `schema.graphql` is extracted from this file); please update `schema.js` with the proposed changes in`schema.graphql` so that the graphql dev server aligns with the latest schema. 

Thank you 😊

tynandebold:
> The changes on the `schema.graphql` file looks good to me.
> 
> one thing to note: the schema outlined in `schema.graphql` file should always be consistent with that outlined in `schema.js` for the dev server ( `schema.graphql` is extracted from this file); please update `schema.js` with the proposed changes in`schema.graphql` so that the graphql dev server aligns with the latest schema.
> 
> Thank you 😊

@studioswong I was going to write this same comment! Then I poked around a bit and it seemed to me that this file isn't actually being used anywhere. Is that true? I can't find an instance of where it's used.

studioswong:
> @studioswong I was going to write this same comment! Then I poked around a bit and it seemed to me that this file isn't actually being used anywhere. Is that true? I can't find an instance of where it's used.

Great question - I've responded to your question on slack but pasting it here again for reference:
"schema.js was set up as part of the mock server initially where it will be consumed as a  schemaLink within the apollo client to be validated within the mock server - this has been changed to use a typical HttpLink since we migrated to integrate with the real API hence it does not seem to be conusmed in the codebase. That said, this file is kept to give us the flexibility in case we hope to validate and test the schema on a local mock server if needed, and should be kept up to date with schema.graphql as a source of truth - I can further supplement this intent in a comment in a future PR."

src/apollo/schema.graphql
```diff
@@ -1,33 +1,30 @@
-scalar JSONObject # this is a custom type used to handle JSON data
-schema {
-  query: Query
-  subscription: Subscription
-}
+"""Generic scalar type respresenting a JSON object"""
```
tynandebold, line 1: Should be `representing` not `respresenting`.

src/apollo/schema.graphql
```diff
@@ -1,33 +1,30 @@
...
-  runsList: [Run]!
-  runMetadata(runIDs: [ID]!): [Run!]!
-  runTrackingData(runIDs: [ID]!, showDiff: Boolean = false): [TrackingDataset]! # showDiff as true will return all tracked fields amongst compared runs
-}
-
-type Subscription {
-  runAdded(runID: ID!): Run!
+  runsList: [Run!]!
+  runMetadata(runIds: [ID!]!): [Run!]!
+  runTrackingData(runIds: [ID!]!, showDiff: Boolean = false): [TrackingDataset!]!
```
studioswong, line 7: Can we keep the comment for context about the `showDiff` variable?

studioswong (approved):
Thanks for aligning the schemas across both files!

package/requirements.txt
```diff
@@ -11,3 +11,4 @@ pandas>=0.24
 sqlalchemy~=1.2
 strawberry-graphql~=0.87.0
 networkx>=1.0
+gitpython~=3.0
```
limdauto, line 14: this should come out of the box with kedro

limdauto (approved):
Great stuff! A minor comment about the dependency

PR #654, opened 2021-11-25
[KED-2988] Fix for plotly when escape button pressed

rashidakanchwala:
## Description

Jira Ticket - https://jira.quantumblack.com/browse/KED-2988

## Development notes

Besides fixing the bug, I have also raised z-index for plotly modal so it comes on top of the global toolbar. 

## Checklist

### DIFF
--- a/src/components/plotly-modal/index.js
+++ b/src/components/plotly-modal/index.js
@@ -19,39 +19,47 @@ const PlotlyModal = ({ metadata, onToggle, visible }) => {
     return null;
   }
   return (
-    <div className="pipeline-plotly-modal">
-      <div className="pipeline-plot-modal__top">
-        <button
-          className="pipeline-plot-modal__back"
-          onClick={onCollapsePlotClick}
-        >
-          <BackIcon className="pipeline-plot-modal__back-icon"></BackIcon>
-          <span className="pipeline-plot-modal__back-text">Back</span>
-        </button>
-        <div className="pipeline-plot-modal__header">
-          <NodeIcon className="pipeline-plot-modal__icon" icon={nodeTypeIcon} />
-          <span className="pipeline-plot-modal__title">
-            {metadata.node.name}
-          </span>
+    <>
+      {' '}
+      {metadata && (
+        <div className="pipeline-plotly-modal">
+          <div className="pipeline-plot-modal__top">
+            <button
+              className="pipeline-plot-modal__back"
+              onClick={onCollapsePlotClick}
+            >
+              <BackIcon className="pipeline-plot-modal__back-icon"></BackIcon>
+              <span className="pipeline-plot-modal__back-text">Back</span>
+            </button>
+            <div className="pipeline-plot-modal__header">
+              <NodeIcon
+                className="pipeline-plot-modal__icon"
+                icon={nodeTypeIcon}
+              />
+              <span className="pipeline-plot-modal__title">
+                {metadata.node.name}
+              </span>
+            </div>
+          </div>
+          <PlotlyChart
+            data={metadata.plot.data}
+            layout={metadata.plot.layout}
+            view="modal"
+          />
+          <div className="pipeline-plot-modal__bottom">
+            <button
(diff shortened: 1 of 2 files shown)


src/components/plotly-modal/index.js
```diff
@@ -19,39 +19,46 @@ const PlotlyModal = ({ metadata, onToggle, visible }) => {
...
-          <BackIcon className="pipeline-plot-modal__back-icon"></BackIcon>
-          <span className="pipeline-plot-modal__back-text">Back</span>
-        </button>
-        <div className="pipeline-plot-modal__header">
-          <NodeIcon className="pipeline-plot-modal__icon" icon={nodeTypeIcon} />
-          <span className="pipeline-plot-modal__title">
-            {metadata.node.name}
-          </span>
+    <>
+      {metadata && (
```
tynandebold, line 23: This check works most of the time but there seems to be a case where it doesn't:

1. Open viz and click on the "Uplift Report Figures" node
2. Expand the Plotly graph to full-size
3. Hit the escape key
4. Click any other node that doesn't have a Plotly chart in the metadata panel

You should see an error. 

I think if you're just a little more specific we should be set:

```suggestion
      {metadata && metadata.plot && (
```

src/components/plotly-modal/plotly-modal.scss
```diff
@@ -21,7 +21,7 @@
   right: 0;
   bottom: 0;
   left: 0;
-  z-index: 3;
+  z-index: 5;
```
tynandebold, line 24: Given we now have a global toolbar, that toolbar should almost always be visible. 

With that in mind, I think we should keep the `z-index` at 3 and then bump over the the `left` value to `80px`, since that's the width of the global toolbar.

There's actually a SCSS variable you can use for this [here](https://github.com/quantumblacklabs/kedro-viz/blob/6206e54eb4f7380d18c95a69496ddd578c78087e/src/styles/_variables.scss#L68).

tynandebold (changes requested):
Great work! Just two small things and then we're good :)

src/components/plotly-modal/plotly-modal.scss
```diff
@@ -20,7 +20,7 @@
   top: 0;
   right: 0;
   bottom: 0;
-  left: 0;
+  left: 80px;
```
tynandebold, line 23: I would suggest changing this hard-coded value to the variable we have set in the `variables` file, like we do [here](https://github.com/quantumblacklabs/kedro-viz/blob/9f0283c3f4af02c94b258311edf1997a1ad4742d/src/components/experiment-tracking/sidebar/sidebar.scss#L10).

tynandebold (approved):
Works like a charm :)

I've left one small comment and I'll approve now. Great work.

src/components/plotly-modal/index.js
```diff
@@ -19,39 +19,46 @@ const PlotlyModal = ({ metadata, onToggle, visible }) => {
...
-          <BackIcon className="pipeline-plot-modal__back-icon"></BackIcon>
-          <span className="pipeline-plot-modal__back-text">Back</span>
-        </button>
-        <div className="pipeline-plot-modal__header">
-          <NodeIcon className="pipeline-plot-modal__icon" icon={nodeTypeIcon} />
-          <span className="pipeline-plot-modal__title">
-            {metadata.node.name}
-          </span>
+    <>
+      {metadata && metadata.plot && (
```
limdauto, line 23: you can potentially move this check to line 18 and short circuit there. IMO it's cleaner that way.

```
if (!visible.plotModal || metadata?.plot) {
    return null
}
```

limdauto (approved):
LGTM just a minor comment

src/components/plotly-modal/index.js
tynandebold: Optional chaining for the win. Forgot this also works in vanilla JS, not just TS.

@rashidakanchwala I agree with Lim here. This is cleaner.

PR #657, opened 2021-11-29
[KED-2965] Return gitBranch data

rashidakanchwala:
## Description

Based on Jira Ticket - https://jira.quantumblack.com/browse/KED-2965

## Development notes

I have added code to sqlite store which when writing to session_store will take the sha value and get the branch name and add it to the db. 

## Checklist


package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -42,10 +43,22 @@ def to_json(self) -> str:
         """Returns session_store information in json format after converting PosixPath to string"""
         session_dict = {}
         for key, value in self.data.items():
-            if _is_json_serializable(value):
+            # Get branch info through git_sha
+            if key == "git":
+                sha = [v for k, v in value.items() if k == "commit_sha"]
```
MerelTheisenQB, line 48: Why do you need to loop through the values? Can't you just do `value.get("commit_sha")`?

MerelTheisenQB: And what would happen if we don't have any "git" data at all or if the process can't complete for some reason?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -42,10 +43,22 @@ def to_json(self) -> str:
...
+                git = subprocess.run(
+                    ("git", "name-rev", "--name-only", sha[0]),
+                    stdout=subprocess.PIPE,
+                    check=True,
+                )
+                branch = git.stdout.decode("utf-8")
+                value["branch"] = branch
                 session_dict[key] = value
```
MerelTheisenQB, line 56: Is the value here always `_is_json_serializable` or should it still go through that check?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -42,10 +43,22 @@ def to_json(self) -> str:
         """Returns session_store information in json format after converting PosixPath to string"""
         session_dict = {}
         for key, value in self.data.items():
-            if _is_json_serializable(value):
+            # Get branch info through git_sha
+            if key == "git":
+                sha = [v for k, v in value.items() if k == "commit_sha"]
+                git = subprocess.run(
```
limdauto, line 49: Kedro comes out of the box with gitpython installed: https://github.com/quantumblacklabs/kedro/blob/master/requirements.txt#L7 -- can you check to see if they any API we can leverage instead of doing subprocess?

limdauto (commented):
I have a general comment about the approach. We can probably leverage gitpython here.

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: Thanks, i also realised that if we do it using git-python. We don't really need 'sha' lookup. Because branch name gets added during kedro-run to the session_store.db and if i do repo.active_branch, I get the branch name they were working from.

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: i made changes to the code. if there's no git then branch is not added at all.

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: @limdauto  - let me know if this works and I can update the tests

rashidakanchwala: Also, should we write branch_name to the repo when doing kedro run. or should we retrieve branch_name from sha during kedro-viz.
Initialy I was doing it on development environment so ofcourse if i tried to run github commands -- it would run on kedro-viz github rather than kedro-project github. But for production mode it would be fine because both would be run from kedro project github

AntonyMilneQB: @limdauto 

Butting in here because Rashida asked me about something else on this PR and then I just spotted this comment...

Even though gitpython is a requirement of kedro, this doesn't mean that the git executable exists. If it doesn't then `import git` will crash. And if the git executable _does_ exist but we're not in a git repo, this code should be written so it doesn't crash.

So unless git is a hard requirement of kedro-viz (which I don't think it should be), we should probably:
* move the `import git` inline
* wrap the whole thing in some error catching code

See https://github.com/quantumblacklabs/private-kedro/pull/1284 🙂

AntonyMilneQB: Worth checking also what happens if `import git` works but you're not in a repo - what does `git.Repo` do then? Again this should return empty data rather than crash.

rashidakanchwala: I can confirm it now works when you're not in any repo. It just doesn't have git in the session data

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -42,6 +42,12 @@ def to_json(self) -> str:
         """Returns session_store information in json format after converting PosixPath to string"""
         session_dict = {}
         for key, value in self.data.items():
+            if key == "git":
+                import git  # pylint: disable=import-outside-toplevel
```
limdauto, line 46: Let's `try...catch` this. The original concern was not about being outside of a git directory, but more like not having git installed in your system.

rashidakanchwala: sure, also if git is not available then it just won't return have git in the session_store dictionary.

limdauto (approved):
LGTM

PR #663, opened 2021-12-02
[KED-2862] & [KED-2863] Add Mutation to GraphQL Endpoint to update Run Details 

rashidakanchwala:
## Description

Jira Ticket - https://jira.quantumblack.com/browse/KED-2862 and https://jira.quantumblack.com/browse/KED-2863

## Development notes

Please note that front-end will always return Notes, Bookmark, and Title. If the user has not specified any title, or notes. Front-end will return the current note, or title. 
Hence in the BE we don't need to validate what the front-end returns except the type of data and just assign it to the db. 

## Checklist


package/kedro_viz/models/sql_models.py
```diff
@@ -20,3 +21,19 @@ class Config:
...
+
+class UserDetailsModel(Base):
+    """Data model to represent user details data specified in a Kedro Viz Session."""
+
+    __tablename__ = "user_details"
+
+    id = Column(String, ForeignKey(RunModel.id), primary_key=True, index=True)
+    bookmark = Column(Boolean)
+    title = Column(String)
+    notes = Column(String)
```
limdauto, line 34: book mark should be `False` by default. Other fields should be nullable.

limdauto (commented):
A few comments but directionally lgtm 💯

package/kedro_viz/api/graphql.py
```diff
@@ -257,6 +263,88 @@ def run_tracking_data(
...
+    def update_run_details(self, run_id: ID, run_input: RunInput) -> Response:
+        """Updates run details based on run inputs provided by user"""
+        session = data_access_manager.db_session
+        run_data = get_runs([run_id])
+        if not run_data:
+            return BadInputType(
+                run_id=run_id, error_message="Given run_id doesn't exist"
+            )
+        
+        run_input.title = run_data[0].title if not run_input.title else run_input.title
```
limdauto, line 309: Let's pull `run_data[0]` out into a variable. Better yet, you can do 

```
try:
    run_data = next(get_runs([run_id])
except StopIteration:
    return BadInputType(...)
```

package/tests/test_api/test_graphql.py
```diff
@@ -427,33 +427,37 @@ def test_graphql_run_list_endpoint_no_dbsession(self, client):
...
             "kedro_viz.data_access.DataAccessManager.db_session",
             new_callable=PropertyMock,
         ) as mock_session:
             mock_session.return_value = example_db_dataset
             response = client.post(
                 "/graphql",
-                json={"query": "{runMetadata(runIds: [1534326]) {id bookmark}}"},
+                json={
+                    "query": """{runMetadata(runIds: ["%s"]) {id bookmark}}"""
+                    % save_version
```
limdauto, line 442: this can use `f-string` instead of `%`

package/tests/test_api/test_graphql.py
```diff
@@ -427,33 +427,37 @@ def test_graphql_run_list_endpoint_no_dbsession(self, client):
...
         with mock.patch(
             "kedro_viz.data_access.DataAccessManager.db_session",
             new_callable=PropertyMock,
         ) as mock_session:
             mock_session.return_value = None
             response = client.post(
                 "/graphql",
-                json={"query": "{runMetadata(runIds: [1534326]) {id bookmark}}"},
+                json={
+                    "query": """{runMetadata(runIds: ["%s"]) {id bookmark}}"""
```
limdauto, line 458: ditto

package/tests/test_api/test_graphql.py
```diff
@@ -522,8 +526,95 @@ def test_graphql_runs_tracking_data_endpoint(
...
+                            __typename
+                            ... on UpdateRunDetailsSuccess {
+                            runDetails
+                            }
+                            ... on BadInputType {
+                            runId
+                            errorMessage
+                            } }
+                            }"""
+                    % save_version
```
limdauto, line 558: let's use f-string please

package/tests/test_api/test_graphql.py
```diff
@@ -522,8 +526,95 @@ def test_graphql_runs_tracking_data_endpoint(
...
+            )
+            assert response.json() == {
+                "data": {
+                    "updateRunDetails": {
+                        "__typename": "UpdateRunDetailsSuccess",
+                        "runDetails": {
+                            "run_id": "2021-11-02T18.24.24.379Z",
+                            "bookmark": False,
+                            "title": "Hello Kedro",
+                            "notes": "There are no notes",
```
limdauto, line 569: nit: the `there are no notes` example is a bit counter intuitive in this positive, happy path.

limdauto (approved):
Awesome. Mostly nits comment. Otherwise LGTM

studioswong (changes requested):
Thanks for setting this up! I haven't test-drive the new mutation endpoints yet, just a quick comment on the error type ( see below). 

It would be great if you could also list the set of error messages for us to expect for the different error type ( does the error message specifically states the input field causing the error amongst the 3 fields within the `RunInput` type? What kind of `errorMessage` should the FE be expecting for bad input types? It would be helpful to list those expected error messages in the QA section so FE can know what to expect as part of our QA work. 

On another note, putting in a 'request change' here to remind the team that this should only be merged in after the latest patch fix release, alongside the other FE PR for mutation to avoid any potential breaking changes, and to allow room for us to further test in its integration with the existing FE work.

studioswong (commented):
I've tested the mutation endpoints via graphiql, so far the initial attempts for the mutation endpoints works 👍 

One suggestion on the error messages: 
is it possible to return more meaningful error messages other than the message below? something such as `cannot input non-string values for \'notes' field`

The FE will ideally do data validation before the submission of the form as well, but this useful error message will be helpful for us to display the error on the FE. 

image

src/apollo/schema.js
```diff
@@ -5,8 +5,23 @@ import GraphQLJSON, { GraphQLJSONObject } from 'graphql-type-json';
 import gql from 'graphql-tag';
 
 const typeDefs = gql`
+  type BadInputType {
+    runId: ID!
+    errorMessage: String!
+  }
```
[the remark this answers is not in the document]
tynandebold: Agree on the name change Susanna suggests above.

tynandebold (approved):
The JS looks good to me! Just one name change that Susanna also suggested and otherwise all good.

package/kedro_viz/api/graphql.py
rashidakanchwala: hi, i wasn't able to get this working. It was throwing me NoneType is not iterable error. I tried to fix it via stackoverflow but couldn't get it to work. Please feel free to take over and push the above.

limdauto:
@studioswong the error that you specifically mentioned happen at the type level and handled by graphql, not our business logic. The one error that we handle ourselves we have included the context, i.e. the run_id that doesn't exist. We can't do much about the validation error at a type level. I'd imagine the error is already quite explicit in this case.

limdauto:
Also regarding merge order, there is no harm in merging right now either. The frontend isn't consuming this backend change, so changes are not affecting users at the moment.

src/apollo/schema.js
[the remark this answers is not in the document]
studioswong: Great points raised both - I started writing my reply but stopped mid way for the release but Tynan has just summed up my mind perfectly.

@limdauto I agree 100% with your suggestion on the union type - the suggestion to rename this to a more generic error type was an intermediate solution to cater for the current situation that both `InvalidRun` and `InvalidInput` errors are returned under the same `BadInputType` ( as confirmed by having tried out both scenerios on graphiql), and your suggestion is the ideal way to go.

I just saw that this has been updated to `updateRunDetailsFailure` - thank you. 

@limdauto I presume your plans for the future union would be ilke this?

```type UpdateRunDetailsFailure = InvalidRunError | UnauthorisedRequestError | InvalidInputError```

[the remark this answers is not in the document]
limdauto: Alright thanks both. Sounds good to me. We use this as intermediate and move to union type later.

studioswong:
Thanks for the new updates @limdauto 👍 

I've tested the mutation and so far it works well for updating the title and notes field. 

2 things I've observed:

1. updating the `bookmark` field with `false` does not update the bookmark field to false ( it does update from `false` to `true` otherwise):

image

2. The error returned doesn't seem to match the `updateRunsDetailsFailure` type as the `runID` field doesn't get returned in the response - see below
image

one non-relevant question - I supposed subscription is not set up yet as part of this ticket? (asking this as I observed that values are't being updated automatically on the app which I think the apollo client on the FE should pick it up out of the box. ) Not that it's needed for this ticket at all 😉

limdauto:
@studioswong thanks. Will check the bookmark.

Re error: There are 2 levels of errors here. The first level is type error, which is what you are seeing. It says the `notes` input field needs to be a string. It doesn't even hit our business logic when a type error occurs as it happens at the protocol layer.

The second level is business logic exception, which is the run_id doesn't exist. Try punching it correct type for notes or leaving it out altogether.

studioswong:
> @studioswong thanks. Will check the bookmark.
> 
> Re error: There are 2 levels of errors here. The first level is type error, which is what you are seeing. It says the `notes` input field needs to be a string. It doesn't even hit our business logic when a type error occurs as it happens at the protocol layer.
> 
> The second level is business logic exception, which is the run_id doesn't exist. Try punching it correct type for notes or leaving it out altogether.
> 
> Re subscription: no, it's done separately but maybe we are talking about different things here. Once you send the mutation and receive the response back, does it not reflect in the UI?

any chance we can have type errors resolved as part of the business logic exception? That way the FE can do error handling according to an aligned error response. Currently it will be much harder to catch type error responses, which in fact would be an error much likely to occur than invalid runID errors. 

Also, I notice that, given that the type errors and invalid ID errors are resolved in different levels, it seems that only `type` errors will be caught in the case that both errors are happening. ( see below)

image

In this case, putting both types of errors to be resolved in the business logic has the added benefit that it can return both error types in case more than 1 type of error exists.

tynandebold:
> > @studioswong thanks. Will check the bookmark.
> > Re error: There are 2 levels of errors here. The first level is type error, which is what you are seeing. It says the `notes` input field needs to be a string. It doesn't even hit our business logic when a type error occurs as it happens at the protocol layer.
> > The second level is business logic exception, which is the run_id doesn't exist. Try punching it correct type for notes or leaving it out altogether.
> > Re subscription: no, it's done separately but maybe we are talking about different things here. Once you send the mutation and receive the response back, does it not reflect in the UI?
> 
> any chance we can have type errors resolved as part of the business logic exception? That way the FE can do error handling according to an aligned error response. Currently the type error response will require a different way of error handling than the response from the invalidID response on the FE, with the 'out of the box' graphql type error much less helpful than the `UpdateRunDetailsFailure` response object ( and unfortunately, type errors are also the more common type of errors as well).
> 
> Also, I notice that, given that the type errors and invalid ID errors are resolved in different levels, it seems that only `type` errors will be caught in the case that both errors are happening. ( see below)
> 
> image
> 
> In this case, putting both types of errors to be resolved in the business logic has the added benefit that it can return both error types in case more than 1 type of error exists.

Don't we lose the for-free benefit of GraphQL's type checking if we push it down into the business logic? That seems really counterintuitive to me. I thought one reason GraphQL was chosen was for it's great out-of-the-box [validation](https://graphql.org/learn/validation/) and so FE/BE wouldn't have to maintain those checks.

And that second example you outline in the screenshot is the expected behavior isn't it? The type checks throw first (plus I think passing `false` in the notes field will be an impossibility to do in the frontend anyways).

studioswong:
> Don't we lose the for-free benefit of GraphQL's type checking if we push it down into the business logic? That seems really counterintuitive to me. I thought one reason GraphQL was chosen was for it's great out-of-the-box [validation](https://graphql.org/learn/validation/) and so FE/BE wouldn't have to maintain those checks.
> 
> And that second example you outline in the screenshot is the expected behavior isn't it? The type checks throw first (plus I think passing `false` in the notes field will be an impossibility to do in the frontend anyways).

yea you're absolutely right about `false` being (almost) impossible to send from the FE - here I am taking the viewpoint purely from the performance of the BE towards more robust error handling regardless of any precautions from the FE because, let's face it, theoretically the FE can change anytime 😆  ( and of course, the FE will have its own form error handling as well.) 

You are right about the 'for-free' benefit of GraphQl's type checking too - I am not advocating for us to scrap it, but to just relax it for the input fields so that it could be pushed on to be resolved and return a `InvaidInput` type response to enable more meaningful action from the FE, such as identifying the input field that has the wrong input type. ( currently the error message `String cannot represent a non-string value` doesn't help the FE identify which input field produced the type error.) 

While we are on the topic, a suggestion for `invalidInput` error type would be kind of like this:

``` type UpdateRunDetailsFailure {
    runId: ID!
    inputField: String!
    errorMessage: String!
  }```

@limdauto I know that this would entail more work and refactor on your side and am happy for us to save this change for a future workshop on graphql error handling down the line.

One more thing - I noticed that sending an empty string for both the `title` and `notes` field will not induce any change on those fields - I might suggest for this to be possible for at least the `notes` field as it essentially allows the user to 'delete' the notes. ( currently there is no 'delete note' button in the design, but the user can simply enter an empty field in the modal for the notes which would mean their hopes to delete the old notes.)

 Happy to approve this once the bookmark issue is resolved 😉

studioswong (approved):
Works well, thanks for making all those changes @limdauto . 
I also have to note how I love falling back to the default title when an empty string gets sent for the title field 👍

PR #664, opened 2021-12-02
KED-3009: refactor Primary Toolbar setup

studioswong:
## Description

resolves [KED-3009](https://jira.quantumblack.com/browse/KED-3009)

This PR is raised as part of the work to implement the toggle changes user journey. 

The current setup of the primary toolbar is mainly catered for the usage on the flowchart; previously it was set up to be reused by the experiment tracking view. However, with the addition of specific buttons for the experiment tracking feature, meanwhile with the need to dynamically render different set of toolbar buttons to cater for the needs of single vs comparison view for experiment-tracking primary toolbar, there is a need to extract the toolbar setup for both views, while setting up a reusable primary toolbar component that is reused amongst the toolbar setup for both views. 

Currently both sidebars for the experiment-tracking view and flowchart view are controlled by the same global state, which is not ideal given they are completely different SPAs with different sets of information. I have furthermore set up the visibility of the sidebar for the experiment-tracking view to be controlled by a separate local state rather than with the global state so that the visibility of both sidebars are controlled separately. 

## Development notes

I have introduced the following in this PR:
1. Refactor the current `primary-toolbar` component to be a reusable component, stripping out the previous custom setup for the flowchart toolbar, such as the redux state setup. 
2. Refactored and introduced more specific tests for the primary-toolbar component
3. Introduced the `flowchart-primary-toolbar` component to house all the specific set of toolbar buttons as well as redux state props for the flowchart view.
4. Set up tests for `flowchart-primary-toolbar`
5. Introduced the `experiment-primary-toolbar` to house all future buttons to be developed for the experiment tracking view. ( Please note I had not introduced any tests given it is only rendering the basic collapsable component - I will further add on more tests for this in my future PR for setting up the toggle-changes user journey.)
6. removed the sidebar component under `experiment-tracking`, given that we are only using the main sidebar for both views. ( this could be introduced back in shall the sidebar grows in complexity, similar to this exercise.)

## QA notes

I had added all relevant tests; please test this locally with your experiment-tracking data to ensure the primary-toolbar works well with both flowchart and experiment tracking views. 

## Checklist


src/components/experiment-wrapper/index.js
```diff
@@ -11,6 +11,7 @@ import './experiment-wrapper.css';
 const MAX_NUMBER_COMPARISONS = 2; // 0-based, so three
 
 const ExperimentWrapper = ({ theme }) => {
+  const [sidebarVisible, setSidebarVisible] = useState(true);
```
tynandebold, line 14: A nice convention for booleans is using **is** at the front of variable names. What do you think about the following?

```suggestion
  const [isSidebarVisible, setIsSidebarVisible] = useState(true);
```

src/components/experiment-tracking/experiment-primary-toolbar/index.js
```diff
@@ -0,0 +1,16 @@
...
+import PrimaryToolbar from '../../primary-toolbar';
+
+export const ExperimentPrimaryToolbar = ({
+  sidebarVisible,
+  setSidebarVisible,
+}) => {
+  return (
+    <PrimaryToolbar
+      visible={{ sidebar: sidebarVisible }}
+      onToggleSidebar={setSidebarVisible}
```
tynandebold, line 11: Can we try and sort our props alphabetically? It makes things easier to read if this convention is followed throughout the codebase. 

There are other instances of where this can be done in this PR, too, but I'll only write it here once.

src/components/primary-toolbar/index.js
```diff
@@ -46,59 +30,9 @@ export const PrimaryToolbar = ({
...
-  onToggleSidebar: (visible) => {
-    dispatch(toggleSidebar(visible));
-  },
-  onToggleTextLabels: (value) => {
-    dispatch(toggleTextLabels(Boolean(value)));
-  },
-});
-
-export default connect(mapStateToProps, mapDispatchToProps)(PrimaryToolbar);
+export default PrimaryToolbar;
```
tynandebold, line 38: Should we rename this file and component to `BasePrimaryToolbar` or `PrimaryToolbarBase`? 

I think that could add a little bit of clarity and qualification, especially since the other`PrimaryToolbar` components (flowchart and experiment) have a descriptor.

src/components/flowchart-primary-toolbar/index.js
```diff
@@ -0,0 +1,89 @@
...
+        />
+        <IconButton
+          ariaLabel="Export graph as SVG or PNG"
+          className={'pipeline-menu-button--export'}
+          onClick={() => onToggleExportModal(true)}
+          icon={ExportIcon}
+          labelText="Export visualisation"
+          visible={visible.exportBtn}
+        />
+      </>
```
tynandebold, line 59: I don't think this needs to be in a fragment (`<> </>`).

tynandebold (changes requested):
Looks good and works as it should! :)

I've left a few comments about naming and other conventions I'd like us to try and start following more thoroughly. Nothing major at all.

One thing I'd add is I'm not convinced this change is strictly necessary at this point in time. We had a mechanism for sharing the `<PrimaryToolbar />` component in place already: 

```jsx
{isExperimentView ? (
  <IconButton
    ariaLive="Edit run details"
    className={'pipeline-menu-button--labels'}
    onClick={() => showRunDetailsModal(true)}
    icon={PencilIcon}
    labelText={`Edit details`}
  />
) : (
  <>
    <IconButton
      ariaLive="polite"
      className={'pipeline-menu-button--labels'}
      onClick={() => onToggleTextLabels(!textLabels)}
      icon={LabelIcon}
      labelText={`${textLabels ? 'Hide' : 'Show'} text labels`}
      visible={visible.labelBtn}
    />
    ...(etc)...
  </>
)}
```

For the icons we have to add for experiment tracking, those would live under the first condition, as in the example above, and not much logic would've resided in this file. That's good, because the file wouldn't have grown too much in scope. I've also started working with the pattern above in the branch for my ticket, so now I'll have to refactor a bit when I merge yours in.

It's definitely not wasted work though, and I think there's an opportunity here for clearer communication among the team when we want to make a change like this that could affect what's being done in each other's tickets.

studioswong:
> One thing I'd add is I'm not convinced this change is strictly necessary at this point in time. We had a mechanism for sharing the `<PrimaryToolbar />` component in place already:
> 
> ```js
> {isExperimentView ? (
>   <IconButton
>     ariaLive="Edit run details"
>     className={'pipeline-menu-button--labels'}
>     onClick={() => showRunDetailsModal(true)}
>     icon={PencilIcon}
>     labelText={`Edit details`}
>   />
> ) : (
>   <>
>     <IconButton
>       ariaLive="polite"
>       className={'pipeline-menu-button--labels'}
>       onClick={() => onToggleTextLabels(!textLabels)}
>       icon={LabelIcon}
>       labelText={`${textLabels ? 'Hide' : 'Show'} text labels`}
>       visible={visible.labelBtn}
>     />
>     ...(etc)...
>   </>
> )}
> ```
> 
> For the icons we have to add for experiment tracking, those would live under the first condition, as in the example above, and not much logic would've resided in this file. That's good, because the file wouldn't have grown too much in scope. I've also started working with the pattern above in the branch for my ticket, so now I'll have to refactor a bit when I merge yours in.
> 
> It's definitely not wasted work though, and I think there's an opportunity here for clearer communication among the team when we want to make a change like this that could affect what's being done in each other's tickets.

Thanks for your review! 

Having been exploring the current setup of the primary toolbar, and especially after initially setting up the additional buttons for experiment-tracking, these are the factors that prompted me to make this change:

1. Seperation of data ingestion setup
Given the different data ingestion setup, Experiment view buttons will consume data and props from a different setup ( hooks + graphql data) instead of the redux setup for flowchart. Currently the setup is tailored for redux data ingestion - it works now given that there is no props to be passed in for the experiment view, but once we start passing in props from hooks this will bring this into a messy hybrid state which I am trying to avoid.

2. Maintainability + seperation of context / logic
As mentioned in the description, within Experiment view there is 2 separate sets of tool buttons for single view and comparison view to be built later on - this setup seems fine now when there is only 1 button, but imagine building on all those different conditions for the toolbar for just the experiment tracking page 😅 This helps us to keep the specific logic for the toolbar separate while allowing us to reuse part of the toolbar setup.

3. Reusability 
As with most react best practices, separating the collapsable primary toolbar out as a dummy reusable component allows reusability. This is also a lesson from having worked on this codebase - currently our component setup is very custom with very low reusability - with the introduction of an extra experiment-tracking page setup, this is a good time for us to refactor our component setup, rather than building on top of the madness of the current setup. 

This is ideally to be done initially during the MVP ( I recall you had the exercise of moving this from the experiment-tracking component to this setup), which I wasn't aware of the above problems it poses until coming down to actually revisit the design and think about the mechanics of the addition of the buttons. 

Noted on the communication front - I thought the benefits and the advantages of this setup is a natural choice, and I wasn't aware that you were working on anything that would touch this part of the code as I did mentioned this change during standup. if you share your working branch, I can also help to resolve any merged conflicts, if that helps? 

Great discussion as always!

src/components/experiment-wrapper/index.js
studioswong: sure, thanks for the suggestion! I'll make the change

src/components/experiment-tracking/experiment-primary-toolbar/index.js
studioswong: This is out of context but great suggestion regardless - this isn't a practice we follow in the codebase but this would be a good suggestion to be brought up in the future team norms / coding best practice discussion? We can then look into setting a PR to tackle this and enforcing this in future PR reviews after the discussion.

src/components/primary-toolbar/index.js
studioswong: In looking at our codebase, this isn't a naming convention we follow - our reusable components adopts the typical naming as is. 

In this case, the descriptor is only added to specific setup (`experiment-tracking-primary-toolbar` and `flowchart-primary-toolbar`), which this setup is self-explanatory enough to differentiate the intent of the Primary toolbar? 

We should also revisit our current component file setup so we can move reusable components into a separate folder, which by then this will be even more explantory.

tynandebold (approved):

PR #665, opened 2021-12-07
[KED-2997] Create rename run and edit notes UI behavior

tynandebold:
## Description

Resolves [KED-2997](https://jira.quantumblack.com/browse/KED-2997).

## Development notes

To do this, I've created a couple of new components:

Front-end : 

- A `<RunDetailsModal/>` component, based on the `<SettingsModal/>` that we already had, except this new one isn't connected in any way to Redux. It does share a lot of styles with the `<SettingsModal/>`, as I'm importing that CSS file.
- A new `<Input/>` component that lives within a new `components/ui` folder. This component actually uses a `<textarea/>` HTML element under the hood, which gives good flexibility. And further, going forward, I think we can start to move some UI-only components into this folder and create any new UI components here, as needed.
- A `<Pencil/>` icon component.

Backend : 
UpdateRunDetailSuccess now returns a Run with the latest changes requested by the user instead of returning a JSONObject as it did previously. 

## QA notes

Check out the branch and click on the Edit details pencil icon or the run metadata title/notes text to trigger the modal.

## Checklist


src/components/ui/input/index.js
```diff
@@ -0,0 +1,69 @@
...
+  useRef,
+  useState,
+} from 'react';
+
+import './input.css';
+
+const MIN_HEIGHT = 20;
+
+const Input = ({
+  characterLimit = false,
```
studioswong, line 14: if we set this to 0 by default then we would not need the seperate check for `isLimitSet`
```suggestion
  characterLimit = 0,
```

src/components/ui/input/index.js
```diff
@@ -0,0 +1,69 @@
...
+  const handleChange = (e) => {
+    setLimitedLengthText(e.target.value);
+  };
+
+  const setLimitedLengthText = useCallback(
+    (text) => {
+      isLimitSet
+        ? setValue(text.slice(0, characterLimit))
+        : setValue(text.slice(0));
+    },
```
studioswong, line 44: following my comment above, if `characterLimit` is set to 0 by default then we could simplify this to 
```suggestion
    (text) => setValue(text.slice(0, characterLimit));
```

src/components/experiment-tracking/run-metadata/index.js
```diff
@@ -20,6 +25,11 @@ const RunMetadata = ({ isSingleRun, runs = [] }) => {
     setToggleNotes({ ...toggleNotes, [index]: !toggleNotes[index] });
   };
 
+  const onTitleOrNoteClick = (id) => {
+    setSelectedMetadataRunId(id);
+    setShowRunDetailsModal(true);
+  };
+
```
studioswong, line 32: Similar to my comment above, I would directly assign the run metadata here and remove the need for the `selectedMetadataRunId` hook.

```suggestion
  const onTitleOrNoteClick = (run) => {
    setSelectedRunMetadata(run);
    setShowRunDetailsModal(true);
  };

```

src/components/experiment-tracking/details/index.js
```diff
@@ -25,20 +35,50 @@ const Details = ({ selectedRuns, sidebarVisible }) => {
...
+
+      setSelectedRunMetadata(metadata);
+    } else {
+      const metadata = runMetadata.find(
+        (run) => run.id === selectedMetadataRunId
+      );
+
+      setSelectedRunMetadata(metadata);
+    }
+  }, [enableComparisonView, runMetadata, selectedMetadataRunId, selectedRuns]);
```
studioswong, line 54: we would likely not need this if we omit the `selectedMetadataRunId` hook and assign the run directly within the click callback

studioswong (changes requested):
Great stuff, the UI works great!

I've added some futher comments, in particular some suggestions of taking out the need of the `selectedMetadataRunId` hook and assigning the run metadata directly on click. 

Adding in a request change to also remind us not to merge this in until the mutation endpoints are merged in as a completed feature for this user journey.

src/components/ui/input/index.js
tynandebold: I see what you're saying but I don't think this is going to work. If slice is set to `slice(0, characterLimit)` when the `characterLimit = 0`, then the method [starts and ends at 0](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice), meaning nothing is returned. So I believe we do need this, the check above, and the ability to set the default value to `false`.

src/components/experiment-tracking/details/index.js
```diff
@@ -1,16 +1,26 @@
...
-const Details = ({ selectedRuns, sidebarVisible }) => {
+const Details = ({
+  enableComparisonView,
+  selectedRuns,
+  setShowRunDetailsModal,
+  showRunDetailsModal,
+  sidebarVisible,
+  theme,
+}) => {
+  const [selectedRunMetadata, setSelectedRunMetadata] = useState(null);
```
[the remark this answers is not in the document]
tynandebold: Good idea. I've changed it to `runMetadataToEdit`. Hope that's better!

src/components/experiment-tracking/details/index.js
```diff
@@ -1,16 +1,26 @@
...
+  const [selectedMetadataRunId, setSelectedMetadataRunId] = useState(null);
```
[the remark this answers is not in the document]
tynandebold: I've removed this one now, as suggested.

src/components/experiment-tracking/run-metadata/index.js
tynandebold: Brilliant suggestion. Thanks for this.

src/components/experiment-tracking/details/index.js
tynandebold: Brilliant suggestion here as well. I've implemented.

src/components/experiment-tracking/run-metadata/index.js
```diff
@@ -34,20 +46,27 @@ const RunMetadata = ({ isSingleRun, runs = [] }) => {
             className={classnames('details-metadata__run', {
               'details-metadata__run--single': isSingleRun,
             })}
-            key={run.title + i} // note: this should revert back to use gitSha once the BE returns the actual value
+            key={run.title + i} // Note: change to gitSha once BE returns the value
```
rashidakanchwala, line 49: hi, BE does return gitSha and also gitBranch

src/components/experiment-tracking/run-details-modal/index.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import Button from '@quantumblack/kedro-ui/lib/components/button';
+import Modal from '@quantumblack/kedro-ui/lib/components/modal';
```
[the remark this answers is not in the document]
tynandebold: Yes, absolutely we can.

src/components/experiment-tracking/run-details-modal/run-details-modal.test.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import RunDetailsModal from './index';
+import Adapter from 'enzyme-adapter-react-16';
+import { configure, mount, shallow } from 'enzyme';
+
+configure({ adapter: new Adapter() });
+
+describe('SettingsModal', () => {
```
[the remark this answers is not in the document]
tynandebold: Good catch. Fixed.

src/components/experiment-tracking/run-metadata/index.js
tynandebold: Good call. I actually changed it to `runId`. That seems safer.

src/components/experiment-tracking/run-details-modal/index.js
```diff
@@ -0,0 +1,96 @@
...
+          <Input
+            defaultValue={runMetadataToEdit?.title}
+            onChange={(value) => onChange('title', value)}
+            size="large"
+          />
+        </div>
+        <div className="pipeline-settings-modal__content pipeline-settings-modal__content--short">
+          <div className="pipeline-settings-modal__header">
+            <div className="pipeline-settings-modal__name">
+              Notes (this is searchable)
```
studioswong, line 63: Let's take out `this is searchable` for now until we have implemented the search feature.

src/components/experiment-tracking/run-metadata/index.js
```diff
@@ -105,9 +119,12 @@ const RunMetadata = ({
                   <td>
                     <p
                       className="details-metadata__notes"
+                      onClick={() => onTitleOrNoteClick(run.id)}
                       style={toggleNotes[i] ? { display: 'block' } : null}
                     >
-                      {sanitiseEmptyValue(run.notes)}
+                      {run.notes !== ''
+                        ? run.notes
+                        : '- Add notes here (Notes are searchable)'}
```
studioswong, line 127: similar to above, let's take out the `Notes are searchable` phrase until we have implemented the search feature?

tynandebold: Removed.

src/components/experiment-tracking/run-details-modal/index.js
tynandebold: It's gone now.

studioswong:
Thanks for the efforts for aligning on the mutation @tynandebold and @rashidakanchwala.

I tested this branch this morning and just now with the new changes that was merged in earlier today, and noticed the following: 

1) The mutation doesn't seem to be working ( see below screenshot.) I am guessing this might be fixed by the latest fix to the BE that @rashidakanchwala might push to this branch? 

image

2) This is likely a problem on the FE side - Once submitting a mutation request that failed, clicking on the rename module for other runs will still display the error message ( 'Couldn't update run details. ....' ) 

3) This is also likely to be a problem on the FE side - On testing this branch earlier without the latest BE changes (i.e when the mutation is still working), I notice that, on having submitted the mutation successfully while having the rename module still on, the run details on the module jumps to the first run on the sidebar. (I would've capture this error in a gif, but am not able to reproduce this given the existing mutation errors. We can test this and revisit this once the mutation problems are all clear.)

studioswong:
2 more comments about the behaviour of the FE: 

1. As per the design [here](https://app.zeplin.io/project/5d5ea7a05efca76a74f7d0ea/screen/6184257874e55631904ad216), the default 'Add here' phrase under the notes input field should not be counted as part of the input characters
image

image

2. Some weird behaviour I observed - On trying to edit the notes and receiving an error, the note editing field persists with the last entry on clicking a new run - see below.
edit-notes

rashidakanchwala (approved):
Hi, looks great and works well. Thanks @tynandebold 

Would like @limdauto to review the BE work. 

Also nit, currently Apply Changes doesn't close modal. It felt more natural to me it would. But there's no requirements of such in the design.

studioswong:
@tynandebold on checking out this branch and running the app I've been getting the following error - looks like it's worth revisiting? 
image

tynandebold:
> @tynandebold on checking out this branch and running the app I've been getting the following error - looks like it's worth revisiting? image

@rashidakanchwala was getting the same thing yesterday. I think you just need to restart all servers/`npm install`/clear cache/etc. It should work after that. Not exactly sure what Rashida did but she got it working after seeing the same error.

studioswong:
yes indeed weirdly an `npm install` does the trick - thank you. 

The mutation is working well now, and I see that the word count and error message issue has been solved - many thanks for that!

One more observation, on having fill in the note editing field and selecting the cancel option, the previous note entry persists on the modal for a different run ( see below gif)
ezgif com-gif-maker (1)

Sorry to be thick - We are almost there!

.circleci/config.yml
```diff
@@ -125,8 +125,8 @@ commands:
       - checkout
       - restore_cache:
           keys:
-            - v${CACHE_VERSION}-dependencies-{{ checksum "package.json" }}
-            - v${CACHE_VERSION}-dependencies-
+            - v${CACHE_VERSION}-{{ arch }}-dependencies-{{ checksum "package.json" }}
+            - v${CACHE_VERSION}-{{ arch }}-dependencies-
```
studioswong, line 129: are these changes relevant to this PR?

tynandebold: Sort of. They're an addendum to what @limdauto did yesterday.

limdauto (approved):
LGTM!

.circleci/config.yml
limdauto: Yea this is needed to separate `linux` cache from `windows` cache in CI.

studioswong (approved):
works well for me now - let's get this in once the CI is all clear 😉

PR #667, opened 2021-12-09
[KED-3012] Build out 'Show Changes' user journey

studioswong:
## Description

Resolves [KED-3012](https://jira.quantumblack.com/browse/KED-3012). 

This ticket is the implementation of the 'show changes' user journey in displaying the blue trend arrows that is displayed under comparison view. 

image

## Development notes

There are a series of related changes to the existing experiment tracking components in enabling this functionality:
- addition of hooks and local states to track both the 'showChanges' and 'PinnedRun' states
- addition of related icons for enabling / disabling show changes and identifying the pinned runs, and the trend arrows
- addition of related buttons in both the run-metadata and run-dataset tables
- addition of related style changes
- related tests

On de-selecting a run on the sidebar that is a selected pinned run, the left-most displayed run will automatically assigned as the pinned run. 

Please note that, while working on this feature, I discovered a bug related to the display of the data on the dataset table, which is caused by the order of the runsIDs in the `selectedRuns` array. I had also put in a fix in this PR to ensure that the runsID within the `selectedRuns` array matches the order of the sequence of display of the experiment tracking details. ( see screenshots explaining the bug below)

value bug 01

value bug 02

## QA notes

I had added related tests to test the new functionality. 

## Checklist

- [xx] Read the [contributing](/CONTRIBUTING.md) guidelines


studioswong:
Please note that currently some tests related to the installation of Kedro-viz dependencies are failing on the CI - see below. 
All JS tests should be passing, as can be verified by running the local tests. 

image

tynandebold (commented):
Hey just a quick comment off the top: the light theme isn't quite right. The icons aren't changing color: 

image

src/components/icons/non-selected-pin.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const NonSelectedPinIcon = ({ className }) => (
```
tynandebold, line 3: I think `UnselectedPinIcon` is stronger than `NonSelectedPinIcon`.

src/components/icons/non-selected-pin.js
```diff
@@ -0,0 +1,14 @@
...
+  <svg className={className} viewBox="0 0 24 24">
+    <path
+      d="M12 4a8 8 0 1 1 0 16 8 8 0 0 1 0-16zm0 2a6 6 0 1 0 0 12 6 6 0 0 0 0-12zm0 4a2 2 0 1 1 0 4 2 2 0 0 1 0-4z"
+      fill="#FFF"
+      fillRule="nonzero"
+      fillOpacity=".55"
```
tynandebold, line 9: From what I've seen it seems like we try to leave stying off icons and handle that with class names. This is why the colors aren't changing when I toggle the theme.

src/components/icons/pin-arrow.js
```diff
@@ -0,0 +1,20 @@
...
+    'm10 4.167 4.583 4.583-1.146 1.146-2.604-2.604v9.375H9.167V7.292L6.563 9.896 5.417 8.75z',
+  // down arrow icon
+  downArrow:
+    'M10 15.833 5.417 11.25l1.145-1.146 2.605 2.604V3.333h1.666v9.375l2.604-2.604 1.146 1.146z',
+};
+
+const PinArrowIcon = ({ className, icon }) => {
+  return paths[icon] ? (
+    <svg className="dataset-arrow-icon" viewBox="0 0 24 24">
+      <path d={paths[icon]} fill="#00BCFF" fillRule="evenodd" />
```
tynandebold, line 15: Don't think this should have a `fill` color.

src/components/icons/pin-arrow.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+export const paths = {
+  // up arrow icon
+  upArrow:
+    'm10 4.167 4.583 4.583-1.146 1.146-2.604-2.604v9.375H9.167V7.292L6.563 9.896 5.417 8.75z',
+  // down arrow icon
```
tynandebold, line 7: Think you can remove these comments, as the object property names are self explanatory.

src/components/icons/selected-pin.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const SelectedPinIcon = ({ className }) => (
+  <svg className={className} viewBox="0 0 24 24">
+    <path
+      d="M12 4a8 8 0 1 1 0 16 8 8 0 0 1 0-16zm0 4a4 4 0 1 0 0 8 4 4 0 0 0 0-8z"
+      fill="#00BCFF"
```
tynandebold, line 7: Remove `fill` color.

src/components/icons/show-changes.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const ShowChangesIcon = ({ className }) => (
+  <svg className={className} viewBox="0 0 24 24">
+    <g fill="#FFF" fillRule="evenodd">
```
tynandebold, line 5: `Fill` color here can be removed too.

src/components/icons/up-arrow.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const UpArrowIcon = ({ className }) => (
+  <svg className={className} viewBox="0 0 24 24">
+    <path
+      d="m10 4.167 4.583 4.583-1.146 1.146-2.604-2.604v9.375H9.167V7.292L6.563 9.896 5.417 8.75z"
+      fill="#00BCFF"
```
tynandebold, line 7: And here.

src/components/experiment-tracking/experiment-primary-toolbar/index.js
```diff
@@ -1,15 +1,31 @@
...
       visible={{ sidebar: sidebarVisible }}
       onToggleSidebar={setSidebarVisible}
-    />
+    >
+      <IconButton
+        ariaLive="polite"
+        className={'pipeline-menu-button--labels'}
+        onClick={() => setEnableShowChanges(!enableShowChanges)}
+        icon={ShowChangesIcon}
+        labelText={`${enableShowChanges ? 'Disable' : 'Enable'} Show Changes`}
```
tynandebold, line 24: We generally do sentence case here, so I think this should be:

```suggestion
        labelText={`${enableShowChanges ? 'Disable' : 'Enable'} show changes`}
```

For example:

image

src/components/experiment-tracking/run-metadata/run-metadata.scss
```diff
@@ -12,6 +15,28 @@
...
+.details-metadata__title {
+  display: flex;
+  justify-content: space-between;
+}
+
+.details-matadata__buttons {
+  list-style: none;
+  display: flex;
+  padding: 0px 0px 0px 0px;
+  margin: 0px 0px 0px 0px;
```
tynandebold, line 27: This can just be `padding: 0;` for brevity. Same for `margin`.

src/components/experiment-tracking/run-metadata/run-metadata.scss
```diff
@@ -12,6 +15,28 @@
   }
 }
 
+.details-metadata__title {
+  display: flex;
+  justify-content: space-between;
+}
+
+.details-matadata__buttons {
```
tynandebold, line 23: The alignment here seems ever so slightly off. 

image

I think they should be aligned to the baseline.

src/components/experiment-tracking/details/index.js
```diff
@@ -38,8 +44,19 @@ const Details = ({ selectedRuns, sidebarVisible }) => {
...
         })}
       >
-        <RunMetadata isSingleRun={isSingleRun} runs={runMetadata} />
-        <RunDataset isSingleRun={isSingleRun} trackingData={runTrackingData} />
+        <RunMetadata
+          isSingleRun={isSingleRun}
+          runs={runMetadata}
+          enableShowChanges={enableShowChanges}
+          pinnedRun={pinnedRun}
+          setPinnedRun={setPinnedRun}
```
tynandebold, line 52: I really think we should just start sorting our props and attributes now, because if we're going to do it in the future anyways we're just adding to our technical debt.

tynandebold (changes requested):
Looking good so far. I don't actually have any metrics locally on my machine, so I can't see anything displayed in the table. I need to ask @datajoely why that is.

Most of my comments right now are about styling and naming, so it should give you a start. I'll have a closer look at the JS once I can actually see this stuff in action.

And a last note about sorting attributes and props: I don't think this suggestion is out of context. If we're going to adopt this standard in the future, which I really think we should, we're only adding to our technical debt now by not doing it. It's a quick and easy thing to start doing and does improve developer experience. I do suggest we starting doing it now.

src/components/experiment-tracking/run-metadata/run-metadata.scss
studioswong: I had made minor adjustments to tackle the tiny discrepancy - please see below. 
image. 

This seem to be caused by the setting of the svg which I've adjust this with a padding for now, please feel free to propose any better way to cater for this small visual change ( I just don't want to dwell on this as a blocker for the release).

src/components/experiment-tracking/run-dataset/index.js
```diff
@@ -66,8 +99,13 @@ function buildDatasetDataMarkup(
   datasetKey,
   datasetValues,
   rowIndex,
-  isSingleRun
+  isSingleRun,
+  pinnedRun,
+  enableShowChanges
 ) {
+  // function to return new set of runData with appropiate pin
```
tynandebold, line 106: Nit: `appropriate` instead of `appropiate`.

src/components/experiment-tracking/run-dataset/index.js
```diff
@@ -66,8 +99,13 @@ function buildDatasetDataMarkup(
   datasetKey,
   datasetValues,
   rowIndex,
-  isSingleRun
+  isSingleRun,
+  pinnedRun,
+  enableShowChanges
```
tynandebold, line 104: Might be worth adding these two parameters to the comment above.

src/components/icons/non-selected-pin.js
tynandebold: Still think we should change this. Nonselected isn't actually a word.

src/components/experiment-wrapper/index.js
```diff
@@ -12,22 +12,39 @@ const MAX_NUMBER_COMPARISONS = 2; // 0-based, so three
...
+  const [pinnedRun, setPinnedRun] = useState();
 
   const { data, loading } = useApolloQuery(GET_RUNS);
 
+  // runs needs to be sorted by time to ensure runIDs get sent to
+  // graphql endpoint in correct order
+  const sortRunByTime = (runs) => {
+    const runsWithTimestamps = runs.map((run) => ({
+      id: run,
+      dateObj: new Date(run.replace('.', ':').replace('.', ':')),
```
tynandebold, line 28: You could probably put this part `run.replace('.', ':').replace('.', ':')` in our `src/utils/date-utils.js` file, as it's also used within a function there.

tynandebold (approved):
I finally got the metrics to show up (thank you!) and it's working great.

I've left a couple of nit comments and then have this one: should the initial compared run always be on the left-most side? When I was testing, the pinned run I was often getting was on the right, so I was initially reading right-to-left and not left-to-right, like so: 

image

To reproduce: 
1. Open exp tracking. 
2. Select the second run in the runs list.
3. Toggle comparison mode on.
4. Select the first run in the runs list.
5. Notice the pinned run is on the right.

That might be the expected behavior? It just felt slightly odd to me. I don't think it should hold up this PR, though it's a UX point we should think about.

cc @GabrielComymQB and Mackay (who doesn't have a GitHub account?).

studioswong:
> I finally got the metrics to show up (thank you!) and it's working great.
> 
> I've left a couple of nit comments and then have this one: should the initial compared run always be on the left-most side? When I was testing, the pinned run I was often getting was on the right, so I was initially reading right-to-left and not left-to-right, like so:
> 
> To reproduce:
> 
> 1. Open exp tracking.
> 2. Select the second run in the runs list.
> 3. Toggle comparison mode on.
> 4. Select the first run in the runs list.
> 5. Notice the pinned run is on the right.
> 
> That might be the expected behavior? It just felt slightly odd to me. I don't think it should hold up this PR, though it's a UX point we should think about.
> 
> cc @GabrielComymQB and Mackay (who doesn't have a GitHub account?).

Hiya thanks for raising this Tynan! good observation! 

Yes indeed in the case that you had shown is an expected behaviour - so the expected behaviour is for the pin to be automatically assigned to the first selected run on the sidebar ( in your case, the second run), and once a pin has been previously assigned, the selected pin on the run will persist as long as it is part of the combo of selected runs. 

say if you have 3 runs, and in your case, if the second run is unselected from this combination, that's when we automatically assign the left-most run ( i.e the earliest run). 

This behaviour has been discussed and approved in a design QA session with @GabrielComymQB  and Mackay; but nonetheless, we can definitely revisit this interaction, especially if this is a problem raised by the user.

src/components/experiment-wrapper/index.js
studioswong: great suggestion - I had incorporated this with some further refactor by extracting the `sortRunByTime` function into `date-utils` as well :)

PR #669, opened 2021-12-09
JS components: use folder name as file name pattern

### LINKED ISSUE #666: Refactor JS implementation of index.js files
## Description

It's frustrating that all our JS-component code lives in `index.js` files. I find that hard when searching for a file to open. Most often, the file I want to open isn't the top choice, but much further down the list, like so (I search for "run-list" and want to open **run-list-card**):

image

## Context

If this is improved it'll save everyone time when searching for and opening files.

## Possible Implementation

All the content in our index files should be moved into a newly-created file, with the filename matching the folder name. Sticking with the `run-list-card` example from above, our new file would be named `run-list-card.js` and contain all the necessary code for that component. Finally, all index files should be nothing more than and `import` and `export`, like so:

```jsx
import RunListCard from './run-list-card';

export default RunListCard
```

This pattern is well established. We even had it in [Kedro UI](https://github.com/quantumblacklabs/kedro-ui/tree/master/src/components/modal).

## Checklist

tynandebold:
## Description

Fixes #666.

## Development notes

See issue #666 for details.

This also removes a couple of unused function in the `src/utils/index.js` file.

## Checklist


limdauto (approved):
Amazing. Thank you!

studioswong:
I know this is already merged in, but one of the really counter-productive effect of this PR is that all git-history related to the files for the entire repo is now gone - perhaps there is a better way to execute this to preserve the git history with a procedure like [this](https://devblogs.microsoft.com/oldnewthing/20190919-00/?p=102904)?

tynandebold:
> I know this is already merged in, but one of the really counter-productive effect of this PR is that all git-history related to the files for the entire repo is now gone - perhaps there is a better way to execute this to preserve the git history with a procedure like [this](https://devblogs.microsoft.com/oldnewthing/20190919-00/?p=102904)?

That's not quite right I don't think. All the history is still there, even the line history. For example, run this in terminal and you'll see explicit line blames from several authors: 

```bash
git blame -C src/components/flowchart/flowchart.js
```

I think the problem is that the history may not be being displayed correctly in certain instances. For me in VSCode, the line history isn't right. It just says the changes were from this PR, and I agree, that isn't great and should be remedied.

PR #688, opened 2022-01-07
Update `make run` to make starting Python server easier

AntonyMilneQB:
## Description
I was just thinking about the development process and thought it would be simplified slightly if instead of needing to copy and paste the awkward 
```
PYTHONPATH=$(pwd)/package python3 package/kedro_viz/server.py <path-to-your-test-project>/new-kedro-project
```
we could just run a `make` command. In fact, `make run` is already defined to do something like this, just it's a bit out of date. I've updated it so now the process of developing on the backend goes something like
```
npm start
make build
export PROJECT_PATH=<path-to-your-test-project>/new-kedro-project
make run
<modify Python files>
<kill existing kedro-viz process>
make run
```

## Checklist


CONTRIBUTING.md
```diff
@@ -131,7 +131,7 @@ npm run lib
...
 
-Before launching a development server with a real Kedro project, you'd need to have [Python](https://www.python.org/)(>=3.7, <3.9) and Kedro installed. We strongly recommend setting up [conda](https://docs.conda.io/en/latest/) to manage your Python versions and virtual environments. You can visit Kedro's guide for installing [conda](https://kedro.readthedocs.io/en/latest/02_get_started/01_prerequisites.html#conda) and [how to get started with Kedro](https://kedro.readthedocs.io/en/latest/02_get_started/02_install.html) for more information.
+Before launching a development server with a real Kedro project, you'd need to have [Python](https://www.python.org/)(>=3.7, <=3.9) and Kedro installed. We strongly recommend setting up [conda](https://docs.conda.io/en/latest/) to manage your Python versions and virtual environments. You can visit Kedro's guide for installing [conda](https://kedro.readthedocs.io/en/latest/02_get_started/01_prerequisites.html#conda) and [how to get started with Kedro](https://kedro.readthedocs.io/en/latest/02_get_started/02_install.html) for more information.
```
AntonyMilneQB, line 134:
```suggestion
Before launching a development server with a real Kedro project, you'd need to have [Python](https://www.python.org/)(>=3.7, <3.9) and Kedro installed. We strongly recommend setting up [conda](https://docs.conda.io/en/latest/) to manage your Python versions and virtual environments. You can visit Kedro's guide for installing [conda](https://kedro.readthedocs.io/en/latest/02_get_started/01_prerequisites.html#conda) and [how to get started with Kedro](https://kedro.readthedocs.io/en/latest/02_get_started/02_install.html) for more information.
```

package.json
```diff
@@ -15,7 +15,6 @@
     "start:app": "PORT=4141 react-scripts start",
     "start:css": "npm run build:css && node-sass src/ -o src/ --watch --recursive",
     "start:lib": "rm -rf lib && babel src --out-dir lib --copy-files --watch",
-    "start:api": "python package/kedro_viz/server.py \"$npm_config_project_path\"",
```
AntonyMilneQB, line 18: This looked like it was trying to do something similar to `make run` but is missing the right `PYTHONPATH`. I couldn't see `$npm_config_project_path` defined anywhere so I figured this was obsolete and nobody used it now. But please say if I'm missing something here!!

CONTRIBUTING.md
```diff
@@ -173,14 +173,18 @@ Build the application with:
...
+As far as the development server for the backend is concerned, you only need to run `make build` once when you first setup the project.
 
 Then launch the server with:
 
 ```bash
-PYTHONPATH=$(pwd)/package python3 package/kedro_viz/server.py <path-to-your-test-project>/new-kedro-project
+make run PROJECT_PATH=<path-to-your-test-project>/new-kedro-project
+```
+This command will launch a Kedro-Viz server at [localhost:4142](http://localhost:4142) and serve data from a real Kedro pipeline located at the project path supplied to the command. From then on, launching the app locally at [localhost:4141](http://localhost:4141) will pull data from the Kedro-Viz server that is running on port 4142. To avoid needing to supply `PROJECT_PATH` every time you call `make run`, you can set it as an environment variable:
+```bash
```
AntonyMilneQB, line 184:
```suggestion

```bash
```

package.json
limdauto: Yes it was an earlier attempt to start everything under `npm run start` but don't think anyone is using it.

limdauto (approved):
LGTM. Thank you

package.json
tynandebold: I might keep it, just in case someone wants to use it, and actually make it usable?

```
"PYTHONPATH=$(pwd)/package python3 package/kedro_viz/server.py \"$npm_config_project_path\""
```

tynandebold (approved):
Nice one, thanks for this.

As we talked about last week, I might also keep the NPM script, but your call!

package.json
AntonyMilneQB: I thought about this a bit more and decided to leave it out. Overall I think it's slightly less confusing to have one way of doing things rather than duplicated command, and I kind of like the logical npm/Makefile split (from the zen of Python, "There should be one-- and preferably only one --obvious way to do it."). 🙂

tynandebold: Fine for me.

PR #689, opened 2022-01-10
[KED-2996] Feature/toggle bookmark behavior

tynandebold:
## Description

Resolves [KED-2996](https://jira.quantumblack.com/browse/KED-2996).

## Development notes

This PR enables all the bookmark functionality. There are three places where you can bookmark a run from:

- The runs list card
- The pipeline toolbar (when not in comparison view)
- From the kebab menu (when in comparison view)

Each area is wired up to now be able to toggle bookmarks.

I also wired up the close icon next to the kebab menu. That was a quick win and easy to do.

## QA notes

I chose not to include the bookmark help text on hover in the runs list. Here's a design screenshot:

image

I did this for two reasons: 

- By default, the help text gets cut off and can't extend outside of the sidebar. There's a lot of overhead here that would have made this seemingly simple change time consuming and, in my opinion, not worthwhile for our MVP approach.
- When I did have it there, though not working perfectly, it felt very intrusive with the tooltip coming up over and over and over again.

Separately, I also ordered the icons like this:

image

And not like this:

image

Isn't the close icon always on an end? I didn't feel right when it was in the middle. 

## Checklist


src/components/experiment-tracking/runs-list/runs-list.scss
```diff
@@ -4,7 +4,7 @@
   bottom: 0;
   left: 0;
   overflow-y: scroll;
-  position: absolute;
+  // position: absolute;
```
rashidakanchwala, line 7: nit - maybe we can remove the comments?

rashidakanchwala (approved):
LGTM! Thanks Tynan!!!

src/components/experiment-tracking/runs-list/runs-list.scss
tynandebold: Of course. This is leftover. Great catch. Removed!

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
```diff
@@ -1,23 +1,47 @@
...
+
+  const toggleBookmark = () => {
+    updateRunDetails({
+      variables: {
+        runId: selectedRunData.id,
+        runInput: { bookmark: !selectedRunData?.bookmark },
+      },
+    });
+  };
+
```
studioswong, line 31: This function is used in 2 places ( both here and `runMetadata.js` ) - I would extract that out to the main experiment wrapper to pass it down to both components instead.

src/components/experiment-tracking/run-metadata/run-metadata.js
```diff
@@ -12,9 +17,42 @@ const sanitiseEmptyValue = (value) => {
...
+  const [updateRunDetails] = useMutation(UPDATE_RUN_DETAILS, { client });
+
+  const toggleBookmark = () => {
+    updateRunDetails({
+      variables: {
+        runId: runId,
+        runInput: { bookmark: !isBookmarked },
+      },
+    });
+  };
```
studioswong, line 31: see above comment

src/components/experiment-tracking/run-metadata/run-metadata.test.js
```diff
@@ -100,14 +100,16 @@ describe('RunMetadata', () => {
...
+  //   const wrapper = mount(
+  //     <RunMetadata
+  //       enableShowChanges={true}
+  //       isSingleRun={false}
+  //       runs={twoRuns}
+  //     />
+  //   );
+
+  //   expect(wrapper.find('.pipeline-menu-button__pin').length).toEqual(4);
+  // });
```
studioswong, line 114: This test can be put back in once the  `useMutation` hook is moved up to the wrapper level.

On another note, please add relevant tests to cover the introduction of the new kebab menu and the bookmark icon.

src/components/experiment-tracking/runs-list-card/runs-list-card.js
```diff
@@ -15,27 +19,47 @@ const RunsListCard = ({
...
-  selectedRuns = [],
+  selectedRunIds = [],
 }) => {
   const { id, timestamp, title = null, bookmark } = data;
   const [active, setActive] = useState(false);
+  const [updateRunDetails] = useMutation(UPDATE_RUN_DETAILS, { client });
   const humanReadableTime = toHumanReadableTime(timestamp);
 
-  const onClick = (id) => {
+  const onClick = (id, e) => {
```
studioswong, line 29: please rename this to something more descrption such as `onToggleBookmarkButton`, etc

src/components/experiment-tracking/runs-list-card/runs-list-card.js
```diff
@@ -52,7 +76,15 @@ const RunsListCard = ({
...
-      {bookmark && <BookmarkIcon className={'runs-list-card__bookmark'} />}
+      {bookmark ? (
+        <BookmarkIcon className={'runs-list-card__bookmark'} />
+      ) : (
+        <BookmarkStrokeIcon
+          className={
+            'runs-list-card__bookmark runs-list-card__bookmark--stroke'
+          }
+        />
+      )}
```
studioswong, line 87: please add a test for testing the expected display of the different icons

studioswong (changes requested):
Thanks for setting this up, the bookmark journey works well in general.

One thing I noticed.  - on toggling 'show changes' off the entire burger menu, including the toggle bookmark buttons, etc, also gets toggled off, while instead they should still exist and only the 'pin run' button should be toggled off. 

I'll continue to go through some of the code and leave further single comments if I see anything.

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
tynandebold: Good idea. Though instead of passing it as a prop I'm going to make it a custom hook so we can pass it around freely as we wish.

src/components/experiment-tracking/run-metadata/run-metadata.test.js
tynandebold: I wish this was true. The test still fails even if the mutation hook is moved. The hook itself still fires in this file, so it doesn't make a difference.

src/components/experiment-tracking/runs-list-card/runs-list-card.js
tynandebold: This function can either toggle the bookmark or fire the `onRunSelection` function, so it isn't strictly about the bookmark. 

I can change it to `onRunsListCardClick`. Is that ok?

src/components/experiment-tracking/runs-list-card/runs-list-card.js
tynandebold: Added.

tynandebold:
> Thanks for setting this up, the bookmark journey works well in general.
> 
> One thing I noticed. - on toggling 'show changes' off the entire burger menu, including the toggle bookmark buttons, etc, also gets toggled off, while instead they should still exist and only the 'pin run' button should be toggled off.
> 
> I'll continue to go through some of the code and leave further single comments if I see anything.

Updated everything you asked for. Lmk if there's anything else :)

src/components/experiment-tracking/run-metadata/run-metadata.test.js
studioswong: I think this problem with `useMutation` would be something we need to resolve within this PR as this seems to restrict our ability to write tests, which I highly doubt it would be designed this way given how many users there are within the react community. 

Throwing a thought here - looking at the error it looks like the test was not able to locate the client, which you might need to wrap the wrapper with an ApolloProvider in the test?

studioswong: Also, it also does seem weird that this problem surfaces in this PR and not in the previous one when we introduced the mutation hook for the first time - I wonder what is the reason behind that?

it might either suggest that there might be a different setup, or perhaps there wasn't the right tests set up to properly test that 😆

src/apollo/mutations.js
```diff
@@ -22,10 +21,8 @@ export const UPDATE_RUN_DETAILS = gql`
 `;
 
 export const useUpdateRunDetails = () => {
-  const [_updateRunDetails, { error, loading, reset }] = useMutation(
-    UPDATE_RUN_DETAILS,
-    { client }
```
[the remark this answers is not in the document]
tynandebold: No it wasn't that. It's just we were't actually mocking the `useUpdateRunDetails` mutation.

src/apollo/queries.js
```diff
@@ -19,6 +19,7 @@ export const GET_RUN_METADATA = gql`
     runMetadata(runIds: $runIds) {
       id
       author
+      bookmark
```
limdauto, line 22: not in scope of your PR but this query has duplicated `id` field. Can we quickly remove it here?

limdauto (approved):
I haven't tested the functionality but the code LGTM!

src/apollo/queries.js
tynandebold: Absolutely. It's gone.

src/apollo/mutations.js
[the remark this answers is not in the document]
studioswong: indeed. I've always wondered why we need to pass in the client here as it's passed on from the `ApolloProvider` anyways, good to delete this here.

src/components/experiment-tracking/run-details-modal/run-details-modal.test.js
```diff
@@ -5,6 +5,24 @@ import { configure, mount, shallow } from 'enzyme';
...
+
+jest.mock('../../../apollo/mutations', () => {
+  return {
+    useUpdateRunDetails: () => {
+      return {
+        reset: mockReset,
+        updateRunDetails: mockUpdateRunDetails,
+      };
+    },
+  };
```
studioswong, line 21: Now that we have set up the mock for this mutation function, let's add the test that actually calls the `updateRunDetails` function ( which is called from `onApplyChanges` ), as well as mocking and testing the error state for the error message (i.e testing the existence of the `run-details-modal-error-wrapper` class) - otherwise there isn't any point of setting up this mock function here.

RELEASE.md
```diff
@@ -8,6 +8,10 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+## Major features and improvements
+
+- Add the toggle bookmark for experiment runs feature. (#689)
```
studioswong, line 13:
```suggestion
- Implemented the 'toggle bookmark' journey that allows bookmarking runs and displaying runs as a separate list . (#689)
```

src/components/experiment-wrapper/experiment-wrapper.js
```diff
@@ -17,63 +17,92 @@ const ExperimentWrapper = ({ theme }) => {
...
+      } else {
+        setSelectedRunIds(data.runsList.map((run) => run.id).slice(0, 1));
+      }
+    }
+  }, [data, selectedRunIds]);
+
+  useEffect(() => {
+    /**
+     * If we return runs and aren't in comparison view, set a single selected
+     * run data object for use in other components.
```
studioswong, line 86: it would be more helpful to explain about what other components are here - I am guessing the run-metadata and run-details component?

studioswong (approved):
LGTM, thanks for getting through with the fix to the tests! 

I've left a few comments which I'll let you fix them before merging it in. For the tests for `run-details-modal`, I'll let you decide on whether to tackle it in this PR or as a separate PR.

PR #692, opened 2022-01-13
[KED-3033] Fix Kedro-viz to work with partial, curry & lambda function 

### LINKED ISSUE #635: [KED-3033] Kedro Viz not working with curry or partial functions
## Description
Hello! I am having an error when clicking on a node in `kedro viz` that has a partial or curry function. Besides the error, the visual interface does not show the inputs or parameters.

```
AttributeError: 'functools.partial' object has no attribute '__closure__'
```
This is the error I get when using partial functions.

```
AttributeError: 'curry' object has no attribute '__closure__'
```
This is the error I get when using curry functions.

I have tried using `update_wrapper` and it is still raising the same error :/ .
## Context
I am trying to run `kedro viz` with a project that uses multiple partial functions,.

## Steps to Reproduce

1. Define partial function on node
2. Run kedro viz
3.  Click on function node
4. Watch logs

## Expected Result
Run as normal.

## Actual Result
The side bar with the parameters and input does not load.

```
```
AttributeError: 'functools.partial' object has no attribute '__closure__'
```
```
AttributeError: 'curry' object has no attribute '__closure__'
```

## Your Environment

## Checklist

### PR DESCRIPTION
**Pull request by rashidakanchwala**
Signed-off-by: Rashida Kanchwala <rashida.kanchwala@quantumblack.com>

## Description

Fixes #635.

This PR was created in response to the bug highlighted in this ticket - https://jira.quantumblack.com/browse/KED-3033. While investigating this issue where error was thrown in the console for partial function, I noticed that for partial, & lambda functions -- the name on the side panel doesn't appear. 

This is because <lambda> and <partial> function names have the '<>' tag and html assumes it as an html tag instead of treating it as a name. 

## Development notes

This PR fixes the three below issues 
- functools.partial  has a no __closure__. error message in the console. Thanks @AntonyMilneQB 
- partial functions do not show input, output, parameter info in metadata panel 
- partial functions /lambda functions name doesn't show in the side panel. 
- 
Screenshot 2022-01-13 at 12 53 37

UPDATED SCREENSHOT 

Screenshot 2022-01-18 at 12 51 50

## Checklist

### CODE REVIEW

**Comment by AntonyMilneQB** [comment]
Nice work! What does the node name of the partial or lambda function appear as now? Is it possible to post a screenshot just so I can see an example?

**Comment by rashidakanchwala** [comment]
> Nice work! What does the node name of the partial or lambda function appear as now? Is it possible to post a screenshot just so I can see an example?

Done. For this PR I have left it as partial, and lambda without the tags. But raising a ticket to figure this out.

**Comment by AntonyMilneQB** [comment]
> > Nice work! What does the node name of the partial or lambda function appear as now? Is it possible to post a screenshot just so I can see an example?
> 
> Done. For this PR I have left it as partial, and lambda without the tags. But raising a ticket to figure this out.

Cool, sounds good. For future reference the `node.name` for these 3 nodes is:
```
<partial>([companies]) -> [companies_out3]
<lambda>([companies]) -> [companies_out]
inner([companies]) -> [companies_out2]
```

whereas the `node.short_name` is:
```
<Partial>
<Lambda>
Inner
```

So I'm guessing we're using the `short_name` here which isn't very descriptive. I'm not sure what the right solution would be though - I think the whole naming of node in general needs to be rethought (both in kedro core and kedro viz). Possibly just using the full `node.name` for lambda/partial/decorated functions is the right thing to do as an easy temporary fix anyway.

**Comment by rashidakanchwala** [comment]
> > > Nice work! What does the node name of the partial or lambda function appear as now? Is it possible to post a screenshot just so I can see an example?
> > 
> > 
> > Done. For this PR I have left it as partial, and lambda without the tags. But raising a ticket to figure this out.
> 
> Cool, sounds good. For future reference the `node.name` for these 3 nodes is:
> 
> ```
> <partial>([companies]) -> [companies_out3]
> <lambda>([companies]) -> [companies_out]
> inner([companies]) -> [companies_out2]
> ```
> 
> whereas the `node.short_name` is:
> 
> ```
> <Partial>
> <Lambda>
> Inner
> ```
> 
> So I'm guessing we're using the `short_name` here which isn't very descriptive. I'm not sure what the right solution would be though - I think the whole naming of node in general needs to be rethought (both in kedro core and kedro viz). Possibly just using the full `node.name` for lambda/partial/decorated functions is the right thing to do as an easy temporary fix anyway.

I completely agree -- partial, lambda doesn't make much sense. I spoke to Lim and he said we need to make a new ticket for this with design involved so I will create that post this fix.

File: package/kedro_viz/models/graph.py [hunk 1]
```diff
@@ -43,6 +43,10 @@ def _strip_namespace(name: str) -> str:
     return re.sub(pattern, "", name)
 
 
+def _strip_tags(name: str) -> str:
+    return name.replace("<", "").replace(">", "")
```
limdauto, line 47: Interesting... I quite like the `<>` to indicate that these are not normal functions.

package/kedro_viz/models/graph.py
```diff
@@ -403,18 +408,21 @@ class TaskNodeMetadata(GraphNodeMetadata):
...
-        code_full_path = Path(inspect.getfile(kedro_node._func)).expanduser().resolve()
-        try:
-            filepath = code_full_path.relative_to(Path.cwd().parent)
-        except ValueError:  # pragma: no cover
-            # if the filepath can't be resolved relative to the current directory,
-            # e.g. either during tests or during launching development server
-            # outside of a Kedro project, simply return the fullpath to the file.
-            filepath = code_full_path
-        self.filepath = str(filepath)
+        if inspect.isfunction(kedro_node.func):
```
limdauto, line 411: Could you put a comment here stating that this is needed because `kedro_node.func` could be partial or curried functions?

limdauto (approved):
Nice one!

package/kedro_viz/models/graph.py
studioswong: One thing I am quite confused as to the presence of  the`<>`tag being the problem behind rendering the name as it is returned as part of a string - I think it is possible for us to keep it within the name?

rashidakanchwala: it is very weird. this is how it looks...
Screenshot 2022-01-18 at 09 52 20

rashidakanchwala: Same occurs with lambda.

rashidakanchwala: in the FE, we read it like this _dangerouslySetInnerHTML={{ __html: label }}_ 

maybe we can change this?

limdauto: Ahhhhhhhhhhh `_dangerouslySetInnerHTML`

studioswong: yea that explains why - not sure why we set the label this way, but if there is no use for this then we can either take the `_dangerouslySetInnerHTML` out, or set a small function on the FE to replace the `<` tag in the string with `&lt;`. ( I might say the latter might be better as it is likely this is set up this way for some reason.)

AntonyMilneQB: @rashidakanchwala does it work if you do this?
```
def _strip_tags(name: str) -> str:
    return name.replace("<", "&lt;").replace(">", "&gt;")
```

AntonyMilneQB: If so then we can easily just do this on the backend and avoid removing the `_dangerouslySetInnerHTML` in case it's there for a good reason.

rashidakanchwala: Hi, I had to make the above change on the FE. because if i did it in the backend, then the metadata panel was messed up. Now, it works fine
Screenshot 2022-01-18 at 12 29 35

studioswong (commented):
@rashidakanchwala Thank you for setting this up and making the changes to expose the `<>` tag, the name shows up well 👍 

2 things I noticed:
1. The input/output information is not showing up for the partial function on my side ( see below ) - this might be due to the function setup on my side too. 
image

2. Search currently doesn't work for the lamba or partial function (see below 2 images) - this requires further consideration within the search UI on the FE. 
image

image


To enable the name to work properly with search might suggest the need to alter the FE.

rashidakanchwala:
> @rashidakanchwala Thank you for setting this up and making the changes to expose the `<>` tag, the name shows up well 👍
> 
> 2 things I noticed:
> 
> 1. The input/output information is not showing up for the partial function on my side ( see below ) - this might be due to the function setup on my side too.
>    image
> 2. Search currently doesn't work for the lamba or partial function (see below 2 images) - this requires further consideration within the search UI on the FE.
>    image
> 
> image
> 
> To enable the name to work properly with search might suggest the need to alter the FE.

Thanks @studioswong  for highlighting this issue. This was because I was converting all the '<' to '&gt' on the FE. And the tags related to highlightText '<b> (line 25 - https://github.com/quantumblacklabs/kedro-ui/blob/master/src/components/search-results/search-results-utils.js) were also getting converted to '&lt' ... This has been solved by moving the tag replacement logic to the BE. And adding dangerouslysetinnerHTML to metadata title. Thanks @tynandebold  for your support!

src/components/metadata/metadata.js
```diff
@@ -82,9 +82,10 @@ const MetaData = ({
                   className="pipeline-metadata__icon"
                   icon={nodeTypeIcon}
                 />
-                <h2 className="pipeline-metadata__title">
-                  {metadata.node.name}
-                </h2>
+                <span
+                  className="pipeline-metadata__title"
+                  dangerouslySetInnerHTML={{ __html: metadata.node.name }}
+                />
```
tynandebold, line 88: Looks like this html element changed from an `h2` to a `span`. It should stay as a `h2`.

tynandebold (approved):
Works great!

Just one small comment about a change to a HTML element that should be switched back before merging :)

studioswong (approved):
works well, well done 😄

PR #699, opened 2022-01-17
Make CircleCI fail on bad formatting

AntonyMilneQB:
## Description

Currently the CI runs `black` and `isort` but won't actually fail if there's formatting errors. I've added the `--check` flag to both so that it will give non-zero exit code if the files aren't formatted correctly.

Locally you should still run `make lint` to fix and check files. I've done this, hence the small changes to whitespace.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ pytest: build
 e2e-tests: build
 	cd package && behave
 
-lint: format-fix pylint-check
+lint: format-fix lint-check
 
 format-fix:
 	isort package/kedro_viz package/tests package/features
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,17 @@ pytest: build
 e2e-tests: build
 	cd package && behave
 
-pylint:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -23,9 +23,17 @@ pytest: build
 e2e-tests: build
 	cd package && behave
 
-pylint:
-	cd package && isort .
+lint: format-fix lint-check
+
+format-fix:
+	isort package/kedro_viz package/tests package/features
```
AntonyMilneQB, line 29: Note I changed `cd package && isort .` to match `black`.

limdauto (approved):
Awesome thank you. I could swear that it used to fail even without `--check` because it modifies the files and produces a dirty git directory or something. Buit this is definitely better.

rashidakanchwala (approved):
LGTM. Useful PR to review since I am learning CircleCI set up :D .. thanks Antony

PR #701, opened 2022-01-17
[KED-3052] Add globalToolbarWidth to constants

tynandebold:
## Description

Fixes [KED-3052](https://jira.quantumblack.com/browse/KED-3052).

## Development notes

I've added on the `globalToolbarWidth` const to the necessary widths and removed it from a place where it wasn't needed.

I also capitalize the use of `Plotly` in the release-notes file and formatted it.

## QA notes

Look at our default data set and expand the `Reporting` pipeline. Before this changed one part of the flowchart looked like this:

image

And now we have:

image

## Checklist

### DIFF
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -59,7 +59,6 @@ export const drawLayerNames = function () {
     layers,
   } = this.props;
 
-  console.log('sidebarWidth: ', sidebarWidth);
   this.el.layerNameGroup
     .transition('layer-names-sidebar-width')
     .duration(this.DURATION)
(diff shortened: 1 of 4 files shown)


src/components/flowchart/draw.js
```diff
@@ -60,10 +59,11 @@ export const drawLayerNames = function () {
     layers,
   } = this.props;
 
+  console.log('sidebarWidth: ', sidebarWidth);
```
studioswong, line 62: please delete

studioswong (commented):
LGTM on initial review - will check out the branch to test this out.

studioswong (approved):
Tested this with the same scenario as listed on the ticket, looks much better now.

There is still a slight cut off on the left with the current zoom setting, but imo I don't think it's anything to sweat - putting it here in case we want to fix this later. 
image

src/components/flowchart/draw.js
tynandebold: Whoops. Thanks. Deleted.

PR #707, opened 2022-01-19
[KED-2948] Fixed params as prefix issue for datasets 

### LINKED ISSUE #621: [KED-3034] If a dataset is prefixed with the word 'params' it is treated as a parameter and only shows on the expanded view
## Description
As users start to experiment with the new `tracking.JSONDataSet` they may name their catalog entries something like `params_random_forest` and Kedro Viz will currently mistakenly count them as a parameter type node when they are in fact a regular data node.

## Steps to Reproduce

1. Name a dataset something prefixed with 'params'
2. Visualise

## Expected Result
Should appear as a data node like the 'hyperparams_linear` node below

## Actual Result
image

Node only visible in the expanded parameter view - also impossible that a parameter should be an output to a node

rashidakanchwala:
Signed-off-by: Rashida Kanchwala <rashida.kanchwala@quantumblack.com>

## Description

Fixes #621.

## Development notes

'params:' and 'parameters' are reserved keywords. If anyone uses them in Kedro. The node will be considered as a parameter. 

This issue has been fixed, so if you rename a dataset to param_<dataset_name>, it will still display as a dataset 

Screenshot 2022-01-19 at 15 32 14

## Checklist


package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,241 +1,241 @@
...
-            True
-            >>> modular_pipelines.has_modular_pipeline("doesnt exist")
-            False
-        """
-        return modular_pipeline_id in self.tree
-
-    def as_dict(self) -> Dict[str, ModularPipelineNode]:
-        """Return the repository as a dictionary."""
-        return self.tree
+"""`kedro_viz.data_access.repositories.modular_pipelines`
```
rashidakanchwala, line 1: No idea, why github is not ignoring this since it's exact same !!

AntonyMilneQB: Mysterious! Maybe pulling `main` in will fix it?

AntonyMilneQB (approved):
Nice fix! Actually I just went to check how we do this in kedro and I think we've made the same mistake there 😅

tynandebold (approved):
Nice one

package/kedro_viz/data_access/repositories/modular_pipelines.py
deepyaman: Glad I'm not the only going crazy. I was looking through the blames and couldn't tell what changed!

PR #708, opened 2022-01-19
Fix bookmark regression

limdauto:
## Checklist

### DIFF
--- a/package/tests/test_api/test_graphql/test_mutations.py
+++ b/package/tests/test_api/test_graphql/test_mutations.py
@@ -6,95 +6,6 @@ from kedro_viz.models.experiments_tracking import RunModel
 
 
 @pytest.mark.usefixtures("data_access_manager_with_runs")
-class TestGraphQLEndpointsWithRuns:
-    def test_graphql_run_list_endpoint(
-        self,
-        client,
-        example_run_ids,
-    ):
-        response = client.post("/graphql", json={"query": "{runsList {id bookmark}}"})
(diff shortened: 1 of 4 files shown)


package/tests/test_api/test_graphql/test_mutations.py
```diff
@@ -5,95 +5,6 @@
...
-                                    "runId": example_run_id,
-                                    "value": "column_seven",
-                                }
-                            ],
-                        },
-                    },
-                ]
-            }
-        }
-        assert response.json() == expected_response
```
limdauto, line 94: These tests were moved to other files but I forgot to delete them in the previous branch

studioswong (approved):
works great, thanks for the fast turnaround of the fix! 👍 

I noticed that you've updated the test to reflect the bookmark mutation, which is great. Any chance we can add in an extra test to test the `bookmark=False` mutation case( i.e the scenario of the bug that triggered this fix)? 

Thanks so much for this!

limdauto:
@studio so previously there is already a test for both value `True` & `False` at the API test level. What I changed was the test data setup. It was passing by mistake, i.e.:

```
original_bookmark_value = False

if new_bookmark_value:
    original_bookmark_value = new_bookmark_value

return original_bookmark_value
```

As you can see, the test will pass because `original_bookmark_value` happened to be equal to `new_bookmark_value` if they are both False. Changing `original_bookmark_value` to `True` makes sure it fails.

PR #709, opened 2022-01-20
[KED-3004] Search MVP experiment tracking

studioswong:
## Description

resolves [KED-3004](https://jira.quantumblack.com/browse/KED-3004).

This PR is the work in setting up the search for experiment tracking runslist. This will now allow users to search through the runs list and return results with runs that either the 'Title', 'gitsha' and 'notes' field matches the search value.

image

image

## Development notes

1. I have refactored the existing search bar from a tailored component to the node-list to a general search bar, and set up related tests.
2. I have refactored existing runs-list and runs-list card setup with the search Logic, and added in tests related to the search feature.
3. I have set up a new test suite for runs-list component as currently there isn't any test suites for that component. 

## QA notes

Please check out the branch and QA the search. I will also sync up with design (@gabrielcomym) to go through the latest implementation. 

## Checklist


src/components/experiment-tracking/runs-list/runs-list.scss
```diff
@@ -1,9 +1,28 @@
...
+  justify-content: space-between;
+  padding: 2em 3.5em 1.2em 3em;
+  position: sticky;
+  top: 0;
+  z-index: 1;
+}
+
+.compare-switch-wrapper__text {
+  font-size: 14px;
+}
```
tynandebold, line 28: These exact styles are also present in the `/sidebar/sidebar.scss` file. Not sure why they're duplicated here?

image

src/components/experiment-tracking/runs-list/runs-list.test.js
```diff
@@ -0,0 +1,64 @@
...
+  },
+  {
+    bookmark: false,
+    id: 'run2',
+    timestamp: new Date('October 15, 2021 03:26:00').toISOString(),
+    title: 'Run 2',
+    notes: 'notes of run 2 is a fish',
+  },
+  {
+    bookmark: false,
```
tynandebold, line 24: Maybe we should also test when a bookmark is `true` and then we get the two-accordions view?

src/components/search-list/search-list.scss
```diff
@@ -0,0 +1,102 @@
+@import '../../styles/variables';
+
+.pipeline-search-list {
+    padding: 2px ($sidebar-padding - 0.4) 0 ($sidebar-padding - 0.4);
+  
+    .kui-searchbar,
```
tynandebold, line 6: The formatting is off on this file. Can you try and save it again?

src/utils/index.js
```diff
@@ -60,3 +60,18 @@ export const changed = (props, objectA, objectB) => {
     objectA && objectB && props.some((prop) => objectA[prop] !== objectB[prop])
   );
 };
+
+const { escapeRegExp } = utils;
+/**
+ * Check whether a piece of text matches the searchvalue, true if no search value given
```
tynandebold, line 66: `search value` is two words and you don't need the `, true if no search value given` in the top line since you specify that in the return line.

```suggestion
 * Check whether a piece of text matches the search value
```

tynandebold (changes requested):
Awesome stuff. Excellent first pass.

A few comments from me, but nothing major. They should be quick changes :)

src/components/experiment-tracking/runs-list-search/runs-list-search.js
```diff
@@ -3,12 +3,12 @@ import SearchBar from '@quantumblack/kedro-ui/lib/components/search-bar';
...
 
 /**
- * Handle Node List Search
+ * Handle Runs List Search
  * @param {function} onUpdateSearchValue Event handler
  * @param {string} searchValue Search text
  * @param {string} theme Light/dark theme for Kedro-UI component
  */
-export const NodeListSearch = ({ onUpdateSearchValue, searchValue, theme }) => {
+export const RunsListSearch = ({ onUpdateSearchValue, searchValue, theme }) => {
```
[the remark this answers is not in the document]
studioswong: good spot - I completely forgot about it

src/components/experiment-tracking/runs-list-card/runs-list-card.js
```diff
@@ -18,12 +22,21 @@ const RunsListCard = ({
...
   const { updateRunDetails } = useUpdateRunDetails();
   const humanReadableTime = toHumanReadableTime(timestamp);
 
+  const isMatchSearchValue = (text) =>
+    searchValue ? textMatchesSearch(text, searchValue) : false;
+
+  const displayValue = (value) =>
+    isMatchSearchValue(value) ? getHighlightedText(value, searchValue) : value;
+
+  const isNotesContainSearchValue = isMatchSearchValue(notes);
```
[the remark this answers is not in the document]
studioswong: thanks for suggesting this! Actually I originally had this set up exactly the way you suggested but then striped this out into the 2 separate checks as I find separating the logic into 2 separate functions more readable ( and also as a result of comments from previous PRs 😆  it's funny how many difference of opinion there is). 

This is the same behind setting out a separate variable for the note check for simplicity and readability in the JSX, without the need for anyone later in to refer to what `isMatchSearchValue` does as well.

src/components/experiment-tracking/runs-list/runs-list.scss
studioswong: good catch, I was about to delete the style under `sideBar` but forgot!

rashidakanchwala (commented):
So I tried the search and it works well. But I noticed the search results number doesn't change. Please see it shows Bookmarked (2) Unbookmarked(5) --> instead of Bookmarked (1) Unbookmarked(1)

Screenshot 2022-01-21 at 09 44 47

tynandebold:
> So I tried the search and it works well. But I noticed the search results number doesn't change. Please see it shows Bookmarked (2) Unbookmarked(5) --> instead of Bookmarked (1) Unbookmarked(1)
> 
> Screenshot 2022-01-21 at 09 44 47

Great catch.

studioswong:
good catch @rashidakanchwala - I'll update this

src/components/experiment-tracking/runs-list/runs-list.test.js
studioswong: indeed - on that note, I have observed that there are currently no tests set up for the run-list for the bookmark journey - it would be great if we can set some tests for that. I will raise a separate ticket for that. 

Your suggested test condition would sound like a perfect test case for the bookmark journey test setup to be dealt with in that ticket - what do you think?

tynandebold: Yeah we could and probably should be robust and create a separate ticket tailored more towards bookmarks. We'll still need a test for the distinct accordions, and I think that could live in this file, as those are part of the runs list.

You can do it so quick too, because if you use `bookmark: true` on one of the runs then the condition on line 38 becomes this, and then that basically checks for two accordions:

```js
expect(wrapper.find('.runs-list__wrapper').length).toBe(2);
```

src/components/experiment-tracking/runs-list/runs-list.scss
```diff
@@ -1,9 +1,38 @@
...
 .runs-list__accordion-header {
   padding: 0.8em 3.5em 0.8em 3em;
 }
+
+.compare-switch-wrapper {
+  align-items: baseline;
+  background: var(--color-bg-3);
+  display: flex;
+  justify-content: space-between;
+  padding: 2em 3.5em 1.2em 3em;
```
tynandebold, line 30: I think this should be `padding: 3.5em 3.5em 1.2em 3em;`. `35px` (plus the little extra from the search component) between the search and compare toggle, not `20px`.

src/components/experiment-tracking/runs-list-card/runs-list-card.js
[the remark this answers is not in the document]
tynandebold: Fine with me. I would still change the variable name to `isSearchValueInNotes` instead of `isNotesContainSearchValue`.

tynandebold (approved):
Looks awesome. I'm approving with a couple of comments that I hope you'll change before merging in (specifically the spacing between the search and the comparison toggle and the `All` label never appearing).

Great stuff!

rashidakanchwala (approved):
Thanks, Susanna! Looks great.

PR #715, opened 2022-01-24
[KED-3049] Update contributors guide to include guidance on DCO

AntonyMilneQB:
## Description
As per https://github.com/kedro-org/kedro/pull/1167 and https://github.com/kedro-org/kedro/pull/1177, now that we're part of LF, we need all commits to be signed-off. I'm not a fan of this, but after spending far too much time on it I've come up with this workflow as the simplest possible one for now.

1. `make sign-off`: this creates a git hook that will add the Signed-off-by line to all commit message. This is a better solution than aliasing git commit for various reasons
2. `require: members: false` for the DCO GitHub app configuration. This means that [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (i.e. GPG or S/MIME signed, _not_ signed-off) by members of kedro-org do not need to have the Signed-off-by line. In practice, this is very helpful because it means that any commits maintainers make on GitHub (applying code review suggestions, merging in `main`, etc.) will get ✅  for DCO check.

## Possible future improvements
- Enable `allowRemediationCommits` to make it easier for users to retroactively sign off commits
- Maybe fork DCO app so that the "Set DCO to pass" button is available to all? (Might defeat the point of the check though...)

## Checklist


studioswong (approved):
awesome work for setting up the docs and the sign off workfow 👍

PR #716, opened 2022-01-25
[KED-2981] Re-write Kedro UI buttons

rashidakanchwala:
## Description

This PR is a part of the Kedro-UI migration ticket. In this PR we migrate the button functionality from Kedro-ui to Kedro-viz 

## Development notes

The Kedro-UI button functionality had lot of styles that we weren't using in Kedro-viz. So I have made that simple in this PR. In [Kedro-UI](https://quantumblacklabs.github.io/kedro-ui/), we have primary button, second button, small button, button with wipe transition. For consistency sake, I have used only one button style (primary) for all buttons on Kedro-viz. Additionally added the size small since it was used in the experiment tracking. I have removed secondary button style (which was also used for experiment tracking) and also removed the wipe transition (used in large pipeline warning) for the sake of consistency and simplicity. 

UPDATE - As per the discussion with Tynan, and Gabriel in the check-in, I have added secondary mode for buttons (since it's used). I have removed wipe transition which was used in large pipelines - and Gabriel is comfortable with it. 
I have also removed disable since we are not using it atm. 

## Checklist


tynandebold:
Just a quick comment off the top: I think we need to keep the secondary button style (and really any other styles that were used in Viz).

For example, the edit runs modal has two button styles:
Screen Shot 2022-01-25 at 10 03 44

Though with your change it looks like this:
Screen Shot 2022-01-25 at 10 02 19

There should be some differentiation between those two.

src/components/experiment-wrapper/experiment-wrapper.js
```diff
@@ -181,7 +181,7 @@ const ExperimentWrapper = ({ theme }) => {
             rel="noreferrer"
             target="_blank"
           >
-            <Button theme={theme}>View docs</Button>
```
tynandebold, line 184: I noticed you're removing the `theme` prop here, which is fine, but didn't remove it on the buttons in the `src/components/export-modal/export-modal.js` file. I wonder why? I think you can remove it there as well can't you?

tynandebold (approved):
Looks great! Just one tiny comment about not removing the `theme` prop in one file, but it's not a blocker :)

src/components/experiment-wrapper/experiment-wrapper.js
rashidakanchwala: yes, i will remove it :) .. thanks!

src/components/button/button.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Generic Kedro Button
+ */
+const Button = ({ children, onClick, size, mode }) => (
+  <span className="kedro kui-button">
+    <button
+      className={classnames(
+        'kui-button__btn',
+        `kui-button__btn--${size}`,
+        `kui-button__btn--${mode}`
```
AntonyMilneQB, line 15: Just curious... what's up with the ` instead of '?

rashidakanchwala: It's called template literals similar to f-string or % in python ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

AntonyMilneQB: Ahh cool, thank you!

studioswong (commented):
I just had an initial skim through and tested the button, looks good! I'll do a more through dive into the code and let you know if I spot anything.

Please update the release notes for this feature in the meantime :)

src/components/button/button.scss
```diff
@@ -0,0 +1,108 @@
...
+}
+
+.kui-theme--dark {
+  --color-button--active: #{rgba($color-light, 0.85)};
+  --color-button--outline: #{$color-light};
+  --color-button__text--hover: #{rgba($color-dark, 0.85)};
+  --color-button__text: #{rgba($color-light, 0.85)};
+}
+
+.kui-button {
```
studioswong, line 28: Please remove references of `kui` and rename all css classnames for this component to either `kedro-button` or just `button`.

studioswong (changes requested):
The code overall looks good! Requesting changes mainly for renaming the button classes as we are moving away from kui references. (I understand we have used `kui-theme` for the theme naming for the overall app given our previous setup and links to kui, which I have put in a ticket to rename those class name to strip away the kui reference( just `theme--light` or `theme--dark` ) and can be done in one go after we have migrated all kui elements.)

studioswong (approved):
Nice one 👏

PR #722, opened 2022-01-28
Document how to use demo project in `make run`

AntonyMilneQB:
## Description

Following https://github.com/kedro-org/kedro-viz/pull/688, which created the `make run` command to make it easier to start the dev server, and https://github.com/kedro-org/kedro-viz/pull/696, which added demo-project to the repo, I've now linked the two together so that `make run` points to demo-project with no other fiddling needed 🎉 

Now the process of developing something on the backend just goes as follows:
```
npm start
make build
make run
```

This will start a dev server pointing to demo-project with full access to the project pipeline, experiment tracking data, etc. No need to faff around setting a `PROJECT_PATH` any more.

I've updated CONRIBUTING with instructions on the above and also to explain the use of demo project better.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,7 +107,7 @@ Alternatively, if you know the JIRA ticket number of the issue that you are fixi
 
 ### JavaScript development
 
-_*Note*: We use the [latest release of v14 Node.js](https://nodejs.org/download/release/latest-v14.x/) in our development environment._
+_*Note*: We suggest using the [latest release of Node.js v14](https://nodejs.org/download/release/latest-v14.x/) in your development environment._
 
 First clone this repo, then download and install dependencies:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -186,13 +186,13 @@ make run PROJECT_PATH=<path-to-your-test-project>/new-kedro-project
 
 #### Launch the development server with the `SQLiteSessionStore`
 
-Kedro-Viz provides a `SQLiteSessionStore` that users can use in their project to enable experiment tracking functionality. If you want to use this session store with the development server, make sure you don't use a relative path when specifying the store's location. For example, to specify the local `data` directory within a project as the session store's location, configure the project's `settings.py` as follows:
+Kedro-Viz provides a `SQLiteSessionStore` that users can use in their project to enable experiment tracking functionality. If you want to use this session store with the development server, make sure you don't use a relative path when specifying the store's location in `settings.py`. For example, `demo-project` specifies the local `data` directory within a project as the session store's location as follows: 
 ```python
 from kedro_viz.integrations.kedro.sqlite_store import SQLiteStore
 SESSION_STORE_ARGS = {"path": str(Path(__file__).parents[2] / "data")}
 ```
 
-`demo-project` includes a `settings.py` file that includes this option. Owing to this coupling between the project settings and Kedro-Viz, if you wish to execute any Kedro commands on `demo-project` (including `kedro run`), you will need to install the Kedro-Viz Python package. To make sure that you are installing the one that you are developing against, run:
+Owing to this coupling between the project settings and Kedro-Viz, if you wish to execute any Kedro commands on `demo-project` (including `kedro run`), you will need to install the Kedro-Viz Python package. To install your local development version of the package, run:
 
 ```bash
 pip3 install -e package
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -189,15 +189,15 @@ As far as the development server for the backend is concerned, you only need to 
 make run
 ```
 
-This command will launch a Kedro-Viz server at [localhost:4142](http://localhost:4142) and serve data from `demo-project`. From then on, launching the app locally at [localhost:4141](http://localhost:4141) will pull data from the Kedro-Viz server that is running on port 4142.
+This command will launch a Kedro-Viz backend server at [localhost:4142](http://localhost:4142) and serve data from `demo-project`. If you wish to also launch the frontend app then execute `npm start` in a separate terminal window. [localhost:4141](http://localhost:4141) will then pull data from the backend Kedro-Viz server that is running on port 4142.
 
-If you wish to point the server to a different Kedro project then you can do so by altering the `PROJECT_PATH` variable:
+If you wish to point the backend server to a different Kedro project then you can do so by altering the `PROJECT_PATH` variable:
 
 ```bash
 make run PROJECT_PATH=<path-to-your-test-project>/new-kedro-project
 ```
 
-> **Note**: Once the development server is launched at port 4142, the local app will always pull data from that server. To prevent this, you can comment out the proxy setting in `package.json` and restart the dev server at port 4141.
+> **Note**: Once the backend development server is launched at port 4142, the local app will always pull data from that server. To prevent this, you can comment out the proxy setting in `package.json` and restart the dev server at port 4141.
 
 #### Launch the development server with the `SQLiteSessionStore`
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -107,7 +107,7 @@ Alternatively, if you know the JIRA ticket number of the issue that you are fixi
 
 ### JavaScript development
 
-_*Note*: We use `node==14` in our development environment._
+_*Note*: We use the [latest release of v14 Node.js](https://nodejs.org/download/release/latest-v14.x/) in our development environment._
```
AntonyMilneQB, line 110: Maybe I don't know what I'm talking about here, but there seem to be many different version of 14.x.y and the exact one does seem to make a difference to the package-lock.json. So I thought best to fix on one particular version.

studioswong: From what I know as long as it's v14 it would work for compiling our app, but I agree that specifying the latest version would help align the package-lock file. 

```suggestion
_*Note*: We suggest using the latest version of [Node.js v.14 (<=14.18.0)](https://nodejs.org/download/release/latest-v14.x/) in our development environment._
```

CONTRIBUTING.md
```diff
@@ -127,99 +127,77 @@ This will serve the app at [localhost:4141](http://localhost:4141/), and watch f
...
 
-```bash
-kedro run
-```
+Alternatively, you can synchronously load one of the mock datasets in `/src/utils/data`. The 'spaceflights' dataset is mainly used as mock data for unit testing.
 
-Now you are ready to launch a new viz's development server with a real Kedro project. First, changing your directory back to kedro-viz:
+Finally, you can use pseudo-random data, which is procedurally-generated on page load, and is often useful for local development. Random data can be seeded with a hash string, which will allow you to replicate a generated layout. You can supply a seed with a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`. If you do not supply a seed, the app will generate a new pseudo-random one every time, and will output it to the browser console in case you wish to reuse it.
```
studioswong, line 151: Nice one - thanks for adding this to bridge the gap! Great one. 

```suggestion
Finally, you can use pseudo-random data, which is procedurally-generated on page load, and is often useful for local development. Random data can be seeded with a hash string, which will allow you to replicate a generated layout. You can supply a seed with a `seed` query string in the URL, e.g. `http://localhost:4141/?data=random&seed=oM4xauN4Whyse`. If you do not supply a seed query to the URL, the app will generate a new pseudo-random seed on every browser refresh, and will output it to the browser console in case you wish to reuse it.
```

studioswong (approved):
Awesome - thanks for this work! 👍  I've left a few comments and suggestions above but nothing majorly blocking.

CONTRIBUTING.md
AntonyMilneQB: Ah, that makes sense. Thanks for explaining @studioswong

AntonyMilneQB: But why should we be suggesting <=14.18.0 rather than the latest 14.18.3 (which is what I've been using and seems to work ok)?

CONTRIBUTING.md
```diff
@@ -127,99 +127,77 @@ This will serve the app at [localhost:4141](http://localhost:4141/), and watch f
...
+Kedro-Viz provides a `SQLiteSessionStore` that users can use in their project to enable experiment tracking functionality. If you want to use this session store with the development server, make sure you don't use a relative path when specifying the store's location in `settings.py`. For example, `demo-project` specifies the local `data` directory within a project as the session store's location as follows: 
 ```python
 SESSION_STORE_ARGS = {"path": str(Path(__file__).parents[2] / "data")}
 ```
 
-### Data sources
-
-Kedro-Viz uses a unique identifier to determine the data source from one of several available sources. You can configure this by appending a query string to the URL, e.g. `http://localhost:4141/?data=random`. Alternatively, you can set it with an environment variable when starting up the dev server:
+Owing to this coupling between the project settings and Kedro-Viz, if you wish to execute any Kedro commands on `demo-project` (including `kedro run`), you will need to install the Kedro-Viz Python package. To install your local development version of the package, run:
```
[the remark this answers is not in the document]
AntonyMilneQB: You are right that there's two options - either install a released version from PyPI (`pip install kedro-viz`) or install your local development version (the command given). However, when you do `make run` it will always use the local development version (since that sets `PYTHONPATH=$(shell pwd)/package`). So for consistency I think it's best to only use the local development version for the session store as well - otherwise your project is using a different version of kedro-viz from the what the backend server uses. I will clarify this 🙂

CONTRIBUTING.md
studioswong: yea good point - yea let's go for <=4.18.3

CONTRIBUTING.md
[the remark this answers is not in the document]
studioswong: sounds great - yea let's just scope this to only mention the installation of the local development version ( which is what this section is mainly about anyways). 👍

Makefile
```diff
@@ -14,6 +14,8 @@ clean:
 	find . -regex ".*/__pycache__" -exec rm -rf {} +
 	find . -regex ".*\.egg-info" -exec rm -rf {} +
 
+PROJECT_PATH ?= demo-project
```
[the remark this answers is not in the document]
studioswong: great one!

package/kedro_viz/server.py
```diff
@@ -108,13 +108,14 @@ def run_server(
     )
     args = parser.parse_args()
 
-    source_dir = bootstrap_project(args.project_path)
+    project_path = (Path.cwd() / args.project_path).absolute()
```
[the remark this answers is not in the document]
studioswong: Niceeee

CONTRIBUTING.md
```diff
@@ -142,99 +142,77 @@ This will serve the app at [localhost:4141](http://localhost:4141/), and watch f
...
-Then launch the server with:
+As far as the development server for the backend is concerned, you only need to run `make build` once when you first setup the project. You can then launch the server with:
 
 ```bash
-make run PROJECT_PATH=<path-to-your-test-project>/new-kedro-project
+make run
 ```
-This command will launch a Kedro-Viz server at [localhost:4142](http://localhost:4142) and serve data from a real Kedro pipeline located at the project path supplied to the command. From then on, launching the app locally at [localhost:4141](http://localhost:4141) will pull data from the Kedro-Viz server that is running on port 4142. To avoid needing to supply `PROJECT_PATH` every time you call `make run`, you can set it as an environment variable:
+
+This command will launch a Kedro-Viz server at [localhost:4142](http://localhost:4142) and serve data from `demo-project`. From then on, launching the app locally at [localhost:4141](http://localhost:4141) will pull data from the Kedro-Viz server that is running on port 4142.
```
tynandebold, line 192: I'm not sure how clear it is at this stage of the instructions that the backend and frontend are now running in tandem and that one is feeding the other.

Perhaps adding in another reference here to running the frontend with `npm start` will hammer home the point that now the backend and frontend are running together?

Do we need to be as explicit as saying something like, "You need to have two terminal windows open..." etc etc?

```suggestion
This command will launch a Kedro-Viz server at [localhost:4142](http://localhost:4142) and serve data from `demo-project`. From then on, launching the app locally with `npm start` at [localhost:4141](http://localhost:4141) will pull data from the Kedro-Viz server that is running on port 4142.
```

tynandebold (approved):
Looks great.

I've added one comment about a potential increase of directional clarity. Curious to hear your thoughts.

limdauto (approved):
Thank you this is amazing

CONTRIBUTING.md
AntonyMilneQB: This is a great point, thank you! I've made it clearer in 99c173a. Note that if you're just doing backend development, you can actually run the backend server at 4242 without the frontend one

tynandebold: Great! Thanks for adding.

PR #727, opened 2022-02-02
[KED-3060] Create viz version graphql endpoint

tynandebold:
## Description

For an upcoming frontend feature, we need a way to programmatically check the user's installed Viz version and the latest one released on PyPI.

## Development notes

I've created a GraphQL query (`version`) that returns the current and latest versions.

## QA notes

To test:

1. Check out the branch
2. Run `make run`
3. Open http://127.0.0.1:4142/graphql
4. Run this query: 

```graphql
query MyQuery {
  version {
    installed
    latest
  }
}
```

The output should look akin to this: 

```json
{
  "data": {
    "version": {
      "installed": "4.3.1",
      "latest": "4.3.1"
    }
  }
}
```

P.S. this is my first Python PR. Please be critical of my code and help me learn :)

## Checklist

### DIFF
--- a/package/kedro_viz/api/graphql.py
+++ b/package/kedro_viz/api/graphql.py
@@ -19,7 +19,6 @@ from typing import (
     cast,
 )
 
-import requests
 import strawberry
 from fastapi import APIRouter
 from strawberry import ID
@@ -27,6 +26,7 @@ from strawberry.asgi import GraphQL
 
 from kedro_viz import __version__
 from kedro_viz.data_access import data_access_manager
+from kedro_viz.integrations.pypi import get_latest_version
 from kedro_viz.models.experiments_tracking import RunModel, UserRunDetailsModel
 
 logger = logging.getLogger(__name__)
@@ -126,10 +126,7 @@ def get_version() -> Version:
     Returns:
         the currently installed and most-recent released version of Viz.
     """
-    package = "kedro-viz"
-    response = requests.get(f"https://pypi.org/pypi/{package}/json")
-    latest_version = response.json()["info"]["version"]
-
+    latest_version = get_latest_version()
     version = Version(current=__version__, latest=latest_version)
     return version
 
@@ -328,10 +325,7 @@ class Query:
         """Query to get data for specific runs from the session store"""
         return get_run_tracking_data(run_ids, show_diff)
 
-    @strawberry.field
-    def version(self) -> Version:
-        """Query to get current and latest Kedro Viz versions"""
-        return get_version()
+    version: Version = strawberry.field(resolver=get_version)
 
     runs_list: List[Run] = strawberry.field(resolver=get_all_runs)
 
--- a/package/kedro_viz/api/graphql.py
+++ b/package/kedro_viz/api/graphql.py
@@ -21,12 +21,13 @@ from typing import (
 
 import strawberry
 from fastapi import APIRouter
+from semver import VersionInfo
 from strawberry import ID
 from strawberry.asgi import GraphQL
 
 from kedro_viz import __version__
 from kedro_viz.data_access import data_access_manager
-from kedro_viz.integrations.pypi import get_latest_version
+from kedro_viz.integrations.pypi import get_latest_version, is_running_outdated_version
 from kedro_viz.models.experiments_tracking import RunModel, UserRunDetailsModel
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


limdauto:
Which screen are we doing this on? If we do it on the flowchart, we might need a rest endpoint. Otherwise we will need to port Apollo client to the flowchart screen, which is actually a good thing but might increase the scope of the ticket.

package/kedro_viz/api/graphql.py
```diff
@@ -305,6 +328,11 @@ def run_tracking_data(
         """Query to get data for specific runs from the session store"""
         return get_run_tracking_data(run_ids, show_diff)
 
+    @strawberry.field
+    def version(self) -> Version:
```
limdauto, line 332: If it's a function without parameter

```
version: Version = strawberry.field(resolver=get_version)
```

tynandebold: Slick.

package/kedro_viz/api/graphql.py
```diff
@@ -119,6 +121,19 @@ def get_runs(run_ids: List[ID]) -> List[Run]:
     )
 
 
+def get_version() -> Version:
```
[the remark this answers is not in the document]
tynandebold: Nice.

tynandebold:
> Which screen are we doing this on? If we do it on the flowchart, we might need a rest endpoint. Otherwise we will need to port Apollo client to the flowchart screen, which is actually a good thing but might increase the scope of the ticket.

It's mocked up on the flowchart but I think it should exist on both routes, flowchart and experiment tracking.

Why would we need a rest endpoint? Our Apollo `<Provider />` wraps the whole app, not just the experiment tracking part. Or is it for another reason?

limdauto:
@tynandebold ah I see. I was under the impression that the Apollo Provider was on the exp route only. Don't mind me then. This is actually fantastic news.

limdauto (approved):
Nice one!

rashidakanchwala (commented):
Look's great Tynan! :) 
i suppose we will just need to update the fe schema and add the tests !

studioswong:
> @tynandebold ah I see. I was under the impression that the Apollo Provider was on the exp route only. Don't mind me then. This is actually fantastic news.

Yep @limdauto the idea behind wrapping the ApolloProvider on app level rather than  on a route is to allow the gradual migration of all endpoints to graphql, with new needs on data to be built out as new graphql queries while slowly stripping away the need for rest in stages ( I am working on a spike doc on the next steps for that as we speak - will share with you once it's drafted. )

src/apollo/schema.graphql
```diff
@@ -49,3 +50,8 @@ union UpdateRunDetailsResponse = UpdateRunDetailsSuccess | UpdateRunDetailsFailu
 type UpdateRunDetailsSuccess {
   run: Run!
 }
+
+type Version {
+  installed: String!
```
studioswong, line 55: naming this as `Current` might be more relevant as I understand there can be multiple versions of a package installed in a python environment ( @limdauto @AntonyMilneQB can correct me if I'm wrong)

studioswong (commented):
just tested and it works well, nice one. 

As @rashidakanchwala mentioned, please add the relevant python tests for this endpoint.

src/apollo/schema.graphql
limdauto: there can't be multiple versions of a package installed in the same python environment. There can be many environments.

studioswong: Thanks for clarifying @limdauto - on that note, I've heard that it is possible to install multiple versions of a package by specifying it to be installed in a different directory other than the default `site-packages`? ( not that it is a common use case at all) Again, I'm no python dev, so you know best 😄 

On the subject, `Current` version to me sounds more intuitive as we normally refer to 'current' and 'latest' versions - curious to hear your thoughts.

tynandebold: The reason I didn't like `current` is because I didn't know what it was describing, and it seemed like it could describe both. As in: is it the _current_ version the user has installed? Or is it the _current_ version that's released on PyPI? 

With `installed` and `latest` I think it's more clear. It might even be clearer if it's `installed` and `latestRelease`.

AntonyMilneQB: I also don't like `current` for the same reason as @tynandebold mentioned. `installed` and `latest` make most sense to me.

With that said, we are already using `current` as the variable name [where we do the CLI nudge](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/launchers/cli.py#L76-L80). So whatever we go for, let's be consistent so we don't have current, latest and installed in the codebase.

AntonyMilneQB: @studioswong you could indeed install different version of the same Python package to different locations. But when we do `from kedro_viz import __version__`, it will be well-defined and import from only one installation of the package. So it does make sense to refer to the "installed Python package" as a unique entity.

AntonyMilneQB (commented):
Awesome work, and top-notch Pythoning 🐍 

I have some minor comments and questions - easier to discuss over a call, but just to list them:
* pedantic suggestions for `get_version()`
* when does this get called?
* question about the schema types and whether some fields are nullable
* `latest < installed` logic
* apollo schema (just for my understanding)

rashidakanchwala (approved):
Looks great Tynan!!! thanks

PR #730, opened 2022-02-07
[KED-2983] Migrate SearchBar component

tynandebold:
## Description

We're migrating the components we use from [Kedro-UI](https://github.com/quantumblacklabs/kedro-ui/), and this is one of those migrations. Here's the [relevant ticket](https://jira.quantumblack.com/browse/KED-2983) (QB-only access).

## Development notes

I've only migrated the [`SearchBar`](https://github.com/quantumblacklabs/kedro-ui/tree/master/src/components/search-bar) component because that's the only one we use. 

Further, I've put this new component in the [`./src/components/ui`](https://github.com/kedro-org/kedro-viz/tree/main/src/components/ui) folder. I think we should adopt this architecture pattern going forward, and even move some of our current UI-only components into this folder, like the `modal` and `button`. This will separate our concerns a bit more and also allow for easier hooking into Viz in the future if we want to start adding other services or components.

Note that there are still some references to `kui` in the styles here (e.g. `kui-input`). That's due to the fact that the Kedro-UI input is a dependency of the `SearchBarRenderer` component. We'll have to move that input over in it's own PR and then resolve this dependency. 

UPDATE: I've also migrated the Kedro-UI `<Input />` component into the `search-bar` UI folder. That input was tightly coupled to the search bar, and we already have a generic input component anyways, so this one exists solely for the purpose of the search bar (for now). In the future, maybe we can combine the two inputs into one, but for now, that shouldn't happen here.

## QA notes

Check out this branch and then compare the behavior and looks to what's on the demo site.

## Checklist


AntonyMilneQB (commented):
I haven't tried it out yet, plus I don't know anything about the code, so take this with a pinch of salt but LGTM!

Just to check: it used to be `kui_searchbar` (no hyphen) but you deliberately made it `search-bar` (with hyphen)? 

I like the idea of putting more stuff in `components/ui`. To a noob like me the number of directories in `components` is a little overwhelming and confusing, so if they could be grouped together a bit that would be nice.

tynandebold:
> I haven't tried it out yet, plus I don't know anything about the code, so take this with a pinch of salt but LGTM!
> 
> Just to check: it used to be `kui_searchbar` (no hyphen) but you deliberately made it `search-bar` (with hyphen)?
> 
> I like the idea of putting more stuff in `components/ui`. To a noob like me the number of directories in `components` is a little overwhelming and confusing, so if they could be grouped together a bit that would be nice.

Thank you!

Yes, that was deliberate. To my knowledge, search bar is in fact two words, so `searchbar` written as such was wrong. Further, it's written as `search-bar` in the [component library](https://github.com/quantumblacklabs/kedro-ui/tree/master/src/components/search-bar), so I think it was a mistake to have it as all one string. We're moving away from `kui` as well, which is why that's gone.

Glad you like the `ui` folder idea, too. Agree `/components` is a little hectic.

rashidakanchwala (approved):
Thanks Tynan ... LGTM! I also like the idea of having a ui folder.

src/components/ui/search-bar/search-bar-renderer.jsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import CloseIcon from '../../icons/close';
+import Input from '@quantumblack/kedro-ui/lib/components/input';
```
studioswong, line 6: This suggests the need to also migrate the `Input` component to our code base as well - @tynandebold have you gotten a chance to check the extent of usage of this component to migrate this as well as part of this ticket?

src/components/ui/search-bar/search-bar-renderer.test.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { shallow, configure } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import SearchBarRenderer from './search-bar-renderer';
+
+configure({ adapter: new Adapter() });
```
studioswong, line 6: is this set up needed for this test?

studioswong (commented):
I've tested the search bar and it behaves as it should, good call for setting up the UI folder.

My main comment relates to the 'Input' component from Kedro-UI that is introduced in this ticket - it would be worth to do some follow up to see the extent in which that component is consumed in the search bar ( it might not require all the functions that was within that input component), to see whether the set of of this component can be incorporated as part of this PR.

src/components/ui/search-bar/search-bar-renderer.jsx
tynandebold: I can see about moving this one too.

src/components/ui/search-bar/search-bar-renderer.test.js
tynandebold: not sure. this file was a 1-to-1 migration. will check

src/components/ui/search-bar/search-bar-renderer.jsx
tynandebold: I've moved it over. It's massively increased the scope of the PR now, but there wasn't much of an alternative.

AntonyMilneQB (approved):
LGTM but do take that with a large pinch of salt :D

studioswong (commented):
one observation - I notice that the migrated file names are `.jsx` instead of `js`?

studioswong (commented):
one small fix before merging in: I think the search-bar-renderer is still a jsx file.

Thank you :)

PR #731, opened 2022-02-07
[KED-2984] Remove Kedro-UI webfont usage

tynandebold:
## Description

We're migrating the components we use from [Kedro-UI](https://github.com/quantumblacklabs/kedro-ui/), and this is one of those migrations, though this time it's a deletion and simplification. Here's the [relevant ticket](https://jira.quantumblack.com/browse/KED-2984) (QB-only access).

## Development notes

The [WebFontLoader](https://github.com/typekit/webfontloader) library we were using from TypeKit/Google hasn't been updated since May 2017. We weren't really leveraging it's useful options either. We we using it to trigger a Redux action if the web font loaded and even if it didn't:

```js
LoadWebFont({
  active: setFontLoaded,
  inactive: setFontLoaded,
});
```

The result of that action would be one indicator to Viz if the flowchart should load or not. Since we made no distinction between a successful/unsuccessful font load, I decided to take it out completely. So I've removed the two instances of the Kedro-UI Webfont business that was present in `src/components/app/app.js`:

```jsx
import '@quantumblack/kedro-ui/lib/styles/app-no-webfont.css';
import LoadWebFont from '@quantumblack/kedro-ui/lib/utils/webfont.js';
```

In it's place, I'm loading the Google Font we use in the `index.html` file:

```html
<link rel="preconnect" href="https://fonts.googleapis.com" />
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
<link
  href="https://fonts.googleapis.com/css2?family=Titillium+Web:wght@400;600&display=swap"
  rel="stylesheet"
/>
```

And bringing in the few lines of CSS we need for the font styles:

```css
.kedro {
  -webkit-font-smoothing: antialiased;
  font-family: 'Titillium Web', sans-serif;
  font-size: 10px;
  font-weight: 400;
  letter-spacing: 0.1px;
  line-height: 1.4;
}
```

I've also removed all references that dealt with testing the `fontLoaded` action/state/etc.

This part of the app is much simpler now, and in a way starts us along a lengthy path moving away from Redux.

## QA notes

Check out the branch and see if you notice anything different with the fonts.

## Checklist

### DIFF
--- a/src/store/initial-state.js
+++ b/src/store/initial-state.js
@@ -17,7 +17,6 @@ export const createInitialState = () => ({
   prettyName: settings.prettyName.default,
   ignoreLargeWarning: false,
   loading: {
-    graph: false,
     pipeline: false,
     node: false,
   },
--- a/public/index.html
+++ b/public/index.html
@@ -10,12 +10,16 @@
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
     <title>Kedro-Viz</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Titillium+Web:wght@400;600&display=swap"
+      rel="stylesheet"
+    />
   </head>
 
   <body>
-    <noscript>
-      You need to enable JavaScript to run this app.
-    </noscript>
+    <noscript> You need to enable JavaScript to run this app. </noscript>
     <div id="root"></div>
   </body>
 </html>
--- a/src/actions/actions.test.js
+++ b/src/actions/actions.test.js
@@ -13,7 +13,6 @@ import {
   TOGGLE_TEXT_LABELS,
   TOGGLE_THEME,
   UPDATE_CHART_SIZE,
-  UPDATE_FONT_LOADED,
   TOGGLE_CODE,
   TOGGLE_MODULAR_PIPELINE_FOCUS_MODE,
   changeFlag,
@@ -30,7 +29,6 @@ import {
   toggleTextLabels,
   toggleTheme,
   updateChartSize,
-  updateFontLoaded,
   toggleFocusMode,
 } from '../actions';
 import {
@@ -272,15 +270,6 @@ describe('actions', () => {
       chartSize,
     };
     expect(updateChartSize(chartSize)).toEqual(expectedAction);
-  });
-
-  it('should create an action to update the state when the webfont has loaded', () => {
-    const fontLoaded = true;
-    const expectedAction = {
-      type: UPDATE_FONT_LOADED,
-      fontLoaded,
-    };
-    expect(updateFontLoaded(fontLoaded)).toEqual(expectedAction);
   });
 
   it('should create an action to change a flag', () => {
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -154,19 +154,6 @@ export function updateZoom(zoom) {
   };
 }
 
-export const UPDATE_FONT_LOADED = 'UPDATE_FONT_LOADED';
-
-/**
- * Update whether the webfont has loaded, which should block the chart render
- * @param {Boolean} fontLoaded Whether the font has loaded
- */
-export function updateFontLoaded(fontLoaded) {
-  return {
-    type: UPDATE_FONT_LOADED,
-    fontLoaded,
-  };
-}
-
 export const TOGGLE_MINIMAP = 'TOGGLE_MINIMAP';
 
 /**
--- a/src/components/app/app.js
+++ b/src/components/app/app.js
@@ -2,10 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Provider } from 'react-redux';
 import 'what-input';
-import '@quantumblack/kedro-ui/lib/styles/app-no-webfont.css';
-import LoadWebFont from '@quantumblack/kedro-ui/lib/utils/webfont.js';
 import configureStore from '../../store';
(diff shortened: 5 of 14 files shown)


rashidakanchwala (approved):
I tested this and it works fine. The font is loading correctly. The refactoring makes sense, I wonder why we used WebFonts if we could easily do this. Thanks @tynandebold

src/store/initial-state.js
```diff
@@ -17,7 +17,6 @@ export const createInitialState = () => ({
   prettyName: settings.prettyName.default,
   ignoreLargeWarning: false,
   loading: {
-    graph: false,
```
studioswong, line 20: This state cannot be deleted as this is also utilised for the display of the spinner during graph calculations ( refer to   `selectors/loading.js` and the tests outlined under `graph.test.js`

studioswong, line 20: This cannot be deleted as this is still used within the `loading` selectors to trigger the loader during graph calculations

studioswong (changes requested):
I've tested this and it generally works well, thanks for proposing a much simplier way to load the fonts and stripping the redux stuff. 

One main concern is with regards to the `loading.graph` state, which is also used to trigger the loading spinner during graph calculation ( see my below comment), which is a super quick fix.

src/store/initial-state.js
tynandebold: Having this in breaks one of the tests in `src/components/app/app.test.js`:

The failure: 

```
App › updates the store › but does not override non-pipeline values
```

It expects `"graph": false` but receives `"graph": true`, and I think that's because we not longer prevent the graph from loading based on the status of the font loading. This is why I took it out.

Should I also remove the other relevant `getGraphLoading` lines from `selectors/loading.js` or leave the graph property in there and get the tests to pass? I just don't know if it's relevant anymore, that's the thing.

limdauto: Yea it's because the layout calculation is triggered in this test (`fontLoaded` was what prevented it before, which feels like a hack at best and maybe a coincidence at worst). I wonder if there is anyway to wait for the calculate to be done before assert? Otherwise we can probably modify the expectation to assert for `"graph": true`

tynandebold: Yes, exactly.

> I wonder if there is anyway to wait for the calculate to be done before assert?

We could maybe try a timeout or something. That doesn't seem great though. Changing to `"graph": true` in the `loading` object solves it. I'd rather do that.

limdauto: I wonder if a search for webfont on kedro-viz slack channel would yield anything interesting? I recall seeing it being discussed on slack.

tynandebold: Good call. Here's the relevant discussion: 

image

The statement from Richard makes me think we don't need this graph check anymore. He says:

> I don’t think it currently does delay booting the whole app, it only delays running a layout calculation and displaying the whole graph. The font is loaded asynchronously while the rest of the app is loaded, but then it doesn’t do layout/graph rendering until after the font is loaded because it’d just have to update it again anyway.

I think we can safely do the layout/graph rendering without messing up the layout because of the font loading. Meaning: we don't have to block it due to fonts.

Plus, now we're using `<link rel="preconnect" ... />` which was essentially a wish of his.

studioswong: I agree there's no point in blocking the loading of the graph due to fonts ( in the old setup, the graph can still load with the fallback font when, say, there's no internet and is unable to retrieve web fonts, so there's no need to delay it). That said though, `loading.graph` is not just used for checking the loading of the fonts - it is still used to check if the `calculateGraph` action is still executing, and hence, display the spinner. 

In regards to the test, I might approach it to mocking the dispatch of the action to update the graph loader rather than changing the value in `prepareNonPipelineState` as it's a once off situation for this test ( to simulate the state when the graph finishes loading) - changing it to `graph: true` within the `prepareNonPipelineState` might have further implications to the entire app. ( I know it's a much more painful way, but it's much better to mock for a one off situation rather than changing the state initialisation logic.)

tynandebold: I think I follow you but I'm not 100% sure. I've made the changes [here](https://github.com/kedro-org/kedro-viz/pull/731/commits/cadc1b3ff7d9f0971f6253179ecd0f583db7b553) by adding a TOGGLE_GRAPH_LOADING test and then removing the non-pipeline values one, because the `graph` value in that doesn't match anymore due to the removal of the `fontLoading`. Let me know if this is what you meant.

limdauto (approved):
LGTM

src/store/initial-state.js
studioswong: just checked the code, and haha nope, this isn't what I meant at all ( esp with the deletion of the test), but regardless it is good to add the test for the reducer. 

what I meant initially was in this test is as follows:

```
test('but does not override non-pipeline values', () => {
      const wrapper = shallow(<App data={demo} />);
      wrapper.setProps({ data: spaceflights });
      // mock the dispatch of the action for graph loading back to false
      expect(getState(wrapper)).toMatchObject(prepareNonPipelineState({}));
    });
  });
```

alternatively, a quick getaround to this ( other than deleting the test) could be just manually assign the nonpipelinestate object to false:

```
test('but does not override non-pipeline values', () => {
      // mock the initial pipeline state except for the graph loading state given the state updates immediately on mount
      const mockNonPipelineState = Object.assign(prepareNonPipelineState({}), {loading:{graph: false}})

      const wrapper = shallow(<App data={demo} />);
      wrapper.setProps({ data: spaceflights });
      expect(getState(wrapper)).toMatchObject(mockNonPipelineState);
    });
  });
```

tynandebold: Ohhhh ok, I get it. Please check! [Commit is here](https://github.com/kedro-org/kedro-viz/pull/731/commits/627c0f57024848307d4d80fe65a7c08a7ae9964c).

studioswong (approved):
great work, thanks for pushing through with the test ⭐

PR #740, opened 2022-02-10
[Chore] Remove unnecessary test

limdauto:
## Description

As @AntonyMilneQB this test does the exact same thing as `test_endpoint_main_no_session_store` so removing it.

## Checklist


AntonyMilneQB:
Thanks for checking this! But I think we should probably remove the other test, `test_endpoint_main_no_session_store`, since that is associated with the fixture `example_api_no_session_store` that only gets used once so could also be removed.

Do we have tests for the main endpoint when there _is_ a session store? The fact there's a test called `test_endpoint_main_no_session_store` makes me feel like there should be something akin to `test_endpoint_main_session_store` somewhere - or is the session store stuff already covered by another set of tests? 🤔

limdauto:
@rashidakanchwala do you have any suggestion about cleaning up these tests? Otherwise I will do another pass and rejig things a bit.

rashidakanchwala (approved):
It makes sense... both tests do the same chk

rashidakanchwala (changes requested):
hi @AntonyMilneQB  is right -- the fixture example_api_no_session_store is the same as example_api. So instead of removing test_endpoint_main.  we can remove the test_endpoint_main_no_session_store for now

in the Kedro-viz backlog, we still have a ticket (KED-2910) to write an integration test for session_store. 
Other sqlite_store unit tests are covered in test_intergration/sqlite folder. 
Also, in test_graphql/conftest -- we are mocking the db_sessions!

limdauto:
@AntonyMilneQB @rashidakanchwala thanks. pushed

rashidakanchwala (approved):
Thanks Lim!!

PR #745, opened 2022-02-15
[KED-2986] Remove references to Kedro-UI

tynandebold:
## Description

This should complete the migration away from Kedro-UI.

## Development notes

The references have been removed in all necessary files and the library has been uninstalled using NPM.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/745"></a>


AntonyMilneQB (approved):
Nice one! 

Are all the old Kedro-UI things in the ui folder now? I was sort of expecting to see more there, but I don't really know what stuff was in Kedro-UI before. https://github.com/kedro-org/kedro-viz/tree/feature/KED-2986-remove-kedro-ui-references/src/components/ui

studioswong:
great call @AntonyMilneQB  - there are a few other components ( such as the `dropdown`, `buttons` and `modal` component) that was migrated over from Kedro-UI, it would be great to move them over to the `ui` folder as well in this PR to conclude the work. 

and thanks @tynandebold for setting this up.

AntonyMilneQB:
If it's easy to do so then I'd say yes, go for it! Or in a separate PR if you prefer, whatever's easiest for you.

AntonyMilneQB:
Nice, thank you!

As part of a separate ticket is there any possibility of grouping other components together into subdirectories? To me `components` looks like a kind of a lucky dip where you could pull anything out from a tiny specific piece (like `large-pipeline-warning`) to a whole chunk of app (like `experiment-tracking`). It seems kind of weird that these are on the same level. But maybe this is how react components are normally arranged? I don't know.

studioswong:
- > Nice, thank you!
> 
> As part of a separate ticket is there any possibility of grouping other components together into subdirectories? To me `components` looks like a kind of a lucky dip where you could pull anything out from a tiny specific piece (like `large-pipeline-warning`) to a whole chunk of app (like `experiment-tracking`). It seems kind of weird that these are on the same level. But maybe this is how react components are normally arranged? I don't know.

yea I completely feel you, that was my first feeling when I started exploring the codebase too. The flat structure of components stems from the previous history with Kedro-Viz mainly being just the flowchart and sidebar with a set of highly customized components just for this usecase, which, given that we have expanded, now would not be applicable. 

The beauty of React ( as compared to other very opinionated frameworks, say Ember) is that it really gives a lot of flexibility to how we structure our folders, and I've seen so many different patterns from different teams and projects according to their team needs. However that said, the way I see we are headed for Kedro-Viz, we can structure our folder as follows:

- top level: `/component`

- second level: contains all higher level folders according to **function** or **feature set** - `utils`, `ui`, `styles`, `store`,  `experiment-tracking` (the folder that houses all customised components just built for this feature) etc. This will also house the entry point of our app, which is `/app`. 

For us this would mean moving all existing customised components for flowchart into its dedicated `/flowchart` folder as well.

- third level: folders containing files that **actually implements the function** of the second level folders - for `/ui` it would house all the reusable components such as `/button`, `/experiment-tracking` would house all components made just for implementing experiment tracking features, etc.

from the third level on that's when we can have the flexibility to build on any file structure that we think suits whatever function the second level decides. 

this is just a very high-level idea of what we can set as our folder structure guideline, variations will of course happen as we move forward 😄

studioswong (commented):
Thanks for migrating all the other elements for kedro-ui @tynandebold , folder looks much better arranged now!

Looking at the logic of the migration we seem to establish the pattern of moving UI components that are 'reusable', I would suggest that `search-list` can be migrated there too along with the `search-bar` as it's also a standalone reusable search component ( we can host this under a `/search` folder for easier management.)

We might be able to move the `plotly-chart` under there too as it can also act as a standalone component, but I'll leave that up to you as this is getting out of scope and just a nice to have 😄 

Overall nice work!

tynandebold:
Great thoughts from both of you. These are solid and sound ideas.

I think a couple of them should be discussed a little bit more and ultimately acted upon and implemented, just not in this PR. It's starting to balloon. Therefore, I'd ask that we finish this up, create a new issue with the ideas stated above, and proceed from there. 

@studioswong if you're happy with the current state it would be great if you can approve this!

STYLE_GUIDE.md
```diff
@@ -123,7 +118,7 @@ Kedro-Viz has both light and dark themes, and every component must be styled app
 
 ![Kedro-Viz Color Styleguide](./.github/img/kedro_colours.png)
 
-The themes are set using the same theme classes that Kedro-UI uses - `kui-theme--light` and `kui-theme--dark` - applied on the top-level container element. To simplify code and avoid increasing CSS specificity, we use CSS custom properties to handle theme colours. We recommend that you use these too where possible - they are defined in the App component.
+The themes are set using two classes - `kui-theme--light` and `kui-theme--dark` - applied on the top-level container element. To simplify code and avoid increasing CSS specificity, we use CSS custom properties to handle theme colours. We recommend that you use these too where possible - they are defined in the App component.
```
studioswong, line 121: I recall this being discussed in a previous PR, we should remove the naming reference of `kui-` in the theme classes as the last piece of the puzzle to complete the Kedro-UI migration work - this could be scoped out in a seperate PR.

studioswong (approved):
looks good, thanks for migrating the components! 

We do have an important task of renaming our theme classes away from `kui-` to complete the kedro-UI migration ( see below comment), but we can scropt this out in a separate PR 😄

STYLE_GUIDE.md
tynandebold: Completely agree. Good call.

PR #746, opened 2022-02-16
[KED-3061] Build out 'upgrade Kedro-Viz' call to action UI journey

studioswong:
## Description

resolves [KED-3061](https://jira.quantumblack.com/browse/KED-3061).

This PR builds out the journey to remind users to update their kedro viz when a new release is made. 

Screenshot 2022-02-16 at 09 57 53

Screenshot 2022-02-16 at 09 58 12

Screenshot 2022-02-16 at 09 58 23

Screenshot 2022-02-16 at 10 00 59

Screenshot 2022-02-16 at 10 01 07

## Development notes

For this PR I have made the following changes:

1. Moved the ApolloProvider from the wrapper level to the app level to allow fetching of data from the graphql endpoint within the wrapper. 
2. Set up the `update-reminder` component as well as related styles
3. Extracted and refactorered the existing command display from metadata panel into a seperate UI component for reusability ( this helps with the goal of improving our codebase for more reusability)
4. Refactored the metadata panel to use the newly set up `command-copier` component and all relevant styles and test
5. Refactored the current logo for the settings panel to display an indicator when there is a new version for upgrade
6. Set up the reminder message in the settings modal.

Note: Given the move of the apollo provider from the wrapper level to the app level, the current tests for the wrapper is not passing given that Jest doesn't seem to render the wrapper component with it is wrapped under the mocked providers for both Apollo and  Redux. ( However we need both mocked providers for the tests as we need both the redux store and the graphql client to pass data into that component.) 

I am currently finding a best solution to fix this, however this is not a common problem given the unique situation of our codebase using both redux and graphql for data ingestion. I will be investigating this in the meantime and will push the best solution to fix this, along with a series of tests. 

## QA notes

To QA this properly technically we would need to install this local version of Kedro-Viz with Kedro as an end to end journey; however, it is currently impossible to QA this way as there is no newer version of Kedro-Viz avaliable for this local version. 

To QA the UI, simply change the values that are passed in the props for the `settings-modal`, `Global-sidebar` components, as well as removing the checks to trigger the reminder in order to trigger the UI. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/746"></a>


src/components/ui/command-copier/command-copier.js
```diff
@@ -0,0 +1,49 @@
...
+          <span
+            className={modifiers('copy-message', {
+              visible: showCopied,
+            })}
+          >
+            Copied to clipboard.
+          </span>
+          <ul className="toolbox">
+            <IconButton
+              ariaLabel="Copy run command to clipboard."
```
AntonyMilneQB, line 37: Is this the caption when I hover over the copy button? I don't actually see it on my screen so not sure when it's meant to appear?

Either way, I think this label is specific to the old use of the copying `kedro run` and doesn't sound right if this is now a generic button also used for the `pip install` command. Maybe just "Copy command to clipboard" or even "Copy to clipboard" would be better.

The variable name `runCommand` should maybe be updated to just to `command` or maybe even more generic since we could potentially copy something else with this button too (like a dataset filepath)?

AntonyMilneQB:
Looks really good! Just a few initial comments playing around with it:
1. It doesn't look like the "Dismiss" button is clickable (like my mouse cursor doesn't change to the hand) - not sure if this is just a problem on my setup (using Gitpod in Chrome) though?
2. If I dismiss the note should it remember that when I refresh the page? If `--autoreload` is enabled then potentially it refreshes quite a lot by itself.
3. Can we change the wording of the text a bit? I think it would be nice to be more consistent with the upgrade message we give on the CLI:
```
WARNING: You are using an old version of Kedro Viz. You are using version 4.3.0; however, version 4.3.1 is now available.
You should consider upgrading via the `pip install -U kedro-viz` command.
You can view the complete changelog at https://github.com/kedro-org/kedro-viz/releases.
```

studioswong:
> Looks really good! Just a few initial comments playing around with it:
> 
> 1. It doesn't look like the "Dismiss" button is clickable (like my mouse cursor doesn't change to the hand) - not sure if this is just a problem on my setup (using Gitpod in Chrome) though?
> 2. If I dismiss the note should it remember that when I refresh the page? If `--autoreload` is enabled then potentially it refreshes quite a lot by itself.
> 3. Can we change the wording of the text a bit? I think it would be nice to be more consistent with the upgrade message we give on the CLI:
> 
> ```
> WARNING: You are using an old version of Kedro Viz. You are using version 4.3.0; however, version 4.3.1 is now available.
> You should consider upgrading via the `pip install -U kedro-viz` command.
> You can view the complete changelog at https://github.com/kedro-org/kedro-viz/releases.
> ```

Thanks for the questions @AntonyMilneQB, all good ones!

1. Indeed I was going to add the css for the hand cursor but forgot while fixing the tests, I'll add that in - nice spot!
2. yes indeed I have not implemented the persist of the 'dismiss' state - I had given it some thought but decided not to implement this for the following reasons:
    a) I'm not sure whether we should save the 'dismiss' state given that, say the user dismiss for version 3.4.1, we still want to remind them for version 3.4.2. 
    b) Persisting the 'dismiss' state means saving that state in localStorage. The current setup of the localstorage (read and write) is tied in with the redux store, and we currently don't have any setup to tie our localstorage to the graphgql cache / local hooks yet ( there is a ticket in the backlog for this [here](https://jira.quantumblack.com/browse/KED-3015)). Given this setup consuming data from the graphql layer rather than via the redux store, the state is determined by a hook and not a state in the redux store, where I would rather having set up the proper localstorage file structure to accomodate the new grapghql data rather than implementing it here. 
   c) It wasn't specified as a requirement from design anyways ( this might require more thinking from design, such as how long the user can dismiss the message, how often we should bug them, etc), hence for simplicity sake I had decided to implement the most straight forward approach. (besides, the whole point of this pop up is to bug them until they upgrade 😆 )
3. For the upgrade message I simply followed the design - perhaps we could liase with design to see their thoughts on the upgrade message?

src/components/ui/command-copier/command-copier.js
studioswong: the `aria-label` is mainly for accessibility 😄  though I agree we can change `run command` to just `command` to allow this to be more specific.

tynandebold:
Agree with Antony, it looks awesome. Really is a killer feature!

Just a quick note on your first point in your `Development notes` that may save you some headaches with the tests: I'm fairly certain you can pass our `client` as an option to the `useApolloQuery` hook, like so:

```jsx
const { data } = useApolloQuery(GET_VERSIONS, {
  client,
});
```

That will then allow you to keep the `<Provider/>` component within `<Wrapper/>`. From there, you shouldn't have to deal with any of these Redux/Apollo clashes that you described. It's probably also simpler and arguably nicer this way, since Redux and Apollo still remain separated.

I'd suggest to perhaps try this before going down that testing and compatibility rabbit hole.

AntonyMilneQB:
@studioswong Thanks for all the explanation! I think that eventually it would be correct to persist the dismiss state in local storage. Reason being that there are two sorts of ways that people run kedro viz (I think):
1. short-lived instance while developing, in which case user is often killing and restarting the `kedro viz` process at a much greater frequency than we release. Hence if the user dismisses once, the reminder will reappear the next time they launch kedro-viz anyway, so they're not going to miss out on a reminder for 3.4.2 release because they dismissed the previous one for 3.4.1.
2. long-lived instances hosted somewhere the user doesn't have access to the server so can't run the `pip install` command anyway. In this case it also makes sense to dismiss the upgrade message once and then it not to appear again.

Understood that this isn't easy to do technically now and that it's not a big issue so very happy to forget about it now. Just to say that at some point in the future if users do find it annoying and/or the stuff in 2b is resolved then I think persisting to local storage would make sense :)

studioswong:
> 

actually now that we are discussing it, once we have set up localstorage with hooks ( which I don't think it'll be hard to set up anyways), we can save the dismissed version reminder in localstorage  as a cross reference so we can show the popup reminder again when a newer version comes up - it'll be a straight forward one to set up. I'll pop that one up in a separate ticket 😄

src/components/global-toolbar/global-toolbar.js
```diff
@@ -21,6 +21,7 @@ export const GlobalToolbar = ({
   onToggleTheme,
   theme,
   visible,
+  isOutdated,
```
tynandebold, line 24: Would just ask if we can sort props alphabetically.

src/components/icons/settings.js
```diff
@@ -12,6 +12,9 @@ const SettingsIcon = ({ className }) => (
       <path d="M2.8 2.05L4.3 2.05 4.55 5.05 2.55 5.05z" />
       <path d="M12.7 11.95L14.2 11.95 14.45 14.95 12.45 14.95z" />
       <path d="M15.5 8.389c0 3.866-3.134 7-7 7s-7-3.134-7-7 3.134-7 7-7 7 3.134 7 7zm-2 0c0-2.762-2.239-5-5-5s-5 2.238-5 5c0 2.761 2.239 5 5 5s5-2.239 5-5z" />
+      {hasReminder && (
+        <circle id="Oval" fill="#FFBC00" cx="16" cy="4" r="4"></circle>
```
tynandebold, line 16: I wonder if we should use the `$color-accent-yellow` variable here instead of hardcoding it?

src/components/settings-modal/settings-modal.js
```diff
@@ -23,6 +23,8 @@ const SettingsModal = ({
   onTogglePrettyName,
   visible,
   flags,
+  isOutdated,
+  latest,
```
tynandebold, line 27: Maybe `latest` can be a little more specific? It's pretty vague as is. Like `latestVizVersion` or something? 

Sort here if possible.

src/components/ui/command-copier/command-copier.js
```diff
@@ -0,0 +1,49 @@
+import React, { useState } from 'react';
+import modifiers from '../../../utils/modifiers';
+import MetaDataValue from '../../metadata/metadata-value';
+import IconButton from '../../ui/icon-button';
+import CopyIcon from '../../icons/copy';
+import './command-copier.css';
+
+const CommandCopier = ({ command }) => {
```
tynandebold, line 8: So good that this is now a separate component 💥

src/components/ui/command-copier/command-copier.scss
```diff
@@ -0,0 +1,59 @@
...
+  font-size: 0.9em;
+
+  &--no-visible {
+    opacity: 0;
+  }
+}
+
+.copy-message {
+  position: absolute;
+  top: 0;
```
tynandebold, line 22: It seems this needs to be `top: 2px;` instead. I noticed that the `Copied to clipboard.` text doesn't sit exactly along the same baseline as the code text that's being copied.

src/components/update-reminder/update-reminder.js
```diff
@@ -0,0 +1,72 @@
...
+              className="close-button"
+              icon={CloseIcon}
+              onClick={() => setDismiss(true)}
+            />
+          </div>
+        </div>
+        <div className="update-reminder-expanded-detail">
+          <p>
+            We're excited to announce that Kedro-Viz {latest} has been released.{' '}
+            <br />
```
tynandebold, line 35: I'd suggest taking out this `<br />`. I think having just the word **released** on its own line looks a bit disjointed. It also creates a fourth line, when the design only has three:

image

src/components/update-reminder/update-reminder.scss
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.update-reminder-expanded-detail {
+  height: 340px;
+  width: 484px;
+  position: absolute;
+  right: 0px;
+  top: 48px;
+  z-index: 2;
+
```
tynandebold, line 82: Remove this empty line?

src/components/wrapper/wrapper.js
```diff
@@ -1,43 +1,68 @@
...
-          <Route path={['/runsList', '/runsList/:id']}>
-            <ExperimentWrapper />
-          </Route>
-        </Switch>
-      </Router>
-    </div>
-  </Provider>
-);
+export const Wrapper = ({ theme }) => {
+  const { data } = useApolloQuery(GET_VERSIONS, { client });
```
tynandebold, line 22: Should we reassign `data` to something more specific, like `vizVersionData`? As this is such a top-level component it could be better to be more explicit about what this is here.

src/components/wrapper/wrapper.js
```diff
@@ -1,43 +1,68 @@
...
+  const [dismissed, setDismissed] = useState(false);
+  const [isOutdated, setIsOutdated] = useState(false);
+  const [latest, setLatest] = useState(null);
```
tynandebold, line 25: I said this is another comment: it could be nice to make `latest` more specific.

src/components/update-reminder/update-reminder.js
```diff
@@ -0,0 +1,72 @@
...
+import IconButton from '../ui/icon-button';
+import CloseIcon from '../icons/close';
+
+import './update-reminder.css';
+
+const UpdateReminder = ({ versions, setDismiss }) => {
+  const [expand, setExpand] = useState(false);
+  const { latest } = versions;
+
+  const command = 'pip install -U kedro-viz && kedro viz';
```
tynandebold, line 12: Guess we need to change this based on what @AntonyMilneQB said?

src/components/update-reminder/update-reminder.scss
```diff
@@ -0,0 +1,113 @@
...
+  right: 0px;
+  top: 0px;
+  background-color: var(--color-header-bg);
+  z-index: 2;
+  font-size: 12px;
+  display: flex;
+  justify-content: space-between;
+
+  .close-button-container {
+    padding-right: 20px;
```
tynandebold, line 56: One thing about this padding is that it makes the close icon clickable when you aren't hovering over the icon. Maybe margin would be better here?

tynandebold (changes requested):
Look amazing and works great! Nice work. 

I left a few comments for some very small changes but none are really blockers. Still waiting on Mackay to iron out any text inconsistencies between the CLI and this work here. Otherwise it's ready to roll.

src/components/update-reminder/update-reminder.js
```diff
@@ -0,0 +1,72 @@
...
+          </p>
+        </div>
+      </>
+    );
+  }
+
+  return (
+    <div className="update-reminder-unexpanded">
+      <button className="kedro" onClick={() => setExpand(true)}>
+        <p>Kedro-Viz {latest} is here </p>
```
tynandebold, line 63: Oh missed this: I think you need a `!` here, per the design: 

image

Might be wise to extract this out to a variable, since it's also duplicated on line 19 in the same file.

AntonyMilneQB:
@studioswong I still don't seem to get the little pointy hand when I tried this out earlier. But I'm also not getting it in other places where I was expecting to see it (like when I hover over nodes in the pipeline flowchart). So I'm guessing this is just a problem with my setup (Gitpod + Chrome) rather than this PR? Curious if other people are seeing the same.

tynandebold:
> @studioswong I still don't seem to get the little pointy hand when I tried this out earlier. I'm guessing this is just a problem with my setup (Gitpod + Chrome) rather than this PR? Curious if other people are seeing the same.
> 
> Edit: just tried again and the pointy hand is back. So I guess this was just a random thing or I was looking at an old version or something. No need to worry about it.

@AntonyMilneQB I see the pointer cursor. I'm running the branch locally, not via Gitpod.

tynandebold:
Another note. I know we touched on this in Slack, but revisiting here: I think we might need that `Update` wording in the banner. I found [this animation video](https://mckinsey.ent.box.com/file/844809983103?s=1ecyle8qfpo5qvjiu6d7nwwyp17tnrdx) that Hamza made which shows `Update` to be the trigger for opening the panel.

studioswong:
> Another note. I know we touched on this in Slack, but revisiting here: I think we might need that `Update` wording in the banner. Mackay found [this animation video](https://mckinsey.ent.box.com/file/844809983103?s=1ecyle8qfpo5qvjiu6d7nwwyp17tnrdx) that Hamza made which shows `Update` to be the trigger for opening the panel.

yea I just saw that animation too - good to have that as reference. 

The wording of 'Update' for that button ( which prompts to expand the panel) is quite confusing ( I recall from a previous design handover session that he mentioned that button being the prompt to directly update the package) - it might be better to call it 'View More' / 'Expand' / 'Details'?

src/components/icons/settings.js
studioswong: I hardcoded this as there are no dynamic assignment of this color hence there isn't any need to adopt a variable here.

Setting the color variable entails setting up another css class as well as the need to set up another variable to pass within the `icon-button` component, which I am trying to avoid.

tynandebold:
> > Another note. I know we touched on this in Slack, but revisiting here: I think we might need that `Update` wording in the banner. Mackay found [this animation video](https://mckinsey.ent.box.com/file/844809983103?s=1ecyle8qfpo5qvjiu6d7nwwyp17tnrdx) that Hamza made which shows `Update` to be the trigger for opening the panel.
> 
> yea I just saw that animation too - good to have that as reference.
> 
> The wording of 'Update' for that button ( which prompts to expand the panel) is quite confusing ( I recall from a previous design handover session that he mentioned that button being the prompt to directly update the package) - it might be better to call it 'View More' / 'Expand' / 'Details'?

I don't remember being in that handover session, so for me that's why I'm fine with `Update`. When I see that word I just see it as a call to action, and within the context of the flow (see the banner -> click `Update` -> see the panel and update instructions) it does makes sense to me.

From what you suggested, I'm also fine with `View more`. The other two don't feel as strong as CTAs for me. Will leave to whatever you/Mackay think is best :)

tynandebold:
image

I don't see this in the settings modal. Are we leaving it out? It would be cool to have, and I think we can do it given the data we get from the endpoint.

studioswong:
yep I just saw this and had added this in the latest commit.

src/components/ui/command-copier/command-copier.scss
studioswong: I've just tried adding your suggested setting but unforunately that wouldn't work - the current text aligns perfectly with how the spacing had laid out for the other fields in the metadata panel as well. ( see below screenshots)

image

image

I had a check and the text do align with the `Copied to clipboard` text do align with the same baseline as the code text. I think what you meant here is that the copy command logo here not aligning with the top of the copy button, which I had adjusted to help with this alignment.

src/components/wrapper/wrapper.js
studioswong: Thanks for the suggestion, I understand your point with the confusion, but it is not technically feasible to be amended here given the data is wrapped exactly under the `data` field in the graphql response and I would need to directly reference that in the JS object destructing here. 

Moreover, the `data` field is not used directly in the component here but rather used to update the values of the internal hooks, so there isn't any need to reassign it to another variable as that variable won't be used / redundant.

studioswong: I believe you meant assigning the data to an alias - I've just done that in my latest commit

tynandebold: Yeah exactly, rename while destructuring.

src/components/settings-modal/settings-modal.js
```diff
@@ -62,6 +64,28 @@ const SettingsModal = ({
...
+              >
+                View release notes
+              </a>
+            </div>
+          ) : (
+            <div className="pipeline-settings-modal__already-latest">
+              <span>&#8226;</span>
+              <span>
+                {' '}
+                You are already on the latest version of Kedro-Viz (
```
tynandebold, line 84: I love this. Thanks for adding!

Don't think `already` needs to be in this line based on the design:

image

studioswong: indeed - I've updated the wording.

AntonyMilneQB (approved):
Just judging this without having seen any of the designs (I can't access them) so purely from a "does it make sense and work for me as a user" standpoint... Looks great, nice work! ⭐ 

I do still think there's value in saying what Kedro-Viz version the user currently has installed if they're not up to date, because this might influence their decision of whether to upgrade (am I way out of date or just a small patch release out?). Also it just seems natural to me to have the version you're running visible somewhere, presumably the settings page (like how most programs have an About page). But not a biggy.

Also should this mentioned in the release notes?

tynandebold (approved):
Great stuff, thanks for all the updates!

I still think this should be solved before merging into main:

https://user-images.githubusercontent.com/16869061/154555619-3a6bd259-ae3e-457e-8d4f-17d5564336fb.mov

I also agree with @AntonyMilneQB about having the current version somewhere in the panel. It gives good context. And lastly, I'm curious what you and Mackay come up with about the CTA text, too. I'd guess that having an element that's a specific CTA instead of only having `Kedro-Viz {latest} is here!` would further nudge users to click. Let us know what you two come up with.

src/components/ui/command-copier/command-copier.scss
studioswong: just saw your message - I see what you meant now. I've put in a fix 😄

PR #757, opened 2022-02-24
[KED-2962] Export run data as CSV

studioswong:
## Description

resolves [KED-2962](https://jira.quantumblack.com/browse/KED-2962).

This PR is to build out the 'export run' feature, which consolidates the selected run metadata and tracking data as shown on the experiment detail panel into a CSV file for download. 

image

## Development notes

I have made the following changes:

-  moved the fetching of the run metadata and the tracking data up in the experiment-tracking wrapper level to avoid redundancy in network requests. 
- set up a set of util functions to format the run metadata and run tracking data into an array for download
- set up the 'export csv' button to expose this function

Please note that, in case you are opening the csv file in excel, the alignment of the value would vary - please note that this is simply a csv export and is not optimized for excel. 

## QA notes

Please try downloading the csv export and check the download file. The name should be name as `rundata-(runID1)-(runID2).csv`, and the csv should contain all data that is displayed in the details view. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/757"></a>


RELEASE.md
```diff
@@ -8,6 +8,10 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+## Major features and improvements
+- Set up of a pop-up reminder to remind users to upgrade Kedro-Viz when a new version is released. (#746)
+- Set up the 'export run' button to allow exporting of selected run data into a csv file for download. (#746)
```
tynandebold, line 13:
```suggestion
- Set up a pop-up reminder to nudge users to upgrade Kedro-Viz when a new version is released. (#746)
- Set up the 'export run' button to allow exporting of selected run data into a CSV file for download. (#757)
```

src/components/experiment-tracking/details/details.js
```diff
@@ -22,21 +17,12 @@ const Details = ({
   showRunDetailsModal,
   sidebarVisible,
   theme,
+  runMetadata,
+  runTrackingData,
+  metadataError,
+  trackingDataError,
```
tynandebold, line 23: Maybe sort with the other props if not too much of a headache?

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
```diff
@@ -60,6 +70,14 @@ export const ExperimentPrimaryToolbar = ({
         visible={enableComparisonView}
         disabled={showChangesIconDisabled}
       />
+      <CSVLink data={exportData} filename={generateCsvFileName(runMetadata)}>
```
tynandebold, line 73: Pedantic suggestion: this function name could use all capitals for CSV to match the `CSVLink` component naming.

```suggestion
      <CSVLink data={exportData} filename={generateCSVFileName(runMetadata)}>
```

src/components/experiment-wrapper/experiment-wrapper.js
```diff
@@ -151,6 +170,8 @@ const ExperimentWrapper = ({ theme }) => {
             setSidebarVisible={setIsSidebarVisible}
             showRunDetailsModal={setShowRunDetailsModal}
             sidebarVisible={isSidebarVisible}
+            runMetadata={runMetadata}
+            runTrackingData={runTrackingData}
```
tynandebold, line 174: Sort, if possible.

src/components/experiment-wrapper/experiment-wrapper.js
```diff
@@ -24,6 +28,21 @@ const ExperimentWrapper = ({ theme }) => {
 
   const { subscribeToMore, data, loading } = useApolloQuery(GET_RUNS);
 
+  // fetch all metadata and tracking data from selected runs
```
tynandebold, line 31:
```suggestion
  // Fetch all metadata and tracking data from selected runs.
```

src/components/experiment-wrapper/experiment-wrapper.js
```diff
@@ -164,6 +185,10 @@ const ExperimentWrapper = ({ theme }) => {
               showRunDetailsModal={showRunDetailsModal}
               sidebarVisible={isSidebarVisible}
               theme={theme}
+              runMetadata={runMetadata}
+              runTrackingData={runTrackingData}
+              trackingDataError={trackingDataError}
+              metadataError={metadataError}
```
tynandebold, line 191: Sort here, too.

src/components/sidebar/sidebar.js
```diff
@@ -61,6 +63,8 @@ export const Sidebar = ({
               showChangesIconDisabled={!(selectedRunIds.length > 1)}
               showRunDetailsModal={showRunDetailsModal}
               sidebarVisible={sidebarVisible}
+              runMetadata={runMetadata}
+              runTrackingData={runTrackingData}
```
tynandebold, line 67: And here.

src/utils/experiment-tracking-utils.js
```diff
@@ -0,0 +1,70 @@
...
+      const dataKeyNames = Object.keys(data);
+      dataKeyNames.forEach((key) => {
+        let keyData = [key];
+        data[key].forEach((datafield) => keyData.push(datafield.value));
+        csvData.push(keyData);
+      });
+      csvData.push([]);
+    });
+  }
+  return csvData;
```
tynandebold, line 57: Could you space things out a bit here to improve readability?

```suggestion
    runTrackingData.forEach((trackingDataset) => {
      const { datasetName, data } = trackingDataset;
      const dataKeyNames = Object.keys(data);
      
      csvData.push([datasetName]);
      
      dataKeyNames.forEach((key) => {
        let keyData = [key];
        
        data[key].forEach((dataField) => keyData.push(dataField.value));
        csvData.push(keyData);
      });
      
      csvData.push([]);
    });
  }
  
  return csvData;
```

src/utils/experiment-tracking-utils.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+/**
+ * Take a the runMetadata list to generate a meaningful file name for csv export
+ * @param {array} runMetadata The set of runMetadata to be exported
+ * @returns A string to be used as the file name
+ */
+export const generateCsvFileName = (runMetadata) => {
```
tynandebold, line 65: This products some pretty crazy file names, such as this: 

```
rundata-2021-12-16T09.46.21.973Z-2021-12-16T09.47.19.553Z-2021-12-16T09.48.12.714Z.csv
```

I don't have a better suggestion for what it should be though. I just find the above really hard to parse. Have you asked Yetu/design if they have any guidance?

tynandebold (approved):
Looks great and works very well. I'm approving straight away.

I've left mostly style-related suggestions and have one thought: the `constructExportData` function is firing quite often. It would be great if it only triggered when the icon is actually clicked. Wonder if you've thought about trying to do that?

src/utils/experiment-tracking-utils.js
studioswong: yea indeed I do share the same thoughts with you about the file name - we haven't discussed it yet and there isn't any particular requirement for this, I thought it would be great to include the run id in the file name which, given the unique structure of our run id, leads to this 'very readable' file name 😆 

The other way is to simply name the file as `rundata.csv` without including the run ids - we can see if anyone complains about this current file name arrangement and revert if needed?

tynandebold: I like your `rundata.csv` idea (should it be `runData.csv` or even `run-data.csv`?). Maybe we aim for simplicity first and then if users want more specificity we iterate.

rashidakanchwala (approved):
nice implementation. thanks @studioswong. Agree changing the filename to a shorter generic name for now.

src/utils/experiment-tracking-utils.js
studioswong: sounds good, `run-data.csv` it is then 😄

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
```diff
@@ -26,6 +31,8 @@ export const ExperimentPrimaryToolbar = ({
     });
   };
 
+  const exportData = constructExportData(runMetadata, runTrackingData);
```
limdauto, line 34: This can be lazy as `CSVLink` accepts a function as well.

Also this can be a good candidate for `useCallback` (or if you want to stay with a static version of the data instead of a function, `useMemo`) so that the function is only re-created when dependencies change, and I'd suggest making runs selection as a dependency, as from the look of it, the requirement is only export data from selected run.

limdauto (commented):
Great stuff! Quick question: have we considered doing this server-side? I think for selected runs a client-side solution works. But a server-side solution might scale better with more requirements and when pagination is involved.

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
tynandebold: Agree, and related to my comment [here](https://github.com/kedro-org/kedro-viz/pull/757#pullrequestreview-892303855).

studioswong:
> Great stuff! Quick question: have we considered doing this server-side? I think for selected runs a client-side solution works. But a server-side solution might scale better with more requirements and when pagination is involved.

thanks for the comments both ! Sorry I just saw these set of comments after merging this in, wasn't my intention to merge this in before replying here 😆 

@limdauto I agree with you 100% on the take with resolving this on the BE - and indeed normally this would've been resolved by the BE. I opted to resolve this on the FE mainly because

a) The arrangement of the CSV data matches with the layout in the FE table, with all data avaliable on the FE, it makes sense ( and easier now) for this to be resolved on the FE ( hence minimising the send the request to BE, and wait for BE to construct the CSV). 
b) It's more work for now to wait for the set up a new endpoint for the Backend to return the data that is actually all avaliable on the FE

I totally hear you on the scaling front, though we can look into setting up a server side solution once we go down the path of scaling this CSV export feature. ( the current feature would work with pagination as well, given that by the time the user chooses to export this data, all the required data would have been already readily avaliable on the FE to be displayed on the experiment details panel.

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
studioswong: Thanks for the recommendations both! very valid points!

I had a look into using `useCallback` but unfortunately it doesn't work with the setup of `CSVLink` as it would generally take in the value of the object that was passed in on initialization. That said, they do have a `asyncOnClick` prop that seems to allow update of the export data, which would then allow the data object passed in to be updated via the `onClick` callback. 

Regardless, I can look into refactoring this and might put in a small follow up PR for this. Will tag you once it's out.

Issue #764, opened 2022-03-07
[BE] Clean up experiment tracking viz code after `session == run` decision


MerelTheisenQB:
## Description
If a user is in an interactive workflow (e.g. ipython or jupyter), uses the `SQLiteStore` to record a session and then spins up `kedro viz`, the graphql endpoint will fail to fetch data because of an assumption that the user is in a `CLI` workflow: 
https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/api/graphql.py#L81

## Solution
Instead of throwing an error we should make this a `.get()` call and just set `None` for the CLI command. In general it's better to have a `.get()` call to fetch elements from a python dict in case the item isn't there like we do here: https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/api/graphql.py#L77 

## Checklist
- [ ] Include labels so that we can categorise your issue

AntonyMilneQB:
From discussions with @MerelTheisenQB, this task should also include tidying up the various different `session_id`/`run_id`/`timestamp` fields. This follows the changes in https://github.com/kedro-org/kedro/issues/1273, which remove the concept of a `run_id` (it's the same as the `session_id`).

This includes the following, although there might be additional tasks too:

* always use the session store `runs.id` in preference to `runs.blob["session_id"]` (they will always contain the same thing, but let's be consistent), e.g. [here](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/api/graphql.py#L82)
* understand how `timestamp` is used - if only on the frontend as an immutable field then we can just remove this field altogether (it's not present in the database) and use `runs.id` instead
* remove the comment from [here](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/data_access/repositories/runs.py#L84) - there's not going to be a `timestamp` field; we'll just use `id` as it currently has
* be clear and consistent with terminology session vs. run. Possibly rename the `runs` table to `sessions` - but for consistency this would also mean renaming everything that is `run_xyz` to `session_xyz` instead; probably easiest just to leave the names as `run` throughout and document somewhere that a kedro viz run = kedro session, and don't use "session" in kedro viz at all
* check what happens if a user saves a session from `kedro jupyter` (by calling `session.close()`) _without doing a run_ - does this appear in experiment tracking [ideally it shouldn't]? Does it break stuff?

limdauto:
[...]

PR #781, opened 2022-03-08
Drop created_at column from session_store.db

AntonyMilneQB:
## Description

`demo-project/data/session_store.db` contained a column `created_at` which was not doing anything (it contained only null values, and it's not referred to anywhere in the codebase). I think it must be a leftover from when Lim was playing around and made the data in the first place.

I've removed the column from the table just to be tidy and so no one has to try and figure out what it's for again in the future.

## Development notes

Ran the following SQL (only supported in recent version of SQLite):
```
ALTER TABLE runs DROP COLUMN created_at;
```

## QA notes

All experiment tracking data is still there and shown on the frontend as before.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/781"></a>


MerelTheisenQB (approved):
👍

studioswong (approved):
Nice spot 👍 

On a side note, I assume this `created_at` field is a one-off thing / manually created field that's only found in the `session_store` db for `demo-project`, and not a field that's generated during the initial set up of the session store?

rashidakanchwala (approved):
Thanks Antony!

AntonyMilneQB:
> 

Yes indeed - I checked by deleting the session_store.db and doing a `kedro run` to make a fresh new one, and the column isn't there.

tynandebold (approved):

limdauto:
@AntonyMilneQB yes, thank you. It was indeed just used for the demo project session store. It's a by product of this spike: https://github.com/kedro-org/kedro-viz/pull/687

AntonyMilneQB:
@limdauto cool, thanks for the explanation and the extra information on schema migration. 

We shouldn't need to record the timestamp any more, since it has now been decided that `run_id = session_id = timestamp` - see https://github.com/kedro-org/kedro-viz/issues/764. Please do take a look and see if you think this is not the right thing to do actually, it would be great if you could just check if we're missing something!

limdauto:
@AntonyMilneQB oh that's great to know. Yea that should simplify things for sure.

Issue #738, opened 2022-02-10
Using Viz as a standalone React component doesn't work out of the box


tynandebold:
## Description
Using Kedro-Viz as a standalone React component doesn't work out of the box.

## Context
I'm trying to use Viz as a standalone component in my Create React App project. I'm getting some errors, and when those are remedied, I don't see the anything in the browser.

## Steps to Reproduce

1. Create a new project with Create React App: `npx create-react-app kedro-viz-standalone-component`
2. `cd` into the project and then install the the Viz dependency:  `npm install @quantumblack/kedro-viz --save`
3. Add a file with some dummy Kedro data ([data example](https://github.com/kedro-org/kedro-viz/blob/main/src/utils/data/spaceflights.mock.json))
4. Update `App.js` to include the Viz component and the data:
```jsx
import "./App.css";
import KedroViz from "@quantumblack/kedro-viz";
import { demo } from "./demo";

function App() {
  return (
    <div className="App">
      <KedroViz data={demo} />
    </div>
  );
}

export default App;
```
5. Run the app with `npm  start`.
6. Observe the compile errors and subsequently the blank webpage.

## Expected Result
The app should run without error and I should see Kedro Viz running locally.

## Actual Result
I received some dependency errors: 
![Screen Shot 2022-02-10 at 10 49 25](https://user-images.githubusercontent.com/16869061/153396547-65c2fd92-2c9f-4a53-a84e-1b5b4cace4af.png)

And when I fixed those and ran the app again, nothing showed up, thought the DOM elements exist:
![Screen Shot 2022-02-10 at 12 16 31](https://user-images.githubusercontent.com/16869061/153407146-1a0e9865-ee98-48c9-b31a-828dd1cb88ea.png)

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version: Google Chrome Version 98.0.4758.80 (Official Build) (x86_64)
* Operating system and version: MacOS 11.6.2 (20G314)
* NodeJS version used (if relevant): 14.17.0
* Kedro version used (if relevant): 4.3.1

## Checklist
- [x] Include labels so that we can categorise your issue

studioswong:
@rashidakanchwala just saw that you had picked up this issue - I had started looking into this alongside issue #758 as this needs to be fixed in order to ship #758 , and to avoid duplicate efforts I can assign this to myself, if that's ok with you?

studioswong:
[...]

PR #784, opened 2022-03-11
[Fix] Update faker package reference

studioswong:
## Description

Resolves  issue #738.

This is to fix the errors with resolving the `faker` dependency when Kedro-Viz is imported as an npm package. 

## Development notes

This error is caused by the packages being declared under `dev-dependencies` instead of `dependencies`, with which I had updated the packages in our `package.json` file. 

I had also updated the reference of the `faker` npm package to `@faker-js/faker` ([see here](https://www.npmjs.com/package/@faker-js/faker)). 

## QA notes

N/A

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/784"></a>


tynandebold:
What do you think about taking this one step further and completely removing these two dependencies and everything that comes into contact with them?

The reason we have them is for the GraphQL mocks, yet we don't use the mocks for any testing or anything else. And if we did, they don't accurately represent what experiment tracking looks like since there isn't any tracking data or metadata, as shown here when the mocks are toggled to `true`:

Screen Shot 2022-03-11 at 15 40 57

Lastly, now that we have the `demo-project` integrated into the repo, that's effectively mocking experiment tracking, and working quite well at that, so this seems like a great opportunity to trim down some unused aspects of the codebase.

studioswong:
yea I thought about that too, and I agree the mocks haven't been updated when we have the backend endpoints.

That said, I do still think there is value in keeping the mocks given its use as a development tool to help us quickly whip up some data without relying on the Backend or for quickly validating some data set up before investing in the backend - I can see this set up still being of use in the future. 

The situation with the demo project that you described here is a different scenario where it requires being spun up as a dev server, and can only serve the data that is returned by the existing backend setup, which is different from the set up of this mock, being completely independent as a FE tool with its intention to mock any potential new BE endpoint before its development. 

As long as this fix solves the error then we should keep this.

README.md
```diff
@@ -59,7 +59,17 @@ There are two ways you can use Kedro-Viz:
...
+
+    ```javascript
+    <div style={{ height: "100vh" }}>
+      <KedroViz
+        data={dataPipeline}
+      />
+    </div>
+    ```
+  
+  Please ensure that the relevant pipeline data is passed as a prop into the component. We also recommend wrapping the `Kedro-Viz` component with a parent HTML/JSX element that has a specified height (as seen in the above example) in order for Kedro-Viz to be styled properly.
```
tynandebold, line 72: We already have a section in the readme for the [standalone component usage](https://github.com/kedro-org/kedro-viz#standalone-react-component-usage). 

You can probably update the second line in the example already in that section to this:

```jsx
const MyApp = () => (
  <div style={{ height: "100vh" }}>
    <KedroViz data={json} />
  </div>
);
```

And then add the note:

> We recommend wrapping the `Kedro-Viz` component with a parent element that has a specified height in order for Kedro-Viz to be styled properly.

tynandebold (approved):
Looks good!

Just one thing about updating the readme: we already have a section for standalone component usage so it's probably better to just update the content there.

I'll approve straight away though, as that's my only comment.

README.md
studioswong: good spot - completely overlooked the section below.

I've updated the section and wording accordingly.

Issue #732, opened 2022-02-08
[Spike] Investigate the migration away from redux state logic for the Kedro-Viz flowchart


studioswong:
## Introduction

The successful setup and transition towards data-fetching with Graphql and state management with react hooks within the experiment-tracking features resulted in a new set of clean and highly readable code that are highly reusable ( compared to the redux setup that is highly entangled with the redux store.)

With third party consumption of the Kedro-Viz flowchart being one of the most discussed use case for Kedro-Viz, this poses the benefits of refactoring the existing redux setup for the flowchart towards local state management with graphql for data fetching. Yet, the current logic of the graph calculations are tightly entangled within the selectors within the redux setup which poses a series of challenges for the refactoring work that requires careful investigation.

This issue is to investigate and explore the different stages for conducting this work in migrating away from the redux setup.

## Background

[Redux](https://redux.js.org/) has been utilized heavily in the state management and data ingestion of the flowchart (see [this section](https://github.com/kedro-org/kedro-viz/blob/main/ARCHITECTURE.md) in the architecture docs) . All local state of the app, as well as the logic of the calculation of flowchart nodes, modular pipeline tree and input to the layout engine, are all managed and tightly coupled within the selectors in the redux setup. 

## Why isn't the redux setup suitable anymore?

The redux setup poses the following problems:

1. This tightly coupled setup of in-app logic within selectors and the redux state **makes it impossible to extract the flowchart, node-list sidebar and all related components for reusability as individual components outside the context of Kedro-Viz**, as a local store will always need to be included with the component. 

2. The redux setup introduces an excessive amount of code and complexity into the codebase - the set of code to initialisation the set up of the local store,  the need to introduce a series of actions, reducers, and selectors, as well as the additional amount of code within the components to consume the local store, all introduces the unnecessary effort to maintain excessive amount of code down the line. ( This is very apparent compared to the amount of set up code with graphql in the new experiment tracking features, which only requires the simple apollo client setup.)

3. It is never good to have two drastically different data ingestion and state management protocols within the same codebase. For sake of simplicity and maintenance down the line, we will have to migrate away from this one way or another.

## Design

Before designing a solution, here are the set of challenges specific to the current state of the app that we need to consider in the design: 

### **Data Ingestion**
[...]

PR #787, opened 2022-03-21
Add display props to standalone React component + expandAllPipelines flag

### LINKED ISSUE #758: [FE] Add display props to standalone React component
A user would like to be able to completely hide some of the page elements that Viz renders. Right now there is the `visible` prop, but that only toggles the those specified elements to their open or closed positions. It doesn't hide them completely.

A prop, or props, should be added to hide the following elements:
- The global toolbar
- The sidebar
- The minimap

The implementation could look something like this:

```jsx
<KedroViz
  data={json}
  display={{
    globalToolbar: false,
    miniMap: false,
    sidebar: false,
  }}
/>
```

#### Further considerations

- When the global toolbar and sidebar are removed, some CSS will have to be adjusted to account for those now-empty spaces.
- The user also mentioned that we should ensure there aren't any polluting styles on high-level and generic HTML elements (e.g. `html`, `body`, `section`, etc.). Since the merge of #731, we don't have any styles like that.
- This should be wrapped into the next release of Viz and will be under the `Major features and improvements` section of our [release notes](https://github.com/kedro-org/kedro-viz/blob/main/RELEASE.md). 

#### Notes

We should probably also fix #738 in tandem with this, since it could impact the usage of Viz when used as a standalone component.

studioswong:
## Description

resolves #758 

This PR adds a set of input props to define the display of the global Toolbar, sidebar and minimap to allow the display of a simplified version of Kedro-Viz to be consumed as a third party import, as shown below. 

image (12)

Given that the sidebar display can be omitted, I had added in a special `expandAllPipelines` prop that, on being specified as `true`, will allow all modular pipelines to default to the expanded view on initial load. I have also enabled this feature behind a feature flag to enable power users to access this feature if needed. 

## Development notes

1) I have added in new fields within the redux store to enable the control of the display props to control the behaviour of the app.
2) within the app, I have refactored the behaviour of various components ( sidebar, global toolbar, etc), to display according to the display states. This also includes changes to the css setup to cater for the display adjustments. 
3) I have altered the ingestion engine to allow ingesting the set of newly defined `display` props for the react-component import use case. 
4) I have added a new flag to allow the `expandAllPipelines` feature to also be accessible for power users.
5) I have altered and updated the related tests for this new set of props and fields in the redux store. 

## QA notes

The best way to test the behaviour of the display props is to manually changes the default values of the display props within `initial-state.js` to test the different setup. 

For the `expandAllPipelines` feature, this can be tested via setting the respective flag to `true`.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/787"></a>

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -36,6 +36,12 @@ export const flags = {
     default: true,
     icon: '🐳',
   },
+  expandAllPipelines: {
+    name: 'Expand all modular pipelines',
+    description: 'Expand all modular pipelines on first load',
+    default: false,
+    icon: '🔛',
+  },
 };
 
 export const settings = {
--- a/src/store/initial-state.js
+++ b/src/store/initial-state.js
@@ -36,10 +36,10 @@ export const createInitialState = () => ({
     modularPipelineFocusMode: null,
   },
   display: {
-    globalToolbar: false,
-    sidebar: false,
+    globalToolbar: true,
+    sidebar: true,
     miniMap: true,
-    expandAllPipelines: true,
+    expandAllPipelines: false,
   },
   zoom: {},
 });
@@ -78,6 +78,7 @@ export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
       state.pipeline.active = state.pipeline.main;
     }
   }
+
   // Cater for expandAllPipelines in component props or within flag
   // expandAllPipelines needs to happen after the deepmerge of localStorage to overwrite
   // any saved visible state of the nodes
@@ -94,6 +95,11 @@ export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
       if (!modularPipelinesIds.includes(nodeId)) {
         newModularPipelineState.visible[nodeId] = true;
       }
+    });
+
+    console.log('entered here', {
+      ...state.modularPipeline,
+      ...newModularPipelineState,
     });
     return {
       ...state,
@@ -118,8 +124,8 @@ export const prepareNonPipelineState = (props) => {
     newVisibleProps['sidebar'] = false;
   }
 
-  if (props.display?.minimap === false) {
-    newVisibleProps['minimap'] = false;
+  if (props.display?.minimap === false || state.display.miniMap === false) {
+    newVisibleProps['miniMap'] = false;
   }
 
   return {
@@ -139,6 +145,8 @@ export const prepareNonPipelineState = (props) => {
  * @return {object} Initial state
  */
 const getInitialState = (props = {}) => {
+  console.log('props', props);
+
   const nonPipelineState = prepareNonPipelineState(props);
   saveState({
     nodeType: {
@@ -147,7 +155,10 @@ const getInitialState = (props = {}) => {
     },
   });
 
-  const expandAllPipelines = nonPipelineState.display.expandAllPipelines;
+  const expandAllPipelines =
+    nonPipelineState.display.expandAllPipelines ||
+    nonPipelineState.flags.expandAllPipelines;
+
   const pipelineState = preparePipelineState(
     props.data,
     props.data !== 'json',
--- a/src/store/initial-state.js
+++ b/src/store/initial-state.js
@@ -36,10 +36,10 @@ export const createInitialState = () => ({
     modularPipelineFocusMode: null,
   },
   display: {
-    globalToolBar: false,
-    sidebar: true,
+    globalToolbar: false,
+    sidebar: false,
     miniMap: true,
-    expandAllPipelines: false,
+    expandAllPipelines: true,
   },
   zoom: {},
 });
@@ -70,13 +70,35 @@ export const mergeLocalStorage = (state) => {
  * @param {object} data Data prop passed to App component
  * @param {boolean} applyFixes Whether to override initialState
  */
-export const preparePipelineState = (data, applyFixes) => {
+export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
   const state = mergeLocalStorage(normalizeData(data));
   if (applyFixes) {
     // Use main pipeline if active pipeline from localStorage isn't recognised
     if (!state.pipeline.ids.includes(state.pipeline.active)) {
       state.pipeline.active = state.pipeline.main;
     }
+  }
+  // Cater for expandAllPipelines in component props or within flag
+  // expandAllPipelines needs to happen after the deepmerge of localStorage to overwrite
+  // any saved visible state of the nodes
+  if (expandAllPipelines) {
+    const modularPipelinesIds = state.modularPipeline.ids;
+    const nodeIds = state.node.ids;
+
+    // filter out the nodeIds that is the same id as the modualrPipelines and assign everything else to true
+    const newModularPipelineState = {
+      visible: {},
+      expanded: modularPipelinesIds,
+    };
+    nodeIds.forEach((nodeId) => {
+      if (!modularPipelinesIds.includes(nodeId)) {
+        newModularPipelineState.visible[nodeId] = true;
+      }
+    });
+    return {
+      ...state,
+      modularPipeline: { ...state.modularPipeline, ...newModularPipelineState },
+    };
   }
   return state;
 };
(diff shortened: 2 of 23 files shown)


datajoely:
🥳

src/config.js
```diff
@@ -36,6 +36,12 @@ export const flags = {
     default: true,
     icon: '🐳',
   },
+  expandAllPipelines: {
+    name: 'Expand all modular pipelines',
+    description: 'Expand all modular pipelines on first load',
+    default: false,
+    icon: '🔛',
+  },
```
tynandebold, line 44: I'm curious what the logic is for putting this here, in the `Experiments` section, versus displaying it in the `General` section of the setting panel?

src/store/initial-state.js
```diff
@@ -108,7 +152,16 @@ const getInitialState = (props = {}) => {
...
+
+  const expandAllPipelines =
+    nonPipelineState.display.expandAllPipelines ||
+    nonPipelineState.flags.expandAllPipelines;
+
+  const pipelineState = preparePipelineState(
+    props.data,
+    props.data !== 'json',
+    expandAllPipelines
+  );
```
tynandebold, line 164: Possible to add a blank line before `return` statements? It's easier to read them I think.

tynandebold (changes requested):
This is excellent, great work :)

I've left some lower-level comments, another about a leftover `console.log`, and a couple other considerations. Nothing major at all (there's some stuff to sort, but I didn't comment on that, since you can do it at your discretion).

I don't see any new tests though. Nothing along the lines of `it('renders without the global toolbar')` or the like. Should we add some tests checking that these elements aren't rendered or have the correct hidden class names?

One larger though here at the end: did you consider shying away from using Redux to store some of this state? I guess for `expandAllPipelines` Redux is ok, since it's also a feature flag, though for `globalToolbar`, `miniMap`, and `sidebar`, it's possible to pass props through the component tree and could be done that way. If we are eyeing moving away from Redux in the future, as you [wrote here](https://github.com/kedro-org/kedro-viz/issues/732), perhaps we aim to avoid adding things to it where possible?

@rashidakanchwala I found a good way to test without having to modify the initial state. Add this in `./container/container.js` and change the values accordingly:

```jsx
<App
  data={getPipelineData()}
  display={{
    globalToolbar: false,
    miniMap: false,
    sidebar: false,
    expandAllPipelines: false,
  }}
/>
```

studioswong:
> One larger though here at the end: did you consider shying away from using Redux to store some of this state? I guess for `expandAllPipelines` Redux is ok, since it's also a feature flag, though for `globalToolbar`, `miniMap`, and `sidebar`, it's possible to pass props through the component tree and could be done that way. If we are eyeing moving away from Redux in the future, as you [wrote here](https://github.com/kedro-org/kedro-viz/issues/732), perhaps we aim to avoid adding things to it where possible?

Thanks for the review! 

Re: Shying away from redux, I completely understand where you are coming from, and is a huge opponent of moving away from it. 

For `globalToolbar`, `miniMap` and `sidebar`, one thing to note is that we still need to update the `visibility` state  for the respective items in the redux store in order for them to align properly, which would be difficult ( and on top of it, very confusing) to do with it bypassing the data ingestion setup and simply by passing this down the component tree. 

Moreover, I would avoid introducing a new way of data congestion for the react props in this PR prematurely without migrating the other props as well as it would simply be confusing down the line - it's better to align the data ingestion of react props as it is now and introduce the refactoring by stages when we as a team is all aligned. 

On another note, I will look into setting up more tests for the display props.

src/store/initial-state.js
```diff
@@ -35,6 +35,12 @@ export const createInitialState = () => ({
     miniMap: true,
     modularPipelineFocusMode: null,
   },
+  display: {
+    globalToolbar: true,
+    sidebar: true,
+    miniMap: true,
+    expandAllPipelines: false,
```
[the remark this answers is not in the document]
studioswong: Hhmm I see what you mean - the main reasoning behind setting up the `expandAllPipelines` field here is to separate / differentiate between the 2 different use case of the `expandAllPipelines` requirement being a react import prop vs bring specified via a flag, with this particular prop aligning to the first use case. 

I've set this state under the `display` object to align with the need to specify the expansion of all modular pipelines in line with the `display.sidebar` prop via the react component import. ( The naming of this prop being `display.expandAllPipelines` rather than `display.modularPipelines` should explain the intent behind this prop in controlling the display intent with expanded pipelines rather than controlling the display of modular-pipelines)

src/store/initial-state.js
```diff
@@ -64,14 +70,42 @@ export const mergeLocalStorage = (state) => {
...
   if (applyFixes) {
     // Use main pipeline if active pipeline from localStorage isn't recognised
     if (!state.pipeline.ids.includes(state.pipeline.active)) {
       state.pipeline.active = state.pipeline.main;
     }
   }
+
+  // Cater for expandAllPipelines in component props or within flag
+  // expandAllPipelines needs to happen after the deepmerge of localStorage to overwrite
+  // any saved visible state of the nodes
```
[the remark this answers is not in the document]
studioswong: This multi-line comment format is mainly reserved for description of functions - I've cleared up this comment though so it will now fit nicely in one line.

src/config.js
studioswong: The `Experiments` section is where we put all our feature flags ( such as `sizeWarning`, and in this case, `expandAllPipelines`), while `prettyNaming` is not hidden behind a feature flag but rather as a general setting.

tynandebold: Right, I see. From the readme: _"Kedro-Viz uses features flags to roll out some experimental features."_

Let me reword my question: is this `expandAllPipelines` behavior an experimental feature?

src/store/initial-state.js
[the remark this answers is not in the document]
tynandebold: I see. So when using Viz as a standalone component, you can do this as you've written it:

```jsx
<KedroViz data={json}  display={{expandAllPipelines: true}} />
```

I'd suggest we don't need to completely align `expandAllPipelines` with how we'd hide the sidebar or mini map. Though related, they act like distinct modifications. So instead of grouping it in with the `display` prop, it could be a first-class, top-level one. In practice, we could do this and it would expand all pipelines:

```jsx
<KedroViz data={json} expandAllPipelines />
```

What do you think?

src/store/initial-state.js
```diff
@@ -64,14 +70,36 @@ export const mergeLocalStorage = (state) => {
...
   const state = mergeLocalStorage(normalizeData(data));
   if (applyFixes) {
     // Use main pipeline if active pipeline from localStorage isn't recognised
     if (!state.pipeline.ids.includes(state.pipeline.active)) {
       state.pipeline.active = state.pipeline.main;
     }
   }
+
+  // Cater for expandAllPipelines in component props or within flag
+  if (expandAllPipelines) {
```
limdauto, line 83: Sorry for butting in but I'd suggest moving this logic inside `normalizeData`. We already have some logic to specifically only mark children of the root node to be visible [here](https://github.com/kedro-org/kedro-viz/blob/bcaa303647febd59ea8c652160d5d52d05e21545/src/store/normalize-data.js#L215-L221), so I'd extend it to mark _all_ nodes to be visible based on a parameter.

studioswong: heya good shout - indeed I was debating whether I should do it there in `normalizeData`, thanks for suggesting and reconfirming this.

src/config.js
studioswong: yep it is 😄  experimental features are features where we don't officially roll out to all users but are accessible to power users for testing. For example, when we did the initial version of modular pipelines, we hid it behind a feature flag to allow it to still be accessible for users while we continue polishing the feature. 

In this case, the 'expandAllPipelines' feature is an experimental feature that we can make accessible to some users given their mention of this in an issue, where we can make this accessible officially as a feature when we have given it some design love in the future.

tynandebold (approved):
Looks great! Thanks for adding those tests, too :)

rashidakanchwala (approved):
Look's great @studioswong. thanks :)

PR #800, opened 2022-03-29
[KED-764] Clean up experiment tracking code

### LINKED ISSUE #764: [BE] Clean up experiment tracking viz code after `session == run` decision

rashidakanchwala:
## Description

Resolves #764 

## Development notes

- session_id is same as run_id. So I have replaced all session_id to run_id 
- timestamp is same as run_id. So I have removed timestamps and changed the FE reference to run_id 
- use get() to read cli commands so that it doesn't break in kedro jupyter notebook 
- updated ReadMe to highlight that Kedro runs means Kedro sessions. 

Please note - I have run a formatter on the README so it has some additional formatting changes. 

## QA notes

@AntonyMilneQB  - I tested this on jupyter notebook - session.run() on the notebook is now recorded as a run in viz (one thing I did find weird is that as soon as I open the notebook, a session run is initiated automatically but we can sync on that offline). Also session.close() doesn't break anything. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/799"></a>

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/800"></a>


README.md
```diff
@@ -36,29 +36,29 @@ Kedro-Viz is an interactive development tool for building data science pipelines
 - 🔬 Focus mode for modular pipeline visualisation
 - 📊 Rich metadata side panel to display parameters, plots, etc.
 - ♻️ Autoreload on code change
-- 🧪 Supports tracking and comparing runs in a Kedro project
+- 🧪 Supports tracking and comparing runs (a.k.a sessions) in a Kedro project
```
tynandebold, line 39: In @AntonyMilneQB's [comment here](https://github.com/kedro-org/kedro-viz/issues/764#issuecomment-1061842270), he says, "...and don't use "session" in kedro viz at all."

Based on that, should we even include this `(a.k.a sessions)` language? I don't want it to be confusing, and I fear it might be.

tynandebold (approved):
This looks good and works well for me :) Everything continues to work on the frontend as I expect it to and it doesn't look like there are any `timestamp` references where there shouldn't be. There is some `timestamp` naming in a few CSS class names, but that's because we're displaying timestamps in those instances, so I'm happy with that.

I left only one comment about about using the language of `(a.k.a sessions)` in the readme.

Would be great if @AntonyMilneQB or even @MerelTheisenQB could double check the Python parts, as I'm not as confident there.

README.md
rashidakanchwala: ah my intention was to clarify the confusion. This is based on the assumption that people using kedro are more familiar with the term kedro sessions and hence when they read 'runs' they might not relate it being the same as sessions. 
but maybe I need to re-word it better. @AntonyMilneQB ?

studioswong (approved):
Tested this both on the dev server and FE and looking good - approving straight away :) 

Some minor comments below

README.md
```diff
@@ -25,7 +25,7 @@ Live Demo: <a href="https://demo.kedro.org/" target="_blank">https://demo.kedro.
 
 ## Introduction
 
-Kedro-Viz is an interactive development tool for building data science pipelines with [Kedro](https://github.com/kedro-org/kedro). Kedro-Viz also allows users to view and compare different runs in the Kedro project. 
+Kedro-Viz is an interactive development tool for building data science pipelines with [Kedro](https://github.com/kedro-org/kedro). Kedro-Viz also allows users to view and compare different runs (a.k.a sessions) in the Kedro project.
```
AntonyMilneQB, line 28:
```suggestion
Kedro-Viz is an interactive development tool for building data science pipelines with [Kedro](https://github.com/kedro-org/kedro). Kedro-Viz also allows users to view and compare different runs in the Kedro project.
```

AntonyMilneQB: I don't think this is important enough to bother mentioning in the main text on kedro-viz.

README.md
AntonyMilneQB: Understand the good intention of clarifying things here but I agree with @tynandebold that is potentially more confusing. The "session only contains one run" clarification is more of a note for developers rather than something to explain to users. So let's remove it from this list completely and just put it in the contributing guide instead - I'll do a commit to add it.

README.md
```diff
@@ -158,8 +156,8 @@ We also recommend wrapping the `Kedro-Viz` component with a parent HTML/JSX elem
 
 Kedro-Viz uses features flags to roll out some experimental features. The following flags are currently in use:
 
-| Flag | Description |
-|------| ------------|
+| Flag        | Description                                                                             |
```
AntonyMilneQB, line 159: Did your formatting tool do this automatically? 😮  Please share your magical method.

README.md
```diff
@@ -116,12 +116,12 @@ SESSION_STORE_CLASS = SQLiteStore
 SESSION_STORE_ARGS = {"path": str(Path(__file__).parents[2] / "data")}
 ```
 
-Once the above set-up is complete, tracking datasets can be used to track relevant data for Kedro runs. More information on how to use tracking datasets can be found [here](https://kedro.readthedocs.io/en/stable/08_logging/02_experiment_tracking.html)
+Once the above set-up is complete, tracking datasets can be used to track relevant data for Kedro runs (a.k.a sessions). More information on how to use tracking datasets can be found [here](https://kedro.readthedocs.io/en/stable/08_logging/02_experiment_tracking.html)
```
AntonyMilneQB, line 119:
```suggestion
Once the above set-up is complete, tracking datasets can be used to track relevant data for Kedro runs. More information on how to use tracking datasets can be found [here](https://kedro.readthedocs.io/en/stable/08_logging/02_experiment_tracking.html)
```

AntonyMilneQB (approved):
Great stuff, thanks for doing this!

What was the change to `session_store.db`? I wasn't expecting anything to change there 🤔 (let me know if you're not sure how to figure this out and we can try together)

Just a couple of other places where I think changes should still be made:
* Remove https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/data_access/repositories/runs.py#L84
* Rename `timestamp` to `run_id` in `load_versioned_tracking_data` and corresponding tests in `test_graph_nodes`

README.md
rashidakanchwala: oh, but it's vs code magic :)

README.md
AntonyMilneQB: Done in [40f6312](https://github.com/kedro-org/kedro-viz/commit/40f6312a75cb110fe58a4d601827ecb69c1ad694).

README.md
AntonyMilneQB: 😮 🪄 🔮

Issue #805, opened 2022-03-31
Tidy e2e test Python environment


antonymilne:
There's various things that are [pip-installed in environment.py](https://github.com/kedro-org/kedro-viz/blob/main/package/features/environment.py#L83-L87). Ideally we shouldn't be putting custom requirements in here. I tried removing them in #804 but it started failing some tests - looks like we need the click specification for testing on older version of kedro 0.16.

Given that the kedro requirements are going to continually keep bumping the version of click we should work out some better way of installing requirements here so that the testing environment is a clean as possible. Or maybe we should just forget about tests for 0.16 altogether?

There's probably some other bits of tidying that would be useful here, e.g. setting `PIP_DISABLE_PIP_VERSION_CHECK` would be much nicer if done [in the CI config, as we do on kedro](https://github.com/kedro-org/kedro/blob/main/.circleci/continue_config.yml#L201-L202).

tynandebold:
We can now let testing of version 16 fall away and only test back one minor version, so 0.17.

AntonyMilneQB:
Even better, after doing #812 we should have e2e and unit tests split into separate jobs. Then it _might_ be possible to completely remove https://github.com/kedro-org/kedro-viz/blob/main/package/features/environment.py and not use any venv at all.

studioswong:
@AntonyMilneQB would it make a difference if we are supporting >= 0.16.6 instead of >=0.16.1 ( as I saw from the tests in `viz.feature`?) The main reason being that we have still been supporting >=0.16.6 since 3.8.0 and was thinking it would be good to at last support the slightly older versions of Kedro so that they could still get any bug fixes we put for the flowchart features. ( there might still be a big group of users on older CSTs that might not have made the move to upgrade to 0.17 yet)

If we are dropping support for 0.16 altogether, this would be worth a discussion with the wider team as that would trigger some other actions, such as notifying the user groups of the drop our support for 0.16 altogether, etc.

AntonyMilneQB:
@studioswong Very fair point. As far as I can tell...
* kedro-viz has specified `kedro>=0.16.0` since kedro-viz 3.8.0 (here's the commit: https://github.com/kedro-org/kedro-viz/commit/add5c717deed6a1d04abff5aad12c0cffe5067a4)
* we have special logic to handle kedro < 0.17.0 but it's the same for all 0.16.x releases, so nothing special about kedro 0.16.6 compared to 0.16.0 or any other 0.16.x release
[...]

PR #815, opened 2022-04-06
[DRAFT] [KED-793] Extend support to Python 3.9+

rashidakanchwala:
## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -260,6 +260,18 @@ jobs:
       CONDA_ENV_PY_VERSION: '3.8'
     steps: [build]
 
+  build_39:
+    executor: circleci_machine
+    environment:
+      CONDA_ENV_PY_VERSION: '3.9'
+    steps: [build]
+  
+  build_310:
+    executor: circleci_machine
+    environment:
+      CONDA_ENV_PY_VERSION: '3.10'
+    steps: [build]
+
   deploy_demo:
     executor: demo_executor
     steps: [deploy_demo]
@@ -277,6 +289,22 @@ jobs:
     working_directory: ~/repo
     environment:
       CONDA_ENV_PY_VERSION: '3.8'
+      NODE_OPTIONS: '--max-old-space-size=8192'
+    steps: [win_build]
+
+  win_build_39:
+    executor: win/default
+    working_directory: ~/repo
+    environment:
+      CONDA_ENV_PY_VERSION: '3.9'
+      NODE_OPTIONS: '--max-old-space-size=8192'
+    steps: [win_build]
+
+  win_build_310:
+    executor: win/default
+    working_directory: ~/repo
+    environment:
+      CONDA_ENV_PY_VERSION: '3.10'
       NODE_OPTIONS: '--max-old-space-size=8192'
     steps: [win_build]
 
@@ -332,10 +360,14 @@ workflows:
   version: 2
   build:
     jobs:
-      # - build_37
+      - build_37
       - build_38
-      # - win_build_37
+      - build_39
+      - build_310
+      - win_build_37
       - win_build_38
+      - win_build_39
+      - win-build_310
       - deploy_demo:
           requires:
             # - build_37
@@ -354,10 +386,14 @@ workflows:
               only:
                 - main
     jobs:
-      # - build_37
+      - build_37
       - build_38
-      # - win_build_37
+      - build_39
+      - build_310
+      - win_build_37
       - win_build_38
+      - win_build_39
+      - win_build_310
 
   release:
     jobs:
--- a/package/setup.py
+++ b/package/setup.py
@@ -41,7 +41,7 @@ setup(
     long_description_content_type="text/markdown",
     license="Apache Software License (Apache 2.0)",
     url="https://github.com/kedro-org/kedro-viz",
-    python_requires=">=3.7, <3.9",
+    python_requires=">=3.7, <=3.10",
     install_requires=requires,
     tests_require=test_requires,
     keywords="pipelines, machine learning, data pipelines, data science, data engineering, visualisation",


package/test_requirements.txt
```diff
@@ -3,7 +3,7 @@ kedro[pandas,spark]>=0.16.0
 kedro-telemetry>=0.1.1  # for testing telemetry integration
 bandit~=1.6.2
 behave>=1.2.6, <2.0
-black==21.5b1
+black==22.3.0
```
AntonyMilneQB, line 6:
```suggestion
black~=22.0
```
to match kedro

package/features/environment.py
```diff
@@ -84,7 +84,7 @@ def _setup_context_with_venv(context, venv_dir):
             "wheel",
             "botocore",
             "PyYAML>=4.2, <6.0",
-            "click<8.0",
+            "click>=8.1.0",
```
AntonyMilneQB, line 87:
```suggestion
            "click<9.0",
```
to match kedro

package/setup.py
```diff
@@ -41,7 +41,7 @@
     long_description_content_type="text/markdown",
     license="Apache Software License (Apache 2.0)",
     url="https://github.com/kedro-org/kedro-viz",
-    python_requires=">=3.7, <3.9",
+    python_requires=">=3.7, <=3.10",
```
AntonyMilneQB, line 44:
```suggestion
    python_requires=">=3.7, <3.9,
```

package/features/viz.feature
```diff
@@ -5,13 +5,6 @@ Feature: Viz plugin in new project
     Background:
         Given I have prepared a config file with example code
 
-    Scenario: Execute viz with Kedro 0.16.1
-        Given I have installed kedro version "0.16.1"
-        And I have run a non-interactive kedro new
-        And I have installed the project's requirements
-        When I execute the kedro viz command "viz"
-        Then kedro-viz should start successfully
-
```
studioswong, line 14: While I notice that we did mentioned we are not supporting Kedro 0.16 in the description, that said I wonder what is the reason behind this - is this due to any technical decision ( i.e failed tests, etc)? 

from last I recall [we are still supporting >= Kedro 0.16.6 ](https://github.com/kedro-org/kedro-viz/blob/main/CONTRIBUTING.md#launch-a-development-server-with-a-real-kedro-project), my only worry is that, if we are stopping support for 0.16, then those users on a slightly older version will not get any bug fixes and updates to the flowchart features. 

Not that I am against dropping support with 0.16 at all, perhaps this might be worth a wider discussion with the team so that we are decide and make any needed updates to our docs / announcement to user groups, etc?

package/tests/test_api/test_apps.py
```diff
@@ -284,23 +284,23 @@ def assert_example_transcoded_data(response_data):
     # compare nodes
     expected_nodes = [
         {
-            "id": "f2b74919",
+            "id": "2a1abe98",
             "name": "Process Data",
             "full_name": "process_data",
             "tags": ["split"],
-            "pipelines": ["__default__", "data_processing"],
-            "modular_pipelines": [],
```
studioswong, line 292: just out of curiosity, I wonder the need behind deleting the `modular_pipelines` field in the test? ( this doesn't seem to relate to the task on hand, hence my curiousity)

.circleci/config.yml
```diff
@@ -219,22 +220,26 @@ commands:
...
+      python_version: 
+        type: string
     steps:
       - checkout
-      - restore_cache:
-          keys:
-            - v0-{{ arch }}-dependencies-{{ checksum "package.json" }}-{{ checksum "package/requirements.txt" }}-{{ checksum "package/test_requirements.txt" }}
-            # fallback to using the latest cache if no exact match is found
-            - v0-{{ arch }}-dependencies-
-      - win_setup_conda
```
studioswong, line 229: just out of curiosity, I wonder what was the reason behind the these steps failing?

.circleci/config.yml
```diff
@@ -4,9 +4,21 @@ orbs:
   win: circleci/windows@2.4.0
 
 executors:
-  circleci_machine:
+  circleci_machine_37:
     docker:
-      - image: cimg/node:14.15
```
studioswong, line 9: any reason why we are taking away the image for `node:14.15`?

studioswong (commented):
Awesome job setting this up - great job for pushing through with this. 

I've left some comments mainly out of my curiosity behind some of the decisions made in this PR. 

In terms of the JS tests still passing on node 16, it is indeed weird that it's not returning any errors for the `build:css` step - for this are you referring to the tests passing after deleting the `node:14.15` docker image on the circleCI config file?

package/features/viz.feature
rashidakanchwala: yes, this was discussed in the backlog grooming session last week. Update on ticket #805

package/tests/test_api/test_apps.py
rashidakanchwala: so windows builds were failing on 3.9, 3.10 due to some python dependency issues. this was because in kedro-viz we import multiple kedro datasets i.e. pandas, and spark which we use for python tests. 
in order for us to have less python dependencies,i removed the spark datasets in the python tests and replaced it with pandas dataset. Hence i had to edit the python tests accordingly.

.circleci/config.yml
rashidakanchwala: it's to do with some caching on python venv. it is a bit confusing to me too :)

.circleci/config.yml
rashidakanchwala: so for linux, we are no longer using conda or python virtual environments. we are just using docker -- and on docker we can have only one image. hence we have a python image with latest node. so if you see my dev notes, currently on circleCI we are running node 16 but surprisingly all the javascript tests are running properly and passing (when they should be failing) 

Anyhow this is a mystery! but for now since the test are actually being processed and running as expected instead of being skipped we have left it as is.

.circleci/config.yml
AntonyMilneQB: Basically the caches currently aren't specific to the version of Python. e.g. when we try running the 3.10 job, the cache for 3.8 might be loaded up. The best way to ensure we have a clean testing environment is just to not cache anything at all so there's no interference. We can then _maybe_ add it back in in the future CI refactor ticket. If we do so then we should include the python version in the cache key so there's a different cache for each Python version though.

The reason I say _maybe_ is that saving and restoring cache on CircleCI is actually pretty slow, so sometimes there's not actually any time saving to be made compared to not caching and instead just making a fresh environment every time.

.circleci/config.yml
```diff
@@ -336,15 +373,23 @@ workflows:
...
+      - win_build_310:
+           python_version: '3.10'
       - deploy_demo:
           requires:
-            # - build_37
+            - build_37
             - build_38
-            # - win_build_37
+            - win_build_37
             - win_build_38
```
AntonyMilneQB, line 393: Can we add all the builds to `requires` please?

package/test_requirements.txt
```diff
@@ -1,13 +1,15 @@
 -r requirements.txt
-kedro[pandas,spark]>=0.16.0  
+kedro>=0.17.0
```
AntonyMilneQB, line 2:
```suggestion
kedro[pandas]>=0.17.0  
```
and then remove the `pandas` and `pyarrow` requirements below. Easiest to just rely on the version that kedro provides (assuming that works).

package/tests/conftest.py
```diff
@@ -120,7 +119,7 @@ def example_transcoded_catalog():
     yield DataCatalog(
         data_sets={
             "model_inputs@pandas": ParquetDataSet(filepath="model_inputs.parquet"),
-            "model_inputs@spark": SparkDataSet(filepath="model_inputs.csv"),
+            "model_inputs@pandas2": CSVDataSet(filepath="model_inputs.csv"),
```
AntonyMilneQB, line 122: Getting rid of the dependence on spark is a great idea, I really like this. Is there anything else in our config.yml that mentions spark that we can get rid of also?

AntonyMilneQB (approved):
Great job getting this all to work, and thanks for writing it all up too! ⭐ 

Just to check: there's not actually any changes to the kedro-viz package itself that we need for compatibility with Python 3.9 or 3.10, right? This is all just changes to tests and CI setup?

Just to be on the safe side, please could you comment out the `restore_cache` that remain ?

package/test_requirements.txt
rashidakanchwala: actually let me try that. it wasn't working initially because hdfs dependency was interfering with tables dependency on 3.9,3.10 (it's very weird that it's passing on kedro). but now that I have removed spark completely, the above might work.

tynandebold (approved):
This is great, thanks so much

package/tests/conftest.py
rashidakanchwala: there is in js tests but it is totally unrelated to this. It is just front-end testing if things work and doesn't have any python dependency

package/test_requirements.txt
rashidakanchwala: @AntonyMilneQB  -- it doesn't work. It fails when installing pytables dependency. I will revert back to having it seperate for now.

PR #832, opened 2022-04-26
Improve performance when many datasets are missing

deepyaman:
Signed-off-by: Deepyaman Datta <deepyaman.datta@utexas.edu>

## Description

If you have a lot of "missing" parameters in your data catalog, Kedro will give you fuzzy match suggestions (if possible) for what that parameter could be. The fuzzy-matching process is slow, but this probably doesn't matter a lot for most cases, since if you have a lot of missing parameters, you're probably doing something wrong. Think it sounds fun to solve the algorithmic problem of looking up multiple strings for approximate matches in a single pool of catalog entries (I assume just LSH is the way), since the one-by-one approach is a bit inefficient, but probably not important.

HOWEVER... if you're using something, like a plugin, where you have a bunch of dynamically added parameters, this can get slow. We had an issue where Kedro-Viz was especially slow (as in, Viz taking 10+ minutes on CI), because such a plugin specifically only adds those parameters before pipeline run. Thing is, Kedro-Viz doesn't even care about the approximate matches, so you're just wasting time fuzzy matching.

Hacky solution: patch out the fuzzy matching for Kedro-Viz in this situation

Longer-term solution: make `_get_datasets` take a `suggest` flag, and call `_get_datasets` in Kedro-Viz with that flag set to False (i.e. what this and the linked Kedro PR do)

Linked: https://github.com/kedro-org/kedro/pull/1482

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/832"></a>


package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -2,11 +2,44 @@
...
+    ) -> AbstractDataSet:
+        if data_set_name not in self._data_sets:
+            raise DataSetNotFoundError(
+                "DataSet '{}' not found in the catalog".format(data_set_name)
+            )
+
+        data_set = self._data_sets[data_set_name]
+        if version and isinstance(
+            data_set, AbstractVersionedDataSet
+        ):  # pragma: no cover
```
deepyaman, line 34: This condition is unreachable in Kedro-Viz, but I left it to be aligned with the original, linked code.

limdauto (approved):
Thank you :star: This makes sense. Also remind me of this a little bit... https://github.com/kedro-org/kedro/pull/953

tynandebold (approved):
Thanks for this, @deepyaman.

Would you mind please adding a line to the [release notes](https://github.com/kedro-org/kedro-viz/blob/main/RELEASE.md) before merging?

deepyaman:
> Thanks for this, @deepyaman.
> 
> Would you mind please adding a line to the [release notes](https://github.com/kedro-org/kedro-viz/blob/main/RELEASE.md) before merging?

@tynandebold Done!

deepyaman:
> Thank you ⭐ This makes sense. Also remind me of this a little bit... [kedro-org/kedro#953](https://github.com/kedro-org/kedro/pull/953)

Yep, very similar, I thought the same before looking into the exact root cause...

PR #838, opened 2022-05-03
Fix: render without routing when using as a React component

tynandebold:

## Description

The Brix team tried to render the component instance of Viz on a page with a route like this: `/post/:id`. That didn't work because of the exact path matching we have [here](https://github.com/kedro-org/kedro-viz/blob/main/src/components/wrapper/wrapper.js#L57-L62). Nothing was rendered.

## Development notes

Now we use the `displayGlobalToolbar` variable as a switch case: if we display to global toolbar, then we also want all other app elements (`UpdateReminder`, `ExperimentTracking`, etc.). If not, we only render the `FlowChartWrapper` component.

## QA notes

You need to test locally in a dummy repo. You can use `npm link` to do that, but it's tedious. I can show you, as I've already done with @studioswong.

## Checklist

### DIFF
--- a/src/components/wrapper/wrapper.test.js
+++ b/src/components/wrapper/wrapper.test.js
@@ -3,13 +3,13 @@ import { mockState, setup } from '../../utils/state.mock';
 
 const { theme } = mockState.spaceflights;
 const mockProps = {
-  theme,
   displayGlobalToolbar: true,
+  theme,
 };
 
 const mockPropsNoGlobalToolbar = {
-  theme,
   displayGlobalToolbar: false,
+  theme,
 };
 
 describe('Wrapper', () => {
@@ -34,9 +34,8 @@ describe('Wrapper', () => {
 
   it('maps state to props', () => {
     expect(mapStateToProps(mockState.spaceflights)).toEqual({
-      loading: false,
-      theme,
       displayGlobalToolbar: true,
+      theme,
     });
   });
 });
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -7,6 +7,7 @@ Please follow the established format:
 -->
 
 ## Major features and improvements
+
 - Added support for Python 3.9 and 3.10. (#815)
 
 ## Bug fixes and other changes
@@ -14,6 +15,7 @@ Please follow the established format:
 - Change route name from `runsList` to `experiment-tracking`. (#820)
 - Update feature flag description to remind the user of the need for page refresh to apply settings. (#821)
 - Fix experiment tracking not showing run details bug on Windows. (#809)
+- Fix rendering of React component instance with custom routes. (#838)
 
 # Release 4.4.0
 
--- a/src/components/global-toolbar/global-toolbar.js
+++ b/src/components/global-toolbar/global-toolbar.js
@@ -21,7 +21,6 @@ export const GlobalToolbar = ({
   onToggleSettingsModal,
   onToggleTheme,
   theme,
-  visible,
 }) => {
   return (
     <>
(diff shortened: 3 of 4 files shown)


studioswong (commented):
Thanks for setting this up - glad that we got a chance to catch up on this,  it looks much better to use the `displayGlobalToolbar` to single out the Brix use case now :) 

It would be great if we can add a test case under `wrapper.test.js` to test this new use case for the npm component import 😄

tynandebold:
> Thanks for setting this up - glad that we got a chance to catch up on this, it looks much better to use the `displayGlobalToolbar` to single out the Brix use case now :)
> 
> It would be great if we can add a test case under `wrapper.test.js` to test this new use case for the npm component import 😄

Great call. I added a test checking if the `SettingsModal` is rendered or not. We already check if the `GlobalToolbar` renders, so this seems suitable. Let me know if you think there's another thing we can check for though.

src/components/wrapper/wrapper.test.js
```diff
@@ -32,11 +32,16 @@ describe('Wrapper', () => {
     expect(container.length).toBe(0);
   });
 
+  it('does not display the settings modal when displayGlobalToolbar is false', () => {
+    const wrapper = setup.mount(Wrapper, mockPropsNoGlobalToolbar);
+    const container = wrapper.find('.pipeline-settings-modal');
+    expect(container.length).toBe(0);
+  });
+
```
studioswong, line 40: In looking at this test, this test doesn't serve any purpose / isn't exactly testing anything as the settings modal won't be displayed even if `displayGlobalToolbar` is true ( `.pipeline-settings-modal` is only triggered when the modal is triggered and not on the existence of the globalToolbar). 

This could be deleted and ideally, we should set up a test to mock the navigation to one of the routes (such as `/experiment-tracking`) and ensuring that the app doesn't get routed when `displayGlobalToolbar` is false. ( the test condition would be something like 'does not perform any routing when displayGlobalToolbar is false').

rashidakanchwala (approved):

src/components/wrapper/wrapper.test.js
```diff
@@ -32,9 +33,15 @@ describe('Wrapper', () => {
...
-    const container = wrapper.find('.pipeline-settings-modal');
+  it('does not perform any routing when displayGlobalToolbar is false', () => {
+    const wrapper = setup.mount(
+      <MemoryRouter initialEntries={['/experiment-tracking']}>
+        <Wrapper />
+      </MemoryRouter>,
+      mockPropsNoGlobalToolbar
+    );
+
+    const container = wrapper.find('.pipeline-ui--experiment-tracking');
```
studioswong, line 44: Thanks for adding this test - this is definitely structured in a much better way.

I've just had a play around with this and tried testing this using `mockProps` instead of `mockPropsNoGlobalToolbar` but unfortunately this still passes, which indicates that there is something missing from this setup. 

This could either be that we are testing with the wrong css class, or something to do with the setup - worth to do a bit of digging around to find out and set up the right test.

tynandebold: I'm struggling on this one. I don't see a good way to test this.

Since we'll only render the `FlowChartWrapper` when we use `mockPropsNoGlobalToolbar`, we won't have access to any routing, so actually my use of `MemoryRouter` doesn't make sense here.

And since we don't have access to that routing or even `react-router` at this stage, I don't see a way to mock navigation to one of the routes using any of the methods you [normally would use](https://v5.reactrouter.com/web/guides/testing) when routing is present on the page or in the component.

If you have a better idea, do please share it, because I can't see it.

studioswong (approved):
Glad that we got to pair and look into this - what weird things we are seeing with the tests!

Thanks for pushing through with this and glad we came to a test that is generic enough to capture the intent of this PR - let's get this in for the release 😄

Issue #840, opened 2022-05-04
Fix: Flowchart does not load for JSON data inputs


studioswong:
The flowchart does not load properly on first load with JSON data inputs, which is an issue that is uncovered / introduced since the removal of [Kedro-UI webfonts ](https://github.com/kedro-org/kedro-viz/pull/731). 

This is due to the web-worker to render graph nodes not being able to be called and dispatched with initial state initialisation and only on further state changes - this affects Kedro-Viz when loading JSON data inputs as all the data are set within the state during initial state initialisation ( while running Kedro-Viz from a project refreshs the state with the REST API calls and hence does not face this problem.)

This ticket is to fix this to ensure the state updates on initial load.

PR #843, opened 2022-05-05
Fix- Flowchart does not load for static data inputs

studioswong:
## Description

fixes[ issue 840](https://github.com/kedro-org/kedro-viz/issues/840). 

Currently the flowchart would not load for static data soures, which causes problems for our npm component usecase. This PR fixes that. 

## Development notes

This problem is mainly due to the fact that the `getGraphInput` action does not get dispatched on first load ( it is only called via web worker which will only be triggered for future state changes. ) I have added an action within the `configureStore` method to ensure this action still gets fired on first load. 

I have also modified the test set up with mocking the `configureStore` function to ensure it properly configures the store. 

Please note that I had not added any tests for this given the lack of a test suite setup for `store.js`, and it might be an overkill to add that just for testing the dispatch of this action( especially this is a quick fix with the aim to be included in the release tomorrow). 

That said, happy to look into adding one for `store.js` in general in a future ticket. 

## QA notes

Try loading the app via one of the static data sources - we should now be able to load the flowchart properly. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/843"></a>


src/store/store.js
```diff
@@ -62,14 +62,21 @@ const saveStateToLocalStorage = (state) => {
  * @param {Object} initialState Initial Redux state (from initial-state.js)
  * @return {Object} Redux store
  */
-export default function configureStore(initialState) {
+export default function configureStore(initialState, dataType) {
```
tynandebold, line 65: Nit: perhaps add this new argument to the function's comment above?

src/store/store.js
```diff
@@ -62,14 +62,21 @@ const saveStateToLocalStorage = (state) => {
...
     window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
   const store = createStore(
     reducer,
     initialState,
     composeEnhancers(applyMiddleware(thunk))
   );
+
+  // dispatch the calculateGraph action to ensure the graph nodes still gets rendered
+  // on initial load if data is loaded via data prop instead of fetching from Rest API
+  if (dataType !== 'json') {
```
tynandebold, line 76: I was testing this out by logging `dataType` and it was always coming back as `undefined`.

tynandebold (changes requested):
This is great and is working as expected, though I think your added `if` condition is always true. Looks like you need to pass a second argument into the `configureStore` function [here](https://github.com/kedro-org/kedro-viz/blob/main/src/components/app/app.js#L27) so the app knows if it should calculate the graph when the store is being configured.

```js
this.store = configureStore(initialState, initialState.dataSource);
```

studioswong:
> This is great and is working as expected, though I think your added `if` condition is always true. Looks like you need to pass a second argument into the `configureStore` function [here](https://github.com/kedro-org/kedro-viz/blob/main/src/components/app/app.js#L27) so the app knows if it should calculate the graph when the store is being configured.
> 
> ```js
> this.store = configureStore(initialState, initialState.dataSource);
> ```

good catch, turns out I forgot to commit this change - I've just pushed it in. 

The change would be `this.store = configureStore(initialState, this.props.data);`

src/store/store.js
studioswong: I've added the description for this parameter - please see my newer commits.

src/store/store.js
studioswong: good catch - this is due to the lack of the prop in `configureStore` as you've pointed before ( still pinching myself for forgetting to commit that change 😆 ) but now that it's pushed in, this would not be coming back as `undefined`.

RELEASE.md
```diff
@@ -14,6 +14,7 @@ Please follow the established format:
 - Change route name from `runsList` to `experiment-tracking`. (#820)
 - Update feature flag description to remind the user of the need for page refresh to apply settings. (#821)
 - Fix experiment tracking not showing run details bug on Windows. (#809)
+- Fix flowchart not showing on intiial load for static data inputs. (#840)
```
tynandebold, line 17: Oh forgot this: don't we usually reference the PR number, not the issue number? So this would be `#843` and not `#840`.

RELEASE.md
```diff
@@ -14,6 +14,7 @@ Please follow the established format:
 - Change route name from `runsList` to `experiment-tracking`. (#820)
 - Update feature flag description to remind the user of the need for page refresh to apply settings. (#821)
 - Fix experiment tracking not showing run details bug on Windows. (#809)
+- Fix flowchart not showing on intiial load for static data inputs. (#843)
```
rashidakanchwala, line 17:
```suggestion
- Fix flowchart not showing on initial load for static data inputs. (#843)
```

rashidakanchwala (approved):
LGTM. thanks @studioswong

rashidakanchwala:
we have this same issue with running kedro-viz on jupyter notebooks and databricks. the first time -- it displays a blank viz and then when u click on something, the viz loads. i am hoping this fix will sort that out too... magically ;)

studioswong:
> we have this same issue with running kedro-viz on jupyter notebooks and databricks. the first time -- it displays a blank viz and then when u click on something, the viz loads. i am hoping this fix will sort that out too... magically ;)

interesting - that sounds like it could be likely a related issue - I imagine both uses cases are passing a json object directly as a prop instead of retrieving data from the server?

tynandebold (approved):
Working great

PR #850, opened 2022-05-16
Simplify dataset `suggest` handling code

AntonyMilneQB:
## Description

https://github.com/kedro-org/kedro-viz/pull/832 and https://github.com/kedro-org/kedro/pull/1482 added the `suggest` argument to `_get_dataset` and made sure it's used in kedro-viz (released in kedro 0.18.1).

#832 also patched `_get_dataset` to manually disable `suggest` if the user has kedro < 0.18.1 so that users of older version of kedro could also benefit from the performance improvement. Now that kedro 0.18.1 is released, we need to add additional test cases to get coverage on these lines by manually setting `KEDRO_VERSION` in a unit test. This is definitely possible but not something we currently do in any unit tests.

Overall I think we should try to minimise the number of switches that depend on `KEDRO_VERSION` in our codebase. In this case I don't think the importance of the change justifies the additional complexity to make it work on older versions of kedro. Given that this is quite a niche, recently fixed issue and just a performance improvement rather than a big bug, I think it's fine to say that you only get the performance improvement if you're on >=0.18.1.

@deepyaman I'm not absolutely set on this, so let me know if this is very offensive to you and I will leave your changes as they were and update the tests to get coverage.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -18,7 +18,7 @@ Please follow the established format:
 - Update feature flag description to remind the user of the need for page refresh to apply settings. (#821)
 - Fix experiment tracking not showing run details bug on Windows. (#809)
 - Fix rendering of React component instance with custom routes. (#838)
-- Improve performance when many datasets are missing. (#832)
+- Improve performance when many datasets are missing (requires `kedro>=0.18.1`). (#832)
 - Fix flowchart not showing on initial load for static data inputs. (#843)
 
 # Release 4.4.0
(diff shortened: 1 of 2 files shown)


MerelTheisenQB (approved):
It's a nice to have in older versions, but I agree it's not worth the complexity of having the logic for matching versions and testing all of that. 

`0.18.1` is a non-breaking version in the `0.18` series so ideally users should just update and hopefully this fix is even more of an incentive to do so 😄

PR #851, opened 2022-05-16
Fix: lambda python functions not rendering correctly on flowchart

### LINKED ISSUE #817: Lambda python functions don't render nicely on the flowchart
## Description
As you can see, nodes which use a lambda (anonymous function) on do not render nicely on the flowchart:

image

However, it does look like they render correctly on the tree panel to the left 🤔

## Steps to Reproduce

Make a node that looks like this:

```python
node(
   func=lambda x : x,
   inputs='x',
   outputs='y'
)
```

## Expected Result

It should look like `<lambda>` much the same as the tree panel

tynandebold:
## Description

Fixes #817.

## Development notes

I'm no longer using `_strip_tags` in the backend, so the `<>` come through to the frontend. I've create a `replaceMatches()` utility function and added that to the `node-list-row.js` file to ensure both `<lambda>` and `<partial>` will render via `dangerouslySetInnerHTML`.

## QA notes

I've added a lambda node so this can be tested. Open the Gitpod URL and expand the `Ingestion` modular pipeline. You'll see the lambda function there.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/851"></a>


AntonyMilneQB:
This looks right on the tree view and metadata side panel to me but looks kind of funky when I search for it...
image

Outside the scope of this PR, but related: did we ever get a ticket created to change these `partial` names to something more descriptive? As per the discussion in https://github.com/kedro-org/kedro-viz/pull/692.

RELEASE.md
```diff
@@ -6,6 +6,10 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+## Bug fixes and other changes
+
+- Fix lambda Python functions not rendering correctly on flowchart. (#851)
```
AntonyMilneQB, line 11:
```suggestion
- Fix lambda and partial Python functions not rendering correctly on flowchart. (#851)
```
The same fix here will work for both cases :)

src/utils/index.js
```diff
@@ -59,3 +59,25 @@ export const changed = (props, objectA, objectB) => {
...
+ * @param {string} str The string to check
+ * @returns {string} The string with or without replaced values
+ */
+const toReplace = {
+  '&lt;': '<',
+  '&gt;': '>',
+};
+
+export const replaceMatches = (str) => {
+  if (str && str.length > 0) {
```
AntonyMilneQB, line 74: Curious: what is this conditional for here? I sort of understand the zero-length string case but not the case where `bool(str)` would evaluate to False - when does that occur?

tynandebold: If you don't have that check you'd get this error: `TypeError: Cannot read property 'length' of undefined`. I've changed it now to this `str?.length > 0` which does the same thing: first checks if there's a string, and if so, proceed to find the length.

RELEASE.md
tynandebold: Nice one. Updated!

tynandebold:
> This looks right on the tree view and metadata side panel to me but looks kind of funky when I search for it... image
> 
> Outside the scope of this PR, but related: did we ever get a ticket created to change these `partial` names to something more descriptive? As per the discussion in #692.

Good catch on the search. I need to update it, but it's a bit tricky since we're putting these partial and lambda names between `< >` symbols, which really conflicts with HTML, since that will render an HTML element.

Can we use something else instead of these to make it easier for ourselves, like a `~` or something? Or is `< >` convention?

AntonyMilneQB:
> > This looks right on the tree view and metadata side panel to me but looks kind of funky when I search for it... image
> > Outside the scope of this PR, but related: did we ever get a ticket created to change these `partial` names to something more descriptive? As per the discussion in #692.
> 
> Good catch on the search. I need to update it, but it's a bit tricky since we're putting these partial and lambda names between `< >` symbols, which really conflicts with HTML, since that will render an HTML element.
> 
> Can we use something else instead of these to make it easier for ourselves, like a `~` or something? Or is `< >` convention?

I'm afraid `<>` is quite a strong convention since this is how Python names anonymous functions:
image


If `<>` is too awkward then I would just not have any special character there at all, i.e. `lambda`. This would be fine I think.

studioswong (changes requested):
Thanks for tackling this.

I haven't reviewed the code yet, but on checking out the branch and poking around it seems that this breaks the search functionality in general:
image

This is due to having taken out the `dangerouslySetInnerHTML` which is needed to implement the bold highlighting during search.

package/kedro_viz/models/graph.py
```diff
@@ -43,10 +43,6 @@ def _strip_namespace(name: str) -> str:
     return re.sub(pattern, "", name)
 
 
-def _strip_tags(name: str) -> str:
-    return name.replace("<", "&lt;").replace(">", "&gt;")
```
tynandebold, line 47: I think the solve for this is to leave the `<>` in place in the backend and then change it to use the HTML entities _only where we need it_ in the frontend.

src/components/node-list/node-list-row.js
```diff
@@ -124,7 +131,9 @@ const NodeListRow = memo(
                 'pipeline-nodelist__row__label--disabled': disabled,
               }
             )}
-            dangerouslySetInnerHTML={{ __html: label }}
+            dangerouslySetInnerHTML={{
+              __html: replaceMatches(label, stringsToReplace),
```
tynandebold, line 135: Replace the `<>` here so we can still use `dangerouslySetInnerHTML` without a problem, specifically when searching.

src/components/search-list/search-list.scss
```diff
@@ -9,10 +9,11 @@
   }
 
   .search-bar .icon--close {
-    top: 18px;
+    height: 20px;
     right: 2px;
+    top: 18px;
     width: 20px;
-    height: 20px;
+    z-index: 1;
```
tynandebold, line 16: You couldn't click the `X` icon to clear the search. This fixes that.

tynandebold:
On the main branch, searching for something with `<` doesn't work. See here:

image

I tried to fix that in this ticket but it turned into a bigger issue. I'll create a separate bug report for that and leave this one here.

rashidakanchwala (approved):
thanks @tynandebold . approving it since we will look at the <> bug in another ticket :)

studioswong (commented):
The PR is definitely in a much better shape now!

one small thing I noticed ( sorry to be a pain)  - when searching for the lambda function with a `<`, the search results displayed gets weird ( some characters gets wiped out):
image

I think this could potentially be tied to having replaced `<lambda` with `&lt;partial&gt;` ? ( though I am not sure at all - this might require a bit of further poke around to find a workaround for that.)

tynandebold:
> The PR is definitely in a much better shape now!
> 
> one small thing I noticed ( sorry to be a pain) - when searching for the lambda function with a `<`, the search results displayed gets weird ( some characters gets wiped out): image
> 
> I think this could potentially be tied to having replaced `<lambda` with `&lt;partial&gt;` ? ( though I am not sure at all - this might require a bit of further poke around to find a workaround for that.)

Good eye. I think that's to be expected at this point, which will be solved by the solution for this [bug ticket](https://github.com/kedro-org/kedro-viz/issues/854).

In the current version, if you have a `<lambda>` function and search for `<lam`, nothing actually shows up. In this case, if you do the same, at least you get a search result, even though the highlighting isn't exactly correct. I would say returning the result you're showing in your screenshot is better than not returning anything at all.

If you're comfortable with that, would be great if we can approve this now and iterate on the fix in the linked ticket above.

studioswong (approved):
Just saw the ticket you mentioned above, all good 👍 

May I suggest to add in an end-2-end test in the upcoming PR for #854  to ensure we capture any errors with the search case for Lambda functions ( since this seems to be a use case that could be easily prone to errors)? 😆  

Given that we are looking into releasing soon, it might be worth to release first before merging this in so this can be released alongside the fix for the search (#854 ) as one complete fix? ( totally up to the team)

AntonyMilneQB (approved):
Definitely good to get this merged, thanks for doing all the iteration on it!

If it turns out the search thing is hard to fix then honestly I think it's fine if we just drop the `<>` altogether and call it plain `lambda` or `partial` everywhere. Also the number of people who will be searching for a lambda function in the is very small. So would be good to fix but not a biggy at all I think.

tynandebold:
> Definitely good to get this merged, thanks for doing all the iteration on it!
> 
> If it turns out the search thing is hard to fix then honestly I think it's fine if we just drop the `<>` altogether and call it plain `lambda` or `partial` everywhere. Also the number of people who will be searching for a lambda function in the is very small. So would be good to fix but not a biggy at all I think.

Good to know. I'll add this comment to the context of the other ticket I created. I do think it'll be a pain to solve and we'd be wise to just drop `<>` all together, especially as it's probably a very small slice of users searching for it, as you said.

PR #852, opened 2022-05-17
 [Fix] Change Types on Metadata Side Panel

### LINKED ISSUE #752: [FE] Change `Types` on Metadata Side Panel
(transfer from Jira, created by @yetudada)

- We still have a mismatch with how terms on Kedro-Viz vs Kedro are called and one of the ways it shows itself it on the Metadata Side Panel.
- When you open the metadata side panel and see the `Type` entry we refer to `data` instead of `dataset` and `task` instead of `node`.
- This issue focuses on updating those entries: change `type: task` to `type: node` for Kedro Node node and remove `type: data` for dataset node. `type: params` remains unchanged.

Screenshot 2021-07-06 at 16 55 37

studioswong:
## Description

resolves #752 . 

## Development notes

I have added a function to translate the naming of the different types of nodes on the FE to return the desired naming. ( The BE still returns the different types of node as `task`, `data` and `params`, while we want to ensure the nodes are displayed as `node`, `dataset` and `params`. 

I have also updated a variable naming ( I have kept `isTaskNode` as it is to avoid confusion for `isNodeNode`), as well as any related tests. 

## QA notes

The node type on the metadata panel should have the newly updated node type naming. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/852"></a>


src/components/metadata/metadata.js
```diff
@@ -36,7 +36,7 @@ const MetaData = ({
   // Hide plot modal when selected metadata changes
   useEffect(() => onTogglePlotModal(false), [metadata, onTogglePlotModal]);
   const isTaskNode = metadata?.type === 'task';
```
tynandebold, line 38: Maybe add a comment after this line to say that this really should be `isNodeNode`, though this `isTaskNode` naming is better, hence we're leaving it, as you said in your PR description?

src/components/metadata/metadata.js
```diff
@@ -53,6 +53,17 @@ const MetaData = ({
...
 
+  // translates the naming for the different types of nodes
+  const translateMetadataType = (metaDataType) => {
+    if (metaDataType === 'task') {
+      return 'node';
+    }
+    if (metaDataType === 'data') {
+      return 'dataset';
+    }
+    return metaDataType;
```
tynandebold, line 64: This could be a good place to use `switch`.

tynandebold (approved):
Looks great, thanks for this.

Just a couple of small things from me, though nothing blocking.

src/components/metadata/metadata.js
AntonyMilneQB: I'd actually leave the `isDataNode` as it was called rather than renaming to `isDatasetNode`. That way everything is consistent and `isTaskNode` would be "correct" already. The `isXNode` has `X` consistent with the backend naming scheme. This way the only inconsistency is the way the names appear on the screen, which is all handled by `translateMetadataType`.

AntonyMilneQB (approved):
Looks good, just a small comment on naming!

I really wish we'd called the generic "node" a "vertex" instead and then had 3 different types of vertex: node, dataset, parameter. But that's not for now 😅

src/components/metadata/metadata.js
studioswong: yea @AntonyMilneQB I agree on this, and was think the same thing, so you kind of confirmed my thoughts!

I'll just revert it back to `isDataNode`.

rashidakanchwala (approved):
LGTM! thanks @studioswong

PR #865, opened 2022-05-24
[KED - 857] Improve telemetry to work with SVG elements

### LINKED ISSUE #857: Improve telemetry to work with SVG elements
At this moment in time, Heap isn't able to read data attributes that exist on SVG elements. We want to be able to better track what users are interacting with on the `<Flowchart />`, and since that entire thing is an SVG, we can't use the data attribute we thought we could. 

Let's come up with a better mechanism. Heap **can** read classes on SVGs, so that could be the way to go right now.

The line of code we need to update is [right here](https://github.com/kedro-org/kedro-viz/blob/e28e25072c42da87549dc798829323325a24c929/src/components/flowchart/draw.js#L162).

I propose we change that line to this (and then delete the line above this one): 

```js
.attr('class', (node) => `pipeline-node pipeline-node--${node.type}`)
```

Then we can define an event in Heap based off of these distinct class names.

Let's be sure to test this locally first. When I was clicking on nodes/datasets/etc the other day and looking in Heap, I was seeing the element with a class name of `.pipeline-node__bg` as the one that was clicked on. We may need to move the proposed line above around a little bit, depending on how it looks in Heap.

rashidakanchwala:
## Description

Resolves #857 

## Development notes

I have added the 'node type' to rectangle as on heap the rect svg is what gets registered. I have also defined multiple events on HEAP (Dataset selected, Task node selected, Parameter node selected, Modular pipelines selected' 

Screenshot 2022-05-24 at 18 34 04

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/865"></a>

### DIFF
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -159,7 +159,6 @@ export const drawNodes = function (changed) {
     enterNodes
       .attr('tabindex', '0')
       .attr('class', 'pipeline-node')
-      .attr('data-heap-event', (node) => `clicked.graph.${node.type}`)
       .attr('transform', (node) => `translate(${node.x}, ${node.y})`)
       .attr('data-id', (node) => node.id)
       .classed(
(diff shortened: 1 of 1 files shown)


tynandebold (approved):
Looks and works great.

Issue #858, opened 2022-05-20
Multiple pipelines do not hide namespaced elements correctly


antheas:
## Description
Consider a project with multiple pipelines in pipeline registry and modular pipelines in those.

When there is a namespaced element that only appears in one of the pipelines, that element is not correctly hidden in the other ones (appears as a blank space).

![image](https://user-images.githubusercontent.com/5252246/169601592-6fbeb8e3-06c0-4b86-8992-3c2590af47f0.png)

## Context
It's a simple visual bug that clutters the UI.

## Steps to Reproduce
To reproduce, create a kedro project with the following pipeline registry:
``` python
"""Project pipelines."""
from typing import Dict

from kedro.pipeline import Pipeline, pipeline, node


def register_pipelines() -> Dict[str, Pipeline]:
    """Register the project's pipelines.

    Returns:
        A mapping from a pipeline name to a ``Pipeline`` object.
    """
    return {"__default__": pipeline([
        node(func=lambda a: False, inputs='tst.a', outputs='d')
    ]), "pipe2": pipeline([
        node(func=lambda a: False, inputs='tst.b', outputs='tst.c')
    ]), }

```

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version: Latest Vivaldi/Chrome
* Operating system and version: Linux
* NodeJS version used (if relevant): NA
* Kedro version used (if relevant): 0.18.1
* Python version used (if relevant): 3.8.10

## Checklist
- [x] Include labels so that we can categorise your issue

tynandebold:
Thanks for this, @antheas. We'll have a look soon!

PR #869, opened 2022-05-26
[Bug Fix] Modular pipelines with the same name on different pipelines do not render corrected in Kedro Viz

rashidakanchwala:
## Description

This bug was found when trying to fix this issue #858 raised by a user. 

Currently if you have multiple kedro pipelines where the namespace (modular pipeline name) is the same it causes an issue on Kedro Viz. 

Let's say you have the below pipeline 
 

```python

    return {"__default__": pipeline([node(func=lambda a: False, inputs='a', outputs='d')
    ,node(func=lambda a: False, inputs='d', outputs='e')],namespace='tst'), 
             "pipe2": pipeline([node(func=lambda b: False, inputs='b', outputs='c')],namespace='tst'),
    }
```
then if you render on Kedro-viz you get the extra empty rows for pipelines __default__ .tst and __pipe2__ .tst (see below screenshots)

Screenshot 2022-05-26 at 12 50 33

Screenshot 2022-05-26 at 12 50 44

this is because in the api, it always includes all children of the modular pipeline (in this instance - tst) from both pipelines instead of from the selected pipeline 

Screenshot 2022-05-26 at 12 51 14

Screenshot 2022-05-26 at 12 51 30

## Development notes

When i went through the code, i realised children are added to the modular pipeline in the repository layer which runs only the first time. I changed this so that children specific to the pipeline selected are added in the mangers.py. Also because modular_pipeline names are the same, i ensure that the modular_pipeline is first made empty and then the children of only the current registered pipeline are added. 

## QA notes

Once implementation is accepted, will write the required tests 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/868"></a>

### DIFF
--- a/package/kedro_viz/data_access/repositories/modular_pipelines.py
+++ b/package/kedro_viz/data_access/repositories/modular_pipelines.py
@@ -213,12 +213,6 @@ class ModularPipelinesRepository:
         # so does the modular pipeline.
         modular_pipeline.pipelines.update(node.pipelines)
 
-        # Since we extract the modular pipeline from the node's namespace,
-        # the node is by definition a child of the modular pipeline.
-        self.add_child(
-            modular_pipeline_id,
-            ModularPipelineChild(id=node.id, type=GraphNodeType(node.type)),
-        )
         return modular_pipeline_id
 
     def has_modular_pipeline(self, modular_pipeline_id: str) -> bool:
--- a/package/kedro_viz/data_access/managers.py
+++ b/package/kedro_viz/data_access/managers.py
@@ -433,6 +433,12 @@ class DataAccessManager:
                 edges.remove_edge(GraphEdge(bad_input, modular_pipeline_id))
                 node_dependencies[bad_input].remove(modular_pipeline_id)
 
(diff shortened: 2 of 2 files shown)


package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -5,8 +5,7 @@
 from kedro_viz.constants import ROOT_MODULAR_PIPELINE_ID
 from kedro_viz.models.graph import (
     DataNode,
-    GraphNode,
-    GraphNodeType,
+g    GraphNode,
```
tynandebold: There's a `g` in here that shouldn't be.

tynandebold (approved):
Visually, this looks great. I've checked it against the standard suite of data sources and nothing looks amiss. Well done!

I'll approve now, though definitely wait for @limdauto or @AntonyMilneQB to check it, too.

limdauto (changes requested):
Thanks for looking into this @rashidakanchwala I'm temporarily blocking this because I have a feeling this is not the best way to solve the problem. I will circle back once I'm done investigating it.

PR #870, opened 2022-05-26
Disable uvicorn logger to remove duplicated messages

AntonyMilneQB:
## Description
Did you ever notice how chatty kedro-viz logging is and that all the messages are duplicated? 
image

Now that we're changing kedro's default logger to something more beautiful in https://github.com/kedro-org/kedro/issues/1512, the duplication is even more stark and makes the output look wack:
image

The reason for this is that `uvicorn` sets up its own custom logger, and then kedro's logging is also there, so messages get duplicated. The solution is to disable `uvicorn`'s logger. I left the log level set to what they use (INFO) so that we still get a lot of messages, just they're no longer duplicated.

After https://github.com/kedro-org/kedro/issues/1512 and this are merged, kedro viz log messages will then look super sexy, like this:
image

There's no need to wait until #1512 is merged to merge this though. We can go ahead with this right away.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/870"></a>


datajoely:
🔥

limdauto (approved):
Danker. I'm guilty of never reading this log

limdauto:
@AntonyMilneQB although could you help me check if you throw an error in one of the endpoint, does the stack trace get logged correctly?

tynandebold (approved):
Excellent

PR #871, opened 2022-05-30
Fix namespace collison when two different registered pipelines have a modular pipeline with the same name

limdauto:
## Description

Fixes https://github.com/kedro-org/kedro-viz/issues/858

## Development notes

The root cause of the issue is that each registered pipeline should have its own modular pipeline repository. Currently, there is one global modular pipeline repository, so if two different registered pipelines have a modular pipeline with the same name, their nodes list will get mixed up.

## QA notes

I have written a unit test and tested it by hand, both using the example in the github issue and the demo project.

## Checklist


package/kedro_viz/data_access/managers.py
```diff
@@ -47,10 +47,15 @@ def __init__(self):
         self.nodes = GraphNodesRepository()
         self.registered_pipelines = RegisteredPipelinesRepository()
         self.tags = TagsRepository()
-        self.modular_pipelines = ModularPipelinesRepository()
-        # Make sure each registered pipeline has a distinct collection of edges.
+
+        # Make sure each registered pipeline has a distinct collection of
```
limdauto, line 51: As you can see, previously, each registered pipeline keeps its own collection of nodes and edges. Using the same pattern for the modular pipelines is therefore natural.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -90,7 +90,9 @@ def get_or_create_modular_pipeline(
         return self.tree[modular_pipeline_id]
 
     def add_input(
-        self, modular_pipeline_id: str, input_node: Union[DataNode, ParametersNode]
+        self,
+        modular_pipeline_id: str,
+        input_node: Union[DataNode, TranscodedDataNode, ParametersNode],
```
limdauto, line 95: Adding this because mypy was failing

package/tests/test_data_access/test_managers.py
```diff
@@ -336,6 +340,30 @@ def test_add_pipelines_with_transcoded_data(
...
+        }
+
+        data_access_manager.add_catalog(DataCatalog())
+        data_access_manager.add_pipelines(registered_pipelines)
+        modular_pipeline_tree = (
+            data_access_manager.create_modular_pipelines_tree_for_registered_pipeline(
+                DEFAULT_REGISTERED_PIPELINE_ID
+            )
+        )
+        assert len(modular_pipeline_tree["tst"].children) == 1
```
limdauto, line 365: Without the fix in this PR, this number will be `3` because it mistook `tst.b` and `tst.c` as part of the `__default__` pipeline which is what the user reported.

package/kedro_viz/data_access/managers.py
rashidakanchwala: thanks, this def feels more natural!

tynandebold (approved):
Works well. Thank you for this!

rashidakanchwala:
hey @limdauto , i was still reviewing it. there is an issue i found when pipe2 is broken on the FE.

limdauto:
@rashidakanchwala oh interesting. Sorry I was too trigger happy. Wanna jump on a call to discuss what you spot?

PR #879, opened 2022-06-01
Feature: click to expand modular pipeline

### LINKED ISSUE #874: Expand namespaced pipeline by double clicking
## Description
I would like to be able to open namespaced pipeline by double clicking on it the pipeline flowchart.

Screen Shot 2022-05-30 at 4 34 11 PM

## Context
This is just a small enhancement to make digging into pipelines faster.

## Possible Implementation
Not sure, but if someone gives me a pointer, I would be happy to do it!

## Possible Alternatives
Perhaps a keyboard shortcut? I definitely think the mouse click is more useufl.

limdauto:
Signed-off-by: Lim Hoang <limdauto@gmail.com>

## Description

Resolves #874.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/879"></a>


limdauto:
@rashidakanchwala sorry could you assign this to someone else? I mentioned to @tynandebold that someone else will need to take over this because I won't have any time soon to polish it. I'm happy to close this too if it creates noise.

tynandebold:
@limdauto it's no trouble. One of us will pick it up. It's not creating any noise. Can you open it up again please?

rashidakanchwala (approved):
Work's great!!! :) thanks @tynandebold!

src/components/flowchart/flowchart.js
```diff
@@ -422,7 +423,11 @@ export class FlowChart extends Component {
    * @param {Object} node Datum for a single node
    */
   handleNodeClick = (event, node) => {
-    this.props.onLoadNodeData(node.id);
+    if (node.type === 'modularPipeline') {
+      this.props.onClickToExpandModularPipeline(node.id);
```
limdauto, line 427: I really think we should listen for double-clicking because it's "hard" to undo the expansion from the graph. You will have to go all the way back to the tree.

limdauto: I think you can rely on `event.detail` for it: https://thewebdev.info/2021/09/25/how-to-handle-double-click-events-in-a-react-component/

tynandebold: Double-clicking here seems a bit anti-pattern though. All other elements in the flowchart, and the entire app, react to single-clicking.

I'd guess users wouldn't naturally double click on a modular pipeline to try and expand it. Am I off in thinking that thought? Curious what you think, as well as @rashidakanchwala @AntonyMilneQB @yetudada.

limdauto: Fair point. I adopted a folder mental model for modular pipeline on the flowchart from @idanov.

* Single click on a folder gives you metadata for that folder on the same screen
* Double click on a folder means you want to get inside that folder

Since we don't have a back button on this potentially quite disruptive action, I'd err on the side of not making it too easy.

AntonyMilneQB: Let me put what I said on discord here also 🙂 
> In my dream world at some point in the future maybe single click could open up the metadata side panel like it does for nodes/datasets (e.g. to display concatenation of experiment tracking dataset within that pipeline). And double click would then be needed to expand the pipeline. But that's quite a way off

I could also imagine us putting stuff like number of nodes/datasets in the the panel for a modular pipeline but that doesn't seem too important or useful to be honest.

For this + @limdauto's argument that it's "hard" to undo the expansion operation it would make more sense for double click to do the expand.

BUT I also 100% agree with @tynandebold that we don't have any double clicking elsewhere on the flowchart so it's a bit inconsistent and not as discoverable for users (and I don't think it's easy to educate them about this - you just have to hope they figure it out really). And I don't think the modular pipeline metadata panel is going to happen for a while anyway. 

Overall both double and single click make sense to me as a user so I don't have any strong preference here. Given the PR here does single click I'd probably just leave it as that for now and see if anyone complains. At least with single click I think we'll have more users discovering and giving feedback on the feature if they don't like it being single click anyway.

AntonyMilneQB (approved):
So happy to have this added!! Thank you to all who worked on it 🎉

src/components/flowchart/flowchart.js
tynandebold: Well said, thank you. I do think if we do double-click we need something to happen on single click.

If we enable the loading of node data when we click on a modular pipeline, this is what we currently get:

image

Is that useful? Maybe not, though it's not terrible. Maybe we could add the modular pipeline's child nodes in there? I don't think that would be difficult.

I'm starting to solutionize a bit now, so unless we want to add this rather-empty metadata panel on single-click of a modular pipeline, I'll keep it single-click to expand the pipeline.

AntonyMilneQB: To me that doesn't look useful enough to be worth exposing. In fact I think it's maybe even worse than just not doing anything else on single click - now it kind of implies that once you've clicked the modular pipeline, that empty metadata panel is all you get and you wouldn't even think of double clicking to expand.

I think the metadata should probably meet some kind of minimal criteria to be worth showing:
* human-readable `type` (not `modularPipeline`)
* some non-trivial information like the child nodes/pipelines or something
* maybe even some kind of "expand modular pipeline" button would belong in there?
... which would overall require quite a bit of extra design thought.

tynandebold: Copy. Single-click it is (for now).

Issue #883, opened 2022-06-02
--params option in CLI commands


IanVlasov:
## Description
Hello everyone! First, I would like to thank you for the great project and all your efforts to enhance it!

And to the feature. I would like to ask if it is possible to add --params option to the CLI commands to run Kedro Viz the same way as Kedro itself:
```
kedro viz --env environment --params param:value
```
## Context
My request is related to the case when you configure your pipeline depending on the parameter file. I know there are cases except mine when people are trying to do so. It would be perfect if pipeline creation repeats the one that introduced in the Kedro package. Currently, I suffer a bit due to lack of --params option among the CLI commands for Kedro Viz because I use it to define the path to the current parameter file that should be taken for pipeline construction. Something like that:
```
kedro run --params path:path/to/parameter/file
```
To make it work, I added a global variable in settings.py which is redefined by the after_context_created hook if the new value for path was passed. This variable is injected to pipeline_registry.py where config loader loads parameters from defined path and then transmit them to create_pipeline functions.

I know that it is a bit hacky way and that Kedro team tries to avoid such cases due to difficulties to maintain such approach, but seems I am not alone who tries making pipeline creation easier for the rest of the team. E.g., in our case, we have a number of data scientists who loves notebooks and are not too good in their programming skills to properly define the whole pipeline they need. Moreover, we are trying to get some speed up in defining their pipelines that can be quite long and diving into code every time you would like to make any small changes doesn't lead to it. That is why we are trying to make their life a bit easier by giving them a more or less predefined pipeline that can be constructed based on their custom parameters.

## Possible Alternatives
Currently, I can directly use globals.yml for my purposes, but the workflow with CLI commands seems to me much easier than modifying and saving files.

Thanks in advance for all your help!

IanVlasov:
I have prepared the [pull request](https://github.com/kedro-org/kedro-viz/pull/884). Sorry, if anything is incorrect: first time contributing to the open-source project

PR #884, opened 2022-06-03
--params option added to CLI commands similar to one that Kedro has

IanVlasov:
Signed-off-by: Yan Vlasov <vlasov.yan.ed@gmail.com>

## Description

Resolve [#883](https://github.com/kedro-org/kedro-viz/issues/883)

## Development notes

- Added additional --params click.option to cli.py and passed it to `run_server_kwargs`
- Added `extra_params` param to server.py and passed it to `kedro_data_loader.load_data`
- Added 'extra_params' param to data_loader.py and passed it to `KedroSession.create`
- Updated cli tests

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/884"></a>


datajoely:
This is really cool!

tynandebold:
Thank you for this, @IanVlasov. We'll have a look at it and get back to you!

cc @AntonyMilneQB @limdauto

IanVlasov:
Hello @AntonyMilneQB, thanks for congratulations! And, yes, you are totally right that this issue is related to the one I asked in kedro project about.
1. I can confirm that the approach with `after_context_created` + global variable works good. At least, it covers our case. I can give a bit more context, probably you will find a bit more elegant solution that could be implemented in kedro: as I have already described in [#883](https://github.com/kedro-org/kedro-viz/issues/883) we have a group of data scientists which life we try to make a bit easier. They do many experiments during every single day, staying more or less inside the same pipeline. So, we defined something like a pipeline template where they have some freedom to play with the parameters or to decide which nodes they need or doesn't need and in what order. Still, we do not give them the full control to define pipelines through yaml files (like it was represented [here](https://github.com/kedro-org/kedro/issues/650) for example). We assume that if they really need to develop something new, then they should really know what exactly they require, so it can be done using python. To speed up their experiments, we would like to give them an opportunity to run several experiments simultaneously from different folders: they define a parameter file, put it in any folder they like, copy it as many times as they want and can run everything from different terminal session modifying their parameters manually or by using `--params` option. Related data for any particular run is written to the folder with the related parameter file, and all experiments are tracked by mlflow (many thanks to [Galileo-Galilei](https://github.com/Galileo-Galilei) and [his awesome extension](https://github.com/Galileo-Galilei/kedro-mlflow)). As I understand, currently, kedro suppots only single run at one time because otherwise we would have collisions inside data folder. Our next step is to write an extension that could do something like grid-search through some set of possible parameters, so we could see it in MLFlow as a nested run. All local data produced by pipeline we leave as something auxiliary if, for example, they see that something goes wrong, and they need to investigate it via jupyter notebooks. Seems it is pretty full description of our particular case. I didn't mention only that for defining proper paths and inputs/outputs to the nodes we use slightly rewritten version of [kedro-wings](https://github.com/tamsanh/kedro-wings), but I am not sure if it is important. Hopefully, this description will be useful for you in furthe development.

2. Yes, that is true. As I use templated config loader and `globals.yml` I can get around it by simply defining `path: path/to/parameter/file` inside globals.yml for the used environment. But it is slightly inconvenient when you run a pipeline itself by using one command (`kedro run --params=path:path/to/parameter/file`) and should manually modify project files to see the visualization in kedro viz instead of write something like `kedro viz --params=path:path/to/parameter/file` where we just change `run` option to `viz`.

3. I can confirm that with these modifications we can see the visualization of the correct parameter file. I tried to modify kedro viz files inside my project (which is definitely not good, but still) and tested it a bit.

MerelTheisenQB:
Another Kedro user reached out to me about this functionality as well, specifically for Experiment Tracking:

"I am running a node `ABC` by passing params `--params user:poorni,env:dev` and DS in my team runs a node `ABC` by passing params `--params user:ds,env:dev`. We are storing the the resultant dataset of node `ABC` in two different locations.

(our catalog entries are like below)
```
ABC:
   filepath: blabla/{user}/{env}/abc.csv
```   
Assume this `abc` dataset is having metrics, I want to see the results which DS has stored in abc dataset. So when i am running kedro viz it shows the results of dataset which got stored under `--params user:poorni,env:dev`  if i wanted to see DS results i need to change in globals.yml file

if i have an option through kedro viz only we can just pass the params `kedro viz --params user:poorni,env:dev`"

In summary, currently she has to change her `globals.yml` in order to see the results for a run where specific parameters were used.

AntonyMilneQB:
@IanVlasov thanks very much for all the extra information, it's very helpful! I'm definitely happy to accept this change in principle then. I'll review now in terms of the actual implementation.

Unrelated to this PR, but re: "only single run at one time because otherwise we would have collisions inside data folder". This is an interesting topic that has definitely been considered by other people before so worth searching around in case you find other ways of doing it that might help you. e.g. https://github.com/kedro-org/kedro/issues/1303. One very simple approach if you're worried about collisions between multiple users running the pipeline is to make users save to different locations, e.g. set `user: ian` in your `conf/local/globals.yml` file and then use `${user}` in your dataset filepaths to make sure that your output doesn't overwrite anyone else's. [Versioning datasets](https://kedro.readthedocs.io/en/stable/data/kedro_io.html#versioning-using-the-yaml-api) might also help.

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -57,7 +61,10 @@ def load_data(
         from kedro_viz.integrations.kedro.sqlite_store import SQLiteStore
 
         with KedroSession.create(
-            project_path=project_path, env=env, save_on_close=False
+            project_path=project_path,
+            env=env,
+            save_on_close=False,
+            extra_params=extra_params,
```
AntonyMilneQB, line 67: We should also make the same change in the two other occurrences of `KedroSession.create` further down this file.

AntonyMilneQB (commented):
This looks great, thank you for your work on it! I've left a few small comments and will be very happy to approve and get this nice new feature in after those are dealt with 🎉

AntonyMilneQB (approved):
Thank you @IanVlasov! I think there's one more occurrence of `KedroSession.create` that needs updating in `data_loader.py` but after that's done please feel free to merge 👍

AntonyMilneQB:
Ahh yes I see it now. Sorry, I was just misreading things on github! Feel free to merge when ready 🙂 And thank you again for this!

rashidakanchwala (approved):
thanks @IanVlasov! I will merge this on your behalf!

PR #887, opened 2022-06-06
Feature: show matplotlib dataset pngs

### LINKED ISSUE #783: Show Matplotlib dataset pngs in the metadata panel
## Description

`PlotlyDataSet` is super cool but not the only way that data scientist make plots. Kedro also supports `MatplotlibWriter` which saves matplotlib plots (or ones from any other library based on matplotlib like seaborn) to a png. It would be awesome if when I click on a matplotlib dataset in kedro-viz it opened up a thumbnail of the image on the metadata side panel, which I could then click to make full screen. Very similar to how plotly datasets are handled, only it's just a static png file rather than an interactive plot.

To me this sounds quite straightforward to implement and would be a really nice feature, but very interested to hear what others think. @rashidakanchwala?

### Slight complication
`MatplotlibWriter` is a bit weird in that it can write to _multiple_ png files. In this case the `filepath` given in the catalog entry will refer to a directory that contains all the png files. Ideally kedro-viz would also be able to handle this case, but I think we would get most of the value just from handling the simple case where `filepath` is a single png file. So I would forget about this for the MVP anyway.

### Possible concern
Would this affect performance in some way if I have a pipeline with lots of plots? I don't know whether kedro-viz would preload all of the png files or whether a png file would only be requested when the user clicks on the dataset and the metadata panel pops up?

tynandebold:
## Description

Resolves #783.

Massive thank you to @rashidakanchwala for helping so much with this one.

## Development notes

There are a lot of changes here, and that's because I've renamed several files to adopt a more generic approach to the modal formerly named `PlotlyModal`. That's now simply `MetadataModal`, since we may want to put other things in there.

## QA notes

I've added a `Matplotlib Image` node to the demo project, so you'll be able to see the implementation if you open Gitpod and take a look at the preview link.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/887"></a>


AntonyMilneQB:
I just tried this out on gitpod but it got stuck installing requirements with
```
ERROR: Cannot install matplotlib==3.5.0 and matplotlib~=3.5.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested matplotlib~=3.5.2
    The user requested matplotlib==3.5.0
```

So I pushed [3c99820](https://github.com/kedro-org/kedro-viz/pull/887/commits/3c998200f48f4cb058c2eb2cb4646b4854702f0c) which should fix this.

As far as I can tell there's no reason that matplotlib should be a Python requirement of the kedro-viz package, is there? In fact I'm not sure whether we really need these either... 🤔 But that's a separate issue.
https://github.com/kedro-org/kedro-viz/blob/895f89450dbd35fdf2d96f723e5a154bd332dded/package/requirements.txt#L8-L9

package/kedro_viz/models/graph.py
```diff
@@ -585,6 +597,20 @@ def __post_init__(self, data_node: DataNode):
...
+            )
+
+            dataset = cast(MatplotlibWriter, dataset)
+            if not dataset._exists():
+                return
+
+            load_path = get_filepath_str(dataset._get_load_path(), dataset._protocol)
+            with open(load_path, "rb") as img_file:
+                b64_string = base64.b64encode(img_file.read())
+                self.image = b64_string.decode("utf-8")
```
AntonyMilneQB, line 612:
```suggestion
            self.image = b64_string.decode("utf-8")
```
Since `img_file` is not used here, it's neatest to leave it outside the context manager (`with ...`).

package/kedro_viz/models/graph.py
```diff
@@ -585,6 +597,20 @@ def __post_init__(self, data_node: DataNode):
             with dataset._fs.open(load_path, **dataset._fs_open_args_load) as fs_file:
                 self.plot = json.load(fs_file)
 
+        if data_node.is_image_node():
+            from kedro.extras.datasets.matplotlib.matplotlib_writer import (
+                MatplotlibWriter,
+            )
+
+            dataset = cast(MatplotlibWriter, dataset)
+            if not dataset._exists():
```
AntonyMilneQB, line 606: Minor suggestion, but it's better practice to use the public methods where they exist. Similarly for the other occurrences in this file which this was copied from.
```suggestion
            if not dataset.exists():

```

package/tests/test_models/test_graph/test_graph_nodes.py
```diff
@@ -384,9 +384,8 @@ def test_partitioned_data_node_metadata(self):
         data_node_metadata = DataNodeMetadata(data_node=data_node)
         assert data_node_metadata.filepath == "partitioned/"
 
-    @patch("builtins.__import__", side_effect=import_mock)
```
AntonyMilneQB, line 387: Why was this removed? Seems to me that we should be able to run the tests without plotly installed (related to the point of removing plotly from requirements.txt).

AntonyMilneQB (commented):
Github is doing something weird so submitting this before it loses my comments... review in progress - will edit it!

package/tests/test_models/test_graph/test_graph_nodes.py
```diff
@@ -426,10 +426,30 @@ def test_plotly_json_dataset_node_metadata(self, patched_json_load):
...
         plotly_json_dataset_node = MagicMock()
         plotly_json_dataset_node.is_plot_node.return_value = True
-        plotly_json_dataset_node.is_metric_node.return_value = False
+        plotly_json_dataset_node.is_image_node.return_value = False
+        plotly_json_dataset_node.is_tracking_node.return_value = False
         plotly_node_metadata = DataNodeMetadata(data_node=plotly_json_dataset_node)
         assert plotly_node_metadata.plot == mock_plot_data
 
+    @patch("base64.b64encode")
+    def test_image_data_node_metadata(self, patched_base64):
```
AntonyMilneQB, line 435: Great job writing a test for this function, which is definitely not easy to do! But even better would be if it's possible to patch `img_file.read`'s return value instead? Maybe there's even some real (very short) image string we could put in here to make it a bit more a real test?

AntonyMilneQB (approved):
This is great and all seems to work perfectly!

I will check the thing about requirements. Other than that, I just have minor comments and suggestions really - let me know if you want to discuss anything further!

package/kedro_viz/models/graph.py
tynandebold: Noted.

src/components/metadata-modal/metadata-modal.js
```diff
@@ -33,19 +35,33 @@ const PlotlyModal = ({ metadata, onToggle, visible }) => {
...
-        view="modal"
-      />
+      {metadata?.plot ? (
+        <PlotlyChart
+          data={metadata.plot.data}
+          layout={metadata.plot.layout}
+          view="modal"
+        />
+      ) : (
+        <div className="pipeline-matplolib-chart">
```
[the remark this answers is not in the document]
tynandebold: Good catch!

src/components/metadata-modal/metadata-modal.scss
```diff
@@ -30,7 +30,8 @@
   height: 100%; /* Full height (cover the whole page) */
   background-color: var(--color-bg-plot);
 
-  .pipeline-plotly-chart {
+  .pipeline-plotly-chart,
+  .pipeline-matplolib-chart {
```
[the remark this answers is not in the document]
tynandebold: Fixed.

package/tests/test_models/test_graph/test_graph_nodes.py
tynandebold: I'm not sure. @rashidakanchwala? Should I add it back in?

rashidakanchwala: so I removed it, because it wasn't actually being used. :S

src/components/metadata-modal/metadata-modal.js
```diff
@@ -33,19 +35,33 @@ const PlotlyModal = ({ metadata, onToggle, visible }) => {
...
+      )}
       <div className="pipeline-plot-modal__bottom">
         <button
           className="pipeline-plot-modal__collapse-plot"
           onClick={onCollapsePlotClick}
         >
           <CollapseIcon className="pipeline-plot-modal__collapse-plot-icon"></CollapseIcon>
           <span className="pipeline-plot-modal__collapse-plot-text">
-            Collapse Plotly Visualization
+            {metadata?.plot
```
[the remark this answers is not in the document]
tynandebold: Is it possible that a node with a `dataset_type` of plotly or matplotlib won't contain and `image` or a `plot`? If it is, I think we'll still need to check for either `image` or `plot`. In fact, this did happen to me yesterday. I added in my new node but didn't do `kedro run`, so the image wasn't there but the node on the flowchart was present, as was the metadata panel.

I agree with everything else. The logic isn't identical and should be. I'll update it.

package/tests/test_models/test_graph/test_graph_nodes.py
tynandebold: Surely. I've added a very small base64 PNG to the test, taken from [here](https://stackoverflow.com/a/36610159).

src/components/metadata-modal/metadata-modal.js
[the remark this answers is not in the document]
AntonyMilneQB: Ah yes, that is a very good point. If you haven't done a `kedro run` yet then `plot`/`image` won't be defined.

In this case I guess there's two bits of logic we need:
* something to check the `dataset_type` (probably using `shortTypeMapping`? Something like `switch shortTypeMapping[dataset_type]` or whatever)
* something to check whether the dataset actually has `image` or `plot` defined

AntonyMilneQB:
OK, tests should all be fixed now 🤞 

In response to [my previous question about requirements](https://github.com/kedro-org/kedro-viz/pull/887#issuecomment-1148279327), `pandas` and `plotly` should remain as core kedro-viz requirements because they're needed for the plot shown in the metadata panel for `MetricsDataSet`. See https://github.com/kedro-org/kedro-viz/pull/554#discussion_r712295761. `matplotlib` shouldn't be there though. The right way to get the tests working without matplotlib installed is using the `import_mock`.

src/components/metadata-modal/metadata-modal.js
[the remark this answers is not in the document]
AntonyMilneQB: Also, curious: is there any reason these explicitly cast to `Boolean`? Seems like some things in metadata.js are and some aren't:
```
  const hasCode = Boolean(metadata?.code);
  const isTranscoded = Boolean(metadata?.originalType);
  const showCodePanel = visible && visibleCode && hasCode;
  const showCodeSwitch = hasCode;
```

rashidakanchwala:
> OK, tests should all be fixed now 🤞
> 
> In response to [my previous question about requirements](https://github.com/kedro-org/kedro-viz/pull/887#issuecomment-1148279327), `pandas` and `plotly` should remain as core kedro-viz requirements because they're needed for the plot shown in the metadata panel for `MetricsDataSet`. See [#554 (comment)](https://github.com/kedro-org/kedro-viz/pull/554#discussion_r712295761). `matplotlib` shouldn't be there though. The right way to get the tests working without matplotlib installed is using the `import_mock`.

I am confused. react-plotly.js is what is used for the plot shown in the FE.

src/components/metadata-modal/metadata-modal.js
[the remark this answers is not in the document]
tynandebold: Good catch. Added those two to this, as suggested: 

```jsx
if (!visible.metadataModal || (!metadata?.plot && !metadata?.image)) {
```

The first two are cast as such because they're being converted from strings. The last two don't need it, since they're created from booleans they'll be that type, too.

rashidakanchwala (approved):
great work on your python PR :) looks great! 

just nit on the json file having a newline. 

also, please if you get a chance to update the release docs with the missing stuff shared on slack :)

PR #896, opened 2022-06-09
Alter tooltip appearance timing

### LINKED ISSUE #889: Alter the timing of the tooltip's appearance
## Description

All tooltips only should appear after 2 seconds, not instantly on hover as it is right now.

Screenshot 2022-05-23 at 10 45 1
 
## Acceptance criteria

- User only sees the tooltip after hovering over a certain icon for 2 seconds.
  - This includes tooltips on the flowchart as well as the ones that appear when hovering over the icons.
  - If 2 seconds feels too long or too short, we can adjust accordingly to where it's comfortable. 

## Checklist

cvetankanechevska:
## Description
Resolves #889 

## Development notes
Added transition-delay property to IconButton & Tooltip components for getting onHover effect with 2s delay.
## QA notes
Here is a short demo video:

https://user-images.githubusercontent.com/88193161/172806452-cc6ea224-5fae-46ff-a184-6ffe235b461d.mov

Q: Hovering over icons in the flowchart we don't have a cursor: "pointer", and some of the nodes are clickable. Should we add that?

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/896"></a>


src/components/ui/icon-button/icon-button.scss
```diff
@@ -52,12 +52,14 @@
...
   left: 50%;
-  display: none;
   margin-top: -1.2em;
   margin-left: 1.6em;
   padding: 0.5em 0.7em;
   font-size: 1.4em;
   white-space: nowrap;
+  display: inline-block;
+  opacity: 0;
+  transition: opacity ease 0.1s;
```
tynandebold, line 62: I think you can add the delay to this line and remove it from line #76.

```suggestion
  transition: opacity ease 0.1s 2s;
```

src/components/ui/tooltip/tooltip.scss
```diff
@@ -19,6 +19,7 @@ $triangle-size: 10px;
   &--visible {
     visibility: visible;
     opacity: 1;
+    transition-delay: 2s;
```
tynandebold, line 22: I think we'll need to add the delay to the [base class](https://github.com/kedro-org/kedro-viz/blob/dc068db7e43d5abe84618085ea567cca6e71008e/src/components/ui/tooltip/tooltip.scss#L16). In testing, if it's only added to the modified class, there's some weirdness when the tooltip text changes and moves to other nodes. This is because the toolip itself is only ever on DOM element on the page and isn't dynamically created.

tynandebold (changes requested):
Fantastic stuff here, and congratulations on the first PR in Viz 🥳 

I've added a couple small comments here, the first more of a suggestion and the second something I think we'll need to change.

Lastly, I find the `2s` duration to be much too long. It seems to take an age for the tooltips to appear. I'd decrease the number to `0.5s` or thereabouts. Curious to hear how it feels to others, especially @GabrielComymQB and @Mackay031.

src/components/ui/icon-button/icon-button.scss
GabrielComymQB: Hi, as I understand, we don't need any delay on MouseLeave
The delay is only onMouseOver, and when we leave it disappears instantly. That's the idea :)

cvetankanechevska: Yes. And for your previous comment, I think you need to launch the server also using: `make run`

GabrielComymQB: Hi, I've just loaded Gitpod and I don't think I can notice any changes from last week. But yeah, let me know and I can have a try from my side when it's time :)

src/components/ui/tooltip/tooltip.scss
tynandebold: @cvetankanechevska I think you still need to make this change. This delay needs to go on the base class of `.pipeline-tooltip`, otherwise the user could see a weird in between state of the tooltip, as seen here:

Screen Shot 2022-06-13 at 13 14 21

GabrielComymQB (dismissed):

tynandebold (approved):
Looks good to me!

Perhaps the delay could be reduced a little bit. Still feels a little too long, though if everyone is happy it's A-ok with me.

cvetankanechevska:
> Looks good to me!
> 
> Perhaps the delay could be reduced a little bit. Still feels a little too long, though if everyone is happy it's A-ok with me.

Just pushed my last change with `0.5s` delay.

rashidakanchwala (approved):
LGTM, thanks @cvetankanechevska

tynandebold:
> > Looks good to me!
> > Perhaps the delay could be reduced a little bit. Still feels a little too long, though if everyone is happy it's A-ok with me.
> 
> Just pushed my last change with `0.5s` delay.

Let's see what Gabriel and Mackay think, too. Gabriel approved at `1s`, so I'd guess he's fine with that.

GabrielComymQB:
Hi, I had approved 1 second but now this doesn't feel right to me. And it's not related to timing. I think there's a bit of the logic of the states that we have to nail. I'm happy to talk about it during or after standup and illustrate what I mean.

GabrielComymQB (approved):
I'm reviewing and approving the time change. If we can please address [my comments ](https://github.com/kedro-org/kedro-viz/commit/595cc309ccb88cdb78e3337cc0af0b3032531129#commitcomment-76064403)in a separate ticket that would be awesome, many thanks!

cvetankanechevska:
> I'm reviewing and approving the time change. If we can please address [my comments ](https://github.com/kedro-org/kedro-viz/commit/595cc309ccb88cdb78e3337cc0af0b3032531129#commitcomment-76064403)in a separate ticket that would be awesome, many thanks!

Awesome, thanks @GabrielComymQB I will create a new ticket for this in Backlog.

PR #898, opened 2022-06-09
[Kedro-viz-890]/run-export-modal

### LINKED ISSUE #890: Run Export shows no modal
## Description

On the experiment tracking page, after clicking export, users have no feedback to their action. This does not follow the proposed design. We should make use of the modal interaction/selection as seen on the reference link. This is the same pattern we use on the pipeline side, when users try to download the pipeline graph.

Screenshot 2022-05-24 at 11 40 1

## Acceptance criteria

- After clicking on the export icon, user is presented with our standard “export modal” 
- User can make the selection between **Export all and close** and **Cancel**.
  - If the user chooses **Export all and close**, we should show the same UX as built in #875.
- After clicking on any of the options the modal closes.
- Another way of leaving the modal is to click anywhere in the background outside the modal.

image 1

Reference link: [https://zpl.io/2G3Gvp3](https://zpl.io/2G3Gvp3)

## Checklist

Huongg:
## Description

Resolves #890: To create a modal to notify users when they download a run in the experiment tracking

## Development notes
ezgif com-gif-maker

## QA notes
There are tests to ensure both buttons work as expected

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/898"></a>


src/components/experiment-tracking/run-export-modal.js/run-export-modal.test.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import RunExportModal from './index';
+import Adapter from 'enzyme-adapter-react-16';
+import { configure, mount, shallow } from 'enzyme';
+import { render, screen } from '@testing-library/react';
+
+// to help find text which is made by multiple HTML elements
```
Huongg, line 7: Is there a place I can put this util function? I think we can re-use it somewhere else later

tynandebold:
Congratulations on your first PR in Viz!!! 🎆 

Just one observation before a full review: I believe the entire button background should turn green on the `success` state, as it does in this modal:

Screen Shot 2022-06-10 at 09 47 42

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
```diff
@@ -70,19 +62,13 @@ export const ExperimentPrimaryToolbar = ({
...
-        <IconButton
-          ariaLabel="Export graph as SVG or PNG"
-          className={'pipeline-menu-button--export'}
-          icon={ExportIcon}
-          labelText="Export run data"
-        />
-      </CSVLink>
+      <IconButton
+        ariaLabel="Export graph as SVG or PNG"
+        className={'pipeline-menu-button--export'}
```
rashidakanchwala, line 67:
```suggestion
        className={'pipeline-menu-button--export-runs'}
```

rashidakanchwala: this is just from heap analytics perspective as Export button in Pipeline visualisation also has the exact same class name

src/components/experiment-tracking/run-export-modal.js/run-export-modal.test.js
rashidakanchwala: i was thinking maybe in  src/components/ui/button we could add a success button?

Huongg: yeah i was thinking the same, what do you think of naming it `button-success`. It sounds weird but if we name it this way, all the button components display next to each other.
eg 
```
- button
- button-success
```

what do you think @rashidakanchwala @tynandebold ?

tynandebold: We should probably generalize it more than that. At some point we'll probably need to supply a `fail` state, too. So maybe we do something like `button-message` or similar.

Also to your original question about the utility function: you could just `export const findTextWithTags` for now, because at the moment I don't think we need it anywhere else, so it may be premature to move it elsewhere. If you want to move it though, somewhere in one of the [utils files](https://github.com/kedro-org/kedro-viz/tree/main/src/utils) would be best, but you'd probably want to create a new file, like `test-utils.js`, so maybe we hold off for now.

src/components/experiment-tracking/experiment-primary-toolbar/experiment-primary-toolbar.js
```diff
@@ -70,19 +62,13 @@ export const ExperimentPrimaryToolbar = ({
...
-      >
-        <IconButton
-          ariaLabel="Export graph as SVG or PNG"
-          className={'pipeline-menu-button--export'}
-          icon={ExportIcon}
-          labelText="Export run data"
-        />
-      </CSVLink>
+      <IconButton
+        ariaLabel="Export graph as SVG or PNG"
```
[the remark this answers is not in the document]
Huongg: updated, thanks @rashidakanchwala. good spot 😄

rashidakanchwala (approved):
thanks @Huongg -- LGTM :)

src/components/experiment-tracking/details/details.js
```diff
@@ -40,13 +44,26 @@ const Details = ({
...
+        )}
+        {showRunExportModal && (
+          <RunExportModal
+            theme={theme}
+            visible={showRunExportModal}
+            setShowRunExportModal={setShowRunExportModal}
+            runMetadata={runMetadata}
+            runTrackingData={runTrackingData}
+          />
+        )}
```
tynandebold, line 65: I asked about this yesterday when you were getting a sense check on this approach, and now looking at this work locally it seems that this _does_ change the behavior of the modals animating in and out.

If you look on the `main` branch you'll see that when they enter there's a fade in and up, and when they disappear it's down and out. We'll need to keep that UX here, too, which means we'll have to find a way not to do this `showRunExportModal &&` or do that and still animate.

Huongg: hey @tynandebold such a good spot, you're right. We should not remove the animation, sorry i didnt realise it yesterday. Actually, i can just use the existing props `visible` to apply the context, rather than check with props to render a specific component 

Let me know what you think about the latest changes? Happy to discuss further

src/components/experiment-tracking/details/details.js
```diff
@@ -40,13 +44,22 @@ const Details = ({
...
+          setShowRunDetailsModal={setShowRunDetailsModal}
+          theme={theme}
+          visible={showRunDetailsModal}
+        />
+        <RunExportModal
+          theme={theme}
+          visible={showRunExportModal}
+          setShowRunExportModal={setShowRunExportModal}
+          runMetadata={runMetadata}
+          runTrackingData={runTrackingData}
```
tynandebold, line 60: Nit (coding style): I try to advise us to alphabetically sort component props, and also when destructuring and even CSS styles, where applicable. It makes readability easier when you know there's an order to things. There's a great [VSCode extension](https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines) that makes this easy.

src/components/experiment-tracking/run-export-modal.js/run-export-modal.js
```diff
@@ -0,0 +1,71 @@
...
+          <CSVLink
+            data={exportData}
+            asyncOnClick={true}
+            onClick={updateExportData}
+            filename="run-data.csv"
+          >
+            <Button
+              onClick={handleClick}
+              mode={isSuccessful ? 'success' : 'primary'}
+            >
```
tynandebold, line 55: Referencing the design, I think we want the `small` size button here. There's a `size` prop you can use.

tynandebold (approved):
Excellent changes. The animations are back! Thank you for that. I'm approving now with some comments and a couple of thoughts:

1. When I toggle through the `RunExportModal` with my keyboard and export the data the modal animation doesn't fire nor does the modal close. It does work on the `RunDetailsModal` though.
2. I like the use of context, I just hope that if we need to react to a failed API request and then show a failed state in the button instead of the success we'll be able to. We will, right?

src/components/experiment-tracking/run-export-modal.js/run-export-modal.js
Huongg: done, thank you

Huongg:
> Excellent changes. The animations are back! Thank you for that. I'm approving now with some comments and a couple of thoughts:
> 
> 1. When I toggle through the `RunExportModal` with my keyboard and export the data the modal animation doesn't fire nor does the modal close. It does work on the `RunDetailsModal` though.
> 2. I like the use of context, I just hope that if we need to react to a failed API request and then show a failed state in the button instead of the success we'll be able to. We will, right?

hey to answer your question as below
1. yeah you're right it doesn't have the animation at the moment when using the keyboard. I guess because it was wrapped around by a CSV link. I'll think of a way to fix it but for now let's merge this ticket and I'll open a new bug for that? 
2. yeah definitely, we just need to introduce a new state in the context to handle the failed state

tynandebold:
> > Excellent changes. The animations are back! Thank you for that. I'm approving now with some comments and a couple of thoughts:
> > 
> > 1. When I toggle through the `RunExportModal` with my keyboard and export the data the modal animation doesn't fire nor does the modal close. It does work on the `RunDetailsModal` though.
> > 2. I like the use of context, I just hope that if we need to react to a failed API request and then show a failed state in the button instead of the success we'll be able to. We will, right?
> 
> hey to answer your question as below
> 
> 1. yeah you're right it doesn't have the animation at the moment when using the keyboard. I guess because it was wrapped around by a CSV link. I'll think of a way to fix it but for now let's merge this ticket and I'll open a new bug for that?
> 2. yeah definitely, we just need to introduce a new state in the context to handle the failed state

Perfect ☺️ 

Yes, let's create a new ticket to track that keyboard bug. And good to know about the failed state. Thanks for confirming. Merge this whenever you can!

PR #905, opened 2022-06-13
Refactor CircleCi  

rashidakanchwala:
## Description

This is the first part of the kedro-viz circleCI refactor. In this PR we refactor the 'build' workflow on circleCI :- 

- we split javascript and python tests to different jobs so they can run in parallel 
- we also split python unit and e2e test and python lint to different jobs so they can run in parallel

Below is what our old circleCI config looked like for the build workflow:-

Screenshot 2022-06-15 at 09 40 37

Below is the refactor:-

Screenshot 2022-06-15 at 09 42 13

Next part of the refactor includes :- 
- caching/persist to workspace for dependencies and builds 
- conditional logic to do only certain tasks for front-end and back-end 
- also add merge_to_main/daily/ and other workflows

Thank you @AntonyMilneQB for all the guidance on how to approach and solve this :) 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/905"></a>


.circleci/config.yml
```diff
@@ -1,33 +1,26 @@
...
-    working_directory: ~/repo
-  circleci_machine_310:
+  docker:
+    parameters:
+      python_version:
+        type: string
     docker:
-      - image: cimg/python:3.10-node
+        - image: cimg/python:<<parameters.python_version>>-node
     working_directory: ~/repo
```
AntonyMilneQB, line 17: Is this `working_directory` needed? What happens if you remove it?

.circleci/config.yml
```diff
@@ -1,33 +1,26 @@
...
-  demo_executor:
-    docker:
-      - image: cimg/python:3.8
 
 commands:
   setup_python_env:
     description: Setup Python environment
```
AntonyMilneQB, line 21: On the kedro we do the following:
* no `description` fields
* make sure any `run` step has a `name` that describes what it does

This overall looks clearer on CircleCI. Please can you do the same here?

.circleci/config.yml
```diff
@@ -1,33 +1,26 @@
...
     steps:
+      - checkout
       - run:
           name: Install Python
           command: python --version
```
AntonyMilneQB, line 26:
```suggestion
```

AntonyMilneQB: This doesn't do anything except show the Python version, which is already given in a following step.

AntonyMilneQB: Can't make a comment there, but looks like we can remove L31 also? Since `-r requirements.txt` is called from `test_requirements.txt` already.

AntonyMilneQB: Would also be nice if we could remove the `cd ..` at the end 🙂

.circleci/config.yml
```diff
@@ -1,33 +1,26 @@
...
     working_directory: ~/repo
-  demo_executor:
-    docker:
-      - image: cimg/python:3.8
 
 commands:
   setup_python_env:
     description: Setup Python environment
     steps:
+      - checkout
```
AntonyMilneQB, line 23:
```suggestion
```
And then just make sure it's in all the jobs.

.circleci/config.yml
```diff
@@ -217,121 +173,133 @@ commands:
...
+    parameters:
+      python_version:
+        type: string
+    executor:
+      name: docker
+      python_version: <<parameters.python_version>>
+    steps:
+      - checkout 
+      - setup_python_env
+      - install_node_dependencies
```
AntonyMilneQB, line 232: Do we need this?

.circleci/config.yml
```diff
@@ -78,6 +71,18 @@ commands:
...
+      - run:
+          name: Run Python tests
+          command: make pytest
+
+  run_e2e_tests:
+    steps:
+      - run:
+          name: Run all end to end tests
+          command: make e2e-tests
+
```
AntonyMilneQB, line 85: These can go straight into the relevant `job`. No need to define it as a separate command.

.circleci/config.yml
```diff
@@ -217,121 +173,133 @@ commands:
...
     parameters:
       python_version:
         type: string
+    executor:
+      name: docker
+      python_version: <<parameters.python_version>>
+    steps:
+      - setup_python_env
+      - install_node_dependencies
+      - run_e2e_tests
```
AntonyMilneQB, line 207: After removing the `run_e2e_tests` commands:
```suggestion
      - run:
          name: Run Python tests
          command: make pytest
```

.circleci/config.yml
```diff
@@ -217,121 +173,133 @@ commands:
...
+  e2e_tests:
     parameters:
       python_version:
         type: string
+    executor:
+      name: docker
+      python_version: <<parameters.python_version>>
+    steps:
+      - setup_python_env
+      - install_node_dependencies
```
AntonyMilneQB, line 206: Do we need this? Is it so that the following`build` stage runs ok?

.circleci/config.yml
```diff
@@ -217,121 +173,133 @@ commands:
...
+    executor: win/default
     steps:
-      - checkout
-      - win_build:
+      - checkout 
+      - win_setup_conda:
           python_version: <<parameters.python_version>>
+      - win_setup_requirements
+      - install_node_dependencies
+      - run_win_python_tests
```
AntonyMilneQB, line 246: Similarly we can just write this one inline, delete the definition of `run_win_python_tests` and reduce the number of redirections in the file.
```suggestion
      - run:
          name: Run Python tests on Windows
          command:
            conda activate kedro-viz; make pytest
```

.circleci/config.yml
AntonyMilneQB: Similarly for several of the other commands actually. Basic rule: if a `command` is just one `run` step long and only used in one `job`, there's no point having a `command` for it. Just put it directly into the `job` instead.

.circleci/config.yml
```diff
@@ -217,121 +173,133 @@ commands:
...
-  win_build_310:
-    executor: win/default
+  build_package: 
     parameters:
       python_version:
         type: string
-    working_directory: ~/repo
+    executor:
+        name: docker
+        python_version: <<parameters.python_version>>
```
AntonyMilneQB, line 283:
```suggestion
        python_version: "3.8"
```
Since this is only used with one Python version, let's not bother with the matrix notation for it. Similarly for any other jobs that are only invoked with one Python version.

.circleci/config.yml
```diff
@@ -217,121 +173,133 @@ commands:
...
+      - test_lib_import
+      - run_eslint
+      - run_javascript_tests
 
-  win_build_310:
-    executor: win/default
+  build_package: 
     parameters:
       python_version:
         type: string
```
AntonyMilneQB, line 280:
```suggestion
```

.circleci/config.yml
```diff
@@ -366,63 +339,71 @@ release_filter: &release_filter
...
+            parameters:
+              python_version: ["3.7", "3.8", "3.9", "3.10"]
+      - javascript_lint_and_tests:
+          matrix:
+              parameters:
+                python_version: ["3.8"]
+      - build_package:
+          matrix:
+            parameters:
+              python_version: ["3.8"]
```
AntonyMilneQB, line 372:
```suggestion
      - build_package
```

AntonyMilneQB (commented):
This is a great start! Just made some initial comments for now - in general I think we can simplify things further by reducing the number of "layers of indirection".

Other things to ponder - will explain tomorrow:
* where should `checkout` go?
* can we remove whole `win_setup_requirements` job?
* where should `build` go?

.circleci/config.yml
```diff
@@ -49,83 +34,20 @@ commands:
...
-          command: make e2e-tests
-
-  cleanup:
-    steps:
-      - run:
-          name: Cleanup
-          command: make clean
-
-  build_npm_package:
+  build_package:
```
AntonyMilneQB, line 39: Let's try to give this and `package_app` names which make it clearer what the difference is. Maybe `npm_build` and `python_package`?

.circleci/config.yml
```diff
@@ -49,83 +34,20 @@ commands:
...
     steps:
       - run:
           name: Build application
```
AntonyMilneQB, line 42: "Build Node application" or something? NPM application? I'm not sure what the right terminology is.

.circleci/config.yml
```diff
@@ -49,83 +34,20 @@ commands:
...
-          command: npm run build
+          command: make build
 
   package_app:
     steps:
       - run:
-          name: Package app
+          name: Build package
```
AntonyMilneQB, line 48: "Make Python package"

.circleci/config.yml
```diff
@@ -49,83 +34,20 @@ commands:
...
           command: make package
-
+      
   deploy_demo:
```
AntonyMilneQB, line 51: This should all be moved to a `job`.

.circleci/config.yml
```diff
@@ -217,123 +102,143 @@ commands:
             python -V
```
AntonyMilneQB, line 102: Same changes here as in the Linux version :)

.circleci/config.yml
```diff
@@ -342,21 +247,28 @@ jobs:
...
       - checkout
       - setup_python_env
       - install_node_dependencies
       - run:
           name: Install twine
           command: python -m pip install -U twine
-      - run:
-          name: Build Python package
-          command: make package
+      - package_app
```
AntonyMilneQB, line 260: Let's put this above the "Install twine" step

AntonyMilneQB (approved):
Amazing work! Just a few small comments but this is a HUGE improvement on what we currently have. Thank you!

The release job actually looks ok to me, so I think the only thing we need to add back in is the `deploy_demo` and consider whether we want a daily scheduled pipeline (not workflow like it used to be).

src/components/experiment-tracking/run-metadata/run-metadata.js
```diff
@@ -132,13 +132,13 @@ const RunMetadata = ({
...
                             ariaLive="polite"
-                            className="pipeline-menu-button--labels"
+                            className="pipeline-menu-button--labels__kebab"
                             icon={KebabIcon}
                           />
                         </HiddenMenu>
                         <IconButton
                           ariaLive="polite"
-                          className="pipeline-menu-button--labels"
+                          className="pipeline-menu-button--labels__close"
```
tynandebold, line 141: Same with this one.

.circleci/config.yml
tynandebold: Maybe `Build React application`?

tynandebold (approved):
Looks and works great! Amazing job. 

Only a couple small things from me; no blockers.

AntonyMilneQB:
⚠️ We shouldn't merge this until #917 goes in because the release jobs don't exist here!

PR #934, opened 2022-06-21
Toggle pretty name only on apply and close

### LINKED ISSUE #877: Settings should be applied to PrettyName only after Apply and Close clicked.
## Description

The below bug is introduced by the 'Apply and Close' feature on settings panel.

## Expected Result
Settings like 'Pretty Name' should be applied only once 'Apply and Close' button is clicked. 

## Actual Result
Currently in the settings panel, settings like 'Pretty Name' gets applied automatically when you toggle the settings.

tynandebold:

## Description

Fixes #877 

## Development notes

A state variable is created, and based on that `boolean`, the pretty name will be on/off when **Apply and close** is clicked.

## QA notes

View the Gitpod link, open the settings panel, and toggle **Pretty Name**.  Nothing should happen until you click on **Apply and close**, just liked the other two toggles.

Also note I removed the `(page reload required)` language, since we're not auto reloading the page when the apply button is clicked.

Further, I noticed that some of the elements didn't match the [design](https://zpl.io/2Z3WqYl), so I fixed that as well.

Before:

image

After:

image

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/934"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -8,7 +8,7 @@ Please follow the established format:
 
 ## Bug fixes and other changes
 
-- General design-debt fixes. (#933)
+- General design-debt fixes. (#933, #934)
 
 # Release 4.7.0
 
--- a/src/components/settings-modal/settings-modal.js
+++ b/src/components/settings-modal/settings-modal.js
@@ -9,8 +9,8 @@ import { getFlagsState } from '../../utils/flags';
 import SettingsModalRow from './settings-modal-row';
 import { settings as settingsConfig } from '../../config';
 
-import Modal from '../ui/modal';
 import Button from '../ui/button';
+import Modal from '../ui/modal';
 
 import './settings-modal.css';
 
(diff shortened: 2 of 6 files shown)


Huongg (approved):
hey @tynandebold  everything looks good to me 😄🕺 
Just one point to consider in the future, maybe at one point, it's worth having a shared typography type across kedro-viz? So it will be more consistent and I assume it's easier to not get the font messed up. 

Not sure if we've done it before but if not, maybe we can work with @GabrielComymQB and @Mackay031 to define the typography types for viz. What do you think?

tynandebold:
> hey @tynandebold everything looks good to me 😄🕺 Just one point to consider in the future, maybe at one point, it's worth having a shared typography type across kedro-viz? So it will be more consistent and I assume it's easier to not get the font messed up.
> 
> Not sure if we've done it before but if not, maybe we can work with @GabrielComymQB and @Mackay031 to define the typography types for viz. What do you think?

Absolutely, this is a great idea. We don't set anything `$fontSize`s anywhere, and it would be in our interest if we did.

GabrielComymQB:
this is something we already have in our Main Figma file and we make use of it.
https://www.figma.com/file/K8WJMMTkvFHn6OZYg7Ypfz/Kedro-Design-Master?node-id=2175%3A45972

however, we are just in the process of updating it and cleaning up everything we don't use. 
this came from the design team library and now we are making it ours.

Will keep you posted.

PR #936, opened 2022-06-22
Fix/improve layers labels and watermark highlights

### LINKED ISSUE #901: Improve layers labels and “watermark” highlights
## Description

When layers are active and the user hovers over the flowchart, the active layers are not lit in the way it should be. The original explorations from two years ago show that the “active layer” (the one on hover) should be highlighted in two ways: the written label that follows the edge of the canvas and also the subtle “watermark” background which highlights the boundaries of the layer.
 
Note: the background highlight does work correctly if you're on the light theme.

## Acceptance criteria

- The “active layer” label that follows the edge of the canvas should be lit in higher contrast. I’m not sure how we’re treating this colour variable at the moment (if it’s a solid coulour or a transparent white). If it is a transparent white. Let’s go .55 alpha for the inactive layers and .85 alpha for the active one.
- The subtle “watermark” background which highlights the boundaries of the layer should have increased contras as shown in the reference link bellow.

Reference links: 

https://zpl.io/blEY77P

## Checklist

Huongg:
## Description
Fixes #901 

## Development notes
ezgif com-gif-maker (3)

## QA notes

I only fix the layer's label, to ensure it gets highlighted when hovering over the layer. 
- I use .5 opacity for the inactive layers name and 1 for the active one, which is slightly different from what you @tynandebold suggested. As I think 0.55 vs 0.85 does not make it look much different but happy to change it back if you prefer 0.55 vs 0.85

For the watermark highlights, waiting for @GabrielComymQB to confirm the colour for both dark and light themes and I'll update them accordingly. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/936"></a>


src/components/flowchart/flowchart.js
```diff
@@ -448,6 +449,44 @@ export class FlowChart extends Component {
...
+  /**
+   * Enable a layer's active state when hovering it, update labelName's active className accordingly
+   * @param {Object} event Event object
+   * @param {Object} node Datum for a single node
+   */
+  handleLayerMouseOver = (event, node) => {
+    node &&
+      this.setState({
+        activeLayer: node.name,
+      });
```
tynandebold, line 461: This is just a conditional, right? It's a bit off from our usual pattern, which would be this: 
```suggestion
    if (node) {
      this.setState({
        activeLayer: node.name,
      });
    }
```

Let me know if there's a good reason to do it this way though.

src/components/flowchart/flowchart.js
```diff
@@ -448,6 +449,44 @@ export class FlowChart extends Component {
...
+
+    const { activeLayer } = this.state;
+    const layerName = document.querySelector(`[data-id="${activeLayer}"]`);
+
+    activeLayer &&
+      layerName &&
+      layerName.classList.add('pipeline-layer-name--active');
```
tynandebold, line 468:
```suggestion
    if (activeLayer && layerName) {
      layerName.classList.add('pipeline-layer-name--active');
    } 
```

src/components/flowchart/flowchart.js
```diff
@@ -448,6 +449,44 @@ export class FlowChart extends Component {
...
+    const layerName = document.querySelector(`[data-id="${activeLayer}"]`);
+
+    activeLayer &&
+      layerName &&
+      layerName.classList.remove('pipeline-layer-name--active');
+
+    node &&
+      this.setState({
+        activeLayer: undefined,
+      });
```
tynandebold, line 487: I'd turn these into `if` blocks, too.

src/components/flowchart/styles/_layers.scss
```diff
@@ -68,3 +68,7 @@
   font-size: 1.6em;
   white-space: nowrap;
 }
+
+.pipeline-layer-name--active {
```
tynandebold, line 72: We may want to add an opacity transition to the `pipeline-layer-name` base class of `350ms` or around there so it matches up more with the layer background opacity transition.

src/components/flowchart/flowchart.js
```diff
@@ -448,6 +449,44 @@ export class FlowChart extends Component {
     node && this.showTooltip(event, node.fullName);
   };
 
+  /**
+   * Enable a layer's active state when hovering it, update labelName's active className accordingly
+   * @param {Object} event Event object
+   * @param {Object} node Datum for a single node
+   */
+  handleLayerMouseOver = (event, node) => {
```
tynandebold, line 457: `event` isn't being used here, though when I remove it locally the highlight doesn't show up anymore. Do you see the same? I'd think you could take it out as an argument, and in the `mouseOut` function as well.

tynandebold (approved):
Looks great! It's much more sleek now than it was before.

I've left some comments, though nothing huge, so I'll approve straight away.

src/components/flowchart/flowchart.js
Huongg: hey yeah it's just a conditional. There's no good reason to do it this way 😄 I'll update the code to make it consistent with the codebase

src/components/flowchart/styles/_layers.scss
Huongg: great suggestion, added 👍

src/components/flowchart/flowchart.js
Huongg: hmm i don't think it will work, as the `event` is always the first parameter of the `mouseEvent` callback, right? And `node` is a data we passed it to this function. I think if we remove `event` and it will think `node` is as `event`. I guess that's why the highlight wouldn't work

GabrielComymQB (approved):

src/components/flowchart/flowchart.js
tynandebold: Makes sense!

PR #939, opened 2022-06-23
Fix/tags active inactive behaviour

### LINKED ISSUE #920: Improve ‘Tags’ active/inactive component behavior
## Description

When no tags are used, the Filter panel subgroup “Tags” is still active and available. It looks broken this way - like nothing is showing for some reason, but it's still interactive.

Screen Recording 2022-06-10 at 15 02 1

## Acceptance criteria

- If no tags are present, the Tags subgroup in the Filter panel should take a disabled state as seen on reference link. 
- User is not able to interact with any elements (blue indicator selector or expand/collapse arrows).
- The text label “Tags” takes an inactive .30 alpha white state.

Reference link:

https://www.figma.com/file/K8WJMMTkvFHn6OZYg7Ypfz/Kedro-Design-Master?node-id=4272%3A5484

image

## Checklist

cvetankanechevska:
## Description

Resolves #920 

## Development notes

Added new check if items list is empty or not, according to that toggle button will behave properly. And text label for that row will be displayed with less opacity.

Here we have disabled button for Tags:

Screenshot 2022-06-23 at 14 08 46

## QA notes

Tests that we already have for `node-list-groups` and `node-list-group` are modified. 
- In `node-list-groups` - I am checking if DOM element has a disabled class for the button, if not we can than simulate click() event.
- In `node-list-group` -  I've added empty array as in mock data from spaceflights we don't have `items` data

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/939"></a>


GabrielComymQB (approved):

tynandebold (approved):
Looks great, and thanks for sorting some props. I'm approving with one thought:

I'd suggest you add a test to check that when `items` is 0 the new class name (`pipeline-type-group-toggle--disabled`) is actually set. I don't think you're doing that at the moment, are you?

cvetankanechevska:
> Looks great, and thanks for sorting some props. I'm approving with one thought:
> 
> I'd suggest you add a test to check that when `items` is 0 the new class name (`pipeline-type-group-toggle--disabled`) is actually set. I don't think you're doing that at the moment, are you?

No, I don't have that check. I will try to add it.

rashidakanchwala (approved):
LGTM

Issue #910, opened 2022-06-15
New platform-dependent `%launch_viz` line magic


AntonyMilneQB:
Ideal outcome:
* make new `%launch_viz` line magic in kedro-viz, so that in any notebook with kedro IPython extension loaded, `%launch_viz` will be available
* `%launch_viz` starts a kedro-viz server and supplies a URL which the user can click on to open kedro-viz in another browser window

Required steps:
- [ ] Make sure the current code is still the best way to launch the server from inside the line magic 
- [ ] _Potentially difficult._ Work out the correct URL to access the kedro-viz instance on various platforms (databricks, sagemaker, etc.).
- [ ] _Potentially difficult._ Work out how to programmatically obtain this URL
- [ ] Work out how to automatically figure out which platform the notebook is running on
- [ ] Output the correct URL or some useful message which might help a user find their kedro-viz instance if we can't figure out the URL ourselves

To consider:
* how/which arguments can the user pass? Ideally all the same flags as used when you do `kedro viz` would be available the same way
* how to kill the server? See note in https://github.com/kedro-org/kedro/pull/1355 for current bug with `%run_viz` where this doesn't happen

tynandebold:
Which platforms do we prioritize at the beginning? Databricks, Sagemaker, and Platform McK seem like the three obvious choices.

AntonyMilneQB:
That sounds like the right choice to me also, though Platform McKinsey uses databricks so I would hope that if it works on there it would work on databricks in general.

I'm currently quite optimistic that if we can get the jupyter-server-proxy working then this will just work straight away on all these platforms 🤞

AntonyMilneQB:
Let's assume there will be two different ways that `%launch_viz` would work:
1. Databricks: use the above. So far tested on Azure; still need to test on AWS and GCP.
2. Jupyter servers: use [jupyter-server-proxy](https://github.com/kedro-org/kedro-viz/pull/941/commits/8d3ba797705a133d4ccf91b0a6d28b10f425c679). So far tested locally; still need to test on Sagemaker.

Next steps:
1. above next steps for databricks method
2. make similar `%launch_viz` that starts process and links to it (find out if there's a programmatic way to get the URL through jupyter-server-proxy)
3. something that detects your platform and switches between the above two methods; starts the process and outputs something useful ("Kedro-Viz started on port X") even if it can't work out what the platform is
4. get `%launch_viz` to take arguments for `--pipeline` etc.
[...]

PR #941, opened 2022-06-26
Fix: Viz not working with proxy URLs

### LINKED ISSUE #916: Kedro viz with proxy does not show additional information
## Description
Additional information is available only for the first couple of seconds when you are open it, and then it just disappears. Look at the picture below. I hided some information. What I am saying about is inside red rectangle.
Screenshot 2022-06-03 at 16 39 01

## Context
We have a remote cluster setup for working where we have jupyter hub installed and all interactions with this remote server are through this hub. So, to get kedro viz there, we had to make it work via this hub by using proxies. I am not sure about all technical details, but I can share with you my conversation with our support guy who was investigating this problem a bit:

> Long story short -- this is an issue with how kedro-viz frontend treats working with the proxy server
> Static javascript that is already built-in in the package is not configured to dynamically modify its paths when served through the proxy server (which is the case for us). On the start of kedro viz, we load an html which has some relative paths, causing all the content to prepend /user/<username>  base path. But all the subsequent requests are done from the loaded js files that don't have  /user/<username> base path prepended as they should. After lots of digging I found out this: https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually. Which I confirmed to myself by finding this: https://github.com/mlflow/mlflow/blob/master/mlflow/server/js/src/setupProxy.js. As you can see, this file being present in their repo is most likely (99%) the reason why mlflow does not have such an issue.

I can confirm that we start MLFlow the same way as Kedro Viz, and it does not have such problems. But still, I cannot be sure that the reason is really in these configurations.

## Steps to Reproduce
I am afraid I cannot properly describe all steps to reproduce this issue, as I have lack of technical knoledges of how jupyterhub works with proxies. Please let me know if you have any questions that I can address to our infrastructure team to help you understand the source of the problem.

## Expected Result
Information should be shown

## Actual Result
Information does not be shown

## Your Environment

* Web browser system and version: does not matter
* Operating system and version: CentOS Linux 7
* Output of jupyter --version command: 
```
  IPython          : 7.29.0
  ipykernel        : 6.5.0
  ipywidgets       : 7.6.5
  jupyter_client   : 7.0.6
  jupyter_core     : 4.9.1
  jupyter_server   : 1.11.2
  jupyterlab       : 3.2.2
  nbclient         : 0.5.5
  nbconvert        : 6.2.0
  nbformat         : 5.1.3
  notebook         : 6.4.5
  qtconsole        : not installed
  traitlets        : 5.1.1
```

## Checklist

tynandebold:
## Description

Fixes #916

## Development notes

If a user runs Viz with a proxied URL, like you may try to do as described in [this issue](https://github.com/kedro-org/kedro-viz/issues/916), the application would load without critical information, and sometimes no information at all. The reason being is that the URL wouldn't pass our [hard-coded route match](https://github.com/kedro-org/kedro-viz/blob/main/src/components/wrapper/wrapper.js#L60).

In this PR, we'll now allow a dynamic pathname to be recognized and matched, thus allowing the flowchart or experiment-tracking pages to render correctly.

Lastly, we'll also sanitize the URL and remove any duplicative `experiment-tracking` strings from appearing in the pathname. We need to guard against this because, again, if a user has a proxy URL or a base URL that isn't `/`, there will be conflicts. The use case: a user runs Viz with [jupyter-sever-proxy](https://jupyter-server-proxy.readthedocs.io/en/latest/index.html), which adds `/kedro_viz/` to the pathname. So Viz loads, then we navigate to the experiment-tracking page, which does load correctly. Yet if the page is refreshed, we'd receive an incorrect result. If we didn't remove `experiment-tracking` from the pathname, we'd get something like `/kedro_viz/experiment-trackingexperiment-tracking`, and that's not desirable.

## QA notes

Run Viz via `jupyter-sever-proxy` by doing the following:

- Checkout the branch
- `cd` in the `demo-project` folder
- `pip install jupyter-sever-proxy`
- Run `kedro viz`
- Click the `Kedro Viz` icon that appears

All functionality within Viz should function as it should.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/941"></a>


src/components/icons/logo.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 const LogoIcon = ({ className }) => (
-  <svg className={className} viewBox="24 25 32 30">
+  <svg viewBox="24 25 32 30">
```
AntonyMilneQB, line 4:
```suggestion
  <svg className={className} viewBox="24 25 32 30">
```
This was just me trying to do something, wasn't meant to stay as a change.

AntonyMilneQB (commented):
This is great! I will test it out later today.

Let's leave out all the Python stuff and icon for now since it shouldn't be needed just to fix the proxy bug. It needs a bit more thought (the stuff I did was just a quick POC) and would be done as part of a modified https://github.com/kedro-org/kedro-viz/issues/910.

package/setup.py
```diff
@@ -50,6 +50,10 @@
     package_data={"kedro_viz": list(files)},
     zip_safe=False,
     entry_points={
+        "jupyter_serverproxy_servers": [
+            # name = packagename:function_name
+            "kedro_viz = kedro_viz.launchers.jupyter:jupyter_server_proxy",
+        ],
```
tynandebold, line 56: @AntonyMilneQB should I remove this, too?

src/components/icons/logo.js
tynandebold: Reverted.

package/setup.py
tynandebold: I've removed it.

src/apollo/config.js
```diff
@@ -7,24 +7,27 @@ import {
...
-    : window.location.host;
+const { host, pathname, protocol } = window.location;
+const sanitizedPathname = replaceMatches(pathname, {
+  'experiment-tracking': '',
+});
+
+const wsHost = process.env.NODE_ENV === 'development' ? 'localhost:4142' : host;
 
-const wsProtocol = window.location.protocol === 'https:' ? 'wss' : 'ws';
+const wsProtocol = protocol === 'https:' ? 'wss' : 'ws';
```
AntonyMilneQB, line 19:
```suggestion
const wsProtocol = protocol === 'https:' ? 'wss:' : 'ws:';
```
Just to be consistent about what "protcol" refers to.

src/components/global-toolbar/global-toolbar.js
```diff
@@ -32,7 +38,7 @@ export const GlobalToolbar = ({
             disabled={false}
             icon={LogoIcon}
           />
-          <NavLink exact to={{ pathname: '/' }}>
+          <NavLink exact to={{ pathname: `${sanitizedPathname}` }}>
```
AntonyMilneQB, line 41:
```suggestion
          <NavLink exact to={{ pathname: {sanitizedPathname} }}>
```
If I got the syntax right...

AntonyMilneQB (commented):
This is great!! But... it doesn't actually seem to work for me 🤔 I'm not sure why because when I check the `window.location` it seems like it should. Probably I've just not set it up correctly locally. Let me show you on Wednesday and hopefully we can figure it out.

I'm also wondering:
* should we make this even more generic to include `search` and `hash` as well as `pathname`? I think this actually might be needed in real scenarios with jupyter server proxy, not just a theoretical possibility. Seems like it's possible to pass a `location` object to these tags rather than needing to reconstruct the `host + pathname + search + hash` string
* I don't quite get where the `/` should be still 😅 I'm sure what you've done is right but I need to think through it again with a fresh head to understand

tynandebold:
> This is great!! But... it doesn't actually seem to work for me 🤔 I'm not sure why because when I check the `window.location` it seems like it should. Probably I've just not set it up correctly locally. Let me show you on Wednesday and hopefully we can figure it out.
> 
> I'm also wondering:
> 
> * should we make this even more generic to include `search` and `hash` as well as `pathname`? I think this actually might be needed in real scenarios with jupyter server proxy, not just a theoretical possibility. Seems like it's possible to pass a `location` object to these tags rather than needing to reconstruct the `host + pathname + search + hash` string
> * I don't quite get where the `/` should be still 😅 I'm sure what you've done is right but I need to think through it again with a fresh head to understand

Oh no, ok, let's sync up today.

AntonyMilneQB (approved):
Awesome stuff. Super excited to get this fixed and what it means we can hopefully do next 💥 

It would be nice if a future iteration also passed the `search` and `hash` (looks like [`Link` at least could handle this easily](https://v5.reactrouter.com/web/api/Link/to-object))

PR #960, opened 2022-07-05
Update outline property for hover adn success state

### LINKED ISSUE #931: Buttons render with weird visible borders in solid state
## Description

The OnHover version of our primary button shows a “grey line” between the boarder and the solid rectangle shape as seen on supporting images.

Screenshot 2022-05-25 at 14 01 1

## Acceptance criteria

- The OnHover state of the button should show as a solid shape. No visible jagged borders or differenc collours.

## Checklist

cvetankanechevska:
Signed-off-by: Cvetanka <cventanka_nechevska@external.mckinsey.com>

## Description

Fixes #931 

## Development notes

Set `outline: none` for hover and success state.

## QA notes

Screenshot 2022-07-05 at 15 34 38

Screenshot 2022-07-05 at 15 35 49

## Checklist

### DIFF
--- a/src/components/ui/button/button.scss
+++ b/src/components/ui/button/button.scss
@@ -75,8 +75,9 @@ $secondary-underline-offset-hover: 4px;
   }
 
   &:hover:not(:disabled) {
-    color: var(--color-button__text--hover);
     background: var(--color-button--outline);
+    color: var(--color-button__text--hover);
+    outline: none;
   }
 
   &:disabled {
@@ -131,7 +132,7 @@ $secondary-underline-offset-hover: 4px;
   animation: scale 0.35s ease;
   background-color: $color-success;
   color: $color-dark;
-  outline: 2px solid $color-success;
+  outline: none;
   width: 172px; // Match the `Apply and close` width to avoid a layout shift.
 }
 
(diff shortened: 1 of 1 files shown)


tynandebold (commented):
I think we need to think about this with a little more detail. Since we're no longer using `outline`, the buttons change height when they're hovered over. That isn't ideal.

Maybe we should use `border` on the base classes now instead of `outline`. What do you think?

Also, please make sure to add this PR to the `release.md` file.

GabrielComymQB (approved):

GabrielComymQB:
> I think we need to think about this with a little more detail. Since we're no longer using `outline`, the buttons change height when they're hovered over. That isn't ideal.
> 
> Maybe we should use `border` on the base classes now instead of `outline`. What do you think?
> 
> Also, please make sure to add this PR to the `release.md` file.

Yes, I completely missed this.

What's happening is that, on hover, we are removing the stroke line border (which I think is an external border) completely and going back to a core rectangle which then will be 2px smaller on all the sides.

The dimension of the buttons should not change. There should be a solid base which never changes in size regardless of the style of the borders varying in different states.

cvetankanechevska:
> I think we need to think about this with a little more detail. Since we're no longer using `outline`, the buttons change height when they're hovered over. That isn't ideal.
> 
> Maybe we should use `border` on the base classes now instead of `outline`. What do you think?
> 
> Also, please make sure to add this PR to the `release.md` file.

I've replaced 'outline' with 'border' and it looks good to me. Thanks for catching this 👍

tynandebold (approved):

GabrielComymQB (approved):

PR #978, opened 2022-07-19
FE changes based on GraphQL Refactor 

rashidakanchwala:
## Description

This PR was created based on the backend refactor ticket. #964 
The backend has now changed and for runTrackingData -- it will now get queries also based on the group (dataset type) the tracking data belongs to i.e. Metrics, JSON Data (and in future Plots)

In this ticket, we have adjusted the front-end to get information from the backend for each dataset. If a particular group (Metrics, JSON Dataset) has no datasets, that group will not be shown in the front-end. Otherwise all run tracking data will now have  another parent in the heirarchy/accordian i.e. type of dataset (Metrics, JSON Data, Plots) 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/978"></a>


src/components/experiment-wrapper/experiment-wrapper.js
```diff
@@ -40,12 +40,22 @@ const ExperimentWrapper = ({ theme }) => {
...
 
+  let runTrackingData = {};
+
+  if (metric.length > 0) {
+    runTrackingData['Metrics'] = metric;
+  }
+
+  if (json.length > 0) {
+    runTrackingData['JSON Data'] = json;
+  }
```
tynandebold, line 57: I don't think this is needed as it recreates the shape of the data that comes directly from the query.

src/components/experiment-tracking/accordion/accordion.scss
```diff
@@ -27,6 +27,10 @@ $toggle-width: 1.15em;
   &--large {
     font-size: 1.6em;
   }
+
+  &--x-large {
+    font-size: 1.9em;
+  }
```
tynandebold, line 33: Looking at the designs I don't see where it calls for this font size. Think we should be able to use what we already have.

src/components/experiment-tracking/run-dataset/run-dataset.test.js
```diff
@@ -3,48 +3,56 @@ import RunDataset from '.';
 import { runs, trackingData } from '../../experiment-wrapper/mock-data';
 import { shallow, mount } from 'enzyme';
 
-const booleanTrackingData = [
-  {
-    datasetName: 'Data Analysis',
-    data: {
-      classWeight: [{ runId: 'My Favorite Sprint', value: false }],
+const booleanTrackingData = {
+  json: [
```
AntonyMilneQB, line 7: Should this be `JSONData` just to match the query?

Also do we want test cases that include metrics?

AntonyMilneQB (approved):
Awesome work! I just tried it out and all works as I expected it to ⭐ 

I can't judge the javascript very well, but my only comment there is about test cases.

One small thing: when you don't have any metrics (for example) in a particular run then the section still appears there on the frontend with empty datasets.
image

There's two things that are empty here:
* a dataset wasn't saved in that particular run: in this case I would just not show that dataset at all
* no datasets in a group (e.g. metrics) were saved at all. In this case I don't know if we should give an empty "Metrics" section or just not show it at all. Probably the latter to be consistent with the above point. But maybe the former gives better discoverability to what are the sections that _could_ exist there if you saved the right sort of dataset.

No need to change anything for this now. I already have a `TODO` to consider changing how this works in the backend code, plus it needs some discussion since maybe breaks the comparison view if we remove missing datasets? But interested to hear if you think we should change this in future.

src/components/experiment-tracking/run-dataset/run-dataset.js
```diff
@@ -44,41 +45,54 @@ const RunDataset = ({
...
   return (
     <div
       className={classnames('details-dataset', {
         'details-dataset--single': isSingleRun,
       })}
     >
-      {trackingData.map((dataset) => {
-        const { data, datasetName } = dataset;
-
+      {Object.keys(trackingData).map((group) => {
```
AntonyMilneQB, line 56: Curious how the order of the sections `Metrics` vs. `JSON data` is determined here. Is it just automatically the same as the order specified in `query getRunTrackingData`?

rashidakanchwala:
> Awesome work! I just tried it out and all works as I expected it to ⭐
> 
> I can't judge the javascript very well, but my only comment there is about test cases.
> 
> One small thing: when you don't have any metrics (for example) in a particular run then the section still appears there on the frontend with empty datasets. image
> 
> There's two things that are empty here:
> 
> * a dataset wasn't saved in that particular run: in this case I would just not show that dataset at all
> * no datasets in a group (e.g. metrics) were saved at all. In this case I don't know if we should give an empty "Metrics" section or just not show it at all. Probably the latter to be consistent with the above point. But maybe the former gives better discoverability to what are the sections that _could_ exist there if you saved the right sort of dataset.
> 
> No need to change anything for this now. I already have a `TODO` to consider changing how this works in the backend code, plus it needs some discussion since maybe breaks the comparison view if we remove missing datasets? But interested to hear if you think we should change this in future.

I agree, @tynandebold  -- earlier in the commits I had broken down runTrackingDatasets to metrics, json and then was comparing if either one was empty not to add that section in the tracking data if it's empty. Can we do something similar to that but maybe more cleaner?

tynandebold (approved):
Looks great!

Mind the merge conflicts when pulling in the latest changes from `main`.

src/components/experiment-tracking/run-dataset/run-dataset.js
rashidakanchwala: query has it :)

Issue #953, opened 2022-07-04
Implement visualizing plots in experiment tracking


tynandebold:
## Description

This ticket focuses on the implementation of the design work created in #938.

The backend will need to recognize the versioned plots (Matplotlib and Plotly Datasets) and the frontend will need to display them, with the behavior outlined in the Figma files:

1. [Finding and exploring plots - single view](https://www.figma.com/file/akJ7QasxvLcgqc8Hxud7QB/Kedro-Design%3A-OUT?node-id=217%3A7286)
2. [Finding and exploring plots - comparison view](https://www.figma.com/file/akJ7QasxvLcgqc8Hxud7QB/Kedro-Design%3A-OUT?node-id=230%3A7796)
3. [Expanded graphs, page variations (1,2 and 3 runs)](https://www.figma.com/file/akJ7QasxvLcgqc8Hxud7QB/Kedro-Design%3A-OUT?node-id=232%3A12150)

## Checklist

- [x] Include labels so that we can categorise your feature request

tynandebold:
### Plan of action

- In `graphql.py`, line 235: look for matplotlib and ploty with version==true
  - there are two plotly datasets that we need to check for here. look in the shortTypeMapping config var
- graphql.py, line 249: needs to become something more generic and what's passed to the FE is either JSON or PNG/JPG. base64 the matplotlib img
  - we want to keep the same data structure as well (key/value pairs). the key would be the filename, and there's only one value, which is the JSON for Plotly or the base64 image for matplotlib
- on the frontend, according to the dataset type, show the appropriate plot or data
- update `shortTypeMapping` and add the two missing dataset, e.g "kedro.extras.datasets.tracking.metrics_dataset.MetricsDataSet"
- break up the tracking datasets into sections: plots, metrics, JSON data. More info below...

#### Plots at the top

We also want to plots to appear at the top of the tracking area. They should go into their own accordion as well, as shown here in the design:

<img width="356" alt="image" src="https://user-images.githubusercontent.com/16869061/177190647-79d7a27c-1aad-4d0c-9c31-e63e5a127337.png">

We'll have three groups: Plots, Metrics, and JSON. Our `runTrackingData` query will now return the following:

```graphql
type GroupedDataset {
  groupedDatasetType: String 
  dataset: TrackingDataset[]
}
```

`TrackingDataset` then becomes:

```graphql
type TrackingDataset {
  datasetName: String
  data: JSONObject
}
```

[...]

Issue #980, opened 2022-07-19
[Spike] Experiment tracking test refactoring


antonymilne:
Noting down ideas from #964 here; will explain better and break into tickets in due course...

**Write lots of tests. Improve the way we do GraphQL tests in general.** Various ideas here:
* https://strawberry.rocks/docs/operations/testing; https://www.ontestautomation.com/writing-tests-for-graphql-apis-in-python-using-requests/; https://github.com/graphql-python/gql/tree/master/tests/starwars; https://github.com/strawberry-graphql/strawberry/blob/main/CHANGELOG.md#01060---2022-04-14
* e2e query tests and unit tests:
  * `router` is very small and doesn't need tests
  * `types` doesn't need testing
  * `schema` methods (query, mutation, subscription) largely delegate to `data_access_manager.get...` (already covered by unit tests) and then call `format` functions on the results if required. Tests are covered by e2e-style query tests that we already have
  * `serializers` should be unit tested - currently missing

We should also improve/create tests for the models layer:
* `flowchart`: those for metadata side panel don't seem quite right but maybe others too. e.g. should we reuse the "real" datasets from graphql conftest.py
* `experiment_tracking`: we should cover the bit marked with `# pragma: no cover`

For adding plots to experiment tracking:
* ~Add query by group as `TrackingDatasetGroup` and produce existing behaviour for metrics and JSON.~ done in #978 
* Try to align `TrackingDatasetModel` and `TrackingDataset`. Consider model for run which would simplify (maybe even remove) `format_run_tracking_data`. How to query by `run_id` correctly?
* Would what it take to get rid of `self.runs[run_id] = {}` and just not return anything when that version of a dataset doesn't exist?
* Might be worth doing a new model for each `TrackingDatasetRun` in strawberry but not a dataclass model. Maybe do this as GraphQL interface with different implementations, serializers for plots, etc.

Important other refactoring:
* Reuse `DataNode` and `DataNoteMetadata` models. There's too much duplication between these and tracking datasets (~also code to get dataset type is in both~ Done in https://github.com/kedro-org/kedro-viz/pull/1232, possibly other things should move to `utils.py`)
* ~Move flowchart import of optional dependencies to same scheme used in experiment tracking~ done in #984 
* Better system for `check_db_session`, e.g. decorator argument that returns empty iterable (could be done automatically from type hint)? Null class?
* Consider whether `is_tracking_dataset` should use `isinstance` instead, but be careful with imports
[...]

PR #984, opened 2022-07-20
Feature/Visualise Plots in Experiment Tracking

rashidakanchwala:
## Description

Resolves https://github.com/kedro-org/kedro-viz/issues/953.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/984"></a>


package/kedro_viz/models/experiment_tracking.py
```diff
@@ -2,16 +2,23 @@
...
+from kedro.io.core import get_filepath_str
+
 from sqlalchemy import Column
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.sql.schema import ForeignKey
 from sqlalchemy.types import JSON, Boolean, Integer, String
 
+from kedro.extras.datasets.plotly import JSONDataSet as PlotlyJSONDataSet
+from kedro.extras.datasets.plotly import PlotlyDataSet
+from kedro.extras.datasets.matplotlib import MatplotlibWriter
```
AntonyMilneQB, line 20: We can't have these as top-level imports because they won't work in a project that doesn't have the right optional dependencies (like matplotlib) installed.

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -78,11 +85,18 @@ def load_tracking_data(self, run_id: str):
...
+            from kedro.extras.datasets import json, tracking, plotly, matplotlib
 
             tracking.JSONDataSet._load = json.JSONDataSet._load  # type: ignore
             tracking.MetricsDataSet._load = json.JSONDataSet._load  # type: ignore
-            self.runs[run_id] = self.dataset.load()
+            plotly.JSONDataSet._load =  json.JSONDataSet._load  # type: ignore
+            plotly.PlotlyDataSet._load = json.JSONDataSet._load 
+            matplotlib.MatplotlibWriter._load = matplotlib_writer_load
+
+            if isinstance(self.dataset, (PlotlyJSONDataSet, PlotlyDataSet, MatplotlibWriter)):
```
AntonyMilneQB, line 96: Let's do a string check of `self.dataset_type` here like `if self.dataset_type in ["kedro.extras.datasets...", "..."]`. It's ugly and we should re-think this in future but it's safer than relying on the imports and consistent with how `get_dataset_type` works.

AntonyMilneQB: In fact, because these are all plot types we could do as `if TRACKING_DATASET_GROUPS[self.dataset_type] is TrackingDatasetGroup.PLOT`. That's much nicer.

package/kedro_viz/data_access/repositories/tracking_datasets.py
```diff
@@ -14,8 +14,9 @@
 
 TRACKING_DATASET_GROUPS = {
     # TODO: add these.
```
AntonyMilneQB, line 16:
```suggestion
```

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -78,11 +85,18 @@ def load_tracking_data(self, run_id: str):
...
+                self.runs[run_id] = {get_filename(self.dataset): self.dataset.load()}
```
AntonyMilneQB, line 97:
```suggestion
                self.runs[run_id] = {self.dataset._filepath.name: self.dataset.load()}
```
Unless there's some reason we need to use `get_filename` as you wrote it?

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -99,3 +113,19 @@ def load_tracking_data(self, run_id: str):
...
+
+def get_filename(dataset: AbstractVersionedDataSet) -> str:
+     dataset_description = dataset._describe()
+     return (dataset_description.get("filepath") or dataset_description.get("path")
+            ).name
+
+
+def matplotlib_writer_load(dataset: AbstractVersionedDataSet) -> str:
+    load_path = get_filepath_str(dataset._get_load_path(), dataset._protocol)
+    with open(load_path, "rb") as img_file:
```
AntonyMilneQB, line 126: This might not work right on protocols outside local (like s3). We should do this instead (also in the metadata side panel):

```suggestion
    with dataset._fs.open(load_path, "rb", **dataset._fs_open_args_load) as img_file:
```

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -99,3 +113,19 @@ def load_tracking_data(self, run_id: str):
...
     return f"{dataset.__class__.__module__}.{dataset.__class__.__qualname__}"
+
+
+def get_filename(dataset: AbstractVersionedDataSet) -> str:
+     dataset_description = dataset._describe()
+     return (dataset_description.get("filepath") or dataset_description.get("path")
+            ).name
+
+
+def matplotlib_writer_load(dataset: AbstractVersionedDataSet) -> str:
```
AntonyMilneQB, line 124:
```suggestion
# "MatplotlibWriter" is in quotes to avoid importing it.
def matplotlib_writer_load(dataset: "MatplotlibWriter") -> str:
```

AntonyMilneQB (commented):
Looks good so far! I just left lots of small suggestions.

We should check very carefully how this works for kedro projects without the relevant dataset dependencies installed - let's catch up later today on that.

package/kedro_viz/models/experiment_tracking.py
rashidakanchwala: ok, in this case i will move TRACKING_DATASET_GROUPS to model layer and import it in the repository layer

package/kedro_viz/models/experiment_tracking.py
rashidakanchwala: this doesn't work. it says dataset does not have any function ._fs.open

AntonyMilneQB: Hmm, I think this should work. Maybe you need `mode="rb"`?

package/kedro_viz/api/graphql/schema.py
```diff
@@ -78,17 +78,21 @@ def run_tracking_data(
...
-            )
-            for dataset in tracking_dataset_models
-        ]
+
+        all_tracking_datasets = []
+
+        for dataset in tracking_dataset_models:
+            runs = {run_id: dataset.runs[run_id] for run_id in run_ids}
+            formatted_tracking_data = format_run_tracking_data(runs, show_diff)
+            if len(formatted_tracking_data) > 0:
```
AntonyMilneQB, line 87:
```suggestion
            if formatted_tracking_data:
```
(check that works!)

package/kedro_viz/data_access/repositories/tracking_datasets.py
```diff
@@ -28,15 +21,9 @@ def __init__(self):
         ] = defaultdict(list)
 
     def get_tracking_datasets_by_group_by_run_ids(
-        self, run_ids: List[str], group: Optional[TrackingDatasetGroup]
+        self, run_ids: List[str], group: TrackingDatasetGroup
```
AntonyMilneQB, line 24: Let's change `group: TrackingDatasetGroup = None` in `def run_tracking_data` to remove the `= None` default value too (assuming it's never called without the `group` argument anywhere on the frontend, which I'm pretty sure it's not).

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -46,11 +48,21 @@ class Config:
...
+    PLOT = "plot"
     METRIC = "metric"
     JSON = "json"
 
 
+TRACKING_DATASET_GROUPS = {
+    "kedro.extras.datasets.plotly.plotly_dataset.PlotlyDataSet": TrackingDatasetGroup.PLOT,
+    "kedro.extras.datasets.plotly.json_dataset.JSONDataSet": TrackingDatasetGroup.PLOT,
+    "kedro.extras.datasets.matplotlib.matplotlib_writer"
+    ".MatplotlibWriter": TrackingDatasetGroup.PLOT,
```
AntonyMilneQB, line 60: Let's do this on one line and just put in `# pylint: disable=line-too-long`.

package/kedro_viz/api/graphql/schema.py
AntonyMilneQB: Also, is the `self.runs[run_id] = {}` still needed in the `TrackingDatasetModel` exception? I guess it might still be required so that the `-` are put in the right place on the frontend when a metric is missing, but would be good to check. (_Don't_ make it `None` like the `TODO` says though. If it's still required then `{}` is correct and please remove the `TODO`)

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -78,11 +90,18 @@ def load_tracking_data(self, run_id: str):
...
             # pylint: disable=import-outside-toplevel
-            from kedro.extras.datasets import json, tracking
+            from kedro.extras.datasets import json, matplotlib, plotly, tracking
 
             tracking.JSONDataSet._load = json.JSONDataSet._load  # type: ignore
             tracking.MetricsDataSet._load = json.JSONDataSet._load  # type: ignore
-            self.runs[run_id] = self.dataset.load()
+            plotly.JSONDataSet._load = json.JSONDataSet._load  # type: ignore
+            plotly.PlotlyDataSet._load = json.JSONDataSet._load  # type: ignore
+            matplotlib.MatplotlibWriter._load = matplotlib_writer_load  # type: ignore
```
AntonyMilneQB, line 99: Just thought of a better way to do this. Move this to the top of the file where imports normally go like this:
```
with suppress(ImportError):
    # tracking datasets do not have load methods defined yet but would be the
            # same as json loader.
            # pylint: disable=import-outside-toplevel
            from kedro.extras.datasets import json, matplotlib, plotly, tracking

with suppress(NameError):
            tracking.JSONDataSet._load = json.JSONDataSet._load  # type: ignore
            tracking.MetricsDataSet._load = json.JSONDataSet._load  # type: ignore
            plotly.JSONDataSet._load = json.JSONDataSet._load  # type: ignore
            plotly.PlotlyDataSet._load = json.JSONDataSet._load  # type: ignore
            matplotlib.MatplotlibWriter._load = matplotlib_writer_load  # type: ignore
```

AntonyMilneQB (approved):
Great work! From the backend side I think there's still a couple of small things to do here, but happy to accept so I'm not blocking it while I'm away!

1. Essential: make sure that this works in a project without any matplotlib datasets and without matplotlib installed. You should still be able to show the experiment tracking screen with metrics/json datasets and nothing should crash. This should all work _without_ adding `matplotlib` to kedro-viz requirements (fine to add to test_requirements though).
2. Important but not essential: try and get it working with `dataset._fs.open` rather than `open`. Otherwise this will only work on local matplotlib datasets and not ones on s3 etc., which is quite a limitation. We can come back and add this later if you can't get it working though.

package/kedro_viz/models/experiment_tracking.py
rashidakanchwala: this doesn't work for matplotlib_writer_load. so i will keep it as is and we can refactor it later.

package/kedro_viz/api/graphql/schema.py
rashidakanchwala: this doesn't work. it breaks everything so i have kept it

comym (approved):

package/kedro_viz/models/experiment_tracking.py
AntonyMilneQB: Played around with this and got it working in [04165de](https://github.com/kedro-org/kedro-viz/pull/984/commits/04165de16f0857c066315ad5d5c76824f0ba96d0).

package/kedro_viz/models/experiment_tracking.py
AntonyMilneQB: Done in [04165de](https://github.com/kedro-org/kedro-viz/pull/984/commits/04165de16f0857c066315ad5d5c76824f0ba96d0).

AntonyMilneQB:
Small notes on the commits from [04165de](https://github.com/kedro-org/kedro-viz/pull/984/commits/04165de16f0857c066315ad5d5c76824f0ba96d0) onwards, since it proved to be a surprisingly big chunk of work...  The metadata side panel was not setup correctly to handle all the different cases (versioned and unversioned datasets, remote storage, live update of dataset previews when another `kedro run` is executed). As part of fixing this, I ended up refactoring quite a bit too, and some of the messy bits of flowchart.py are much better now.

1. The metadata side panel now _nearly*_ works for all these cases
2. The management of (a) optional dependencies like `matplotlib` and (b) the dataset loading functions are now handled the same way in both experiment tracking and flowchart. This code has been centralised to data_loader.py so we have much less duplication than before
3. We no longer need `load_latest_tracking_data`. Instead we do `dataset.release` to fix #573, which is much nicer and works for all dataset types

I've updated https://github.com/kedro-org/kedro-viz/issues/980 since some of those things are now fixed and I've also discovered some other things that should be fixed in the future.

*The reason I say nearly is explained in...

AntonyMilneQB (approved):

Huongg (approved):

Issue #882, opened 2022-06-01
Zoom is reset each time a node or dataset is selected


kadeshoe5:
## Description
Kedro-viz resets the current zoom window size and position to the default values whenever a node or dataset is selected in the visualization. 
## Context
It makes it difficult to navigate larger pipelines when I am zoomed in on a specific area and have to constantly re-zoom in when I change which node to inspect

## Steps to Reproduce

1. Run Kedro-Viz
2. Zoom to a node
3. Select node

## Expected Result
Kedro should not zoom out if all dependancies are within the current zoom window. Or there should be a setting that disables this behavior.

## Your Environment
* Web browser system and version: Chrome Version 101.0.4951.64 
* Operating system and version: Mac OSX Monterey 12.4
* Kedro version used (if relevant): 18.0

AntonyMilneQB:
I've noticed this before and agree it's annoying. Personally I don't see why opening the metadata panel should change the zoom on the flowchart at all, even if all the node's dependencies aren't on screen. I'm not even sure what sort of pipelines the current approach would benefit 🤔  So unless I'm missing something, +1 from me for just removing the zoom-reset-on-click behaviour altogether.

tynandebold:
Hey @kadeshoe5 we just merged a PR that will fix this. I'll let you know when the next release is out and hopefully you can give it a try!

Thank you for raising the issue.

PR #988, opened 2022-07-28
Zoom is reset each time a node or dataset is selected

cvetankanechevska:
Signed-off-by: Cvetanka <cventanka_nechevska@external.mckinsey.com>

## Description

Fix: #882 

## Development notes

Not updating the view coordinates when the rightSidebar is open or closed.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/988"></a>


comym:
Hi, here are my comments:

I like how the zoom in/out mechanism is now independent of the opening and closing from the metadata panel. It works perfectly.

However a detail we are missing is to make sure the selected node is always visible and centred to the reduced viewport. Sometimes you select a node and when the metadata panel opens, it goes over the top and you can't see your selected node anymore. Please check this gif where I show it happening:

Screen Recording 2022-07-28 at 10 27 20

In sum, the selected node should be always visible. For that, we don't need to play with zooming in or out, we just make sure we centre the selected node this way:

Screenshot 2022-07-28 at 10 28 47

And this is what I mean when I say "the reduced viewport" to the flowchart:

Screenshot 2022-07-28 at 10 36 56

Many thanks

tynandebold (changes requested):
This is a great start. I agree with Gabriel's comments. We need to ensure we center the selected node.

I think you'll have more success by editing the code [here](https://github.com/kedro-org/kedro-viz/blob/5aebdfb43332b20334edcf5011bba4ce43886dd1/src/components/flowchart/flowchart.js#L389-L399) than by doing your `if` statement.

Just playing around with it a bit this morning I had some success with this:

```js
const transform = viewTransformToFit({
  offset,
  focus,
  viewWidth: chartWidth,
  viewHeight: chartHeight,
  objectWidth: graphWidth,
  objectHeight: graphHeight,
  minScaleX: 0.2,
  minScaleFocus: this.props.visibleMetaSidebar
    ? this.props.chartZoom.scale
    : 0.3,
  focusOffset: this.props.visibleMetaSidebar ? 0.01 : 0.8,
});
```

Another case we need to think about is what happens when you're zoomed in quite a bit and we open a modular pipeline. In that case, I think we need to zoom out a bit, since a lot of things are expanding. I have an idea how to do that. We can talk about it if you'd like.

Huongg:
hey @cvetankanechevska I checked briefly, one thing I notice that the `this.props.visibleMetaSidebar` does not get updated when you close the panel directly. If you can trace it back to where this props gets updated, and why it's not updated when you click the close button directly, it might help to solve the issue. 
 What do you think?

tynandebold (approved):

Huongg (approved):
👍 LGTM

Issue #932, opened 2022-06-20
Micro-animation of Runs details (right hand side)


tynandebold:
## Description

All the copy and content of the Run details on the on the rights hand side of screen appear to be very stiff and jumpy - Labels, expandable groups, columns and rows.

 
![Screenshot 2022-05-23 at 14 02 1](https://user-images.githubusercontent.com/16869061/174599600-c45d0e78-f8de-4ce7-9a6e-c1c03eed5bc1.png)

## Acceptance criteria

- All the copy and content of the Run details should animate as per the Principle prototype when the “compare” toggle is activated and when any layout changes happen on the right hand side of screen. Labels, copy, rows, columns should appear/disappear with ease. Slick. Please see reference link to the prototype.

Reference link: https://mckinsey.box.com/s/w3clia4k5njijhjkl60tall8qnpgabwi

## Checklist
- [x] Include labels so that we can categorise your issue

Huongg:
hey @tynandebold @GabrielComymQB I've tried to summarise all the behaviour for this page. Can you confirm they're correct? Also, I have some questions (**in bold**) along with this too. Let me know if you think it's easier to go through them on call?


### Comparison mode on:
- First title slight fade-in from the right-hand side and title on the left slightly fades out

![image](https://user-images.githubusercontent.com/32060364/178712333-429b4745-a8ff-4e09-ab7b-e013df056cfb.png)
![image](https://user-images.githubusercontent.com/32060364/178712391-4e6382de-2a40-4c6f-a725-c0495aa5ec5e.png)

- At the same time, both sides resize


### Chose more than 1 run to compare
- A set of buttons appears first on the baseline run
![image](https://user-images.githubusercontent.com/32060364/178712475-f812f0ad-ef66-4d11-927d-42ba55c6af85.png)

- Then the second run slightly fades in
![image](https://user-images.githubusercontent.com/32060364/178712518-f73a828d-61bb-4401-9dd5-b367ea558ba0.png)


**There was some movement in the metrics panel but I'm not sure what the exact behaviour is?**


### Toggle differences

**In the prototype, when toggle differences, the metrics seem to be changed too, but on my local host, it only hides or shows the arrow. Do we need to implement this separately?**
 
![ezgif com-gif-maker (7)](https://user-images.githubusercontent.com/32060364/178712121-588f8934-a8d9-4a6e-9b72-f48192dacf4c.gif)



### Turn the comparison mode off
[...]

PR #993, opened 2022-07-29
Chore/micro animation

Huongg:
## Description

Fixes [#932](https://github.com/kedro-org/kedro-viz/issues/932)

## Development notes

Main changes include:
- Now there is a shared animation that currently lives in `run-metadata/animation.css`. This probably needs to move somewhere else. 
- The structure of `run-metadata` is slightly changed now. I separated all the labels (eg: Created By, GitSSH, etc) as a a new column from the run's value. Before it was in the same column of the first run's value
- I used TransitionGroup and CSSTransition to handle the animation of when adding or removing a run from the list. The same structure applies to both `run-metadata` and `run-dataset`

```
 <TransitionGroup .....>
        {runs.map((data, index) => {
          return (
              <CSSTransition
                key={data.runId}
                timeout={300}
                classNames="...."
                enter={isSingleRun ? false : true}
                exit={isSingleRun ? false : true}
              >
                ...children
              </CSSTransition>
            );
          })}
      </TransitionGroup>
```

There's currently a bug for a single view as when you select a different run, the TransitionGroup thought that you added a new run to the comparison view, therefore I need to add the 2 extra props below to make sure it won't do it. If we want to include animation for a single run view, i think it should be done in a separate ticket

```
enter={isSingleRun ? false : true}
exit={isSingleRun ? false : true}
```

Currently, I've combined metaData and dataSet into master data to ensure both sets of data get displayed in sync. This will be implemented from the BE in the future 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/993"></a>


cvetankanechevska:
Hey @Huongg, transition look great! 🎉  

Just one small css issues that I found, here is the screenshot:
Screenshot 2022-08-01 at 19 09 07

src/components/experiment-tracking/run-metadata/run-metadata.scss
```diff
@@ -46,55 +45,90 @@
...
 
 .details-metadata__run td {
   font-size: 1.4em;
   padding-bottom: 16px;
   vertical-align: top;
   word-wrap: break-word; // For breaking potential long text, such as as git branch, title, etc.
+}
 
-  &:first-of-type {
-    width: 230px;
```
tynandebold, line 72: I'd be careful about removing any of these widths and not adding them back where they need to be in your new markup. Things are looking a little too wide.

Your branch:

image

Demo site:

image

src/components/experiment-tracking/run-metadata/animation.scss
```diff
@@ -0,0 +1,83 @@
+$animation-timing: 0.2s;
+
+// animation for the first run when switching the comparison view
+.details-dataset__value,
+.details-dataset__value-header,
+.details-metadata__run--first-run,
+.details-dataset,
+.details-metadata {
+  transform: translateX(0);
+
```
tynandebold, line 10: Nit: I'd remove all the empty spaces between style properties in this file.

src/components/experiment-tracking/run-metadata/run-metadata.js
```diff
@@ -98,109 +101,132 @@ const RunMetadata = ({
...
+                      />
+                      <HiddenMenu isBookmarked={run.bookmark} runId={run.id} />
+                      <IconButton
+                        ariaLive="polite"
+                        className="pipeline-menu-button--labels__close"
+                        icon={CloseIcon}
+                        labelText="Remove run"
+                        labelTextPosition="bottom"
+                        onClick={() => onRunSelection(run.id)}
+                      />
```
tynandebold, line 206: These shouldn't be here when comparison mode is active but only one run is selected.

src/components/experiment-tracking/run-metadata/run-metadata.scss
```diff
@@ -46,55 +45,90 @@
...
   &:last-of-type {
-    min-width: 375px;
-    width: 50%;
+    width: 100%;
   }
 }
 
+.details-metadata__table tbody {
+  display: flex;
+}
```
tynandebold, line 97: I don't think this is needed.

tynandebold (changes requested):
This is looking great! It's so close. Most of my comments are small and should be easy to remedy. 

A couple others:
- I see three console errors in the web inspector. Can we fix those please?
- It doesn't look like the `Value` text label in the tracking data area is animating and I think it should.
- When you have three runs and the middle one is deselected it fades out beautifully, as it should. The run to the left that's still present doesn't seem like it slides over to take the place of the one that just faded out. That might be a tough one. Let me know if that's still a WIP.

Amazing work!

tynandebold:
Some style comments before the code review...

1. When comparison view is on and I have three runs but have to scroll horizontally to see all runs, the metadata container doesn't go full width:

    image

1. When comparison view is on and I have three runs, the space isn't equal between run one and two and run two and three. There's also a misalignment between the value label and the value itself:

    Screen Shot 2022-08-10 at 15 19 34

1. The general spacing is a bit off between the demo site and your branch:

    Demo: Screen Shot 2022-08-10 at 15 21 52

    Local: Screen Shot 2022-08-10 at 15 22 04

tynandebold (changes requested):
Great stuff, ever closer!

Left some comments that I think do need addressing. Let me know if any of it doesn't make any sense :)

Mackay031:
Also, on the screens here — https://github.com/kedro-org/kedro-viz/pull/993#issuecomment-1210753577, the 'baseline' icon is missing and the order is different on icons. Apologies if pointing out the obvious, just want to make sure we capture.

tynandebold:
@Mackay031 I may have turned off `showChanges` at some point when I was taking the screenshots. Can't remember. It may not anything you need to fix though Huong.

Huongg:
thank you for the comments @tynandebold , let me address it quickly

Huongg:
hey thank you for your comments @tynandebold , will get them fixed today. Also @Mackay031 apart from what Tynan mentioned, what do you think about the recent animation, eg timing. Do you think it's close enough to merge?

src/components/experiment-tracking/run-metadata/run-metadata.js
```diff
@@ -98,109 +101,135 @@ const RunMetadata = ({
...
-            className={classnames('details-metadata__run', {
-              'details-metadata__run--single': isSingleRun,
-            })}
-            key={run.id}
-          >
-            <table className="details-metadata__table">
+      <table className="details-metadata__table">
+        {runs.map((run, i) => (
+          <React.Fragment key={run.id + i}>
+            {i === 0 ? (
```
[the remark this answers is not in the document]
Huongg: oh yeah this is odd....i might have committed something by accident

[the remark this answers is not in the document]
Huongg: fixed now 😄

src/components/experiment-tracking/run-metadata/run-metadata.scss
```diff
@@ -1,18 +1,28 @@
...
+  content: '';
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: calc(100% + 4em);
+  height: 100%;
+
+  background-color: var(--color-bg-1);
+
+  z-index: -1;
```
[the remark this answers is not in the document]
Huongg: yeah sorry i forgot to remove this stuff 😄

src/components/experiment-tracking/run-metadata/run-metadata.test.js
```diff
@@ -49,29 +49,37 @@ describe('RunMetadata', () => {
...
     expect(wrapper.find('.details-metadata').length).toBe(1);
-    expect(wrapper.find('.details-metadata__run').length).toBe(3);
+    expect(wrapper.find('.details-metadata__run').length).toBe(4);
   });
 
-  it('renders a single-run view', () => {
+  it('renders a first run for when theres a single run', () => {
     const wrapper = shallow(
       <RunMetadata
+        enableComparisonView={true}
```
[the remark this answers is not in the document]
Huongg: yup im going to make some extra tests now

Mackay031:
Hey, here are a few comments. I am just focusing on the animation and not any other little glitches ie. disappearing images.

Adding this video as a reference [https://mckinsey.box.com/s/7q5bk3tollunl1t9gpzrdbzh072513g1](url)

- @7s Columns animate in smoothly, both header and content together. Looks great. The only comments would be spacing between columns could be reduced a bit. 
- @15s Could you line up the left side of images to left side of column (the image of run 2 and run 3 don't line up with the word 'value'.
- @17s the header and contents still move separately, but its better then before when the header and contents were from different run for a split second.
- @21s things get a bit confusing. Run 2 starts moving and Run 3 jumps across in first step. Then they both jump across in a second step. The runs should have the same behaviour and move in 1 step.

Hope this all makes sense...

Huongg:
> 

hey @Mackay031 thanks for the feedback. I see your points. We discussed in standup this morning that we will fix point 1 and 2 but prob not point 3 and 4. At the moment, due to technical constraints it might be very hard to fix point 3 and 4, and @comym did agree that we would need to think more about a solution for it, it could be using lazy loading or it could be something else we're not 100% sure yet, but it will be done in a separate ticket.

Huongg:
hey team i've made some fixes for the alignment, let me know what you think? Here is the screenshot of all the 3 runs, i changed the background to be lighter so its easier to notice the difference. 

Screenshot 2022-08-15 at 12 09 59

src/components/experiment-tracking/run-dataset/run-dataset.js
```diff
@@ -40,23 +43,27 @@ const resolveRunDataWithPin = (runData, pinnedRun) => {
...
 }) => {
+  if (!trackingData) {
+    return null;
+  }
+
   return (
-    <div
-      className={classnames('details-dataset', {
-        'details-dataset--single': isSingleRun,
-      })}
```
tynandebold, line 54: You might still need this `details-dataset--single` class name. I'm seeing that the single-run view on your branch is closer together than we we had before (yours is the top, demo is the bottom):

Screen Shot 2022-08-15 at 11 24 04

tynandebold (approved):
Looks and works great! I'm approving with two additional comments:

1. There was some style logic to deal with `single` runs that you've removed. I took a before and after screenshot that I put in the inline comment. Have a look at that. I don't mind how you have it now. Maybe @comym feels differently though.
2. I think we need a little more space between the highest-level accordion sections here (where my red mark is).: 
    Screen Shot 2022-08-15 at 11 22 55

    If you check the demo site there's more space there.

Overall though I think it's ready to go! Well done 💥

tynandebold:
> hey team i've made some fixes for the alignment, let me know what you think? Here is the screenshot of all the 3 runs, i changed the background to be lighter so its easier to notice the difference.
> 
> Screenshot 2022-08-15 at 12 09 59

The third run looks ever so slightly off, though the first two look great. Take a look here (using my Finder window as a ruler):

image

src/components/experiment-tracking/run-dataset/run-dataset.js
Huongg: ah yeah good point, I've used the "comparison view" classnames instead the "single-view". Should be updated now

Huongg:
> Looks and works great! I'm approving with two additional comments:
> 
> 1. There was some style logic to deal with `single` runs that you've removed. I took a before and after screenshot that I put in the inline comment. Have a look at that. I don't mind how you have it now. Maybe @comym feels differently though.
> 2. I think we need a little more space between the highest-level accordion sections here (where my red mark is).:
>    Screen Shot 2022-08-15 at 11 22 55
>    If you check the demo site there's more space there.
> 
> Overall though I think it's ready to go! Well done 💥

hey thank you for taking a look at this. I've addressed point 2, as it was missed a classname thats why there wasnt the right spacing. Thanks for noticing it. 

For point 1, i'm not sure if i understood you correclty. Is it part of this comment? `You might still need this details-dataset--single class name. I'm seeing that the single-run view on your branch is closer together than we we had before (yours is the top, demo is the bottom)` If so i did update the width between comparison view and non comparison view.

rashidakanchwala (approved):
Lookg amazing!! Way to go @Huongg !!

comym:
Hi @Huongg 

Alignment and spaces look alright on my side. The only thing that does not look right is that I'm still seeing run showing big gaps on the layout instead of showing the placeholder rectangle saying "No plots to show"
Screenshot 2022-08-15 at 16 32 50
Screenshot 2022-08-15 at 16 34 39


We should always include it when the run doesn't show a graph for that specific dataset.
Can you please double-check?

Apart from that, I'd like to stress that @Mackay031 points that we are leaving out of this release are very important points and should be addressed as separate design debt tickets. 

**Issue 1** - _The header and contents still move separately, but its better then before when the header and contents were from a different run for a split second._

This one we are going to have to address as a team and we are open to suggestions. We should have a solution for that cause the current microanimation is clearly not working when splitting header from content. This is also related to the delay in loading the data as we've talked about before. Let's split this to a different ticket and find a solution for the next iteration.

**Issue 2 -** Inconsistent behaviour. The way we're moving the columns around - when removing them from comparisson mode - are different. When you remove column 1 for instance, the columns move in a certain way, whereas if you remove column 2 the columns move in a different way. This has nothing to do with the topic of "animation or not animation". This is about keeping a consistent look and feel.

We can document the design debt tickets on Figma as we've been doing and share it with you.

Many thanks

Huongg:
hey @comym, for the first issue, for some reason on the gitpod site, you'd need to change the url to `4142` in order for it to load all the charts (as in the screenshot below). 
Screenshot 2022-08-16 at 10 35 20

For point 2, i can't reproduce it and I guess it could be some issue related to the internet speed, hence at that point it wasnt loading yet. Have you tried it again and it still happens?

Huongg:
I agree with the other points that you and @Mackay031 mentioned, Will get some tickets done for those issues with Tynan 👍

comym (approved):

PR #998, opened 2022-08-01
Remove support for Kedro v16

### LINKED ISSUE #967: Remove support for Kedro v16
Since Kedro is now on v18, we can remove support for Kedro v16. 

Our standard is to support the last two major versions of Kedro.

tynandebold:

## Description

Resolves #967.

Would this be a breaking change if a user is still using v16 of Kedro?

## Development notes

Removed one `else` statement and bumped the version in the requirements file.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/998"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,10 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+## Major features and improvements
+
+- Remove support for Kedro v16. (#998)
+
 # Release 4.7.2
 
 ## Bug fixes and other changes
--- a/package/kedro_viz/integrations/kedro/data_loader.py
+++ b/package/kedro_viz/integrations/kedro/data_loader.py
@@ -123,10 +123,3 @@ def load_data(
                 session_store_location = session_store.location
 
         return context.catalog, context.pipelines, session_store_location
-
-    # pre-0.17 load_context version
-    # pylint: disable=no-name-in-module
-    from kedro.framework.context import load_context  # type: ignore
-
-    context = load_context(project_path=project_path, env=env)
-    return context.catalog, context.pipelines, None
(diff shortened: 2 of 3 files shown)


MerelTheisenQB (approved):
LGTM! 👍  This change should probably also go in the release notes, and yes it's technically a breaking change for anyone who is using viz with Kedro <0.17.0, but if we're confident that (almost) nobody is using such an old version it could go in a non-breaking release.

AntonyMilneQB (approved):
Agree with @MerelTheisenQB re: breaking release. 

Personally I would make a breaking release for this to be on the safe side. Following this change, if a user runs a project with kedro 0.16.x and updates their kedro-viz then not only would kedro-viz not work but it would also update their kedro version to 0.17+, which would  either result in dependency conflict or just break everything for them. Unless it's a very tiny minority of users on 0.16 we should absolutely avoid this.

However, is this change exciting enough to warrant a whole new kedro-viz 5 release?  It's a trivial change to make, and maintaining support for 0.16 isn't at all a burden for kedro-viz at the moment, so I don't feel like we're currently being held back by it. Hence I would put this in a `develop` or `5.0.0` branch and wait for some other changes that would justify a release of v5. Up to you though - there's nothing technically wrong with releasing 5.0.0 just to deprecate a version of kedro. Just it doesn't feel very interesting if the only feature of a big new release is "you can't use kedro 0.16 any more" 😅 

Whatever happens, I think we should update the following note in CONTRIBUTING and RELEASE:
> **Note**: Kedro-Viz>=3.8.0 will not work with projects created with Kedro<=0.16.6. Please consider migrating your project to Kedro>=0.17.0 before you develop against the latest version of Kedro-Viz.

tynandebold:
Great points. Thanks, both.

Antony, the next release will have the plots-in-experiment-tracking work in it (#984). Would you say both of those things together warrant a v5 release?

tynandebold:
Here are v16 downloads of Kedro. Can't say how many of them are Viz users. Probably 20-25%, based on what we know from Heap.

image

PR #1003, opened 2022-08-08
Fix/--save-file when YAML has different parameters. 

### LINKED ISSUE #794: Save file sorting code can cause issues if keys aren't all strings
User reported the following error:

```
    run_server(**run_server_kwargs)
  File "/miniconda/envs/customerone/lib/python3.7/site-packages/kedro_viz/server.py", line 88, in run_server
    Path(save_file).write_text(res.json(indent=4, sort_keys=True)) <-------------ERROR here
  File "pydantic/main.py", line 490, in pydantic.main.BaseModel.json
  File "/miniconda/envs/customerone/lib/python3.7/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  ...
    items = sorted(dct.items(), key=lambda kv: kv[0])
TypeError: '<' not supported between instances of 'str' and 'bool'
kedro.framework.cli.utils.KedroCliError: '<' not supported between instances of 'str' and 'bool'
Run with --verbose to see the full exception
Error: '<' not supported between instances of 'str' and 'bool'
```

This is the offending logic. Having checked with @limdauto this was introduced to ensure that you get the same JSON file every time:

https://github.com/kedro-org/kedro-viz/blob/39524c0f6d323775d891527eea276aa563b3be7f/package/kedro_viz/server.py#L86

@musabilal  reported this issue, my hypothesis is that this was caused by boolean key in some YAML configuration within the project. We weren't able to prove this, but we were able to prove that if we removed '`sort_keys=True`' it would work.

## Steps to Reproduce

I think (but wasn't able to prove) doing parameters like this + trying to save JSON will cause the sort error:

```yaml
true: some string
```

JSON exported to disk

⬆️ Stack trace above

## Proposed solution:

```python
try:
   Path(save_file).write_text(res.json(indent=4, sort_keys=True)) 
except TypeError:
   Path(save_file).write_text(res.json(indent=4, sort_keys=False)) 
```

rashidakanchwala:
## Description

Fixes #794 

If you add an entry in parameters.yml like 

```yml
true: 1
1:1 
```

then JSON response throws an error when you sort_keys because it expects a `str` type key but it gets `boolean` or `number`. 

To fix this we can either:
- stringify the keys first and then save the string keys or else 
- we remove the sort_keys.

Update - we remove the sort keys only if there's a Type error thrown

Please see below API/MAIN response -- the unsorted and sorted

Screenshot 2022-08-08 at 11 58 47

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1003"></a>


tynandebold (approved):
Nice fix! It works for me and saved the file just fine when I added the parameter `true: 1`.

I can't really comment on if we do actually want or need to sort these keys. I didn't notice any difference in the interface when they weren't sorted, so maybe it's not needed. 

@AntonyMilneQB mentioned [here](https://github.com/kedro-org/kedro-viz/issues/794#issuecomment-1078848145) that, "It would be nice if we could keep the json file order the same every time." It's not clear to my why that's needed though. Maybe for diffing purposes? The rendering of the JSON in Viz should always be the same regardless. I'm probably missing something here though.

Anyways, I'll approve it now but please also wait for Antony to review.

AntonyMilneQB:
I don't think it's correct that `sort_keys` only sorts the top level keys:
```
json.dumps({"key": {"b": 1, "a": 1}}, sort_keys=False)
# gives '{"key": {"b": 1, "a": 1}}'

json.dumps({"key": {"b": 1, "a": 1}}, sort_keys=True)
# gives '{"key": {"a": 1, "b": 1}}'
```
... and so `sort_keys` is doing a bit more than it would seem here.

I think it would be nice, but definitely not essential, to sort these keys. As per @tynandebold I don't think it shouldn't affect the functioning of kedro-viz in any way. The reason it would be nice to have is based on what @datajoely  said in #794:
> Having checked with @limdauto this was introduced to ensure that you get the same JSON file every time:

Diffing purposes is the only reason I can think of where this would matter, so it's not too compelling. Just a nice to have really.

Overall I'd opt for the solution Joel suggested which maintain sort order where possible:
```
try:
   Path(save_file).write_text(res.json(indent=4, sort_keys=True)) 
except TypeError:
   # Keys of incomparable types (e.g. string and int) cannot be sorted.
   Path(save_file).write_text(res.json(indent=4, sort_keys=False))
```

I'm also curious though: what is it that stringifies all the key names in the metadata side panel? Where does that happen?

AntonyMilneQB:
@tynandebold those are all string keys anyway though. How about the one with key name `1`?

tynandebold:
Seems like it still comes back the same. They're all strings from the API: 

image

This is my yml:

```yaml
split_options:
  test_size: 0.2
  random_state: 3
  target: 'price'
  1: true
```

AntonyMilneQB:
Ah ok. We don't do this conversion on the backend as far as I can tell, so I think it must be done automatically by FastAPI when the response gets sent to the frontend. But then I would expect the same stringification to happen (or at least to be easy to do) when we save to file, because that's just based on `get_default_response()`. It would be nice to get to the bottom of this...

AntonyMilneQB (approved):

Issue #831, opened 2022-04-19
Databricks and kedro-viz


AntonyMilneQB:
Just writing up where we stand on databricks + kedro-viz. This is partly for the sake of documentation and partly to identify next steps/bugs that need fixing, which will be turned into their own issues in due course.

### Best way to get kedro-viz working
This cunning method piggybacks off shiny and reaches us from databricks via @datajoely - thanks to all involved there!
1. Make sure kedro-viz and kedro are both installed in the same scope (notebook vs. cluster). i.e. if you `%pip install kedro` from inside your notebook then you should also `%pip install kedro-viz` from inside your notebook. If your cluster comes with kedro installed on it as a library already then you should also [add kedro-viz as a cluster library](https://docs.microsoft.com/en-us/azure/databricks/libraries/cluster-libraries)
2. Execute `kedro viz --no-browser --host=0.0.0.0` to launch kedro viz on port 4141 (or some other port if you like, via `--port`). You can do this with `%sh kedro ...` from inside a notebook
3. Run the following cell and click "Shiny app URL" to open a browser window
```
%r
library(shiny)
runExample("01_hello")
```
4. The URL will look something like `https://something.azuredatabricks.net/blah/random/stuff/xxxx/`, where `xxxx` is the port that shiny is using. Change this to `4141` and kedro viz should appear

### Can you use `%run_viz`?
Basically no. You can try to do, but you'll likely need to first:
1. Run `session.close` as in https://github.com/kedro-org/kedro-viz/issues/811. This is the ultimate cause of the mysterious timeout errors we've seen, just the real error message gets swallowed up somewhere
2. If you get some message about `ContextualVersionConflict`, it's due to a conflict between pip-tools and pip. You can either `pip install -U pip` or `pip uninstall pip-tools` (won't work if it's a cluster library). 

After that you'll get a "working" instance of `%run_viz`, but all it will output at the end is `<IPython.core.display.HTML object>`. This renders nicely in Jupyter notebooks but not in databricks.

### Things to fix
- [ ] Doing the above actually gives me a blank kedro-viz screen on the pandas-iris starter, although going to `api/main` shows the data is there. **UPDATE**: this seemed to also be the case on Sagemaker. And in fact it looked like the right data _wasn't_ in fact there if you clicked on nodes and datasets.
[...]

PR #1012, opened 2022-08-12
Fix %run_viz on Databricks

### LINKED ISSUE #1011: %run_viz on Databricks should launch Kedro-viz in a new window
## Description

This is the first part of #910 to get launch_viz/run_viz to work on Databricks

## Context
@AntonyMilneQB has figured a solution to launch Kedro-viz in a new window on Databricks - https://github.com/AntonyMilneQB/kedro-launch-viz/blob/main/kedro_launch_viz/__init__.py

In this ticket, we are going to integrate the above solution in Kedro-viz. 

Note - instead of embedding the viz within the notebook and giving the users a not so great kedro-viz experience we are going to launch it in a new window to give them the full experience

## Checklist

AntonyMilneQB:
## Description
Closes #1011.

To do:
- [x] Fix/write unit tests
- [x] Update RELEASE.md
- [x] ~Test on AWS (and ideally GCP also) if possible~ Failed to do this :( Hopefully it works (I think there's a good chance it does), but if not then anyone who tries it should report it as a bug in future and then we can fix it!
- [x] Update databricks docs on Kedro (no need for the R shiny hack any more! 🎉 )
- [x] Remove `package/kedro_viz/html` after testing and before merging
- [x] Get @jmholzer @noklam @rashidakanchwala @tynandebold to manually test and understand

Screen Recording 2022-08-12 at 9 05 43 am

## QA notes

⚠️ So far tested on Azure; still need to test on AWS (important) and GCP (less important). 

## How to efficiently develop with Kedro-Viz on Databricks

After much trial and error, I have come up with a much more streamlined way to iterate on code being developed for Databricks. This should help to make the development loop much faster since there's no need to restart the cluster or manually handle repos this way 🎉 

1. Make a branch for your work
2. Run `make build`, `git add -f package/kedro_viz/html` and push to GitHub. This is temporarily needed while developing on your branch so that you can `pip install` from GitHub but should _not_ remain there when you merge to `main`
3. On Databricks, make sure that `kedro-viz` and `kedro` are _not_ installed as cluster libraries.
4. In your Databricks notebook, run (fill out `NAME-OF-BRANCH`):
```
%pip uninstall -y kedro-viz
%pip install git+https://github.com/kedro-org/kedro-viz.git@NAME-OF-BRANCH#subdirectory=package
```
> **Warning**. Remember there's quite a bit of confusion around differently-scoped `pip install`ed packages. See https://github.com/kedro-org/kedro-viz/issues/831. In short, use `%pip` (_not_ `%sh pip`) if you want to install notebook-scoped and ensure that `kedro` and `kedro-viz` are installed with the same scope (cluster or notebook).
5. To make a test project if one doesn't already exist:
```
%sh test -d iris || yes "" | kedro new -s pandas-iris
```
6. Then load up the Kedro IPython extension, make sure you're pointing to the right project path and do as you please:
```
%load_ext kedro.extras.extensions.ipython 
%reload_kedro iris
%run_viz
```
7. Whenever you make changes to your branch, all you need to do is push to GitHub and then re-run your notebook. This will `pip install` the latest changes to the branch directly from GitHub. No need to restart the cluster or clone repos any more.
8. Make sure you remove the `package/kedro_viz/html` folder before merging to `main`.

> **Note**. It seems like using the Databricks repos feature would be a smoother development process, but it's not. Every time you make a change to your branch you would need to pull the repo and reinstall on cluster, which means restarting the cluster every time (=slow). So don't try doing it that way...

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1012"></a>


package/kedro_viz/launchers/jupyter.py
```diff
@@ -112,30 +152,32 @@ def run_viz(port: int = None, line=None, local_ns=None) -> None:
...
+        _display_databricks_html(port)
+    else:
+        wrapper = f"""
+                <html lang="en"><head></head><body style="width:100; height:100;">
+                <iframe src="http://{host}:{port}/" height=500 width="100%"></iframe>
+                </body></html>"""
+        display(HTML(wrapper))
+
+
+# TODO: add arguments
```
AntonyMilneQB, line 183:
```suggestion
```

package/kedro_viz/launchers/jupyter.py
```diff
@@ -89,15 +91,53 @@ def _allocate_port(start_at: int, end_at: int = 65535) -> int:
...
     )
 
 
+def _is_databricks() -> bool:
+    return "DATABRICKS_RUNTIME_VERSION" in os.environ
+
+
+def _get_databricks_object(name: str):
+    """Gets object called `name` from the user namespace."""
+    return IPython.get_ipython().user_ns.get(name, None)
```
AntonyMilneQB, line 109:
```suggestion
    return IPython.get_ipython().user_ns.get(name)
```

package/kedro_viz/launchers/jupyter.py
```diff
@@ -89,15 +91,53 @@ def _allocate_port(start_at: int, end_at: int = 65535) -> int:
...
+    return f"https://{browser_host_name}{path_name}"
+
+
+def _display_databricks_html(port: int):
+    url = _make_databricks_url(port)
+    displayHTML = _get_databricks_object("displayHTML")  # pylint: disable=invalid-name
+    if displayHTML is not None:
+        displayHTML(f"""<a href="{url}">Open Kedro-Viz</a >""")
+    else:
+        print(f"Kedro-Viz is available at {url}")
```
AntonyMilneQB, line 138:
```suggestion
        displayHTML(f"""<a href="{url}">Open Kedro-Viz</a>""")
    else:
        print(f"Kedro-Viz is available at {url}")
```

noklam:
@AntonyMilneQB I vaguely remember some discussion about using `jupyter-proxy` or something similar but seems this is not the final solution. Could you briefly explain how this roughly works now?

AntonyMilneQB:
> @AntonyMilneQB I vaguely remember some discussion about using `jupyter-proxy` or something similar but seems this is not the final solution. Could you briefly explain how this roughly works now?

Yes, absolutely! I will explain more in the meeting this afternoon but in short `jupyter-server-proxy` does not seem to be the right approach on Databricks (it will still be useful elsewhere though, e.g. probably Sagemaker). Databricks already has a proxying mechanism built in. This is actually exactly what the [R shiny workaround](https://kedro.readthedocs.io/en/stable/deployment/databricks.html#running-kedro-viz-on-databricks) relied on, and what we're doing now is very similar to that but less hacky. The only differences are:
* we programmatically find the URL (`_make_databricks_url`) so there's no need to actually launch R shiny to find the right URL
* we launch kedro viz as a subprocess rather than `%sh kedro viz`, which is much cleaner
* these are built into `%run_viz` and automatically done when you're on Databricks

package/kedro_viz/launchers/jupyter.py
```diff
@@ -89,15 +91,53 @@ def _allocate_port(start_at: int, end_at: int = 65535) -> int:
...
+    url = _make_databricks_url(port)
+    displayHTML = _get_databricks_object("displayHTML")  # pylint: disable=invalid-name
+    if displayHTML is not None:
+        displayHTML(f"""<a href="{url}">Open Kedro-Viz</a>""")
+    else:
+        print(f"Kedro-Viz is available at {url}")
+
+
 # pylint: disable=unused-argument,missing-type-doc
 def run_viz(port: int = None, line=None, local_ns=None) -> None:
```
noklam, line 142: The `local_ns` is not used anywhere.

noklam (approved):
Nice work, really tricky to get the Databricks stuff work. I have tested  on databricks and it works correctly.

package/kedro_viz/launchers/jupyter.py
jmholzer: Are `line` and `local_ns` required in the function signature so that this works as a line magic, though they are unused?

jmholzer (approved):
This all looks great to me, I tested it on Databricks on Azure without problems. Still awaiting the resources to do a check on AWS.

package/kedro_viz/launchers/jupyter.py
AntonyMilneQB: Yeah, I think what @jmholzer says here is right. I tried removing them before and IPython raised some sort of "invalid line magic" exception.

noklam: This isn't documented anywhere, but I think this is the bit why `local_ns` is needed, I am not sure what it does. This is also an unofficial way to extend `kedro`'s line magic by adding things into `kedro.line_magic` in `setup.py`

```
    for line_magic in load_entry_points("line_magic"):
        register_line_magic(needs_local_scope(line_magic))
        logger.info("Registered line magic '%s'", line_magic.__name__)  # type: ignore
```

AntonyMilneQB: Ah yes, you're right. The `local_ns` is required due to the `needs_local_scope` on the kedro ipython extension side. This means _any_ line magics registered with the `kedro.line_magic` entrypoint would need to have this argument.

Previously this was actually used to get the project path as in `local_ns["project_path"]`. In fact, I'd hazard a guess that was why `needs_local_scope` was added in the kedro ipython extension in the first place. Definitely doesn't feel right to me that we're forcing all line magics registered with `kedro.line_magic` to require that argument. And, as you say, this is completely undocumented (as is the existence of that entrypoint to begin with...)

To be honest, now I think about it again I'm not sure why we even expose this entrypoint rather than just letting extensions (like kedro-mlflow, which is probably the only thing besides kedro-viz that has ever used this entrypoint) register their own line magics as in `%load_ext kedro_mlflow`. I guess it's slightly more convenient since then you don't need to do multiple `%load_ext`, but it feels a bit weird to me... Let me make a separate issue for it on the kedro repo anyway.

AntonyMilneQB: @noklam I've removed `line` as an argument since it wasn't doing anything. In also looked like you could do `%run_viz 5000` before to set a custom port, but actually that wouldn't have worked because `port` would be the string `"5000"`.

We've never advertised `%run_viz` as taking arguments (and until now this wouldn't have worked). I'm not going to document or advertise it because I don't want to have to maintain it as a feature, since ultimately we probably want `%run_viz` to properly handle many arguments in the same sort of way that `%reload_kedro` can now. It might be handy to have this available e.g. if databricks start locking down ports of something. So let's bear in mind in case anyone wants to do it on databricks, for now at least you can do `%run_viz 5000`... but this is not officially condoned since it's kind of an experimental feature anyway 😀

AntonyMilneQB: FYI @jmholzer ☝️

AntonyMilneQB: Also FYI for @noklam: https://github.com/kedro-org/kedro/issues/878.

src/components/flowchart/styles/_variables.scss
```diff
@@ -48,7 +48,7 @@
   --node-labeltext-fill: #{$color-text-dark};
   --edge-stroke: #666;
   --edge-arrowhead-fill: #cacaca;
-  --layer-fill: #072030;
+  --layer-fill: red;
```
AntonyMilneQB, line 51:
```suggestion
  --layer-fill: #072030;
```

PR #1015, opened 2022-08-16
Feature/cleanup metadata panel

### LINKED ISSUE #962: Clean up metadata side panel for datasets
There's several small niggles I've always had with the metadata panel for datasets.
image

1. Why do we show `Type: dataset` at all (also applies to nodes)? It's obvious from the icon and information provided that's what it is.
2. Dataset Type is always too long to fit in, when really all that people care about is the last part after the final `.`. Can we make it so that we only show that part but then on mouseover it shows the whole path? (The full path currently shows on mouseover)
3. Similarly (though more debatable) for File Path.
4. Why do we need `Pipeline` field? Is it just the same as the registered pipeline selected?

tynandebold:
## Description

Resolves #962.

## Development notes

This PR achieves these three things:

1. `Dataset Type`: show only the part after the final `.` but allow the whole path to be visible when hovering over the text.
2. Remove the trailing slash from the `File Path` text field.
3. Remove the `Pipeline` field, as it's always precisely the same as the registered pipeline dropdown value.

## QA notes

Take a look at the GitPod link and check against the three points in the section above.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1015"></a>


cvetankanechevska (approved):
Looks good to me!

rashidakanchwala (approved):
LGTM! thanks!

PR #1030, opened 2022-08-23
Add nice plots to kedro demo website for experiment tracking

### LINKED ISSUE #1009: Create nice plots for the kedro demo website for experiment tracking
## Description

This ticket is to add some really nice and attractive plots/images to our demo website to show how powerful the new experiment tracking features are. 

## Context
Doing this will help us demo the best of Kedro-viz to our users and show them what is possible with plot intergration. 

## Possible Implementation
Work with design/data science to come up with nice sensible charts. 
Also loop in with vizx team to see if we can use some of the plots/charts they have.  

## Checklist

rashidakanchwala:
## Description

Resolves #1009 

## Development notes

As discussed with @noklam , I have done the below 
- replace the bar cancellation breakdown bar chart with a horizontal bar chart on feature importance. The feature scores keep changing randomly for every run 
- the cancellation breakdown chart still exists but it is not versioned hence it doesn't show up in the experiment tracking and only in the flowchart 
- the matplotlib single line plot is now replaced with confusion matrix which also changes randomly for each run. The confusion matrix uses seaborn which needs to be pip installed. I have added it to the demo-project requirements.txt 
- the price histogram remains the same, though i have made sure that the plot keeps changing randomly for every run. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1030"></a>


demo-project/src/demo_project/pipelines/reporting/pipeline.py
```diff
@@ -20,7 +21,7 @@ def create_pipeline(**kwargs) -> Pipeline:
             node(
                 func=make_cancel_policy_bar_chart,
                 inputs="prm_shuttle_company_reviews",
-                outputs="cancellation_policy_breakdown",
+                outputs="rcancellation_policy_breakdown",
```
noklam, line 24: typo?

demo-project/src/demo_project/pipelines/reporting/nodes.py
```diff
@@ -90,6 +95,36 @@ def make_price_analysis_image(model_input_table: pd.DataFrame) -> PIL.Image:
...
+    random_actuals = []
+    random_predicted = []
+    for i in range(0, 11):
+        actual = random.randint(0, 1)
+        predicted = random.randint(0, 1)
+        random_actuals.append(actual)
+        random_predicted.append(predicted)
+
+    data = {"y_Actual": random_actuals, "y_Predicted": random_predicted}
+    plt.style.use("dark_background")
```
noklam, line 120: Is this because it looks nicer on the website? It's pretty minor but I found the color scheme a bit unnatural

demo-project/src/demo_project/pipelines/reporting/nodes.py
```diff
@@ -90,6 +95,36 @@ def make_price_analysis_image(model_input_table: pd.DataFrame) -> PIL.Image:
...
+    df = pd.DataFrame(data, columns=["y_Actual", "y_Predicted"])
+    confusion_matrix = pd.crosstab(
+        df["y_Actual"],
+        df["y_Predicted"],
+        rownames=["Actual"],
+        colnames=["Predicted"],
+        margins=True,
+    )
```
noklam, line 128: I can't leave comment on the confusion matrix, I think it should just have [0,1] there, `all` doesn't fit in there it could be just a 2x2 matrix.

noklam (commented):
Thanks for making this PR! Could you also add a screenshot of how the `feature importance` chart looks like? From GitHub I can only see the JSON.

demo-project/src/demo_project/pipelines/reporting/nodes.py
rashidakanchwala: yes, it is just so that it looks good on Kedro-viz dark theme

demo-project/src/demo_project/pipelines/reporting/nodes.py
rashidakanchwala: ok makes sense, i will convert it to 2x2 then

tynandebold (approved):
These look sublime, great work! Approving with two comments:

- On the flowchart, should the `reporting.cancellation_policy_breakdown` have a plot? It did before this change and still have the chart icon on the node, but there's no plot if the Metadata panel.
- When I click on the `reporting.cancellation_policy_grid` node I see this error warning in the server logs:
    ```
    2022-08-25 15:12:15,644 - kedro.io.core - WARNING - 'exists()' not implemented for 'ImageDataSet'. Assuming output does not exist.
    ```

rashidakanchwala:
> These look sublime, great work! Approving with two comments:
> 
> * On the flowchart, should the `reporting.cancellation_policy_breakdown` have a plot? It did before this change and still have the chart icon on the node, but there's no plot if the Metadata panel.
> * When I click on the `reporting.cancellation_policy_grid` node I see this error warning in the server logs:
>   ```
>   2022-08-25 15:12:15,644 - kedro.io.core - WARNING - 'exists()' not implemented for 'ImageDataSet'. Assuming output does not exist.
>   ```

Thanks for spotting this. I did on purpose to have a case where you have plots but you have not versioned them so they don't show up in experiment tracking. But this is also a bug. I will work on fixing this or maybe temporarily remove it and add a ticket to fix it.

demo-project/src/demo_project/pipelines/feature_engineering/nodes.py
```diff
@@ -99,3 +100,19 @@ def joiner(spine_df: pd.DataFrame, *dfs: pd.DataFrame) -> pd.DataFrame:
     # Confirm that the number of rows is unchanged after the operation has completed
     assert spine_df.shape[0] == merged_dfs.shape[0]
     return merged_dfs
+
+
+def create_feature_importance(data: pd.DataFrame):
```
AntonyMilneQB, line 105:
```suggestion
def create_feature_importance(data: pd.DataFrame) -> pd.DataFrame:
```

demo-project/src/demo_project/pipelines/feature_engineering/nodes.py
```diff
@@ -99,3 +100,19 @@ def joiner(spine_df: pd.DataFrame, *dfs: pd.DataFrame) -> pd.DataFrame:
...
+    feature_name = []
+    feature_score = []
+    for i in range(0, 15):
+        feature_name.append("feature_" + str(i))
+        feature_score.append(random.randint(0, 999) / 1000)
+    feature_importance_df = pd.DataFrame(
+        {"Features": feature_name, "Score": feature_score}
```
AntonyMilneQB, line 112:
```suggestion
        {"Feature": feature_name, "Score": feature_score}
```

demo-project/src/demo_project/pipelines/feature_engineering/nodes.py
```diff
@@ -99,3 +100,19 @@ def joiner(spine_df: pd.DataFrame, *dfs: pd.DataFrame) -> pd.DataFrame:
...
+    )
+    random_number = random.randint(0, 100) / 200 + random.randint(0, 100) / 200
+    feature_importance_df["Score"] = feature_importance_df["Score"].apply(
```
AntonyMilneQB, line 115: I don't understand why there's so many different `random.randint`s in this function to generate a single random number? 🤔 Why do we need anything more than just the original `feature_score` random number on L110?

demo-project/src/demo_project/pipelines/reporting/nodes.py
```diff
@@ -56,7 +58,10 @@ def make_price_histogram(model_input_data: pd.DataFrame) -> go.Figure:
         BaseFigure: Plotly object which is serialised as JSON for rendering
     """
     price_data_df = model_input_data[["price", "engine_type"]]
-
+    random_limit = random.randint(0, random.randint(5, 9)) * 1000
+    price_data_df["price"] = price_data_df["price"].apply(
+        lambda x: (x + random.randint(0, random_limit) * random.randint(0, 1))
```
AntonyMilneQB, line 63: Similarly I'm a bit confused here about why we need to combine different numbers. What are we actually trying to achieve for the final `price` column?

demo-project/src/demo_project/pipelines/reporting/nodes.py
```diff
@@ -90,6 +95,35 @@ def make_price_analysis_image(model_input_table: pd.DataFrame) -> PIL.Image:
     return pil_table.image
 
 
+def create_feature_importance_plot(features: pd.DataFrame):
```
AntonyMilneQB, line 98: Type hint for return type here would be good! Similarly L110.

demo-project/src/demo_project/pipelines/reporting/nodes.py
```diff
@@ -90,6 +95,35 @@ def make_price_analysis_image(model_input_table: pd.DataFrame) -> PIL.Image:
...
+    sorted_features = features.sort_values(by="Score", ascending=True)
+    plotly_object = px.bar(
+        sorted_features,
+        x="Score",
+        y="Features",
```
AntonyMilneQB, line 103:
```suggestion
        y="Feature",
```

demo-project/src/demo_project/pipelines/reporting/nodes.py
```diff
@@ -90,6 +95,35 @@ def make_price_analysis_image(model_input_table: pd.DataFrame) -> PIL.Image:
...
+
 def create_matplotlib_chart(companies: pd.DataFrame):
-    plt.plot([1, 2, 3], [4, 5, 6])
+    random_actuals = []
+    random_predicted = []
+    for i in range(0, 11):
+        actual = random.randint(0, 1)
+        predicted = random.randint(0, 1)
+        random_actuals.append(actual)
+        random_predicted.append(predicted)
```
AntonyMilneQB, line 117: This sort of thing can be done much more concisely as a list comprehension:
```
    random_actuals = [random.randint(0, 1) for _ in range(11)]
```

Even better, because we can use numpy here you can avoid and lists and just vectorise the whole thing:
```
    random_actuals = np.random.randint(2, size=11)
```

demo-project/src/docker_requirements.txt
```diff
@@ -1,3 +1,4 @@
 kedro[pandas.CSVDataSet,pandas.ExcelDataSet, pandas.ParquetDataSet, plotly.PlotlyDataSet, matplotlib.MatplotlibWriter]==0.18.1
 scikit-learn~=1.0
 pillow~=9.0
+seaborn==0.11.2
```
AntonyMilneQB, line 4:
```suggestion
seaborn~=0.11.2
```

AntonyMilneQB (approved):
These look much more interesting than the old plots, thank you!

The code that generates the plots currently feels not very Pythonic/pandorable to me. It's not a big problem because it's only needed to generate some nice looking plots for the demo, but it might be a good exercise to try and improve it 🙂 e.g. use `numpy` rather than Python's `random` module: https://stackoverflow.com/questions/32752292/how-to-create-a-dataframe-of-random-integers-with-pandas

demo-project/src/demo_project/pipelines/reporting/nodes.py
rashidakanchwala: done.thanks for this

demo-project/src/demo_project/pipelines/feature_engineering/nodes.py
rashidakanchwala: you are right. initially I was reading from a csv hence I had the randomness and then i just created an in memory array. so i don't need to repeat it. i have removed this

comym (approved):

demo-project/src/demo_project/pipelines/reporting/nodes.py
```diff
@@ -56,7 +58,12 @@ def make_price_histogram(model_input_data: pd.DataFrame) -> go.Figure:
         BaseFigure: Plotly object which is serialised as JSON for rendering
     """
     price_data_df = model_input_data[["price", "engine_type"]]
-
+    p1 = np.random.randint(100, 300) / 1000
+    p2 = np.random.randint(100, 400) / 1000
```
AntonyMilneQB, line 62:
```suggestion
    p1 = np.random.uniform(0.1, 0.3)
    p2 = np.random.uniform(0.1, 0.4)
```

Or you could just generate a 3-dimensional probability distribution like this:
```
p = np.random.dirichlet([1, 1, 1])
```
(FYI this works because a Dirichlet can be used to generate set of probability distributions. Try it out to see what that does. `sum(p)` will always be 1 with each value value positive. This is basically what defines a probability distribution)

cvetankanechevska (approved):

PR #1041, opened 2022-08-31
Feature/lazy loading

### LINKED ISSUE #1005: Explore lazy loading results in the experiment tracking view
## Description
Related to #932 

Sometimes when the data is loading, or when the internet speed is not quite good, our viz page can look quite jumpy (with the animation), and it's hard to tell if something went wrong or just because the internet is bad and the content is still loading 😄. For example with the plot graphs, it can take quite a bit of time for it to load. 

To improve users we can create some simple lazy loading for the page while the data is still loading

## Possible Implementation
I've used [react-content-loader](https://www.npmjs.com/package/react-content-loader). The package is very small and very easy to implement. You can simply create a loader component for each main component (section). 

It's a mix of different shapes using svg, something as simple as the below

```
const MyLoader = () => (
  <ContentLoader viewBox="0 0 380 70">
    {/* Only SVG shapes */}    
    <rect x="0" y="0" rx="5" ry="5" width="70" height="70" />
    <rect x="80" y="17" rx="4" ry="4" width="300" height="13" />
    <rect x="80" y="40" rx="3" ry="3" width="250" height="10" />
  </ContentLoader>
)
```

## Possible Alternatives
I haven't used any other tool yet so open to more suggestions from the team

## Checklist

Huongg:
## Description

Fixes #1005 

## Development notes

I've created 2 loaders components for both meta-data and tracking-data. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1041"></a>


comym:
hi @Huongg just had a look at it now

 It was actually hard to keep the lazy loading running for long to observe it in detail even when switching to a slow 3g connection.

It looks alright to me and my two comments are:

We should try to make a better alignment/positioning of the rectangles to align as perfect as possible to the text data positioning when it appears.

The screenshot below shows the two overlayed. You can see they are not aligned and what happens is when we suddenly move from rectangles to text, there's a jump in positioning. Basically, the expectation of something appearing at a certain position is not matched and there's the impression that the content is moving. I'm aware the rectangles are just a representation of possible content with lines and spacing in between and might never match it perfectly depending on the data we're loading. However, especially the vertical alignment of the columns is important for us to nail.

Have a look:

Screenshot 2022-09-01 at 13 08 44
Screenshot 2022-09-01 at 13 08 35
Screenshot 2022-09-01 at 13 08 27


Lastly, the colours of the pulsating rectangle are alright, however the fact that it's visible every time you click on a run for a split second makes everything in my opinion annoyingly jumpy. Cause it ends up not being something that only occasionally happens, but it happens every time. The only time that it does not happen is when you go back to a run that you actually loaded previously because it's already loaded.

I thought making the lazy loading rectangles have a more subtle colour would help at minimizing this effect and make everything look less jumpy, but I don't think it worked. We can keep the colours as it is for now. I do think it's better though. The contrast between background and foreground is less now and better for the eyes.

I don't know how to proceed from here. If going back to the idea that @tynandebold mentioned before, of creating some sort of delay - in a less hacky way - so users don't see the lazy loading appearing every single time you click on something, but only when is really necessary.

Technically, I leave it to you guys, but that would be the way to go in my opinion. Let's keep the conversation going and I'm happy to hear from the team what the best solution could be. Thanks!

rashidakanchwala (approved):
Looks great, i can see the alignment is correct in comparison mode. thanks @Huongg

src/components/experiment-tracking/run-dataset/run-dataset-loader.js
```diff
@@ -0,0 +1,56 @@
...
+      <rect width="50" height="16" x={x} y={y + GAP} />
+      <rect width="180" height="16" x={x} y={y + GAP * 2} />
+    </>
+  );
+};
+
+export const RunDatasetLoader = ({ theme }) => (
+  <div className="details-dataset">
+    <ContentLoader
+      viewBox="0 0 1000 1000"
```
tynandebold, line 31: Does the height need to be so tall here? I see a lot of empty space at the bottom of the page. I think you can get away with `625` or so.

```suggestion
      viewBox="0 0 1000 625"
```

src/components/experiment-tracking/run-dataset/run-dataset.js
```diff
@@ -1,15 +1,40 @@
...
+        theme === 'dark'
+          ? experimentTrackingLazyLoadingColours.foregroundDarkTheme
+          : experimentTrackingLazyLoadingColours.foregroundLightTheme
+      }
+      speed={2}
+    >
+      <rect width="150" height="16" x={x} y={y + length * 2} />
+    </ContentLoader>
+  );
+};
```
tynandebold, line 36: What's the thinking behind composing this component in this file? To me, it would be more natural if this lived in your `/run-dataset/run-dataset-loader.js` file. Then you'd just have one new import in this file and all the new loader composition is kept separate.

src/components/experiment-tracking/run-dataset/run-dataset.js
```diff
@@ -134,6 +163,7 @@ const RunDataset = ({
  * @param {function} setShowRunPlotsModal callbak function to show runplot modal
  */
 function buildDatasetDataMarkup(
+  index,
```
tynandebold, line 166: Are you using this anywhere? It doesn't seem like you are and if not you can remove it.

src/components/experiment-tracking/run-dataset/run-dataset.js
```diff
@@ -162,25 +194,30 @@ function buildDatasetDataMarkup(
...
-                      index === 0 && enableComparisonView,
-                  })}
+            {datasetValues.map((data, index) => {
+              return (
+                <CSSTransition
+                  key={data.runId}
+                  timeout={300}
+                  classNames="details-dataset__value-animation"
+                  enter={isSingleRun ? false : true}
+                  exit={isSingleRun ? false : true}
```
tynandebold, line 204: I don't see anything new in this section or the new code below, so you could revert back to the implicit return if you'd like.

src/components/experiment-tracking/run-metadata/run-metadata.js
```diff
@@ -9,10 +10,43 @@ import KebabIcon from '../../icons/kebab';
...
+            }
+            speed={2}
+          >
+            <MetadataLoader x={x} />
+          </ContentLoader>
+        </td>
+      </tr>
+    </tbody>
+  );
+};
```
tynandebold, line 48: Same comment here as my earlier one: I think it makes more sense to compose this in your loader file and then import and use it here.

src/components/experiment-tracking/details/details.js
```diff
@@ -45,6 +70,19 @@ const Details = ({
...
 
+  if (showSingleRunLoader) {
+    return (
+      <div
+        className={classnames('kedro', 'details-mainframe', {
+          'details-mainframe--sidebar-visible': sidebarVisible,
+        })}
+      >
+        <RunMetadataLoader theme={theme} />;
+        <RunDatasetLoader theme={theme} />;
```
tynandebold, line 81:
```suggestion
        <RunMetadataLoader theme={theme} />
        <RunDatasetLoader theme={theme} />
```

Those semicolons are rendering.

tynandebold (changes requested):
This is looking great and really brings the feel of the app up to a new level! 

I'm going to request changes instead of approving just because of the number of comments I have. None of them are major though, so they should be able to be remedied without much effort.

A few other points I noticed while testing:

1. At one point I saw the tracking data loader rows were very short:
    Screen Shot 2022-09-05 at 14 01 31
2. In comparison mode, the metadata and tracking data aren't lining up. Maybe this is still a WIP?
    Screen Shot 2022-09-05 at 17 42 19
3. The colors of the light theme have very subtle and for me a hard-to-see contrast. Should we tweak that?
4. I would speed up the animation. Maybe just use the default value, which is think is 1.2, unless design has specifically asked for this :)

src/components/experiment-tracking/run-dataset/run-dataset-loader.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import ContentLoader from 'react-content-loader';
+import { experimentTrackingLazyLoadingColours } from '../../../config';
+
+import './run-dataset.css';
+
+const GAP = 36;
```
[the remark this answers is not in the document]
tynandebold: That way we only have to maintain this in one place :)

src/components/experiment-tracking/details/details.js
```diff
@@ -26,10 +28,33 @@ const Details = ({
...
+      } else {
+        setShowLoader(false);
+      }
+
+      if (isRunDataLoading && !enableComparisonView) {
+        setShowSingleRunLoader(true);
+      } else {
+        setShowSingleRunLoader(false);
+      }
+    }, 100);
```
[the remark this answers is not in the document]
Huongg: hey yeah sorry i might have tried 1s then forgot to update the comment. i think 100 is good enough 👍

src/components/experiment-tracking/details/details.js
```diff
@@ -26,10 +28,33 @@ const Details = ({
...
   showRunPlotsModal,
   sidebarVisible,
+  isRunDataLoading,
+  newRunAdded,
   theme,
 }) => {
   const [runMetadataToEdit, setRunMetadataToEdit] = useState(null);
   const [runDatasetToShow, setRunDatasetToShow] = useState({});
+  const [showSingleRunLoader, setShowSingleRunLoader] = useState(false);
+  const [showLoader, setShowLoader] = useState(false);
```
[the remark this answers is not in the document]
Huongg: hey good point, i renamed it `showRunLoader` what do you think of the name?

src/components/experiment-tracking/run-dataset/run-dataset.js
Huongg: originally i was thinking it should live inside the parent component as it only uses there but what you suggested also makes sense. have moved to the loader component

src/components/experiment-tracking/run-dataset/run-dataset.js
Huongg: good spot, i dont think i need to use it anymore. 👍

src/components/experiment-tracking/run-dataset/run-dataset-loader.js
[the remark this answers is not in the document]
Huongg: 👍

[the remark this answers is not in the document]
Huongg: I decided to move to the config file actually. As I store other colours values there too

Huongg:
> 3\. The colors of the light theme have very subtle and for me a hard-to-see contrast

Hiya, thanks for your comments. For points 1 and 2, I didn't push my the latest last night so what you saw was prob the older version, sorry. I've asked Gabriel about the colour you mentioned in point 3, happy to tweak it. And I also remove the speed to set it back to the default value. 

For other comments, I've addressed them and left some comments with my changes 😄

src/components/experiment-tracking/run-dataset/run-dataset-loader.js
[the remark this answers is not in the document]
tynandebold: Good idea.

src/components/experiment-tracking/details/details.js
```diff
@@ -26,10 +28,33 @@ const Details = ({
   showRunExportModal,
   showRunPlotsModal,
   sidebarVisible,
+  isRunDataLoading,
+  newRunAdded,
```
tynandebold, line 32: Nit: sorting.

src/components/experiment-wrapper/experiment-wrapper.js
```diff
@@ -193,6 +197,8 @@ const ExperimentWrapper = ({ theme }) => {
               theme={theme}
               showRunExportModal={showRunExportModal}
               setShowRunExportModal={setShowRunExportModal}
+              isRunDataLoading={isRunDataLoading}
+              newRunAdded={newRunAdded}
```
tynandebold, line 201: Nit: sorting.

tynandebold (approved):
Fantastic work! Thank you for making the changes.

Huongg:
thank you @tynandebold. 

@comym just reminded me about removing the sliding effect...I completely forgot we agreed we would do that. I've removed the sliding in effect to make it work nicer with the lazy loading now.

tynandebold:
> thank you @tynandebold.
> 
> @comym just reminded me about removing the sliding effect in the comparison view...I completely forgot we agreed we would do that. I've removed the sliding in effect to make it work nicer with the lazy loading now.

I actually think we should keep the sliding effect. Given that it's a safe bet the user is very rarely going to see the loading components, don't we still want the new data to slide in when it appears?

In those rare instances when something is taking ages to load, yes the user will see the loading indicators and then the data will slide in, but that will be a very small percentage of total views. Again, the majority of the time we won't see the loading indicators, and if we don't see them and then we remove the slide in, we lose that lovely animation and moment of delight.

RELEASE.md
```diff
@@ -17,6 +17,7 @@ Please follow the established format:
 - Upgrade to `dart-scss` and update the recommended Node version to v16. (#1026)
 - Fix the export functionality of experiment tracking data. (#1033)
 - Color cleanup and refactor. (#1007)
+- Lazy loading for the experiment tracking. (#1041)
```
tynandebold, line 20:
```suggestion
- Add lazy loading to experiment tracking. (#1041)
```

Can you also remove the line above please? This one:

```
- Color cleanup and refactor. (#1007)
```

We don't need to add that because it's not something a user would see or notice.

Huongg: oh do we not need to mention it? if so i can remove it 👍 thanks for letting me know

tynandebold: No we don't need to mention the color work in the release notes. Only things that would be noticeable to the user or allow them to change the way they interact with Viz (a new Node version, for example) should go in there.

cc @cvetankanechevska

comym:
Hi, I've just checked it with the increased delay time and the sliding effect added back

Let's keep the sliding effect, as it is. Let's keep it.

However I still see the header and content columns moving up and down as described here 
https://github.com/kedro-org/kedro-viz/issues/1005#issuecomment-1238180258

Pretty much there
That's all for now, many thanks

Huongg:
awesome thank you @comym @cvetankanechevska and @tynandebold. Lets keep the sliding in effect then. I've updated the delay timing to 200ms so the gitpod looks in sync with our local version, but also not too slow for data to load

For your feedback about the layout shifting @comym, thanks for noticing it. My latest changes should address them now. You can now review again on the gitpod 😄

src/config.js
```diff
@@ -23,6 +23,15 @@ export const codeSidebarWidth = {
   closed: 0,
 };
 
+export const experimentTrackingLazyLoadingColours = {
+  backgroundLightTheme: '#DADADA',
+  foregroundLightTheme: '#EAEAEA',
+  backgroundDarkTheme: '#071d28',
+  foregroundDarkTheme: '#20313a',
+};
```
cvetankanechevska, line 31: Just one small adjustment, is this formatted with prettier? If so I think all color values will be lower cases.

cvetankanechevska:
> awesome thank you @comym @cvetankanechevska @rashidakanchwala and @tynandebold. Lets keep the sliding in effect then. I've updated the delay timing to 200ms so the gitpod looks in sync with our local version, but also not too slow for data to load
> 
> For your feedback about the layout shifting @comym, thanks for noticing it. My latest changes should address them now. You can now review again on the gitpod 😄

Great work Huong! 🎉

comym (approved):
Hi @Huongg looks and feels great. I'm approving it.

As the last thing, I just realised that we were not using our current colour palette for the lazy loading gradient colours. 

Please make sure to follow:

Dark theme:
Start: Slate-600
End: Slate-200

Light theme:
Start: Grey-200
End: Grey-100

Many thanks :)

src/config.js
Huongg: ah yeah thanks for spotting it, im meant to update with your colour variables but completely forgot again.

Issue #1043, opened 2022-09-02
Handle ImageDataSets on Kedro-viz


rashidakanchwala:
## Description

We recently started supporting Matplotlib plots on Kedro-viz which are basically static pngs and use dataset MatplotlibWriter. It makes sense to extend this support to other kedro datasets like ImageDataSet which also use png.

Recently, we also implemented a new load method for plotly, matplotlib and tracking datasets. At the moment, if you try to click on an ImageDataSet it gives you the following Warning 

`kedro.io.core - WARNING - `exists()` not implemented for `ImageDataSet`. Assuming output does not exist.` 

Maybe implementing the above will also fix this issue. 

To replicate the above error -- run the kedro-viz demo project locally and click on cancellation_policy_grid node. 

## Context
People who use ImageDataSets can also start seeing their images on Kedro-viz

## Checklist

- [x] Include labels so that we can categorise your feature request

AntonyMilneQB:
I think this is a good idea and could be easily added as a Quick Win, but I'm pretty sure that `ImageDataSet` is way less popular than the ones we've added the previews for already, so I don't think it is high priority. We could easily implement it following the current scheme, but the code would add another case to the logic in `flowchart.DataNode`, which is already quite messy.  As per https://github.com/kedro-org/kedro-viz/issues/980 and the comments in code that say ` TODO: improve this scheme`, I don't like how we currently have lots of different switches for `self.dataset_type == "kedro.extras.datasets.matplotlib.matplotlib_writer.MatplotlibWriter"` etc. So maybe we should figure out a better way to refactor all this. Or, even more ambitious, we would try to solve the more general issue here: https://github.com/kedro-org/kedro-viz/issues/907.

So yeah, definitely makes sense to add this. It's just a question of whether we want to add it as a quick fix or figure out a more general solution.

Also note:
* the `WARNING` message should no longer be emitted after #1044 
* the demo-project uses a custom `demo_project.extras.datasets.image_dataset.ImageDataSet` which is not the same as kedro's builtin dataset type and crucially doesn't expose an `exists` method. So if we implement this for `ImageDataSet` then we'd need to change the dataset type to the builtin one for the image to show up

rashidakanchwala:
I think what you said makes a lot of sense. I will also leave this issue to be reviewed for backlog grooming but I also think ImageDataSets minor. Maybe we can create a case for this issue when there's user demand !!

PR #1044, opened 2022-09-05
Remove "exists() not implemented" warning

AntonyMilneQB:
## Description

This will remove the WARNING message written about in https://github.com/kedro-org/kedro-viz/issues/1043. We no longer check for the existence of a dataset if it's not a type that we're trying to load.

This is just a quick fix without any tests since, as per my comments in #1043, this whole `DataNodeMetadata` loading scheme is a bit of a mess and ripe for a refactor and improved tests.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1044"></a>

### DIFF
--- a/package/kedro_viz/models/flowchart.py
+++ b/package/kedro_viz/models/flowchart.py
@@ -580,6 +580,14 @@ class DataNodeMetadata(GraphNodeMetadata):
         if not data_node.is_free_input:
             self.run_command = f'kedro run --to-outputs="{data_node.full_name}"'
 
+        # Only check for existence of dataset if we might want to load it.
+        if not (
+            data_node.is_plot_node()
+            or data_node.is_image_node()
+            or data_node.is_tracking_node()
+        ):
+            return
+            
         # dataset.release clears the cache before loading to ensure that this issue
         # does not arise: https://github.com/kedro-org/kedro-viz/pull/573.
         dataset.release()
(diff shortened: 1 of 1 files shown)


package/kedro_viz/models/flowchart.py
```diff
@@ -580,6 +580,14 @@ def __post_init__(self, data_node: DataNode):
...
             self.run_command = f'kedro run --to-outputs="{data_node.full_name}"'
 
+        # Only check for existence of dataset if we might want to load it.
+        if not (
+            data_node.is_plot_node()
+            or data_node.is_image_node()
+            or data_node.is_tracking_node()
+        ):
+            return
+
```
noklam, line 590: Would it be better to have it as a property or `DataNode`? something like `is_loadable`? At some point there may be other data that we want to load too, for example, a preview pane for a table.

AntonyMilneQB: Yes, absolutely this would make sense 👍 But I don't want to add a new property until we've figure out how we clean this up in general - I commented more on this in https://github.com/kedro-org/kedro-viz/issues/1043.

tynandebold (approved):
The outcome looks good. I don't see the warning message anymore.

rashidakanchwala (approved):
thanks!

noklam (approved):

PR #1049, opened 2022-09-08
Fix Windows CI

AntonyMilneQB:
## Description

Silly Windows doesn't understand `find`, which was being executed by `make build` (through `clean`). We now only use this command in `make package`, which is only run on Linux.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1049"></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 .PHONY: package
 
-
 clean:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,11 +1,12 @@
 .PHONY: package
 
+
```
AntonyMilneQB, line 3:
```suggestion
```

tynandebold (approved):
Nice one!

PR #1056, opened 2022-09-08
Fix Kedro-viz to work with strawberry 0.128.0

rashidakanchwala:
## Description

A small fix in mutation UpdateRunDetails based on the latest strawberry release that introduced a breaking change  (https://github.com/strawberry-graphql/strawberry/releases/tag/0.128.0)

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1056"></a>

### DIFF
--- a/package/kedro_viz/api/graphql/schema.py
+++ b/package/kedro_viz/api/graphql/schema.py
@@ -129,7 +129,7 @@ class Mutation:
             updated_run.bookmark,
             updated_run.notes,
         )
-        return UpdateRunDetailsSuccess(updated_run)
+        return UpdateRunDetailsSuccess(run=updated_run)
 
 
 @strawberry.type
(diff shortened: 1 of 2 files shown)


tynandebold (approved):
Makes sense to me :)

Would be helpful if you could add a bit more detail in the PR description, e.g. a link to the documentation and how you found that change "based on the latest strawberry release," as you wrote.

AntonyMilneQB (approved):
Looks good, thanks for the fix! Is this the only case where we instantiate a strawberry type without specifying the argument name? I see all the tests pass now, but please do check there's no other cases that need to be updated in the code.

Agree with @tynandebold that it would be nice to have a bit more context here, e.g. here is the release that introduced the breaking change: https://github.com/strawberry-graphql/strawberry/releases/tag/0.128.0.

I see strawberry have upgraded to a new version of GraphiQL. When I go to https://demo.kedro.org/graphql I just get "Loading..." now. I haven't tried it locally but would be good to see if this still works.

PR #1062, opened 2022-09-12
Fix pretty name for transcoded datasets

### LINKED ISSUE #859: Pretty-print names do not work in namespaced transcoded datasets
## Description
When there is a dataset that uses transcoding (etc. `abc@a`), its name does not get pretty printed properly.

Both datasets below are namespaced. `tst.a@a`, `tst.b`. `tst.a` doesn't remove the namespace.
image

## Context
Simple UI bug.

## Steps to Reproduce
Create a kedro project with the following registry:

``` python
"""Project pipelines."""
from typing import Dict

from kedro.pipeline import Pipeline, pipeline, node


def register_pipelines() -> Dict[str, Pipeline]:
    """Register the project's pipelines.

    Returns:
        A mapping from a pipeline name to a ``Pipeline`` object.
    """
    return {"__default__": pipeline([
        node(func=lambda a: False, inputs='tst.a@a', outputs='tst.b')
    ])}
```

## Your Environment

* Web browser system and version: Latest Vivaldi/Chrome
* Operating system and version: Ubuntu 22.04
* Kedro version used (if relevant): 0.18.1
* Python version used (if relevant): 3.8.10

## Checklist

rashidakanchwala:
## Description

Resolves #859 

## Development notes

I realised that there was a bug; for transcoded datasets we did not strip the namespace like we did for the rest so I have added that now. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1062"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,12 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+# Release 5.
+
+## Bug fixes and other changes
+
+- Fix pretty name for transcoded datasets. (#1062)
+
 # Release 5.1.1
 
 ## Bug fixes and other changes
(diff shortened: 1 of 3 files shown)


AntonyMilneQB:
Looks good, but I think a test would be a good idea here and should be simple to write.

AntonyMilneQB (approved):
Nice!

PR #1086, opened 2022-09-16
Dependency update

rashidakanchwala:
## Development notes

Most dependencies didn't break anything. A few did -- hence below are the changes made 

- on circleCI, we have pinned the python 3.10 to python 3.10.6 because it's failing with mypy. A fix will be released soon and then we will revert it back to python 3.10 (https://github.com/python/mypy/issues/13627) 
- pylint & flake8 had some new linting requirements such as 
   - any sort of 'with open' file should have an encoding set (multiple places) 
   - there should be a timeout specified for responses (jupyter.py - line 76)
- fastAPI now requires us to return None if there's no value in dataclass properties. So we are doing that (flowchart.py), Also we must return all properties specified in responses. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1086"></a>


package/kedro_viz/launchers/jupyter.py
```diff
@@ -12,7 +12,7 @@
 
 import IPython
 import requests
-from IPython.core.display import HTML, display
+from IPython.display import HTML, display
```
AntonyMilneQB, line 15: Why did this change?

package/test_requirements.txt
```diff
@@ -1,21 +1,21 @@
...
-matplotlib~=3.3.4
-mypy~=0.930
-psutil==5.8.0  # same as Kedro for now
-pylint~=2.8.2
-pytest~=6.2.0
-pytest-asyncio~=0.17.2
-pytest-cov~=2.11.1
-pytest-mock~=3.6.1
+bandit~=1.7
+behave>=1.2, <2.0
```
AntonyMilneQB, line 5:
```suggestion
behave~=1.2
```
Just to be consistent with how we write it elsewhere.

AntonyMilneQB (approved):
Looks great, thank you for all this and the clear write-up! ⭐ 

In the future I think we might like to consider other configuration options like `assignees`, `reviewers` and `versioning-strategy`. But happy to leave dependabot running for now, see how it works, and modify the configuration as we go along if other options would suit us better.

package/kedro_viz/launchers/jupyter.py
rashidakanchwala: new IPython version update - https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html

Issue #749, opened 2022-02-22
Valid YAML when serlialised to JSON can produce something unrenderable


tynandebold:
(transfer from Jira, created by @datajoely)

Apparently this is valid YAML:

```yml
hello: .Inf 
```
This is also a valid Python float:

![image-2022-01-27-09-35-27-344](https://user-images.githubusercontent.com/16869061/155152758-891d44bf-f970-4031-9fe0-3328c5ed43b8.png)

However it isn't valid JSON. You end up with a blank Kedro-Viz:

![image-2022-01-27-09-35-40-358](https://user-images.githubusercontent.com/16869061/155152776-baeaa48d-36a4-4de5-986b-45aeeceac0ef.png)

GeorgySk:
Is this issue really resolved? I'm getting a `ValueError: Out of range float values are not JSON compliant ` when using `.inf` in the latest version. Is this an expected behavior? If it is then, probably, it should be mentioned in the docs.

AntonyMilneQB:
I think what's happened here is that it's fixed in `main` but not yet released. I'm not sure what the plan is for doing a release - @tynandebold? If you need it urgently then it is possible to install the unreleased kedro-viz but it's a bit of a faff.

tynandebold:
We'll do a release late this week or early next.

@GeorgySk let us know if you need this sooner.

PR #1087, opened 2022-09-16
Refactor /main response and save_file result

jmholzer:
## Description
This PR attempts a solution to two problems:

1. It resolves https://github.com/kedro-org/kedro-viz/issues/749 (including cases with arbitrarily nested values).
2. It ensures the same response is received from the /main endpoint and from specifying a save_file in server.py.

With regards to 1., it has the disadvantage that it will encode both .nan and .inf as null, I think it would be far better for the user if it encoded these as the strings "NaN" and "Infinity", respectively.

We could implement this by using a custom encoder for float with fastapi.encoders.jsonable_encoder, though we will need to upgrade fastapi to 0.73.0 for this feature. Since this seems to break a few tests, I will open a separate ticket for this.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1087"></a>

### DIFF
--- a/package/kedro_viz/api/rest/responses.py
+++ b/package/kedro_viz/api/rest/responses.py
@@ -1,8 +1,11 @@
 """`kedro_viz.api.rest.responses` defines REST response types."""
 # pylint: disable=missing-class-docstring,too-few-public-methods
 import abc
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union, Any
 
+import orjson
+from fastapi.encoders import jsonable_encoder
+from fastapi.responses import ORJSONResponse
 from pydantic import BaseModel
 
 from kedro_viz.data_access import data_access_manager
@@ -81,8 +84,7 @@ class DataNodeAPIResponse(BaseGraphNodeAPIResponse):
 
 
 NodeAPIResponse = Union[
-    TaskNodeAPIResponse,
-    DataNodeAPIResponse,
+    TaskNodeAPIResponse, DataNodeAPIResponse,
 ]
 
 
@@ -246,10 +248,8 @@ def get_default_response() -> GraphAPIResponse:
         data_access_manager.get_default_selected_pipeline().id
     )
 
-    modular_pipelines_tree = (
-        data_access_manager.create_modular_pipelines_tree_for_registered_pipeline(
-            default_selected_pipeline_id
-        )
+    modular_pipelines_tree = data_access_manager.create_modular_pipelines_tree_for_registered_pipeline(
+        default_selected_pipeline_id
     )
 
     return GraphAPIResponse(
@@ -267,3 +267,15 @@ def get_default_response() -> GraphAPIResponse:
         modular_pipelines=modular_pipelines_tree,
         selected_pipeline=default_selected_pipeline_id,
     )
+
+
+class ORJSONResponse(ORJSONResponse):
+    @staticmethod
+    def write_to_file(content: Any) -> bytes:
+        encoded_content = jsonable_encoder(content)
+        return orjson.dumps(
+            encoded_content,
+            option=orjson.OPT_INDENT_2
+            | orjson.OPT_NON_STR_KEYS
+            | orjson.OPT_SERIALIZE_NUMPY,
+        )
--- a/package/kedro_viz/api/rest/responses.py
+++ b/package/kedro_viz/api/rest/responses.py
@@ -1,7 +1,7 @@
 """`kedro_viz.api.rest.responses` defines REST response types."""
 # pylint: disable=missing-class-docstring,too-few-public-methods
 import abc
-from typing import Dict, List, Optional, Union, Any
+from typing import Any, Dict, List, Optional, Union
 
 import orjson
 from fastapi.encoders import jsonable_encoder
@@ -84,7 +84,8 @@ class DataNodeAPIResponse(BaseGraphNodeAPIResponse):
 
 
 NodeAPIResponse = Union[
-    TaskNodeAPIResponse, DataNodeAPIResponse,
+    TaskNodeAPIResponse,
+    DataNodeAPIResponse,
 ]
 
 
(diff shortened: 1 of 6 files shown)


jmholzer:
This (draft) PR is solution to two problems:

1. It resolves #749 
2. It ensures the same response is received from the /main endpoint and from specifying a `save_file` in server.py.

With regards to 1., it has the disadvantage that it will encode both `.nan` and `.inf` as `null`, I think it would be far better for the user if it encoded these as the strings "NaN" and "Infinity", respectively.

We could implement this by using a custom encoder for `float` with `fastapi.encoders.jsonable_encoder`, though we will need to upgrade fastapi to 0.73.0 for this feature. Since this seems to break a few tests, I will open a separate ticket for this.

package/requirements.txt
```diff
@@ -10,3 +10,4 @@ pandas>=0.24
 sqlalchemy~=1.2
 strawberry-graphql>=0.99.0, <1.0
 networkx>=1.0
+orjson~=3.8.0
```
AntonyMilneQB, line 13:
```suggestion
orjson~=3.8
```
No need to pin to the 3.8.x line (or probably to 3.x at all tbh... but that's how we normally do things at the moment).

package/kedro_viz/api/rest/router.py
```diff
@@ -27,7 +28,7 @@
 )
 
 
-@router.get("/main", response_model=GraphAPIResponse)
+@router.get("/main", response_class=EnhancedORJSONResponse)
```
AntonyMilneQB, line 31:
```suggestion
@router.get("/main", response_class=EnhancedORJSONResponse, response_model=GraphAPIResponse)
```

Don't we still want the `response_model`?

AntonyMilneQB (commented):
This looks great, thank you for all the many different attempts and discussions about it!

My instinct is that this would be better as `default_response_class` on the `FastAPI` instantiation itself rather than just on the `main` endpoint. This would ensure consistency across the whole app. e.g. when you retrieve parameters data in the metadata side panel, it might come from a different endpoint (?).

tynandebold:
> This looks great, thank you for all the many different attempts and discussions about it!
> 
> My instinct is that this would be better as `default_response_class` on the `FastAPI` instantiation itself rather than just on the `main` endpoint. This would ensure consistency across the whole app. e.g. when you retrieve parameters data in the metadata side panel, it might come from a different endpoint (?).

Yes, the data that populates the metadata side panel is a different endpoint than `/api/main`. That endpoint is `/api/nodes/:node_id`. There's also an endpoint for the pipeline registry dropdown: `/api/pipelines/:pipeline_id`.

package/requirements.txt
jmholzer: Fixed in [49ba250](https://github.com/kedro-org/kedro-viz/pull/1087/commits/49ba25053c441c3ec90268d0cbe0d710fc2c4f7a).

package/kedro_viz/api/rest/router.py
jmholzer: You're right, fixed in [207a718](https://github.com/kedro-org/kedro-viz/pull/1087/commits/207a7181087acc9439056c47fc870f88ce3ae76d).

jmholzer:
> My instinct is that this would be better as `default_response_class` on the `FastAPI` instantiation itself rather than just on the `main` endpoint.

Good call, fixed in [72f898a](https://github.com/kedro-org/kedro-viz/pull/1087/commits/72f898a7317ac9897dadd572ff79551028679569)

package/kedro_viz/api/rest/responses.py
```diff
@@ -267,3 +270,22 @@ def get_default_response() -> GraphAPIResponse:
...
+    @staticmethod
+    def encode_to_human_readable(content: Any) -> bytes:
+        """A method to encode the given content to JSON, with the
+        proper formatting to write a human-readable file.
+
+        Returns:
+            A bytes object containing the JSON to write.
+
+        """
+        encoded_content = jsonable_encoder(content)
```
AntonyMilneQB, line 285: Minor preference to put this outside this class. This would be consistent with how FastAPI handles the endpoint, where (as I understand it) it calls `jsonable_encoder` and _then_ the `Response`. Certainly there's no explicit call to `jsonable_encoder` in the `render` method of `JSONResponse` etc. anyway so it feels a bit off here.

AntonyMilneQB (approved):
Very happy to approve this in principle even though it's still a draft 👍 Thank you for working through all the many different options for the solution! This turned out to be surprisingly subtle and tricky.

Just a couple of notes:
1. I'm actually a bit surprised that all the tests are passing. I was kind of expecting the responses to change a bit due to altering `JSONResponse` to `ORJSONResponse` across the whole app. e.g. maybe the keys would be ordered differently or something. If there's no tests to fix then that's great, but please do give it some manual testing to make sure it's doing what we think it should!  
2. Now we've upgraded to FastAPI 0.73 you should now be able to customise `jsonable_encoder` to translate `.nan`/`.inf` (and any others??) as strings rather than just `null`. Personally I'm not too bothered about this, since it happens so rarely, but if you think it's an improvement then please do go ahead with it! Ideally this would be done in the most generic way possible ("call `str` on anything which orjson translates to `null`") rather than explicitly ("replace `.nan` with NaN and `.inf` with Inf") but I don't feel strongly about that either, since I'm not sure there's any other cases beyond nan/inf that we need to care about.

package/kedro_viz/api/rest/responses.py
jmholzer: You're right, fixed in [4485b93](https://github.com/kedro-org/kedro-viz/pull/1087/commits/4485b9390dd8241e79f2da028aac5260745c8b69).

jmholzer:
Thanks for the review @AntonyMilneQB.

In regards to your individual points:
1. I have gone through some manual testing with the demo project, I could not see anything that was broken or different.
2. I'll open a separate issue for this, I feel it deserves one as it is another atomic piece of work to implement this and it merits a different priority to the other work in this PR.

PR #1096, opened 2022-09-26
Fix modular pipeline Train Evaluation on demo_project 

### LINKED ISSUE #751: [FE] Inconsistent highlight state for input/output nodes in the sidebar
(transfer from Jira, created by @studioswong)

Under focus mode, the input and output nodes relating to the selected modular pipeline should not be highlighted (see below). 

image-2022-01-18-09-36-35-645

Some output nodes for the 'train evaluation' pipeline were in a highlighted state (see below). The input nodes stays in the intended non-highlight state. 

Having tested around this project it seems that this highlighted state is unique to those nodes as pictured - this is likely due to an edge case / minor bug introduced from the collapsed modular pipeline calculations. 

image-2022-01-18-09-32-26-535

rashidakanchwala:
## Description

Resolves #751 

## Development notes

The above issue talks about inconsistent highlighting of nodes in the sidebar. However the nodes are highlighted correctly. it is the placement of the nodes outside the modular pipeline 'Train Evaluation' which is the issue. I have fixed the above by fixing the namespace in the demo_project. 

for e.g. full name of the r2_score node was train_evaluation.r2_score_linear_regression instead of train_evaluation.**linear_regression**.r2_score_linear_regression

~~This is fixed now. Once you run the above, you will now see Hyperparams and R2score nodes within the nested train_evaluation pipelines and no longer outside of it ~~

**UPDATE** - After discussion with Nok on best practices on Kedro; he said renaming nodes was not a common practice so I have removed that. This also eliminates the problem we had before because when you don't explicitly specify output name, the namespace gets applied correctly to the output nodes. 

As a result of the above -- if you now have pretty naming turned on and you focus on the 'Train Evaluations' pipeline -- u will see two nodes called R2_Score and two nodes called 'Experiment Params' (which is expected result) -- if the user wants to see which is which they will have to turn off pretty-name which will show the namespace and highlight the differences

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1096"></a>


tynandebold (approved):
This looks good to me. When focus mode is active, that modular pipeline is now consistent with the display of the other three.

tynandebold:
I think this needs another look. If you click on any of those four nodes under `Train Evaluation`, nothing populates the metadata panel. There was certainly information in there before.

rashidakanchwala:
> I think this needs another look. If you click on any of those four nodes under `Train Evaluation`, nothing populates the metadata panel. There was certainly information in there before.

Yes it did!! thanks for catching that, i had to update the dataset name in catalog yml and also do som data folder clean up. it shld work now.

tynandebold (approved):
Very nice!

noklam (approved):
For the record, I think renaming nodes for a different semantic is fine, but in this case the semantic is exactly modular pipeline, and sticking with the hierarchy with `x.y.z` reflect preserve this hierarchy.

`train_evaluation.hyperparams_linear_regression` -> `train_evaluation.linear_regression.hyperparmeters`

PR #1100, opened 2022-09-27
Fix/remove tags from panel

### LINKED ISSUE #1092: Remove "Tags" from datasets
> **Note**. Good starter task for anyone wanting to get more familiar with kedro-viz!

Along similar lines to https://github.com/kedro-org/kedro-viz/issues/962...

Currently every single graph node on kedro-viz has a Tags field:
image

In reality, however, only kedro nodes can have tags. Datasets, transcoded datasets and parameters should not have this field.

1. Move `tags` from being a field in the base model `flowchart.GraphNode` to only in the kedro node model `flowchart.TaskNode`
2. Remove `tags` from instantiations of the non-`TaskNode` models, e.g. calls to `GraphNode.create_parameters_node` should no longer have a `tags` argument (you'll see it's currently just set to `set()` anyway)
3. Remove tags field from the frontend for everything but a task node

Huongg:
## Description
Fixes #1092 

## Development notes

This fix is to only show the tags field from the frontend if it's a task node. We won't remove it from the BE as it will impact the flowchart logic. You can find the conversation [here ](https://github.com/kedro-org/kedro-viz/issues/1092#issuecomment-1259455021)which we discussed specifically about how it would impact the flowchart

If we're all happy with this fix, we can close the [PR 1099](https://github.com/kedro-org/kedro-viz/pull/1099/files), which include changes from BE

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1100"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,6 +17,7 @@ Please follow the established format:
 - Fix pretty naming for transcoded datasets. (#1062)
 - Keep graph visible when switching pipelines. (#1063)
 - Add a transition to run details in experiment tracking. (#1065)
+- Remove tags field from datasets and parameters in flowchart panel. (#1100)
 
 # Release 5.1.1
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,7 +17,7 @@ Please follow the established format:
 - Fix pretty naming for transcoded datasets. (#1062)
 - Keep graph visible when switching pipelines. (#1063)
 - Add a transition to run details in experiment tracking. (#1065)
-- Remove tags field from datasets and parameters in flowchart panel. (#1100)
+- Remove Tags field from dataset and parameter nodes when viewed in metadata panel. (#1100)
 
 # Release 5.1.1
 
--- a/src/components/metadata/metadata.js
+++ b/src/components/metadata/metadata.js
@@ -190,12 +190,14 @@ const MetaData = ({
                   visible={isTaskNode}
                   value={metadata.outputs}
                 />
-                <MetaDataRow
-                  label="Tags:"
-                  kind="token"
-                  commas={false}
-                  value={metadata.tags}
-                />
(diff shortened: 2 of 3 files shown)


Huongg:
> Thank you for working through this! Sorry the original BE change didn't work out as expected but I hope it was educational anyway 🙈

no worries at all. it was a good learning to make changes from the BE side :)

tynandebold (approved):
Looks good!

Approving now with a question: do we still have a test that checks if the tags are shown? The two you changed are checking that they're _not_ there. If there isn't a test to check if they do exists maybe you could add one?

Huongg:
> Looks good!
> 
> Approving now with a question: do we still have a test that checks if the tags are shown? The two you changed are checking that they're _not_ there. If there isn't a test to check if they do exists maybe you could add one?

hey @tynandebold yup we already have a test for tags, specifically for TaskNode only

```
    it('shows the node tags', () => {
      const wrapper = mount({
        nodeId: splitDataTaskNodeId,
        mockMetadata: nodeTask,
      });
      const row = rowByLabel(wrapper, 'Tags:');
      expect(textOf(rowValue(row))).toEqual(['Split']);
    });
```

RELEASE.md
```diff
@@ -17,6 +17,7 @@ Please follow the established format:
 - Fix pretty naming for transcoded datasets. (#1062)
 - Keep graph visible when switching pipelines. (#1063)
 - Add a transition to run details in experiment tracking. (#1065)
+- Remove tags field from datasets and parameters in flowchart panel. (#1100)
```
tynandebold, line 20: To be a bit more consistent with the language elsewhere in this file:

```suggestion
- Remove Tags field from dataset and parameter nodes when viewed in metadata panel. (#1100)
```

PR #1103, opened 2022-09-28
fix/keyboard-on-run-export-modal

### LINKED ISSUE #926: RunExportModal animation doesn't fire and the modal doesn't close while using keyboard
## Description
When toggling through the RunExportModal with the keyboard and exporting the data the modal animation doesn't fire nor does the modal close

## Steps to Reproduce
ezgif com-gif-maker (2)

## Checklist

Huongg:
Signed-off-by: huongg <huongg1409@gmail.com>

## Description

Fixes #926 

## Development notes

I realised the issue is because there are two `clickHandlers`, and they both apply for `CSVLink` and `Button`, hence to make sure all the events fired at the same time, I combine all the handlers together and pass it to `CSVLink` only. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1103"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -18,6 +18,7 @@ Please follow the established format:
 - Keep graph visible when switching pipelines. (#1063)
 - Add a transition to run details in experiment tracking. (#1065)
 - Remove Tags field from dataset and parameter nodes when viewed in metadata panel. (#1100)
+- Fix keyboard support for exporting experiment runs. (#1103)
 
 # Release 5.1.1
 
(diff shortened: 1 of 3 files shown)


tynandebold (approved):
Well done!

One thing while you're in here: the button focus state is inconsistent. It looks like this in the export run data modal...

image

...but it should look like this:

image

Would you mind please fixing it?

Huongg:
hey thank you @tynandebold . I've made some small changes to update the styling (it feels a bit hacky though). Let me know what you think?

So because we have 2 layers for the button, one is the `CSVLink` to handle export data, and one is the `Button` to ensure it has the same stylling like other buttons. So I kind of need to remove the focus style from `Button` and re-applies it on the `CSVLink` instead.

tynandebold:
It looks good to me :) thanks for fixing that.

cvetankanechevska (approved):

PR #1104, opened 2022-09-28
Show 'MemoryDataSet' if string to shorten is falsy

deepyaman:
## Checklist

### DIFF
--- a/src/actions/nodes.js
+++ b/src/actions/nodes.js
@@ -19,7 +19,7 @@ export const TOGGLE_NODES_DISABLED = 'TOGGLE_NODES_DISABLED';
 /**
  * Toggle a selected group of nodes' visibility on/off
  * @param {Array} nodeIDs The nodes' unique identifiers
- * @param {Boolean} isDisabled Whether the node should be visible
+ * @param {boolean} isDisabled Whether the node should be visible
  */
 export function toggleNodesDisabled(nodeIDs, isDisabled) {
   return {
@@ -59,7 +59,7 @@ export const ADD_NODE_METADATA = 'ADD_NODE_METADATA';
 
 /**
  * Toggle whether to display the loading spinner
- * @param {boolean} loading True if pipeline is still loading
+ * @param {Object} data The node's data
  */
 export function addNodeMetadata(data) {
   return {
--- a/src/components/metadata/metadata.js
+++ b/src/components/metadata/metadata.js
@@ -60,14 +60,14 @@ const MetaData = ({
   }
 
   // translates the naming for the different types of nodes
-  const translateMetadataType = (metaDataType) => {
-    if (metaDataType === 'task') {
+  const translateMetadataType = (metadataType) => {
+    if (metadataType === 'task') {
       return 'node';
-    } else if (metaDataType === 'data') {
+    } else if (metadataType === 'data') {
       return 'dataset';
     }
 
-    return metaDataType;
+    return metadataType;
   };
 
   const onCloseClick = () => {
@@ -85,7 +85,7 @@ const MetaData = ({
   };
 
   const shortenDatasetType = (string) => {
-    return string?.split('.').pop();
+    return string ? string.split('.').pop() : 'MemoryDataSet';
   };
 
   return (
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -179,8 +179,7 @@ Before launching a development server with a real Kedro project, you'd need to h
 The Kedro-Viz repository comes with an example project in the `demo-project` folder. This is used on the [public demo](https://demo.kedro.org/). To use it in your development environment, you need to install both the Kedro-Viz dependencies and a minimal set of dependencies for the demo project:
 
 ```bash
-pip3 install -r package/test_requirements.txt
-pip3 install -r demo-project/src/docker_requirements.txt
+pip3 install -r package/test_requirements.txt -r demo-project/src/docker_requirements.txt
 ```
 
 Now build the application with:
--- a/src/actions/modular-pipelines.js
+++ b/src/actions/modular-pipelines.js
@@ -3,7 +3,7 @@ export const TOGGLE_MODULAR_PIPELINE_ACTIVE = 'TOGGLE_MODULAR_PIPELINE_ACTIVE';
 /**
  * Toggle a modular pipeline item's highlighting on/off (or array of modular pipelines)
  * @param {string|Array} modularPipelineIDs Modular pipeline id(s)
- * @param {Boolean} active True if modualr pipeline(s) active
+ * @param {Boolean} active True if modular pipeline(s) active
  */
 export function toggleModularPipelineActive(modularPipelineIDs, active) {
   return {


src/actions/nodes.js
```diff
@@ -19,7 +19,7 @@ export const TOGGLE_NODES_DISABLED = 'TOGGLE_NODES_DISABLED';
 /**
  * Toggle a selected group of nodes' visibility on/off
  * @param {Array} nodeIDs The nodes' unique identifiers
- * @param {Boolean} isDisabled Whether the node should be visible
+ * @param {boolean} isDisabled Whether the node should be visible
```
deepyaman, line 22: How do you determine whether `boolean` should be capitalized or not, or does it not matter? I made this consistent with the rest of the files, but it's not always consistent with other files.

src/actions/nodes.js
```diff
@@ -59,7 +59,7 @@ export const ADD_NODE_METADATA = 'ADD_NODE_METADATA';
 
 /**
  * Toggle whether to display the loading spinner
- * @param {boolean} loading True if pipeline is still loading
+ * @param {Object} data The node's data
```
deepyaman, line 62: I don't have much context here; feel free to suggest a better description, if necessary.

src/components/metadata/metadata.js
```diff
@@ -85,7 +85,7 @@ const MetaData = ({
   };
 
   const shortenDatasetType = (string) => {
-    return string?.split('.').pop();
+    return string ? string.split('.').pop() : 'MemoryDataSet';
```
deepyaman, line 88: I think this fixes the issue, but I'm not sure whether it's expected that `undefined` gets passed in here?

AntonyMilneQB: I think this is not expected (at least by me anyway). I'd expect the dataset type here to be `kedro.io.core.MemoryDataSet`.

src/actions/nodes.js
tynandebold: It should probably be capitalized everywhere: https://developer.mozilla.org/en-US/docs/Glossary/Boolean

src/components/metadata/metadata.js
tynandebold: For some reason the responses for memory datasets are empty. [See here](https://demo.kedro.org/api/nodes/c08c7708), compared with the response for a [CSV dataset](https://demo.kedro.org/api/nodes/aed46479).

Given that, this does fix the issue. All other datasets return what we expect from the endpoint.

tynandebold (approved):
Approving this now because it does fix the stated issue.

We should create a ticket and investigate why the responses for `MemoryDataSet`s are always empty. Or could you take a look at that in this same PR?

src/actions/nodes.js
deepyaman: I did see that, but then `String`, etc. should all be capitalized, too. Guess can do it later.

src/components/metadata/metadata.js
deepyaman: > For some reason the responses for memory datasets are empty. [See here](https://demo.kedro.org/api/nodes/c08c7708), compared with the response for a [CSV dataset](https://demo.kedro.org/api/nodes/aed46479).

I think this may be the case only for `MemoryDataSet`s not explicitly listed in the catalog. Let me test using the API (thanks for showing that!) as soon as my other issue resolves itself:


https://user-images.githubusercontent.com/14007150/193087503-14f548b2-4363-4c63-b297-3f5d8ab698d6.mov

deepyaman: > For some reason the responses for memory datasets are empty. [See here](https://demo.kedro.org/api/nodes/c08c7708), compared with the response for a [CSV dataset](https://demo.kedro.org/api/nodes/aed46479).

I think this may be the case only for `MemoryDataSet`s not explicitly listed in the catalog. Let me test using the API (thanks for showing that!) as soon as my other issue resolves itself:

193087503-14f548b2-4363-4c63-b297-3f5d8ab698d6

deepyaman: @tynandebold Just to sanity-check since I'm new to Viz...

If you explicitly add a catalog entry for a `MemoryDataSet`, does the `/api/nodes` call also give you and `Internal Server Error`? Maybe some `pydantic` validation error?

deepyaman:
> We should create a ticket and investigate why the responses for `MemoryDataSet`s are always empty. Or could you take a look at that in this same PR?

Happy to take a look; doesn't matter to me whether it's part of this issue or another. I realized I was looking at the GraphQL API stuff on the frontend, but I should be looking at the REST API under `package/`, so I think I know where to pick up debugging. (But if this sounds totally off, let me know; I feel like I understand very little about how Kedro-Viz is structured, especially with the JS frameworks...)

src/components/metadata/metadata.js
tynandebold: I'm afraid I don't know the answer to this one.

tynandebold:
> > We should create a ticket and investigate why the responses for `MemoryDataSet`s are always empty. Or could you take a look at that in this same PR?
> 
> Happy to take a look; doesn't matter to me whether it's part of this issue or another. I realized I was looking at the GraphQL API stuff on the frontend, but I should be looking at the REST API under `package/`, so I think I know where to pick up debugging. (But if this sounds totally off, let me know; I feel like I understand very little about how Kedro-Viz is structured, especially with the JS frameworks...)

Let's keep it all together in this PR then.

You're spot on, you'll need to look at the REST endpoints (`/kedro-viz/package/kedro_viz/api/rest`) for this. For context, we use REST for the flowchart view in Viz and GraphQL for experiment tracking.

src/components/metadata/metadata.js
tynandebold: This is also where my knowledge of Kedro starts to break down. I tried to add a catalog entry for a `MemoryDataSet` but got an error:

```
DataSet 'train_evaluation.random_forest.experiment_params' must only contain arguments valid for the constructor of 'kedro.io.memory_dataset.MemoryDataSet'.
```

If you can instruct me here I can sanity check for you on my end.

deepyaman: You can add a `MemoryDataSet` explicitly like:

```yaml
train_X:
  type: MemoryDataSet
```

I would probably do it for one of the datasets that isn't explicitly registered (e.g. `train_X` and every other dataset other than the input data in `pandas-iris`), but not for parameters. The only other thing you could specify in the catalog entry is `copy_mode`, but that's not necessary.

:)

tynandebold: Nice thank you. I get a 500 error if I do that:

```
GET http://localhost:4141/api/nodes/cae2d1c7 500 (Internal Server Error)
```

Here's the server error:

```
2022-09-30 11:14:49,229 - uvicorn.access - INFO - None:0 - "GET /api/nodes/cae2d1c7 HTTP/1.1" 500
2022-09-30 11:14:49,229 - uvicorn.error - ERROR - Exception in ASGI application
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 372, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/uvicorn/middleware/message_logger.py", line 82, in __call__
    raise exc from None
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/uvicorn/middleware/message_logger.py", line 78, in __call__
    await self.app(scope, inner_receive, inner_send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/fastapi/applications.py", line 199, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/routing.py", line 580, in __call__
    await route.handle(scope, receive, send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/routing.py", line 241, in handle
    await self.app(scope, receive, send)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/starlette/routing.py", line 52, in app
    response = await func(request)
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/fastapi/routing.py", line 224, in app
    response_data = await serialize_response(
  File "/Users/tynandebold/opt/anaconda3/lib/python3.8/site-packages/fastapi/routing.py", line 127, in serialize_response
    raise ValidationError(errors, field.type_)
pydantic.error_wrappers.ValidationError: <unprintable ValidationError object>
```

deepyaman: @tynandebold Thanks for verifying! I thought I must be doing something wrong earlier in my setup since I was getting such errors, but seems this is an actual issue, so I will investigate. :D

deepyaman:
@tynandebold @AntonyMilneQB @rashidakanchwala I guess this is no longer necessary in lieu of #1113, right?

PR #1109, opened 2022-10-01
change to Uppercase

nikotome:
## Description

Changes to the param file where requested, simple small to uppercase modification

## Checklist

### DIFF
--- a/src/actions/nodes.js
+++ b/src/actions/nodes.js
@@ -5,7 +5,7 @@ export const TOGGLE_NODE_CLICKED = 'TOGGLE_NODE_CLICKED';
 
 /**
  * Update the value of the currently-active clicked node
- * @param {string|null} nodeClicked The node's unique identifier
+ * @param {String|null} nodeClicked The node's unique identifier
  */
 export function toggleNodeClicked(nodeClicked) {
   return {
@@ -33,7 +33,7 @@ export const TOGGLE_NODE_HOVERED = 'TOGGLE_NODE_HOVERED';
 
 /**
  * Update the value of the currently-active hovered node
- * @param {string|null} nodeHovered The node's unique identifier
+ * @param {String|null} nodeHovered The node's unique identifier
  */
 export function toggleNodeHovered(nodeHovered) {
   return {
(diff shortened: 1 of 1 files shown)


src/actions/nodes.js
```diff
@@ -70,7 +70,7 @@ export function addNodeMetadata(data) {
 
 /**
  * update node metadata on selection, loading new data if it has not been previously called
- * @param {string} nodeID node id of clicked node
+ * @param {String} nodeID node id of clicked node
  * @return {function} A promise that resolves when the data is loaded
```
deepyaman, line 74:
```suggestion
 * @return {Function} A promise that resolves when the data is loaded
```
I think these should also be changed; see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function.

deepyaman (changes requested):
Hi @nikotome! Thanks for taking a stab at this. I think there are several other files under `src/actions` where this would need to be done (in the same way you've done for `nodes.js`.

Also, see the Details on the DCO error in CI to see how to pass that check.

nikotome:
Change done, thanks. I will look into the other files mentioned.

tynandebold:
Hey @nikotome I think there was still some more work to do here if you're still up for helping out? I can assist in any way that I can. Just reach out. Thank you!

nikotome:
> Hey @nikotome I think there was still some more work to do here if you're still up for helping out? I can assist in any way that I can. Just reach out. Thank you!

Sure, it was great creating my first real life commit. I have a question. I was going to update my previous branch and create a new pull request with the next batch of changes. Would that be the right procedure? Or do I have to create a new branch?

deepyaman:
> > Hey @nikotome I think there was still some more work to do here if you're still up for helping out? I can assist in any way that I can. Just reach out. Thank you!
> 
> Sure, it was great creating my first real life commit. I have a question. I was going to update my previous branch and create a new pull request with the next batch of changes. Would that be the right procedure? Or do I have to create a new branch?

@nikotome If you push additional changes to the same branch, it will automatically update your PR. :)

PR #1113, opened 2022-10-03
Fix `MemoryDataSet` not displaying on metadata panel

### LINKED ISSUE #1093: Give `MemoryDataSet` as dataset type
> **Note**. Good starter task for anyone wanting to get more familiar with kedro-viz!

Along similar lines to https://github.com/kedro-org/kedro-viz/issues/962...

Currently if you have a `MemoryDataSet` then kedro-viz will just give "Dataset type: -". Really this should be a `MemoryDataSet`.
image

It would be worth checking how other dataset types from `kedro.io` render in kedro-viz currently, e.g. `PartitionedDataSet`. I'm not sure why we're not showing `MemoryDataSet` currently so it might be a problem for other "core" dataset types also.

rashidakanchwala:
## Description

Resolves #1093 

## Development notes

@deepyaman , @tynandebold  -- I have made an attempt to fix this issue at the backend. Let me know if this is good approach. 

Basically, in the kedro-viz backend we add datasets to the node based on the catalog entries. If there's no catalog entry for a dataset, we just leave it blank. I have fixed this part by adding MemoryDataSet() to a dataset without any catalog entry. 

Also MemoryDataSets don't have any file path which was a sort of required field in the ResponseAPI for DataNodes (which is why u got the pydantic error), so I have made that optional.

## QA notes

I will add the necessary tests once everyone's ok with this fix. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1113"></a>


package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -57,6 +57,9 @@ def get_dataset(self, dataset_name: str) -> Optional[AbstractDataSet]:
                 dataset_obj = self._catalog._get_dataset(dataset_name)
         except DataSetNotFoundError:  # pragma: no cover
             dataset_obj = None
+        # if dataset has no catalog entry, it is a MemoryDataSet
+        if not dataset_obj:
+            dataset_obj = MemoryDataSet()
```
AntonyMilneQB, line 62: Is this any different from the below?
```suggestion
            dataset_obj = MemoryDataSet()
```

AntonyMilneQB (commented):
This feels like the right fix to me but is exactly what @deepyaman tried here and apparently didn't solve it? https://github.com/kedro-org/kedro-viz/issues/1093#issuecomment-1260281827

package/kedro_viz/data_access/repositories/catalog.py
rashidakanchwala: it work's if you just add it in the exception. the reason why it flickered was because filepath is none for MemoryDataSets and it is a required field in DataNodeResponseAPI which now I have made Optional. 

I will make the above change :)

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -2,8 +2,9 @@
 centralise access to Kedro data catalog."""
 # pylint: disable=missing-class-docstring,missing-function-docstring,protected-access
 from typing import Dict, Optional
+from joblib import Memory
```
deepyaman, line 5:
```suggestion
```

package/kedro_viz/data_access/repositories/catalog.py
deepyaman: I'm happy with this. Yeah, I just didn't understand/get to finding what was happening with the validation error, but this solves it.

demo-project/conf/base/catalog_05_model_input.yml
```diff
@@ -2,3 +2,6 @@ model_input_table:
   type: pandas.ParquetDataSet
   filepath: ${base_location}/05_model_input/model_input_table.pq
   layer: model_input
+
+X_train:
+  type: MemoryDataSet
```
deepyaman, line 7:
```suggestion
```
I guess we can revert it?

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -55,8 +56,9 @@ def get_dataset(self, dataset_name: str) -> Optional[AbstractDataSet]:
                 dataset_obj = self._catalog._get_dataset(dataset_name, suggest=False)
             else:  # pragma: no cover
                 dataset_obj = self._catalog._get_dataset(dataset_name)
+        # if dataset has no catalog entry, it is a MemoryDataSet
         except DataSetNotFoundError:  # pragma: no cover
```
deepyaman, line 60: Wonder if this should be covered in tests, given that it does make a functional difference.

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -55,8 +56,9 @@ def get_dataset(self, dataset_name: str) -> Optional[AbstractDataSet]:
                 dataset_obj = self._catalog._get_dataset(dataset_name, suggest=False)
             else:  # pragma: no cover
                 dataset_obj = self._catalog._get_dataset(dataset_name)
+        # if dataset has no catalog entry, it is a MemoryDataSet
```
deepyaman, line 59: Nit: I think this is bordering on a redundant comment (the code seems really self-explanatory), but I suppose one could argue that it adds a bit of value. Not strong feelings, just a hunch.

package/kedro_viz/data_access/repositories/catalog.py
rashidakanchwala: yes, i will write the tests now !

tynandebold (approved):
It's working nicely. Great job!

Issue #1036, opened 2022-08-26
Disconnected/Incomplete graph in structures with nested modular pipelines


DrDaDe:
## Description
As soon as I try to generate a modular pipeline (with a namespace) whose components themselves contain a modular pipeline (with a namespace), kedro viz generates a completely incorrect and disconnected graph though kedro processes all nodes as expected.

## Context
We have a data ETL project with a quite large number of steps. We want to use kedro, including kedro viz to illustrate the pipeline for stakeholders. To simpify the visualisation, we want to use modular pipelines in order to be able to use expand/collapse in kedro viz. Due to the quite involved process, we need to be able to collapse to more than one level. However, up to now is has been impossible for us to get the correct visualisation in kedro viz if namespacing goes beyond level 1.

## Steps to Reproduce

Here is a minimum example of a pipeline which works in kedro but fails to produce the correct result in kedro viz.
Explanation: download + preprocession should be contained within one namespace "subsub_pipeline".
This should be entirely contained within the namespace "sub_pipeline", which consists of the above download+preprocession and an additional merge with another data source.
The last step (no namespace) should use the result of the sub_pipeline and apply a final function to it.
```python

old_input_data = "old_preprocessed_data"
merged_data = "merged_data"
downloaded_data = "downloaded_data"
preprocessed_data = "preprocessed_data"
final_data = "final_data"

def create_pipeline(**kwargs) -> Pipeline:
    sub_sub_pipeline = pipeline(
        [node(download,
              ["params:api"],
              downloaded_data),
         node(preprocess,
              [downloaded_data],
              preprocessed_data)
         ],
        namespace="subsub_pipeline",
        inputs={},
        parameters={"params:api"},
        outputs={preprocessed_data}
    )

    sub_pipeline = pipeline(
        [
            sub_sub_pipeline,
            node(merge,
              [old_input_data, preprocessed_data],
              merged_data)
         ],
        namespace="sub_pipeline",
        inputs={old_input_data},
        parameters={"params:api"},
        outputs={merged_data}
    )


    new_pipeline = pipeline(
        [   node(
                func=postprocess,
                inputs=[merged_data],
                outputs=final_data
            )
        ]
    )
    return sub_pipeline + new_pipeline
```

To make reproduction simpler, here are the actual contents of the node functions
```python
import pandas as pd
import requests
from kedro.pipeline import Pipeline, node, pipeline


def download(url : str) -> pd.DataFrame:
    dataset = []
[...]

PR #1119, opened 2022-10-04
Fix/disconnected graph with nested modular pipelines

jmholzer:
## Description

This PR resolves [#1036](https://github.com/kedro-org/kedro-viz/issues/1036). A minimally-invasive change to the code now causes nested modular pipelines to render correctly.

## Development notes

I replicated the project in which the error was first seen ([#1036](https://github.com/kedro-org/kedro-viz/issues/1036)). I was able to reproduce the error seen by the user reporting the bug:

Screenshot 2022-10-05 at 11 43 52

After making my changes, the project renders correctly in Viz:

Screenshot 2022-10-05 at 11 52 22

I have additionally (manually) tested the following projects, all of which also render correctly:
- Viz Demo Project
- Spaceflights
- Iris

After the current fix, there is still a problem in rendering where one of the datasets is initially rendered outside of the modular pipeline to which it belongs. After expanding and closing the pipeline, this error in rendering is no longer present. I have discussed this with @tynandebold. It is not clear whether this is a problem with the backend tree-generation code or with the rendering code on the frontend. I will open a separate issue to investigate this further.

A big problem in the investigation of this ticket was the backend code, which in my view has a few problems and is potentially a source of further, unreported bugs (e.g. the rendering issue discussed above). I will open a separate ticket concerning a refactor of this.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1119"></a>


limdauto:
Thanks for the fix and for looking into refactoring the modular pipeline tree generation code. If you need any context in how/why something was done in a certain way I would love to be able to help.

jmholzer:
> Thanks for the fix and for looking into refactoring the modular pipeline tree generation code. If you need any context in how/why something was done in a certain way I would love to be able to help.

Thanks so much @limdauto, really appreciate it!

rashidakanchwala (approved):
Look's good to me. definitely solves the problem of disconnected graphs. 

i tested it and the thing I noticed is the sidebar now has two sub_pipeline.preprocesed data and on the first load you do a see a disconnected sub_pipeline.preprocessed data node which disappears when you expand and collapse. So definitely there's an issue with the namespacing that needs to be tackled as a part of another refactor ticket. 

see images below on what I mean 

Screenshot 2022-10-06 at 12 08 49

Screenshot 2022-10-06 at 12 08 59

tynandebold (approved):
Looks great. Thank you for this!

noklam (approved):
Thanks for the fix! I am not gonna block this PR since I don't know too much about the viz.

Is there any test for the generated tree structure? I guess we will need some tests for this before we start any refactoring work.

PR #1107, opened 2022-09-30
Fix/focus mode modular pipeline highlight

cvetankanechevska:
## Description

Resolves #921 

## Development notes

Added one more boolean property to `createInitialPipelineState` , to track when FocusModeIcon is hovered and create new action in the store for updating this value.

For the style I reused the class we already have for when node is clicked and all other nodes in the flowchart are pushed in the background with some opacity (class: `pipeline-node--faded`).

## QA notes
You can open the Gitpod link, and check the hover effect on FocusModeIcon in the left side node list.

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1107"></a>

PR #1131, opened 2022-10-10
Fix/Faded nodes not part of clicked dataset

cvetankanechevska:
## Description
A small fix in flowchart, faded nodes not part of clicked dataset. Found by @tynandebold , after this PR was merged: #1107.

## Development notes
Remove 'nodeActive' from the condition that was created for adding 'faded' class when hovering on FocusModeIcon because nodes were not updated properly, when some dataset is clicked.

Before:
Screenshot 2022-10-10 at 16 55 07

After:
Screenshot 2022-10-10 at 16 54 17

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1131"></a>

### DIFF
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -269,7 +269,7 @@ export const drawNodes = function (changed) {
       );
   }
 
-  if (changed('nodes', 'nodeActive', 'hoveredFocusMode')) {
+  if (changed('hoveredFocusMode', 'nodes')) {
     allNodes.classed(
       'pipeline-node--faded',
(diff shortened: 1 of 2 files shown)


tynandebold (commented):
It's definitely better but still not back to the original. I see some flashing when I click on a node and the app makes the switch from normal to faded.

At times I also see this when I click on a node but don't move my mouse:

image

The nodes should be greyed out, like the edges (lines between nodes), but they aren't yet. Only when I move my mouse do they become faded.

tynandebold (approved):
Looks good!

Huongg (approved):
LGTM too 👍

PR #1138, opened 2022-10-19
Feature/url parameters

### LINKED ISSUE #1117: Create URL parameters for each element/section in kedro-viz
## Description
Parameters are placeholders in the URL that begin with a colon. 

## Context
With this feature, every time user clicks on a node in the flowchart, or a run in the experiment tracking, it will reflect on the URL, for example:
`https://demo.kedro.org/experiment-tracking/run-id`
or
`https://demo.kedro.org/flowchart/selected:modular_pipeline_id` (we would need to think and agree on the path for workflow as its more complicated than the experiment tracking

Once we have this URL parameter, we can do other things such as store the path in local storage so when the user re-enters the URL, it will open where it was left before, or it can also be helpful for this idea[ here ](https://github.com/kedro-org/kedro-viz/issues/1116)

## Possible Implementation
This can be achieved using React router (which we already have in our kedro-viz). Since there will be a lot of different paths, maybe best if we can have `routes.config` in which we define different path like this

```
export const routes = {
  flowchart: {
      main: '/flowchart',
      selectedModularPipeline: '/flowchart/selected:modular_pipeline_id',
      focusedModularPipeline: '/flowchart/focused:modular_pipeline_id',
      selectedDataset: '/flowchart/selected:dataset_id',
    },
   experimentTracking: {
   .....
   }
}
```

## Checklist

Huongg:
## Description
Fixes #1117 

I use Query parameters rather than URI parameters, hence all of them are combined into a single string as below

```
main: '/',
focusedNode: `/?${pipelineActive}&${params.focused}:id`,
selectedNode: `/?${pipelineActive}&${params.selected}:id`,
```

As discussed with the engineer and design team, the URL parameter should cover a few scenarios, such as
- when selecting/clicking on a single node , the URL will include`selected_id:...`
- it also recognises whether the node belongs to a modular pipeline or not, if it does then expand the modular pipeline list
- when focusing on a modular pipeline the URL will include `focused_id:...`
- we don't update the parameter when switching between different pipeline views but we do include the view in the URL parameter when the user performs an action. eg `?pipeline_id=Modelling%20stage&selected_id=cae2d1c7`
- it also works when you click on the flowchart

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1138"></a>


src/components/metadata/metadata.js
```diff
@@ -73,6 +75,8 @@ const MetaData = ({
   const onCloseClick = () => {
     // Deselecting a node automatically hides MetaData panel
     onToggleNodeSelected(null);
+    // and reset the URL to '/'
+    toFlowchartPage();
```
tynandebold, line 79: This works great when the X icon is selected. though when you select the background of the Flowchart when the Metadata panel is open to close the panel the URL isn't reset. You'll need to reset it there as well.

src/components/pipeline-list/pipeline-list.js
```diff
@@ -30,7 +33,11 @@ export const PipelineList = ({
         onOpened={() => onToggleOpen(true)}
         onClosed={() => onToggleOpen(false)}
         width={null}
-        onChanged={onUpdateActivePipeline}
+        onChanged={(value) => {
+          onUpdateActivePipeline(value);
+          // reset the URL to '/' when switching between different view
```
tynandebold, line 38: Nit: capitalize.

```suggestion
          // Reset the URL to '/' when switching between different view
```

tynandebold (commented):
So far so good! This is amazing and I love that you've added this. Well done.

I'm just adding some initial comments now and then will do another review once you push the changes you're adding based off what I mentioned to you offline a couple of hours ago :)

src/components/metadata/metadata.js
Huongg: hey yeah i included this point now, thank you 😄

src/routes.config.js
```diff
@@ -0,0 +1,17 @@
+export const params = {
```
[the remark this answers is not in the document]
Huongg: ah yeah i like your suggestion actually. It will work well if i move them all in the config 👍

src/components/wrapper/wrapper.js
```diff
@@ -24,6 +24,18 @@ export const Wrapper = ({ displayGlobalToolbar, theme }) => {
...
+  // only when the page is reloaded
+  const [reload, setReload] = useState(true);
+  useEffect(() => setReload(true), []);
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setReload(false);
+    }, 200);
+
+    return () => clearTimeout(timer);
+  }, []);
```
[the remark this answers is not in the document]
Huongg: ah yeah it should be false 👍 good spot thank you

src/routes.config.js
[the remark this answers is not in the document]
tynandebold: I'm seeing they're still in a separate file. Just curious if you're going to leave them in here?

src/selectors/metadata.js
```diff
@@ -58,7 +58,7 @@ export const getClickedNodeMetaData = createSelector(
     nodeTranscodedTypes,
     nodeRunCommand
   ) => {
-    if (!nodeId) {
+    if (!nodeId || Object.keys(nodeType).length === 0) {
```
tynandebold, line 61: Why is this `||` check needed? Maybe I'm missing something...

src/utils/hooks/use-generate-pathname.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * To generate different pathnames based on each action
+ * E.g.: click on a node, or focus on a modular pipeline
+ * or to reset the pathname to the main page
+ */
+export const useGeneratePathname = () => {
+  const history = useHistory();
+
+  const localStorage = window.localStorage.getItem(localStorageName);
+  const activePipeline = JSON.parse(localStorage)?.pipeline?.active;
```
tynandebold, line 14: I think you need to read from the URL here first, and if the URL isn't there then you could also get this from `state.pipeline.active` if you want to. 

Rashida and I were passing around URLs to each other while reviewing and we were seeing that if my `localStorage` pipeline value didn't match what was in her sent URL then the behavior wouldn't work as we intend it. That's why I think we need to read from the URL first, and then move on to what's on the user's machine.

If you use Redux instead of `localStorage`, you'd probably have to pass that as an argument into your hook and then use it here. I don't see any issue with `localStorage` either though. They'll always be in sync.

src/utils/hooks/use-generate-pathname.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const toFlowchartPage = useCallback(() => {
+    const url = generatePath(routes.flowchart.main);
+    history.push(url);
+  }, [history]);
+
+  const toSelectedNode = useCallback(
```
tynandebold, line 21: I think I mentioned this is the demo on Friday: could we also have a function here much like these other three called something like `toSelectedPipeline` where we only add the `pipelineId` to the path and then get something like this on the frontend?

image

Or is this what you were saying about us always needing **both** the `pipelineId` and `id` in the URL?

src/utils/hooks/use-generate-pathname.js
```diff
@@ -0,0 +1,49 @@
...
+    (item) => {
+      const url = generatePath(routes.flowchart.selectedNode, {
+        pipelineId: activePipeline,
+        id: item.id,
+      });
+      history.push(url);
```
tynandebold, line 27: I'm noticing now that we're pushing history to the browser, which is normally fine, but when I interact with the forward or back buttons to change the URL the page doesn't update. Given your code, that makes sense, because you're only trigger that on `reload`. 

So I think we either need to allow the page to update when using the navigation buttons or not push the URLs into the browser history. I think my preference is for the former. What do you think?

src/utils/hooks/use-redirect-location.js
```diff
@@ -0,0 +1,65 @@
+import { useMemo } from 'react';
+import { useLocation, matchPath } from 'react-router-dom';
+import { routes, params } from '../../routes.config';
+
+/**
+ * to trigger different actions based on the current location in the flowchart
+ * and this hook is only called when the page is reload
+ */
```
tynandebold, line 8:
```suggestion
/**
 * To trigger different actions based on the current location in the Flowchart.
 * This hook is only called when the page is reloaded.
 */
```

src/utils/hooks/use-redirect-location.js
```diff
@@ -0,0 +1,65 @@
...
+    if (matchedFocusedNode && Object.keys(modularPipelinesTree).length > 0) {
+      const modularPipelineId = search.split(params.focused)[1];
+      const modularPipeline = modularPipelinesTree[modularPipelineId];
+
+      onToggleFocusMode(modularPipeline.data);
+      onToggleModularPipelineActive(modularPipelineId, true);
+      onUpdateActivePipeline(decodedPipelineId);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [reload]);
```
tynandebold, line 64: To my previous comment about using the browser navigation forward/back buttons: if you add `search` to the dependency array it helps, though it isn't perfect.

tynandebold (changes requested):
This is amazing! I'm still loving it. I have some comments that need to be addressed before merging. Let me know if anything is unclear or if you want to go over anything.

One big thing: if I open Viz in an incognito window and type a `nodeId` that doesn't exist in the URL and visit the page the whole app breaks. We should try and avoid that: 

image

Lastly, I sometimes see this random error in the console:

```
index.js:1 Warning: Cannot update a component (`Connect(MiniMap)`) while rendering a different component (`FlowChartWrapper`). To locate the bad setState() call inside `FlowChartWrapper`, follow the stack trace as described in https://fb.me/setstate-in-render
    in FlowChartWrapper (created by Connect(FlowChartWrapper))
    in Connect(FlowChartWrapper) (at wrapper.js:81)
    in Route (at wrapper.js:80)
    in Switch (at wrapper.js:79)
    in Router (created by BrowserRouter)
    in BrowserRouter (at wrapper.js:66)
    in ApolloProvider (at provider.js:7)
    in GraphQLProvider (at wrapper.js:65)
    in div (at wrapper.js:57)
    in Wrapper (created by Connect(Wrapper))
    in Connect(Wrapper) (at app.js:65)
    in Provider (at app.js:64)
    in App (at container.js:12)
    in Container (at src/index.js:5)
```

src/utils/hooks/use-redirect-location.js
Huongg: yeah good point let me try that

src/components/flowchart/flowchart.js
```diff
@@ -446,8 +446,11 @@ export class FlowChart extends Component {
   handleNodeClick = (event, node) => {
     if (node.type === 'modularPipeline') {
       this.props.onClickToExpandModularPipeline(node.id);
+      // To reset URL to '/' when expanding modularPipepline as we dont track this action
+      this.props.toFlowchartPage();
```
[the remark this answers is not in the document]
Huongg: hmm yeah we dont need to, I've updated this as I thought the expanding is still an action, and if we dont update the URL to reflect this maybe we should set this to default, but I also have no issue to not update it back to the default.

src/selectors/metadata.js
tynandebold: Got it, thanks!

Huongg:
> This is amazing! I'm still loving it. I have some comments that need to be addressed before merging. Let me know if anything is unclear or if you want to go over anything.
> 
> One big thing: if I open Viz in an incognito window and type a `nodeId` that doesn't exist in the URL and visit the page the whole app breaks. We should try and avoid that:
> 
> image
> 
> Lastly, I sometimes see this random error in the console:
> 
> ```
> index.js:1 Warning: Cannot update a component (`Connect(MiniMap)`) while rendering a different component (`FlowChartWrapper`). To locate the bad setState() call inside `FlowChartWrapper`, follow the stack trace as described in https://fb.me/setstate-in-render
>     in FlowChartWrapper (created by Connect(FlowChartWrapper))
>     in Connect(FlowChartWrapper) (at wrapper.js:81)
>     in Route (at wrapper.js:80)
>     in Switch (at wrapper.js:79)
>     in Router (created by BrowserRouter)
>     in BrowserRouter (at wrapper.js:66)
>     in ApolloProvider (at provider.js:7)
>     in GraphQLProvider (at wrapper.js:65)
>     in div (at wrapper.js:57)
>     in Wrapper (created by Connect(Wrapper))
>     in Connect(Wrapper) (at app.js:65)
>     in Provider (at app.js:64)
>     in App (at container.js:12)
>     in Container (at src/index.js:5)
> ```

ah great point, I keep telling myself I need to do some error handling but keep forgetting about it and doing something else.  Do we have an oopps page or Not found page in kedro-viz that i re-use for these scenarioes? @tynandebold

tynandebold:
I think you could use [this file](https://github.com/kedro-org/kedro-viz/blob/main/src/components/pipeline-warning/pipeline-warning.js) with the `isEmptyPipeline` messaging :)

src/utils/hooks/use-generate-pathname.js
Huongg: hey yeah i think we need both in the URL as we're using query parameter. So if you just update the pipelineId, the URL will miss the other id `http://localhost:4141/?pipeline_id=Modelling%20stage&selected_id=`

Unless theres away to show `&selected_id=` conditionally in the URL

src/utils/hooks/use-generate-pathname.js
Huongg: ah yeah good point, i didnt think about that scenario, will update this 👍

Huongg: hmm actually after looking at this again, it's a bit unclear to me. Maybe best we can sync up and you can show me this case in more details? @tynandebold

src/components/node-list/index.js
```diff
@@ -174,6 +184,12 @@ const NodeListProvider = ({
     }
   };
 
+  const handleToggleModularPipelineExpanded = (expanded) => {
+    onToggleModularPipelineExpanded(expanded);
+    // To reset URL to '/' when expanding modularPipepline as we dont track this action
+    toFlowchartPage();
+  };
```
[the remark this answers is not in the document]
Huongg: done 👍

src/utils/hooks/use-redirect-location.js
Huongg: hey so my update for this issue include: 
- pass "search" to the dependency array so it can track when it changed
- update value null to both `onLoadNodeData` and `onToggleFocusMode` when the URL is matched the main page
```
if (matchedFlowchartMainPage) {
      onLoadNodeData(null);
      onToggleFocusMode(null);
    }
```
- ensure we also set `onToggleFocusMode(null)` when selectedNode path is matched, and the other way round for when focusedNode path is matched
```
// Reset the focus mode to null when when using the navigation buttons
      onToggleFocusMode(null);


// Reset the node data to null when when using the navigation buttons
      onLoadNodeData(null);
```

Huongg:
hey @tynandebold @rashidakanchwala I've got the error message working now, since I dont have the design yet but the message looks something like this

Screenshot 2022-10-26 at 15 09 08
Screenshot 2022-10-26 at 14 49 00


As you can see in my changes, I have the `useValidateUrl` to do the API call, however it wouldn't work with the `/api/pipelines/id` because as soon as the end point is wrong for pipelines it will throw error straight away, so it doesnt even get to load the page at all to do the fetch call

Screenshot 2022-10-26 at 15 13 07


and in the `useRedirectLocation` i still need to check if the ID is existed or not anyway, **so maybe to simplify it I can just handle this error directly in the useRedirectLocation , like this commit I did before** [here](https://github.com/kedro-org/kedro-viz/pull/1138/commits/b3dea01160a475cd74c80da8ec011d449b9352af)

Huongg:
> That error message is looking really good! Well done.
> 
> Why does the page crash with an incorrect pipeline ID but not an incorrect node ID? In any event, I think how you handled that before in your previous linked commit can work. The main thing is that the app doesn't crash like in your third screenshot above.

hey sounds good, I've gone back to how I did it before, the error messages should now cover all cases, incl invalid pipeline ID

tynandebold (approved):
Amazing! Approved. Great work on this one, Huong. Thanks so much for seeing it through.

I have one worry that some of the URL params we're adding may potentially mess with some of the jupyter-server-proxy stuff @AntonyMilneQB worked on that I helped him with, though it's hard to tell right now. I don't remember how to test that at the moment, so I think for now it'll have to be ok!

src/components/pipeline-warning/pipeline-warning.js
```diff
@@ -63,6 +68,26 @@ export const PipelineWarning = ({
...
+            {
+              'pipeline-warning--sidebar-visible': sidebarVisible,
+            }
+          )}
+        >
+          <h2 className="pipeline-warning__title">
+            Oops, the URL is not correct
+          </h2>
+          <p className="pipeline-warning__subtitle">
+            {`It seems like it has "${errorMessage}"`}
```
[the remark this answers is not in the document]
Huongg: hey yeah that sounds good to me too 👍

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -2,42 +2,108 @@ import React from 'react';
...
+    pipelines,
+    reload
+  );
+
+  if (invalidUrl) {
+    return (
+      <div className="kedro-pipeline">
+        <Sidebar />
+        <MetaData />
+        <PipelineWarning errorMessage={errorMessage} invalidUrl={invalidUrl} />
```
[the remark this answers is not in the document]
Huongg: hey the reason I separate them with if/else statements so it won't render the Flowchart + any other things, cause other wise the flowchart will still show up in the background

rashidakanchwala (changes requested):
Hi @Huongg  --- looks' greatt!! <3  

i just came across another error handling issue, when we type something incorrect in the focus mode then it's a blank screen. 

try http://localhost:4141/?pipeline_id=__default__&focused_id=featured_engineering instead of http://localhost:4141/?pipeline_id=__default__&focused_id=feature_engineering

Huongg:
> Hi @Huongg --- looks' greatt!! <3
> 
> i just came across another error handling issue, when we type something incorrect in the focus mode then it's a blank screen.
> 
> try http://localhost:4141/?pipeline_id=__default__&focused_id=featured_engineering instead of http://localhost:4141/?pipeline_id=__default__&focused_id=feature_engineering

hey @rashidakanchwala sorry i didnt push my latest changes to it, it should be fixed now with the invalid Modular Pipeline

rashidakanchwala (approved):
Thanks @Huongg .. super excited abt this feature release :)

Huongg:
I've updated the copy for the error message suggested by @stichbury. @comym @Mackay031 if you're happy can i get approval for this PR please? 

Screenshot 2022-11-01 at 11 45 42

comym (approved):

PR #1139, opened 2022-10-20
Fix behaviour for icons on the utility bar

### LINKED ISSUE #909: Fix behaviour for icons on the utility bar
- Make tooltips for icons on the utility bar behave as snappy as the one on the flowchart

Suggested workflow: 

1. First trigger = 1s (works correctly) 
2. Series of trigger events = no need for 1s delay, make it snappy as they are at the moment in the flowchart

Reported by @GabrielComymQB, here is detailed explanation: https://github.com/kedro-org/kedro-viz/pull/896#commitcomment-76064403

cvetankanechevska:
## Description

Fixes #909 

## Development notes
Changed the way tooltips on IconButtons are displayed. Instead of hovering the element and changing the visibility from the css, I've created a local state which is updated accordingly `onMouseEnter` and `onMouseLeave` events.

When a user triggers the `onMouseEnter` event for the first time in the current session, that is saved in the localStorage. So only in this scenario we will have 2s delay for displaying the tooltip, after that tooltips are displayed immediately.

**Q:** What is missing is when should the localStorage be cleared?

**Tests:**

Last three tests for the IconButton component are failing, so I will need help with that part.

I think the problem is here:

```
{labelText && isTooltipVisible && (
  <span
    className={classnames(
       'pipeline-toolbar__label',
        `pipeline-toolbar__label-${labelPosition}`
    )}
  >
     {labelText}
  </span>
)}
```
      
as now we are having plus `isTooltipVisible`, so for these tests to pass we need to specify that `isTooltipVisible=true`.

## Checklist

### DIFF
--- a/src/components/ui/icon-button/icon-button.js
+++ b/src/components/ui/icon-button/icon-button.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import './icon-button.css';
@@ -25,11 +25,28 @@ const IconButton = ({
 }) => {
   const Icon = icon;
 
+  const [isTooltipVisible, setIsTooltipVisible] = useState(false);
+
   const labelPosition = labelPositionTypes.includes(
     labelTextPosition.toLowerCase()
   )
     ? labelTextPosition.toLocaleLowerCase()
     : 'right';
+
+  const showTooltip = () => {
+    if (localStorage.getItem('delayShow') === null) {
+      window.localStorage.setItem('delayShow', true);
+      setTimeout(() => {
+        setIsTooltipVisible(true);
+      }, 2000);
+    } else {
+      setIsTooltipVisible(true);
+    }
+  };
+
+  const hideTooltip = () => {
+    setIsTooltipVisible(false);
+  };
 
   return visible ? (
     <Wrapper container={container}>
@@ -43,9 +60,11 @@ const IconButton = ({
         data-heap-event={dataHeapEvent}
         disabled={disabled}
         onClick={onClick}
+        onMouseEnter={showTooltip}
+        onMouseLeave={hideTooltip}
       >
         {Icon && <Icon className="pipeline-icon" />}
-        {labelText && (
+        {labelText && isTooltipVisible && (
           <span
             className={classnames(
               'pipeline-toolbar__label',
--- a/src/components/ui/icon-button/icon-button.scss
+++ b/src/components/ui/icon-button/icon-button.scss
@@ -62,14 +62,13 @@ $triangle-size: 7px;
   display: inline-block;
 
   font-size: 1.4em;
-  opacity: 0;
+  opacity: 1;
 
   padding: 0.5em 0.7em;
   pointer-events: none;
 
   position: absolute;
 
-  transition: opacity ease 0.1s;
   white-space: nowrap;
 
   // to ensure the tooltip will show on the top of the bookmark dropdown
@@ -86,9 +85,7 @@ $triangle-size: 7px;
 
   .pipeline-icon-toolbar__button:hover &,
   [data-whatinput='keyboard'] .pipeline-icon-toolbar__button:focus & {
-    opacity: 1;
     pointer-events: all;
-    transition-delay: 1s;
   }
 
   &::after {


Huongg:
hey @cvetankanechevska great work 💃  , just a quick thought, rather than using the state to show the tooltip or not, do you think we can use that to add the classnames instead? Then modify the styling based on its classnames. It will help with the failed tests and also good for the page performance as we don't need to keep re-rendering it. something like this 

```
{labelText && (
  <span
    className={classnames(
       'pipeline-toolbar__label',
        `pipeline-toolbar__label-${labelPosition}`
    ), { ['pipleline-toolbar__label-visible']: isTooltipVisible}
  >
     {labelText}
  </span>
)}
```

I think we can re-set the storage whenever the page is reload? what do you think @cvetankanechevska ? 

Also @tynandebold you suggested using redux store before but I think local storage is better as we want to be able to reuse this in the experiment tracking, and i dont think redux store would work with the experiment tracking?

src/components/ui/icon-button/icon-button.js
```diff
@@ -25,12 +25,29 @@ const IconButton = ({
...
+
   const labelPosition = labelPositionTypes.includes(
     labelTextPosition.toLowerCase()
   )
     ? labelTextPosition.toLocaleLowerCase()
     : 'right';
 
+  const showTooltip = () => {
+    if (localStorage.getItem('delayShow') === null) {
+      window.localStorage.setItem('delayShow', true);
```
Huongg, line 38: hey i think we will need to have some naming convention for the storage here rather than just `delayShow`. Something like `kedro-viz-tooltip-show`, maybe best to discuss with the team?

cvetankanechevska:
> hey @cvetankanechevska great work 💃 , just a quick thought, rather than using the state to show the tooltip or not, do you think we can use that to add the classnames instead? Then modify the styling based on its classnames. It will help with the failed tests and also good for the page performance as we don't need to keep re-rendering it. something like this
> 
> ```
> {labelText && (
>   <span
>     className={classnames(
>        'pipeline-toolbar__label',
>         `pipeline-toolbar__label-${labelPosition}`
>     ), { ['pipleline-toolbar__label-visible']: isTooltipVisible}
>   >
>      {labelText}
>   </span>
> )}
> ```
> 
> I think we can re-set the storage whenever the page is reload? what do you think @cvetankanechevska ?
> 
> Also @tynandebold you suggested using redux store before but I think local storage is better as we want to be able to reuse this in the experiment tracking, and i dont think redux store would work with the experiment tracking?

Hey @Huongg , I've tried what you have suggested but other things regarding styling are not into place. I think it's easier in this case to modify the tests for this component, rather than styling.

Huongg:
hey @cvetankanechevska hope the styling issue is fixed now after our call, just a quick note, can you have 1 or 2 tests to cover this change, please

comym (changes requested):
Hi,

The triggering mechanism of the tool tip works very well at the beginning. As intended. 
The tool tip does not show at first and then only starts showing when the user hovers on it for more than X milliseconds. This is good.

However, the tooltip is always on after that, and it never goes back to the default state.
The default state is not showing the tooltip straight away.

What I'm saying is, the idea is to keep the interaction and timings aligned to the way we do when interacting with the flowchart elements.

The icon tooltips have to also go back to the default state at some point, and not stay indefinitely on.

Hope that makes sense. If not, I can do a quick demo. Just drop me a message on Slack if you have any questions.

src/components/ui/icon-button/icon-button.test.js
```diff
@@ -1,6 +1,15 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
 import IconButton from '.';
 import { setup } from '../../../utils/state.mock';
 import MenuIcon from '../../icons/menu';
+import { act } from '@testing-library/react';
+
+// jest.mock('react', () => ({
```
Huongg, line 9: if this is not used can we remove this?

src/components/ui/icon-button/icon-button.test.js
```diff
@@ -58,4 +67,49 @@ describe('IconButton', () => {
 
     expect(wrapper.find('.pipeline-toolbar__label-right').length).toBe(1);
   });
+
+  it('show tooltip', () => {
+    jest.useFakeTimers();
+    const setIsTooltipVisible = jest.fn();
+    jest.useFakeTimers();
```
Huongg, line 74: why do you need to set `useFakeTimers()` twice?

Huongg (approved):
it's working fine for me, the tooltip behaviour does get reset every time i refresh the page :) good job. I like the tests too

src/components/ui/icon-button/icon-button.test.js
cvetankanechevska: Yes, thanks for noticing this. It's added by mistake.

src/components/ui/icon-button/icon-button.test.js
cvetankanechevska: Yes, thanks for noticing this, it's added by mistake.

src/components/ui/icon-button/icon-button.test.js
cvetankanechevska: Yes, I cleared all comments and unused parts.

cvetankanechevska:
> Hi,
> 
> The triggering mechanism of the tool tip works very well at the beginning. As intended. The tool tip does not show at first and then only starts showing when the user hovers on it for more than X milliseconds. This is good.
> 
> However, the tooltip is always on after that, and it never goes back to the default state. The default state is not showing the tooltip straight away.
> 
> What I'm saying is, the idea is to keep the interaction and timings aligned to the way we do when interacting with the flowchart elements.
> 
> The icon tooltips have to also go back to the default state at some point, and not stay indefinitely on.
> 
> Hope that makes sense. If not, I can do a quick demo. Just drop me a message on Slack if you have any questions.

Hey @comym, thanks for this. I made some changes. Can you please recheck again ?

src/components/ui/icon-button/icon-button.js
```diff
@@ -24,13 +24,38 @@ const IconButton = ({
...
+      timeout = setTimeout(() => {
+        setIsTooltipVisible(true);
+      }, 1000);
+    } else {
+      setIsTooltipVisible(true);
+    }
+  };
+
+  const hideTooltip = () => {
+    setTimeout(() => {
```
Huongg, line 52: hey @cvetankanechevska   I think it would be nice to have a comment here to explain why the local storage needs to be reset this way, so whoever works on this component next will have some context.

tynandebold (approved):
Works great! Nice job bringing this all the way.

comym (approved):

Issue #760, opened 2022-02-28
Offer "expand all" and store what was expanded across page refreshes


pascalwhoop:
## Description
The UI always collapses sub-pipelines in an effort to simplify the pipeline. But this makes it hard to see the whole picture easily. 
I have 2 nodes
`A` and `B` and 4 datasets
```
layer_1_in_A
layer_1_in_B
layer_2_out_A
layer_2_out_A

```
(made up names) and the UI collapses them and puts them _next_ to the functions. Only when I expand all layers in the UI is the structure as expected

## Context
It would be nice to have the possibility to keep the pipeline expanded between page reloads. This way, I can use `--autoreload`  and watch the pipeline's progress while developing and also not have to manually expand every single subpipeline / layer combo. 

## Checklist

- [ ] Include labels so that we can categorise your feature request

datajoely:
So @pascalwhoop is it just persisting the expand state between runs - or would you like an 'expand all' button?

pascalwhoop:
I think both is valuable, what do you think? I just started a new project. 2 functions, 5 datasets already 5 things to expand
![image](https://user-images.githubusercontent.com/2284951/155979422-19a3555d-c74c-47ef-9586-2186c62e0930.png)

studioswong:
Hello @pascalwhoop , thank you for reporting this issue, this is a very valid problem and we completely hear you!  👍 

We are currently in the midst of revisiting our local storage setup which will enable us to efficiently persist the expanded state of the flowchart on refresh as suggested here - this will be in one of our upcoming sprints, and we will let you know once we have shipped this in our future release.

datajoely:
I would personally love a button to expand everything to the max, but that's not as high priority

limdauto:
Yea, this has been asked before. I can hack in an expand all button. Shouldn't be too much work. Might need some help sorting out a new icon though.

studioswong:
Hi @pascalwhoop - we are currently implementing an 'expand all' experimental feature that would be accessible via feature flags, which would be accessible in our next release - we will provide you with more details of accessing this feature once the new release is out.

tynandebold:
This feature is now released! Toggle the `Expand all modular pipelines` flag on in the settings panel. 

<img width="895" alt="image" src="https://user-images.githubusercontent.com/16869061/168624231-d46eae30-5790-430d-9e3d-2a7851ba53ca.png">

PR #1143, opened 2022-10-24
Highlight dataset nodes via a flag in the settings panel

tynandebold:

## Description

During an open-source interview on Friday, one of our users said something along the lines of, "I'd like datasets to stand out more on the visualization so they're distinct and easier to see." This PR adds this request behind a feature flag.

This is the quote from the user:

> "It will be nice to see an even bigger difference between the function and the dataset. So to see the distinction, a clear distinction when you look at the visualization. Cause currently the shape is slightly different and you have a symbol but it would be nice maybe to have different colors or something like that even to make it clear. Sometimes when you look at, we have projects in which we have 20 data sets and you end up with a lot of functions in datasets. If you are looking at quickly high level, it's a bit difficult to understand what is a function and what is an output. "

## Development notes

I've added a new experiment to the settings modal to turn this feature on. All it does is add another class name to the flowchart nodes and updates the style accordingly.

## QA notes

This is what it looks like when enabled:

image

To test, view the Gitpod URL and play around with it. I'm sure some improvements can be made by design (cc @comym and @Mackay031 ). This is just a first pass.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1143"></a>


comym:
Hi @tynandebold 

I think this is great as an experiment. I also understand this is the first pass and design improvements have to be done. It's good to see this might be a quick win.

However, I'd love to hear from @yetudada her opinion on that. Cause I believe we don't need to ship extra functionalities just because a single user mentioned he would like to have them. Otherwise, every "feature shopping list" can potentially become a flag.
I just wanted to make this point out and reflect on it.

However, if we feel this is interesting to be explored I'm happy to go for it, and then we can track the flag and see if that's something that is well used.

Overall, that's super!

rashidakanchwala (approved):
Thanks, LGTM!

tynandebold:
@comym I get your points and am happy you raised them.

@yetudada what do you think of Gabriel's comments above?

yetudada:
@tynandebold I agree with @comym on this one for the following reasons:

 -  Users must put in the effort to enable a feature flag; they have to go to the settings panel and turn on the feature. With that in mind, I prefer using the flags to quickly ship higher-value features more commonly heard from our users (not just one). 
 - This design needs designer input because it assumes you're not colour-blind. Unfortunately, we already set up our sprint objectives. I would need powerful motivation to break the sprint to recommend doing this. 

I like that there was quick experimentation to solve this problem for a user, but I would have preferred to see this ticketed and added to the backlog instead.

tynandebold:
Understood. All valid points. Thank you for the response! Two things from me:

1. I do think we've shipped things in the past (e.g. [here](https://github.com/kedro-org/kedro-viz/issues/760) and [here](https://github.com/kedro-org/kedro-viz/issues/874)) that have only been requested by one or two users, excluding members of our team.
2. I chose a color that's already utilized in Viz to account for this potential color-blindness problem and maybe even avoid heavy designer input. My thinking was that (hopefully) it's been thought of here, so it should work if used again in a slightly different manner.
    image

Perhaps a feature flag isn't the way to go, and I agree with that. I'm not sure what's best, as it's really the only mechanism we have for something like this. And to be clear, I don't want to break the sprint and reprioritize anything. I apologize if that's how this came across.

For me, this PR is more of a response to what @idanov and I were raising last week: taking a leap of faith to try and build new things. This seemed like a low-risk example of how we could do that.

PR #1152, opened 2022-11-07
Feature/parallel chart

### LINKED ISSUE #1154: Build out single-run journey for parallel-coordinates plot
This is a child ticket of #1146 and is focused on the single-run journey only.

Huongg:
## Description

Resolves #1154 

Single run for parallel coordinates chart using dummy data

## Development notes

Features and function that include in this PR:
- Matching the design for the chart in single run mode

WIP:
Screenshot 2022-11-14 at 20 08 18

Design:
Screenshot 2022-11-14 at 15 13 32

- Hover interaction between the run list and the chart itself. For example, when you hover over a line in the chart, or from the run list, it will highlight the line and labels of that run, however, all the related label values in each axis won’t be highlighted.
- Hover interaction always works for the Y-axis (metrics name)

- When the run hovers, its line always comes on top of any other lines. 

What does not include in this PR: the Tooltip, the chart in comparison mode, light theme

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1152"></a>


src/components/experiment-tracking/parallel-coordinates/parallel-coordinates.js
```diff
@@ -0,0 +1,215 @@
...
+  return (
+    <div className="parallelCoordinates">
+      <svg
+        ref={ref}
+        style={{
+          height,
+          width,
+        }}
+      >
+        {/* {graphKeys.map((graph) => (
```
Huongg, line 124: I can't find a way to make this Yaxis work in the component yet, still WIP

comym:
Hi,

Here are my comments:

1 - Background colour is wrong. It should be Slate-900 all throughout and not have multiple divisions of colours as you see here. The top tab bar should blend with the background of the whole page.

Screenshot 2022-11-14 at 15 55 34


2 - Is it my impression or are we using 2px lines in the graph? It all looks chunkier. Not sure, just wanted to double-check. We should use 1px lines.


Screenshot 2022-11-14 at 15 57 01

3 - There's some sort of tick/dash mark on top of the first metric label.

Screenshot 2022-11-14 at 15 57 32


4 - The bottom margin is too narrow. We should give it more room. The metrics axis should shorten and the bottom margin increase in size.

Screenshot 2022-11-14 at 16 00 13


5 - The lines are all crossing on top of the metric axis. This should not happen. Axis, value labels and their complementary lines should always be on top. Lines should be at the back and label values and tick marks on top.

Screenshot 2022-11-14 at 16 04 25

Screenshot 2022-11-14 at 16 04 42

Screenshot 2022-11-14 at 16 16 12

Screenshot 2022-11-14 at 16 16 18



6 - One of my main concerns is, it's still very fiddly to hover over any of the lines in the graph. 

7 - When hovering on the card, the lines are flickering when moving the mouse inside the respective cards. This should not happen.

Screen Recording 2022-11-14 at 16 09 54


8 - Top and bottom labels of the axis range should not be this bright and they should not be semi-bold.

Screenshot 2022-11-14 at 16 13 55


9 - Font sizes for selected run values and non-selected ones are different. It should be the same.

Screenshot 2022-11-14 at 16 22 28


10 - On hover, run labels show to be inconsistent. Sometimes highlighted, in a different colour, or non-existent.

Screenshot 2022-11-14 at 16 26 43


11 - On hover, neither the highlighted metric axis nor the run lines should be white FFFFFF. 

Screenshot 2022-11-14 at 16 29 04



Please use the provided designs for reference. All colours and styles are there. 
We'll be working on the light version for that too. 

Thanks

Huongg:
hey @comym thanks for taking a look at this PR and thanks for the feedback. Very helpful as I've been looking at this everyday and it's hard to spot these issues. 

I belive I've addressed your comments with more details as below:

**1. Background colour is wrong**. Yes noted this is wrong but it's something defined from the code base a long time ago. Will double check with Tynan about this issue. 

2. Line width is now set back to the default value which is 1px lines.

3. Those ticks on the top of the chart have now been removed

4. I've increased the margin for the chart bottom

5. The lines are now rendered at the bottom of the chart, so that the labels and other lines can stay on the top

6. I see your concern about the behaviour of hovering over the line. We haven't found the perfect solution yet, I'd suggest including this item to the next iteration. What do you think @tynandebold ?

7. This is now fixed, when hovering over the run card the lines wont be disappeared/appeared again

8. I've doubled checked and updated all the colours according to the figma files. Let me know if you spot anything else? 

9. Font sizes are now also updated with the figma file. However, on the selected run, the values do come with a font-weight: 600, which will make it looks slightly different from other values. Is it correct?

comym:
Hi @Huongg 

Thanks for the comments and iteration. My feedback is:

1 - Lines are still crossing on top of the axis and label values.

Screenshot 2022-11-15 at 11 17 57


2 - The top and bottom range value labels seem to be too bright. Make sure they are Grey-300 in their original state.

Screenshot 2022-11-15 at 11 22 17

3 - The top and bottom range value labels seem to have a different size/weight than the other labels in the range.

Screenshot 2022-11-15 at 11 24 49

4 - The bottom margin is not right. The way I see it here, it now leaves a giant gap at the bottom. The margins should be proportional to the viewport and always show a balanced layout this way. At the moment the margins are varying and affecting the layout depending on the screen size.

Screenshot 2022-11-15 at 11 32 50


5 - To **your question** number 9. It's correct. The weight of the selected runs anchor point values has a different weight and colour than the rest.


Many thanks

tynandebold:
@comym I think I've addressed all your concerns from your previous comment. Please have a look at let me know what you think!

comym:
Hi,

After checking Gitpod these are my current comments:

1 - When hovering over a run line, the plot highlights some random values across different metrics and positions. This should not happen. Only the related values to the on-hover run should show.

Screenshot 2022-11-23 at 13 00 31


2 - Please double-check if the colour of the text values is the same as the line in the background when hovering over a run. They **should not** be the same. The different colours allow the text on top to be readable. The design shows the colours to be different as seen below. 

Screenshot 2022-11-23 at 16 09 26


3 - When hovering over a metric name, the current build only highlights the axis line and some values. This is right but this is also only one of the changes that should occur in this interaction. What we're missing here is to knock back (decrease in contrast) all the other metric columns that are not being highlighted, so that the highlighted one can pop to the front. Also in this state, we knock back many of the other lines and labels. Basically eliminating all the unnecessary noise at that point. See the design for reference.

Screenshot 2022-11-23 at 16 11 39


4 - Uneven margins/padding. The current version has improved the big gap we had before, however it's still unbalanced. In the image below, if we consider the two margins/padding in blue to be the same, we can still see there's a big difference between the top and bottom - marked in red.  We should even it out. Please double-check the designs for reference.

Screenshot 2022-11-23 at 16 15 21


5 - I've recently checked Gitpod again and I've seen some variations of it. Now showing only two metric columns. Not sure if that's working in progress but I'll comment anyway just in case.

Metric names are gigantic and label values have too many decimal values. What's happening is that they are being caped by some sort of mask or area limit as seen below. We should not show all these decimal values and shorten them somehow. I'm happy to talk about it and explore solutions if this is going to be a reality.

Screenshot 2022-11-23 at 16 21 02


6 - Overlapping labels. As marked in red below. There's a mix of very long labels and labels with fewer decimal values. These values are unreadable. I'm happy to talk about it and explore solutions if this is going to be a reality.

Screenshot 2022-11-23 at 16 23 38


If you have any questions just reach out and we can talk about it in detail instead of writing long text blocks here.

Many thanks!

tynandebold (approved):

PR #1195, opened 2022-12-07
Feature/final touches on metrics plots

### LINKED ISSUE #1146: Finish the parallel coordinates metrics plot
Things still to do (some taken from comments in #1152):

**Must complete:**

- [x] When runs are selected, the runs list card needs to show the same shapes as the lines on the metrics chart (diamond, circle, triangle).
- [x] Highlighted run lines should always appear on top of everything else.
- [x] When hovering over a run line, the plot highlights some random values across different metrics and positions. This should not happen. Only the related values to the on-hover run should show. 
- [x] Very long metric axis labels aren't displaying nicely. We need to break them to two lines or cap them at a character limit.
- [x] Limit the number of runs when viewing metrics plots so we don't impact performance too much.

**Nice to complete:**

- [ ] Overlapping metric value labels. No known solution for this at this time. Perhaps better highlighting on hover.
- [x] When hovering over a metric name, the current build only highlights the axis line and some values. This is right but this is also only one of the changes that should occur in this interaction. What we're missing here is to knock back (decrease in contrast) all the other metric columns that are not being highlighted, so that the highlighted one can pop to the front. Also in this state, we knock back many of the other lines and labels. Basically eliminating all the unnecessary noise at that point. See the design for reference.
    - Note: we may forgo this for the initial release.

tynandebold:
## Description

Resolves #1146 

## Development notes

See the checklist of what was worked on in the linked issue.

## QA notes

View the branch locally or checkout the Gitpod link to compare the work to the design.

## Checklist


Huongg (approved):
all working fine to me 👍

PR #1199, opened 2022-12-12
Handle 'null' values on metric plots

rashidakanchwala:
## Description

Currently, if we have null values in the data -- they are not handled correctly on the front-end. 
This is mainly around :- 
1. value.toFixed() on both plots where it throws the error that toFixed() cannot be applied to null 
2. also for time-series -- nulls weren't handled whilst drawing linePath. 

in this ticket, both of the above are fixed. 

## Checklist


Huongg (approved):
good spot thanks Rashida 😄

Issue #1205, opened 2022-12-21
Make Kedro-Viz work with `kedro-datasets`


merelcht:
## Description
From Kedro `0.18.4` we’re using datasets from `kedro-datasets` and so the import paths are different. `kedro_datasets.plotly.PlotlyDataSet` instead of `kedro.extras.datasets.plotly.plotly_dataset.PlotlyDataSet` etc. If a user is using Kedro `0.18.4` with `kedro-datasets` viz doesn't detect plot datasets correctly, experiment tracking doesn't work and other things where paths have been hard-coded in viz. 

## Context
We need to make Kedro viz support `kedro-datasets` and keep it compatible with using datasets from Kedro itself at least for the 5.x series. Perhaps in 6.0 we’d want to remove compatibility with datasets inside kedro

tynandebold:
This ticket will focus on fixing this specific issue in a way that isn't too generic.

PR #1214, opened 2023-01-12
Make Viz compatible with `kedro-datasets`

merelcht:
## Description

Fixes https://github.com/kedro-org/kedro-viz/issues/1205

## Development notes

- I've made the matching of import paths more generic so that both datasets from `kedro.extras` and `kedro-datasets` get detected correctly
- Viz will try to import `kedro-datasets` but if not installed, it will revert to using `kedro.extras.datasets`

## QA notes

I've manually tested this by spinning up the development server as described here: https://github.com/kedro-org/kedro-viz/blob/main/CONTRIBUTING.md#launch-a-development-server-with-a-real-kedro-project

Tried that everything works with `kedro==0.18.3` and also with `kedro==0.18.4` + `kedro-datasets==1.0.1`

## Checklist


rashidakanchwala (approved):
LGTM thanks @merelcht

tynandebold:
Don't forget to add a line to the release notes please :)

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -51,13 +51,12 @@ class TrackingDatasetGroup(str, Enum):
     JSON = "json"
 
 
-# pylint: disable=line-too-long
 TRACKING_DATASET_GROUPS = {
```
AntonyMilneQB, line 54:
```suggestion
# The keys will match a dataset type with any prefix, e.g. 
# plotly.plotly_dataset.PlotlyDataSet would include the following:
# kedro.extras.datasets.plotly.plotly_dataset.PlotlyDataSet
# kedro_datasets.plotly.plotly_dataset.PlotlyDataSet
# my.custom.path.to.plotly.plotly_dataset.PlotlyDataSet
TRACKING_DATASET_GROUPS = {
```

Just for my future reference so I don't forget what we've done here really...

(please check the comment is correct based on my understanding of the code!)

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -111,3 +114,9 @@ def load_tracking_data(self, run_id: str):
 
 def get_dataset_type(dataset: AbstractVersionedDataSet) -> str:
```
AntonyMilneQB, line 115:
```suggestion
def get_full_dataset_type(dataset: AbstractVersionedDataSet) -> str:
    """e.g. kedro.extras.datasets.plotly.plotly_dataset.PlotlyDataSet or kedro_datasets.plotly.plotly_dataset.PlotlyDataSet"""
```
Maybe? Sort of depends how `dataset_type` is used elsewhere as to whether it's worth doing.

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -67,14 +66,15 @@ class TrackingDatasetModel:
...
     # dataset is the actual dataset instance, whereas dataset_type is a string.
-    # e.g. "kedro.extras.datasets.tracking.metrics_dataset.MetricsDataSet"
+    # e.g. "kedro_datasets.tracking.metrics_dataset.MetricsDataSet"
     dataset: AbstractVersionedDataSet
     dataset_type: str = field(init=False)
     # runs is a mapping from run_id to loaded data.
     runs: Dict[str, Any] = field(init=False, default_factory=dict)
 
     def __post_init__(self):
         self.dataset_type = get_dataset_type(self.dataset)
```
AntonyMilneQB, line 76: Is this variable actually still used anywhere now in experiment tracking?

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -111,3 +114,9 @@ def load_tracking_data(self, run_id: str):
 
 def get_dataset_type(dataset: AbstractVersionedDataSet) -> str:
     return f"{dataset.__class__.__module__}.{dataset.__class__.__qualname__}"
+
+
+def get_dataset_module_class(dataset: AbstractVersionedDataSet) -> str:
+    class_name = f"{dataset.__class__.__qualname__}"
+    _, dataset_type, dataset_file = f"{dataset.__class__.__module__}".rsplit(".", 2)
+    return f"{dataset_type}.{dataset_file}.{class_name}"
```
AntonyMilneQB, line 122: The naming is kind of confusing given we already call the "full path" `dataset_type` (which is probably already confusing, but it's also in the flowchart part of kedro-viz source, so probably let's leave it that way for now).
```suggestion
def get_abbreviated_dataset_type(dataset: AbstractVersionedDataSet) -> str:
    """e.g. plotly.plotly_dataset.PlotlyDataSet"""
    abbreviated_module_name = ".".join(dataset.__class__.__module__.split(".")[-2:])
    return f"{abbreviated_module_name}.{dataset.__class__.__qualname__}"
```

package/kedro_viz/models/flowchart.py
```diff
@@ -466,37 +470,39 @@ def __post_init__(self):
             self._get_namespace(self.full_name)
         )
 
+    @staticmethod
```
AntonyMilneQB, line 473: I'd refactor this as per suggestions in experiment_tracking.py.

AntonyMilneQB (approved):
Looks good, thanks for the fix! Just a few minor suggestions to make things a bit clearer.

Definitely would be nice to get @jmholzer to review also before merging in case he has any other comments since he's also been thinking about this.

jmholzer:
> Definitely would be nice to get @jmholzer to review also before merging in case he has any other comments since he's also been thinking about this.

I've read through the code and I definitely think the approach is the right one. I'll add a review today 🙂.

package/kedro_viz/models/experiment_tracking.py
merelcht: Yes, I tried removing this which caused errors. I think it's used for the schema in graphql.

src/utils/short-type.js
```diff
@@ -1,6 +1,8 @@
 import { shortTypeMapping } from '../config';
 
-const getShortType = (longTypeName, fallback) =>
-  shortTypeMapping[longTypeName] || fallback;
+const getShortType = (name, fallback) => {
+  const longTypeName = name?.split('.').slice(-3).join('.');
```
AntonyMilneQB, line 4: Isn't this basically just doing what your new `get_abbreviated_dataset_type` does? Why don't we just pass the abbreviated dataset type to the frontend instead now we have it in the backend.

package/kedro_viz/models/experiment_tracking.py
AntonyMilneQB: I suspect what's happening here is we're passing the full `self.dataset_type` to the frontend, which is then just being turned back into the abbreviated form - see https://github.com/kedro-org/kedro-viz/pull/1214/files#r1071164866.

The full dataset type is used in the frontend in metadata panel in the flowchart view but not in experiment tracking (I think). So there's a good simplification we could make here if all we really need in experiment tracking is the full dataset type.

Feel free to leave as it is for now, but should be an easy tidy up for someone to do in a follow up PR if I'm right.

src/utils/short-type.js
rashidakanchwala: makes sense. because nowhere in the front end we show the full name.

rashidakanchwala: makes sense, i have removed the origin get_dataset_type class and renamed get_dataset_module as get_dataset_type so now everywhere we pass around the shorten dataset name that excludes 'kedro.extras.datasets'

package/kedro_viz/models/experiment_tracking.py
rashidakanchwala: we no longer have this. as get_dataset_module_class is the new 'get_dataset_type' e.g. pandas.csv_dataset.CSVDataset

package/tests/test_api/test_graphql/test_queries.py
```diff
@@ -7,6 +7,21 @@
...
+    from kedro_datasets import tracking, plotly, matplotlib  # isort:skip
+except ImportError:
+    from kedro.extras.datasets import (  # Safe since ImportErrors are suppressed within kedro.
+        matplotlib,
+        plotly,
+        tracking,
+    )
+
+
+def _get_dataset_type(dataset):
```
jmholzer, line 20: It would be good to have a comment here about what this function does, and how and why it is different from the function with the same name in `package/kedro_viz/models/flowchart.py`, as this may get confusing.

package/tests/test_api/test_rest/test_reponses.py
```diff
@@ -8,6 +8,19 @@
...
 
+try:
+    from kedro_datasets import pandas  # isort:skip
+except ImportError:
+    from kedro.extras.datasets import (  # Safe since ImportErrors are suppressed within kedro.
+        pandas,
+    )
+
+
+def _get_dataset_type(dataset):
```
jmholzer, line 19: This function has been declared twice with the same body (here and in `test_queries`), I am fine with the DAMP approach, but ditto my comment.

jmholzer (approved):

package/tests/test_api/test_rest/test_reponses.py
rashidakanchwala: so we don't have a utils file where we can have some common test functions -- one function is for flowchart and the other is for experiment tracking hence there is duplicity.

package/tests/test_api/test_graphql/test_queries.py
rashidakanchwala: u r right. we don't need to call this here. we can jsut import it

rashidakanchwala: u right these two functions are not required in the test files (test_queries, and test _responses). in order to assert the correct result I am just going to use string "pandas.csv_dataset.CsvDataSet" just the way it was done before

Issue #1228, opened 2023-01-20
Kedro Viz command raised MemoryDataSet error (5.2.0)


YoroiiHatemaki:
Hey, we've been successfully using kedro-viz 5.1.1 in our kedro project and today I've tried to switch to kedro-viz 5.2.0. Suddenly, the "kedro viz" command raises an error and fails to complete. 

This is the error: 
`kedro.io.core.DataSetError: Data for MemoryDataSet has not been saved yet.
kedro.framework.cli.utils.KedroCliError: Data for MemoryDataSet has not been saved yet.
Run with --verbose to see the full exception
Error: Data for MemoryDataSet has not been saved yet.
`
Switching back to kedro-viz 5.1.1 allows to successfully run "kedro viz" immediately. Any ideas how we can fix that or what may be causing it? Thanks in advance ^^

tynandebold:
Thanks for reporting the issue @YoroiiHatemaki. I think this is the same problem as #1226. What do you say @jmholzer?

For now, please continue to use version `5.1.1` of Kedro-Viz until we release a fix for this problem.

jmholzer:
> Thanks for reporting the issue @YoroiiHatemaki. I think this is the same problem as #1226. What do you say @jmholzer?
> 
> For now, please continue to use version `5.1.1` of Kedro-Viz until we release a fix for this problem.

It's definitely related, probably closely enough that it's the same problem, though I still haven't been able to reproduce #1226, I'll keep looking into both of these.

jmholzer:
Hey @YoroiiHatemaki, thanks for the report. Unfortunately I can't reproduce the error locally with a simple pipeline, could you please show us your catalog?

YoroiiHatemaki:
Hey @jmholzer, sadly I'm not allowed to show the whole catalog. However, I've found that we have 3 memory datasets in the catalog, and one of them is used both as input and as output in one of the nodes. My guess is that kedro-viz 5.1.1 allowed you to generate a visualization without checking whether all memory datasets have been saved, while kedro-viz 5.2.0 specifically checks for that. Do you think that this might be the issue?

AntonyMilneQB:
I think this is probably unrelated to #1226 and instead to do with https://github.com/kedro-org/kedro-viz/pull/1113 instead.

KikiCS:
[...]

PR #1231, opened 2023-01-24
Fix unhandled exception when viz is run with undefined parameters

jmholzer:
## Description
Resolves [#1228](https://github.com/kedro-org/kedro-viz/issues/1228)

## Checklist


RELEASE.md
```diff
@@ -11,6 +11,7 @@ Please follow the established format:
 ## Bug fixes and other changes
 
 - Visual fixes for metrics plots in experiment tracking. (#1229)
+- Fix bug causing user to be unable to start Kedro-Viz if undefined parameters are used. (#1231)
```
AntonyMilneQB, line 14:
```suggestion
- Fix bug that prevented Kedro-Viz from loading when a project contains undefined parameters. (#1231)
```

package/tests/test_models/test_flowchart.py
```diff
@@ -228,6 +229,20 @@ def test_create_non_existing_parameter_node(self, patched_warning):
             [call("Cannot find parameter `%s` in the catalog.", "non_existing")]
         )
 
+    @patch("logging.Logger.warning")
```
AntonyMilneQB, line 231: Should we also test the `AttributeError` case (presumably caused by `self.kedro_obj = None`)? Just a thought - up to you.

AntonyMilneQB (approved):
Great job tracking this down and fixing it! LGTM but it would be great if you could add a bit of documentation either to this PR or as comments just so we have your current understanding of how things work recorded somewhere since it wasn't at all clear to me looking through the code. e.g. it seems that `CatalogRepository.get_dataset` also handles parameters (based purely on the fact that this fixes the bug), can `self.kedro_obj` still be `None`, etc.

On second thoughts, maybe a slightly clearer fix would be to handle the case of parameters and undefined datasets differently in `CatalogRepository.get_dataset` rather than lumping them together as `MemoryDataSet`?
```
        except DataSetNotFoundError:
            if is_parameter(dataset_name): # function is somewhere in kedro-viz, but maybe doesn't make sense to use it here depending on where it's defined...
                dataset_obj = None   
            else:
                 dataset_obj = MemoryDataSet()
```

But then on third thoughts, that would mean that `self.kedro_obj.load` never executes in `parameter_values` which doesn't sound right 🤔

Basically I think this is all good, but would be great to just explain things a bit more somewhere now that you've untangled it since there's clearly scope for confusion here!

RELEASE.md
tynandebold: `contains` to `contained` to be past-tense consistent?

tynandebold (approved):
Brilliant, thank you!

rashidakanchwala (approved):
LGTM :)

package/tests/test_models/test_flowchart.py
jmholzer: Good thought, this case is covered by the test directly above 🙂. I added comments to both tests clarifying their purpose in [786d23e](https://github.com/kedro-org/kedro-viz/pull/1231/commits/786d23e0f1cb3ad2d442a56153e619a7097d9e3e).

jmholzer:
Thanks for the quick review and for all your help and guidance with this issue.

> Basically I think this is all good, but would be great to just explain things a bit more somewhere now that you've untangled it since there's clearly scope for confusion here!

I added an explanatory comment to the except clause in `ParametersNode.parameters_value` 🙂.

RELEASE.md
jmholzer: Changed in [cf96f46](https://github.com/kedro-org/kedro-viz/pull/1231/commits/cf96f463092a0279b45e1e56d247576937416d1f).

PR #1232, opened 2023-01-24
Add a forgiving `get_dataset_type` implementation to the `models` package

### LINKED ISSUE #1226: KedroCliError raised on launching Viz with certain datasets.
## Description
Running Viz can cause an error on launch when certain datasets are included in the pipeline:

```
kedro.framework.cli.utils.KedroCliError: not enough values to unpack (expected 3, got 1)
Run with --verbose to see the full exception
Error: not enough values to unpack (expected 3, got 1)
```

An initial investigation shows that it's either (or both) _get_dataset_type (flowchart.py) or get_dataset_type (experiment_tracking.py ) that could be causing the issue.

## Context

This stops users launching Viz and is high priority.

jmholzer:
## Description
Resolves #1226.

## Checklist

### DIFF
--- a/package/kedro_viz/models/experiment_tracking.py
+++ b/package/kedro_viz/models/experiment_tracking.py
@@ -12,6 +12,8 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.sql.schema import ForeignKey
 from sqlalchemy.types import JSON, Boolean, Integer, String
 
+from .utils import get_dataset_type
+
 logger = logging.getLogger(__name__)
 Base = declarative_base()
 
@@ -111,9 +113,3 @@ class TrackingDatasetModel:
             )
             self.runs[run_id] = {}
         self.dataset._version = Version(None, None)
-
-
-def get_dataset_type(dataset: AbstractVersionedDataSet) -> str:
-    class_name = f"{dataset.__class__.__qualname__}"
-    _, dataset_type, dataset_file = f"{dataset.__class__.__module__}".rsplit(".", 2)
-    return f"{dataset_type}.{dataset_file}.{class_name}"
--- a/package/kedro_viz/models/experiment_tracking.py
+++ b/package/kedro_viz/models/experiment_tracking.py
@@ -53,7 +53,7 @@ class TrackingDatasetGroup(str, Enum):
     JSON = "json"
 
 
-# Map dataset types to their group
+# Map dataset types (as produced by get_dataset_type) to their group
 TRACKING_DATASET_GROUPS = {
     "plotly.plotly_dataset.PlotlyDataSet": TrackingDatasetGroup.PLOT,
     "plotly.json_dataset.JSONDataSet": TrackingDatasetGroup.PLOT,
--- a/package/kedro_viz/models/experiment_tracking.py
+++ b/package/kedro_viz/models/experiment_tracking.py
@@ -53,11 +53,7 @@ class TrackingDatasetGroup(str, Enum):
     JSON = "json"
 
 
-# The keys will match a dataset type with any prefix, e.g.
-# plotly.plotly_dataset.PlotlyDataSet would include the following:
-# plotly.plotly_dataset.PlotlyDataSet
-# plotly.plotly_dataset.PlotlyDataSet
-# my.custom.path.to.plotly.plotly_dataset.PlotlyDataSet
+# Map dataset types to their group
 TRACKING_DATASET_GROUPS = {
     "plotly.plotly_dataset.PlotlyDataSet": TrackingDatasetGroup.PLOT,
     "plotly.json_dataset.JSONDataSet": TrackingDatasetGroup.PLOT,
--- a/package/kedro_viz/models/utils.py
+++ b/package/kedro_viz/models/utils.py
@@ -1,10 +1,5 @@
 """`kedro_viz.models.utils` contains utility functions used in the `kedro_viz.models` package"""
-
-import logging
-
 from kedro.io import AbstractDataSet
-
-logger = logging.getLogger(__name__)
 
 
 def get_dataset_type(dataset: AbstractDataSet) -> str:
--- a/package/kedro_viz/models/utils.py
+++ b/package/kedro_viz/models/utils.py
@@ -10,6 +10,19 @@ logger = logging.getLogger(__name__)
 def get_dataset_type(dataset: AbstractDataSet) -> str:
     """Get the type of a dataset as a string-- the abbreviated name of the module to
     which ``dataset`` belongs, joined with the name of its class.
+    ::
+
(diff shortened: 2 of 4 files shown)


package/kedro_viz/models/utils.py
```diff
@@ -0,0 +1,25 @@
+"""`kedro_viz.models.utils` contains utility functions used in the `kedro_viz.models` package"""
```
AntonyMilneQB, line 1: Very good innovation to make this file. There's quite a few candidates for other things that should be moved into this file at some point.

package/kedro_viz/models/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    """Get the type of a dataset as a string-- the abbreviated name of the module to
+    which ``dataset`` belongs, joined with the name of its class.
+
+    Args:
+        dataset: The dataset object to get the type of
+
+    Returns:
+        The abbreviated type of the dataset as a string.
+    """
+    if not dataset:
```
AntonyMilneQB, line 20:
```suggestion
    if dataset is None:
```

Not sure if this suggestion would be an improvement, but just something to thinking about... It's less safe/general but I'm not sure this will ever be called with something falsey other than `None`?

package/kedro_viz/models/utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+from kedro.io import AbstractDataSet
+
+logger = logging.getLogger(__name__)
+
+
+def get_dataset_type(dataset: AbstractDataSet) -> str:
+    """Get the type of a dataset as a string-- the abbreviated name of the module to
+    which ``dataset`` belongs, joined with the name of its class.
+
```
AntonyMilneQB, line 13: Some clarifying examples here would be nice.
```suggestion
# When dataset is an instance of any of the following, the returned type is 
# be "plotly.plotly_dataset.PlotlyDataSet".
# kedro.extras.datasets.plotly.plotly_dataset.PlotlyDataSet
# kedro_datasets.plotly.plotly_dataset.PlotlyDataSet
# my.custom.path.to.plotly.plotly_dataset.PlotlyDataSet
# Any module paths that contain only a package and not a subpackage
# will is as is, e.g. a dataset of type package.PlotlyDataSet gives "package.PlotlyDataSet"

```

There's a comment above `TRACKING_DATASET_GROUPS` in experiment_tracking.py which is now out of date and confusing. Please could you change/remove that also?

jmholzer: I agree, some examples would be a really nice addition, I added these in [5c5765d](https://github.com/kedro-org/kedro-viz/pull/1232/commits/5c5765d1b2475327dbad07ab2947ee8abb6226ef). I tried to express the behaviour of the function through examples, since I found it hard to get the wording of descriptions both accurate and clear.

I replaced the out-of-date comment in `experiment_tracking.py` ([6cff8e9](https://github.com/kedro-org/kedro-viz/pull/1232/commits/6cff8e929a8a26914fe080b93645ac1403cbe19c)), let me know if you think the replacement is helpful.

package/kedro_viz/models/utils.py
jmholzer: I think this is a good point, it makes things a little less safe, but I agree with you that this is a good change.

If we ever pass anything falsey to `get_dataset_type` that isn't `None`, it would mean something upstream is broken, which we would want to know about, even if it means having an unhandled exception in a future release.

Changed in [7edd9c7](https://github.com/kedro-org/kedro-viz/pull/1232/commits/7edd9c71ab2733d37ed9639e274e2613ed1457f4).

package/kedro_viz/models/utils.py
jmholzer: I totally agree, I don't think it's worth a ticket, but it's something to keep in mind as we make more changes to `flowchart.py` and `experiment_tracking.py`.

jmholzer: I totally agree, I don't think it's worth a ticket, but it's something to keep in mind as we make more changes to `flowchart.py` and `experiment_tracking.py`.

AntonyMilneQB: I'll add it to the mega-ticket https://github.com/kedro-org/kedro-viz/issues/980 where I keep track of this sort of thing and will at some point break it down into more manageable tasks or chip away at it bit by bit. Looks like I'd already put de-duplicating the `get_dataset_type` code there so we're definitely on the same wavelength!
> * Reuse `DataNode` and `DataNoteMetadata` models. There's too much duplication between these and tracking datasets (also code to get dataset type is in both)

package/kedro_viz/models/experiment_tracking.py
```diff
@@ -51,11 +53,7 @@ class TrackingDatasetGroup(str, Enum):
     JSON = "json"
 
 
-# The keys will match a dataset type with any prefix, e.g.
-# plotly.plotly_dataset.PlotlyDataSet would include the following:
-# plotly.plotly_dataset.PlotlyDataSet
-# plotly.plotly_dataset.PlotlyDataSet
-# my.custom.path.to.plotly.plotly_dataset.PlotlyDataSet
+# Map dataset types to their group
```
AntonyMilneQB, line 56:
```suggestion
# Map dataset types (as produced by get_dataset_type) to their group
```

package/kedro_viz/models/utils.py
AntonyMilneQB: This looks great now, nice idea to do it like a doctest rather than my wordy and not very clear text 👍

package/kedro_viz/models/utils.py
```diff
@@ -0,0 +1,38 @@
...
+import logging
+
+from kedro.io import AbstractDataSet
+
+logger = logging.getLogger(__name__)
+
+
+def get_dataset_type(dataset: AbstractDataSet) -> str:
+    """Get the type of a dataset as a string-- the abbreviated name of the module to
```
AntonyMilneQB, line 11:
```suggestion
    """Get the type of a dataset as a string: the abbreviated name of the module to
```
(very minor I know... it just looked weird to me)

jmholzer: You're right, fixed in [89f04d7](https://github.com/kedro-org/kedro-viz/pull/1232/commits/89f04d7c912cff11e00dbdf095a5d54ae8d67e05).

package/kedro_viz/models/experiment_tracking.py
jmholzer: Added in [4774238](https://github.com/kedro-org/kedro-viz/pull/1232/commits/4774238a726340a700f532bec40747187d20aebe).

AntonyMilneQB (approved):
🎉 Once the user has confirmed this fixes it then I'm very happy to release 👍

tynandebold (approved):
Fantastic!

Gentle reminder to add a line to the release notes.

PR #1250, opened 2023-02-01
Add safeguard to check that npm package has been built

AntonyMilneQB:
It's unlikely to ever happen again, but just to be sure... 5.1.1 was a patch release necessitated by an innocuous-looking but actually disastrous change I made to the `Makefile`.

I've put a check in so that it's no longer possible to run `make package` without the Javascript having successfully built first. I've put the test in the `Makefile ` rather than CircleCI config so that, in addition to preventing any future releases failing in this way, it'll protect us against accidentally distributing a .whl file without the required html (I've made this mistake before also, e.g. when trying things out on databricks).

Tested locally and all seems to work well.
## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
-.PHONY: package
+.PHONY: package build
 
 package:
 	find . -regex ".*/__pycache__" -exec rm -rf {} +
 	find . -regex ".*\.egg-info" -exec rm -rf {} +
+	test -f package/kedro_viz/html/index.html || (echo "Built npm package not found; packaging process cancelled."; exit 1)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,8 +1,9 @@
-.PHONY: package
+.PHONY: package build
```
AntonyMilneQB, line 1: This should have always been there. It means you won't have to do `make build -B` to avoid the "`build` is up to date" error message; just `make build` will always work now.

tynandebold (approved):
Very nice, thank you. Tested it locally and it worked perfectly.

Issue #1244, opened 2023-01-31
Address deprecation warnings coming from SQLAlchemy 1.4


astrojuanlu:
## Description
https://github.com/kedro-org/kedro-viz/pull/1222, shipped in kedro-viz 5.2.0 and 5.2.1, correctly capped SQLAlchemy, but didn't address the deprecation warnings mentioned in https://github.com/kedro-org/kedro/issues/2202. In other words, deprecated features are still in use, it's just that pinning `sqlalchemy <2.0` averted breakage.

These are the offending lines:

https://github.com/kedro-org/kedro-viz/blob/76ea3b08eaf7f30c2f0635d540cb158c44531acb/package/kedro_viz/models/experiment_tracking.py#L11-L18

As far as I know, SQLAlchemy 1.4 is a transition release and it should be possible to make the code forward compatible, see https://docs.sqlalchemy.org/en/20/changelog/migration_20.html and https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declarative_base for a migration path that should be compatible with both. Since `sqlalchemy` is already pinned to `>=1.4`, this should have no further consequences.

## Context
I was following the spaceflights tutorial yesterday and spotted this warning:

```
[01/31/23 09:26:59] WARNING  /Users/juan_cano/.micromamba/envs/kedro310/lib/python3.10/site-packages/kedro_viz/models/experiment_tracki warnings.py:109
                             ng.py:18: MovedIn20Warning: [31mDeprecated API features detected! These feature(s) are not compatible with                
                             SQLAlchemy 2.0. [32mTo prevent incompatible upgrades prior to updating applications, ensure requirements                  
                             files are pinned to "sqlalchemy<2.0". [36mSet environment variable SQLALCHEMY_WARN_20=1 to show all                       
                             deprecation warnings.  Set environment variable SQLALCHEMY_SILENCE_UBER_WARNING=1 to silence this                         
                             message.[0m (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)                                                 
                               Base = declarative_base()  
```

## Steps to Reproduce

1. Install the spaceflights dependencies
2. Run `kedro viz`
3. Spot warning in terminal output

## Expected Result
No deprecation warning.

## Actual Result
Deprecation warning is present.

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

```
% pip list | grep kedro
kedro                0.18.4
[...]

PR #1222, opened 2023-01-17
Pin sqlalchemy dependency

merelcht:
Signed-off-by: Merel Theisen <merel.theisen@quantumblack.com>

## Description

Since a couple of days users who have `kedro-viz` installed see a message like this:

```
/Users/merel_theisen/Projects/kedro-viz/package/kedro_viz/models/experiment_tracking.py:16: MovedIn20Warning: Deprecated API features detected! These feature(s) are not compatible with SQLAlchemy 2.0. To prevent incompatible upgrades prior to updating applications, ensure requirements files are pinned to "sqlalchemy<2.0". Set environment variable SQLALCHEMY_WARN_20=1 to show all deprecation warnings.  Set environment variable SQLALCHEMY_SILENCE_UBER_WARNING=1 to silence this message. (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
```

whenever they're running a Kedro command. This is annoying, distracting and can even seem intimidating to new users who don't know if this is an error or can be ignored. 

## Development notes

I've added an upper bound to the `sqlalchemy` dependency like suggested in the warning message. SQLAlchemy 2.0 hasn't been released yet, but when it is we should remove this upper bound again.

## QA notes

NA

## Checklist

- [ ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #1251, opened 2023-02-01
Update SQLAlchemy import path

AntonyMilneQB:
Closes https://github.com/kedro-org/kedro-viz/issues/1244. As @astrojuanlu correctly observed, we haven't actually gotten rid of the deprecation warning, we just pinned the version. This alters the import path to actually fix it.

In fact, the version was already pinned before: https://github.com/kedro-org/kedro-viz/pull/1222 didn't really do anything because the version specifier `~=1.4` is essentially the same as `>=1.4, <2.0`. So I'm just reverting that to the `~=` we had before, because it's more common in our codebase. Looking at the SQLAlchemy migration guide we could unpin this completely and just do `>=1.4` but I've left the pin there for now just to be on the safe side (in general, I'd like to move towards more open-ended upper bounds, but that's for another time).


astrojuanlu (approved):
Thanks @AntonyMilneQB! Changes LGTM

jmholzer (approved):
Looks good, did you do a manual check that the deprecation warning is gone after these changes?

AntonyMilneQB:
@jmholzer yep, I did and it's gone.

rashidakanchwala (approved):

astrojuanlu:
so, as @AntonyMilneQB spotted, there seems to be a problem with the types:

```
package/kedro_viz/models/experiment_tracking.py:11: error: Module "sqlalchemy.orm" has no attribute "declarative_base"  [attr-defined]
```

(and then a bunch of errors that cascade from there). This is essentially an instance of https://github.com/dropbox/sqlalchemy-stubs/issues/250.

On the other hand, sqlalchemy 2.0 started shipping its own types https://github.com/sqlalchemy/sqlalchemy/issues/6810 and I _think_ `sqlalchemy.orm` is already covered, but we can't switch to it just yet.

The options are:

- don't do anything until we upgrade to sqlalchemy 2.0. users will keep seeing the deprecation warning. the other problem is that this would require "jumping" version without a smoother transition path.
- `# type: ignore` this error until we upgrade to sqlalchemy >= 2.0. introduces a small amount of risk because we would be "blind" to type errors around this area for a while. hopefully there are unit tests for it though?
- not sure there's a third option.

AntonyMilneQB:
I find the warning sufficiently annoying that I think we should prioritise getting rid of it, since it appears not just when you run `kedro viz` but also when you run `kedro jupyter` commands if you have kedro-viz installed (due to the `run_viz` line magic being registered). Hence I'd be in favour of option 2. Risk of introducing bugs as a result of this seems very low, but we should keep an eye on sqlalchemy-stubs so that if there's a fix we can remove the ignore.

Issue #1252, opened 2023-02-03
Improve the `Pretty name` feature


tynandebold:
## Description

The `Pretty name` feature falls short in once specific area:

- When the feature is turned on, the sidebar search will only allow the user to search for pretty-named things. For example, if one searches for `feature_engineering.feat_derived_features` nothing will be returned, even though `Feat Derived Features` exists and is the same thing.
    - The same is true in the opposite direction.

## Possible Implementation

1. We should allow users to search for **both** pretty-named values and normal-named values.

## Checklist

- [x] Include labels so that we can categorise your feature request

PR #1254, opened 2023-02-06
Feature/improve metadata panel node naming

tynandebold:
## Description

Resolves the metadata-panel part of #1252. 

@idanov had the great idea of always showing all node names (pretty name/original name) in the metadata panel. This does that.

## QA notes

Look at the Gitpod link and see the update in the panel. 

We may need to decide on what the text should be for the labels. I've gone with `Pretty node name` and `Original node name`. 

image

@AntonyMilneQB you care a lot about language. Do you have any thoughts here?

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -11,6 +11,7 @@ Please follow the established format:
 ## Major features and improvements
 
 - Enable the display of json objects with `react-json-viewer` in experiment tracking. (#1236)
+- Always show pretty and original node names in the metadata panel. (#1254)
 
 ## Bug fixes and other changes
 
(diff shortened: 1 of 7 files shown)


AntonyMilneQB:
I think `Original node name` is a little bit odd - maybe just `Node name` would do? But if you want to emphasise how it's not the pretty version then it's ok, I don't have a better suggestion than "Original".

tynandebold:
The point was to have something descriptive in each scenario, this being the other one (when `Pretty name` is on):

image

rashidakanchwala (approved):
LGTM! thanks @tynandebold . I can't think of a better name ... this makes sense.

Huongg (approved):
it looks good to me too, thanks @tynandebold

PR #1264, opened 2023-02-20
Shorten Transcoded Dataset type on Metadata Panel

### LINKED ISSUE #1262: Trascoded Original and Transcoded Types can be further shortened.
## Description

The dataset type for a dataset on the metadata panel is shortened so now it only shows 'CSVDataset' or 'ParquetDataset' or 'PlotlyDataset' etc. However for Transcoded datasets we still show the long name so this ticket is to change that. 

## Actual Behavior

Screenshot 2023-02-17 at 09 30 59

## Expected Behavior

in the above screenshot, the types should be changed and represented as below

Original Type - ParquetDataSet
Transcoded Type - ParquetDataSet

rashidakanchwala:
## Description

Resolves #1262 

## QA notes

Ingestion.Int_Typed_Shuttles is a transcoded dataset so you can check before and after for it. 

## Checklist


Huongg (approved):
LGTM thanks @rashidakanchwala

src/components/metadata/metadata.js
```diff
@@ -89,8 +89,12 @@ const MetaData = ({
...
   };
 
-  const shortenDatasetType = (string) => {
-    return string?.split('.').pop();
+  const shortenDatasetType = (value) => {
+    const isList = Array.isArray(value);
+
+    return isList
+      ? value.map((val) => val.split('.').pop())
+      : value?.split('.').pop();
```
tynandebold, line 97: Maybe just this? If you want to set a variable, I'd use `isArray` for the name. `list` is from Python.

```suggestion
    return Array.isArray(value)
      ? value.map((val) => val.split('.').pop())
      : value?.split('.').pop();
```

tynandebold (approved):
Looks good. I left one comment.

Why do we have both of these types though? Do they ever display different values?

image

Issue #1218, opened 2022-06-17
Allow users to collaborate while using experiment tracking


NeroOkwa:
## Description

This is the **third** highest priority issue resulting from the [experiment tracking adoption user research](https://github.com/kedro-org/kedro/issues/1390). Users want to be able to write their experiments to storage that is not on their local computer and share their experiments with other team members. 

This is important as it enables a team of users to collaborate and see each other's results as they iterate on a pipeline, compared to the experience of being limited to one user's local machine. Hence, this is a deciding factor for the adoption of Kedro Experiment Tracking. 

This pain point also came up in the  [experiment tracking user testing sessions](https://github.com/kedro-org/kedro/issues/1627):

> "It was easier for me just to put MLflow in the cloud and report there. And I thought that the native Kedro one was still very coupled. I didn't know how to share the database across many data scientists."
 
### What is the problem?

Users can only perform a model run on a local machine, making it difficult to collaborate on a project with the rest of their team because experiment results are on multiple computers.

Additionally, users have raised other concerns: 

> "I can't store my experiment tracking data on S3, even though we are running everything else on S3."

### Who are the users of this functionality?

Users are primarily data scientists, and data engineers are secondary users. 

### Why do our users currently have this problem?

We designed it this way to launch a simpler experiment tracking in Kedro, even though the predecessor of Experiment Tracking in Kedro (PerformanceAI) had this functionality. 

Currently, users can only store their runs on a local machine via SQLite:

> "For now, we'll keep it simple, and team members will track only their own experiments."

## What is the impact of solving this problem?

It would be possible to view all user experiments across a team in one place and also solve an outstanding adoption issue for Kedro Experiment Tracking:  

> "If we could write our metrics files to like an S3 bucket and then run experiment tracking pointing at that S3 bucket, **that simplifies our workflow in many different ways. It would make Kedro experiment tracking just as easy**, if not easier than MLflow for us."

> "You might train one model locally on your computer... Having all those experiments in one place as a single source of truth is powerful." 

### Option 1 
 
_Open the browser and see it_ - This involves having a server you can connect your Kedro-Viz and other users' Kedro-Viz to. Then you can share from that server.

> _Implication:_ That server introduces many complications and must be running all the time. We also need to decide who hosts the server, us or the user. This is related to the [shareable URL work, which is another Kedro-Viz project](https://github.com/kedro-org/kedro-viz/issues/1116). 

### Option 2

[...]

PR #1286, opened 2023-03-14
Collaborative Experiment Tracking V 1.0 

rashidakanchwala:
## Description

This PR addresses Issue 3 (#1218) from the user research for Experiment tracking. The update enables users to store their session_store and tracking data in the cloud.

Here's an overview of the Collaborative Experiment Tracking implementation:

- In the settings.py, user specifies the S3 bucket location, which triggers the upload of their session_store.db (a SQLite database) to the cloud using fsspec. This process is executed via the SQLiteStore._upload() function during a Kedro run, when a user creates a new experiment. 

- When users execute Kedro-viz, session_store.db files from all other users are downloaded to the user's local machine through the SQLiteStore._download() function. 

- These downloaded databases are then merged with the user's current local session_store.db through SQLITEStore._merge() function . As a result, the local session_store.db contains not only the user's experiments, but also those conducted by other team members.

- Every user collaborating on a particular experiment tracking project consistently maintains a copy of everyone's experiments, both locally and in the cloud. This synchronization is achieved through the SQLiteStore._sync function (which basically downloads, merges and uploads the session_store.db) 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1286"></a>


demo-project/src/demo_project/settings.py
```diff
@@ -9,7 +9,8 @@
 from kedro_viz.integrations.kedro.sqlite_store import SQLiteStore
 
 SESSION_STORE_CLASS = SQLiteStore
-SESSION_STORE_ARGS = {"path": str(Path(__file__).parents[2] / "data")}
+SESSION_STORE_ARGS = {"local_path": str(Path(__file__).parents[2] / "data"),
```
AntonyMilneQB, line 12: Does `local_path` get used anywhere? I don't see it 🤔 

We need to think about how to make this backwards compatible with argument names - will explain on Monday.

package/kedro_viz/data_access/repositories/runs.py
```diff
@@ -117,9 +117,16 @@ def create_or_update_user_run_details(
...
                     run_id=run_id, title=title, bookmark=bookmark, notes=notes
                 )
-                session.add(user_run_details)
+                try:
+                    session.add(user_run_details)
+                except Exception as e:
+                    session.rollback()
+                finally:
+                    session.close()
+
```
AntonyMilneQB, line 126: According to my (limited) understanding of sqlalchemy I thought this shouldn't be needed because it's in the `with self._db_session_class.begin()` context manager - why is it needed?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -24,6 +30,21 @@ def get_db(session_class: sessionmaker) -> Generator:
...
+
+    except Exception as exc:  # pylint: disable=broad-except
+        unique_id = uuid.uuid4()
+        logger.warning(
+            """Something went wrong with getting the username. Generated unique id %s for 
+                 for user's database name  Exception: %s""",
+            unique_id,
+            exc,
+        )
+        return f"{unique_id}.db"
```
AntonyMilneQB, line 45: Not sure whether making a unique id here is probably not a good idea because it gets re-generated every time you run kedro-viz so you might end up with lots of files on s3 🤔 It might be better to just suggest here that people set the environment variable `KEDRO_SQLITE_STORE_USERNAME` in this case. Not sure though, let's think about it.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -24,6 +30,21 @@ def get_db(session_class: sessionmaker) -> Generator:
         database.close()
 
 
+def _get_dbname():
+    try:
+        return f"{getpass.getuser()}.db"
```
AntonyMilneQB, line 35: Let's first of all check env var `KEDRO_SQLITE_STORE_USERNAME` and use that in preference to `getpass.getuser()`.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -32,6 +53,31 @@ def _is_json_serializable(obj: Any):
         return False
 
 
+def _add_to_local_storage(database_name, database_timestamp):
+    try:
+        with open("db.info", "r") as file:
```
AntonyMilneQB, line 58: What does the contents of this file look like?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -59,11 +110,99 @@ def to_json(self) -> str:
                 session_dict[key] = str(value)
         return json.dumps(session_dict)
 
+    def on_commit_sync(self, _):
```
AntonyMilneQB, line 113: What is the `_` for here?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -59,11 +110,99 @@ def to_json(self) -> str:
...
         session_store_data = RunModel(id=self._session_id, blob=self.to_json())
-        database.add(session_store_data)
-        database.commit()
+        try:
+            database.add(session_store_data)
+            database.commit()
+        except Exception as e:
+            pass
+        finally:
+            database.close()
```
AntonyMilneQB, line 130: Can we do this using a context manager instead?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -59,11 +110,99 @@ def to_json(self) -> str:
...
+
+    def download(self) -> List[str]:
+        """Download all the dbs from an s3 locations"""
+        protocol, _ = get_protocol_and_path(self._remote_path)
+        fs = fsspec.filesystem(protocol)
+        databases = fs.glob(f"{self._remote_path}/*.db")
+        databases_location = []
+        for database in databases:
+            database_name = os.path.basename(database)
+            database_timestamp = fs.info(database)["LastModified"]
```
AntonyMilneQB, line 149: Does this `LastModified` thing work outside s3?

Even if it does I think it _might_ be a bit fragile overall and we should consider open the database up and actually look at the timestamp of the most recent row instead of relying on when the file was written. Let me think about it...

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -59,11 +110,99 @@ def to_json(self) -> str:
...
+            isUpdated = _is_db_updated(database_name, database_timestamp)
+            if isUpdated:
```
AntonyMilneQB, line 151:
```suggestion
            if _is_db_updated(database_name, database_timestamp):
```

Too much javascript 😛

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -59,11 +110,99 @@ def to_json(self) -> str:
...
+                            row_dict = row._asdict()
+                            session_store_data = RunModel(**row_dict)
+                            try:
+                                database.add(session_store_data)
+                                database.commit()
+                            except Exception as e:
+                                database.rollback()
+                                pass
+                            finally:
+                                database.close()
```
AntonyMilneQB, line 188: Can we use the context manager?

AntonyMilneQB (commented):
Just a few initial comments for now - I need to look through it all again and think about it some more on Monday. This looks like an awesome start though, great work!

It would be very useful if you could write briefly at a high level what the overall scheme is here, like how do you use db.info and work out which rows to merge? It's not exactly what I was imagining we'd do (which isn't necessarily a problem, just I don't understand it fully).

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -64,6 +87,69 @@ def save(self):
...
+                s3f.write(file.read())
+
+    def _download(self) -> List[str]:
+        """Download all the dbs from an s3 bucket"""
+        protocol, _ = get_protocol_and_path(self._remote_path)
+        fs = fsspec.filesystem(protocol)
+        databases = fs.glob(f"{self._remote_path}/*.db")
+        databases_location = []
+        for database in databases:
+            database_name = os.path.basename(database)
```
merelcht, line 112: I'd recommend using [Pathlib](https://docs.python.org/3/library/pathlib.html) for file path handling. We use this in core Kedro as well and it's more robust across operating systems.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -64,6 +87,69 @@ def save(self):
...
+            with fs.open(database, "rb") as file:
+                db_data = file.read()
+            db_loc = f"{self._path}/{database_name}"
+            with open(db_loc, "wb") as temp_db:
+                temp_db.write(db_data)
```
merelcht, line 117: In the `_upload` method you use the nested format for reading and writing. Why not do the same here?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -64,6 +87,69 @@ def save(self):
...
+                            row_dict = row._asdict()
+                            session_store_data = RunModel(**row_dict)
+                            try: 
+                                database.add(session_store_data)
+                                database.commit()
+                            except Exception:
+                                database.rollback()
+                    
+            temp_engine.dispose()
+            os.remove(db_loc)
```
merelcht, line 148: What does this do?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -64,6 +87,69 @@ def save(self):
...
+
+    def sync(self):
```
merelcht, line 150: It would be helpful to add a comment here explaining what this method does exactly.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -64,6 +87,69 @@ def save(self):
...
+            database_name = os.path.basename(database)
+            with fs.open(database, "rb") as file:
+                db_data = file.read()
+            db_loc = f"{self._path}/{database_name}"
+            with open(db_loc, "wb") as temp_db:
+                temp_db.write(db_data)
+            databases_location.append(db_loc)
+        return databases_location
+
+    def _merge(self, databases_location: List[str]):
```
merelcht, line 121: I find this method quite hard to read. Adding comments would probably make it easier to follow.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -64,6 +87,69 @@ def save(self):
...
+                    if table_name == "runs":
+                        data = database_conn.execute(table_obj.select()).fetchall()
+                        for row in data:
+                            row_dict = row._asdict()
+                            session_store_data = RunModel(**row_dict)
+                            try: 
+                                database.add(session_store_data)
+                                database.commit()
+                            except Exception:
+                                database.rollback()
```
merelcht, line 145: Do we need to add an exception message here, so if something goes wrong it's easier to debug?

merelcht (commented):
I've left some initial comments and questions, but I think this already looks very good!

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: makes sense.

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: Basically we first download all the db files from s3 as only then you can connect to them. After you have merged the runs of each db to the user's local session_store.db you delete this files.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -24,23 +30,40 @@ def get_db(session_class: sessionmaker) -> Generator:
...
         """Returns location of the sqlite_store database"""
         return Path(self._path) / "session_store.db"
 
-    def to_json(self) -> str:
+    @property
+    def remote_location(self) -> Path:
+        """Returns location of the sqlite_store database"""
+        return self._remote_path
+
+    def _to_json(self) -> str:
```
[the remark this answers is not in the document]
rashidakanchwala: yup

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +84,112 @@ def to_json(self) -> str:
...
+        databases_location = []
+        # Download each database to a local filepath
+        for database in databases:
+            database_name = Path(database).name
+            with fs.open(database, "rb") as file:
+                db_data = file.read()
+                db_loc = f"{self._path}/{database_name}"
+                with open(db_loc, "wb") as temp_db:
+                    temp_db.write(db_data)
+            databases_location.append(db_loc)
```
antonymilne, line 132: Similarly to above comment, can we just use `fsspec.get` or similar to do this? https://filesystem-spec.readthedocs.io/en/latest/copying.html

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +84,112 @@ def to_json(self) -> str:
...
+        fs = fsspec.filesystem(protocol)
+        # Find all the databases at the remote path
+        databases = fs.glob(f"{self._remote_path}/*.db")
+        databases_location = []
+        # Download each database to a local filepath
+        for database in databases:
+            database_name = Path(database).name
+            with fs.open(database, "rb") as file:
+                db_data = file.read()
+                db_loc = f"{self._path}/{database_name}"
```
antonymilne, line 129: This `db_loc` won't work well as a string because there's a `/` in it. You should  use `Pathlib` instead for manipulating filepaths 🙂

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +84,112 @@ def to_json(self) -> str:
...
+        # Connect to the user's local session_store.db
+        engine, session_class = create_db_engine(self.location)
+        Base.metadata.create_all(bind=engine)
+        database = next(get_db(session_class))
+
+        temp_engine = None
+        # Iterate through each downloaded database
+        for db_loc in databases_location:
+            if temp_engine:
+                temp_engine.dispose()
```
antonymilne, line 159: What is this for?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +84,112 @@ def to_json(self) -> str:
...
+                            session_store_data = RunModel(**row_dict)
+                            try:
+                                database.add(session_store_data)
+                                database.commit()
+                            except Exception as e:
+                                database.rollback()
+                                logging.exception(f"Failed to add runs to the {database}: {e}")
+            # Close the connection to the downloaded database and delete it
+            temp_engine.dispose()
+            os.remove(db_loc)
```
antonymilne, line 180: Do we actually need to remove the databases? Might just be easier to leave them there, since it's one less thing that could go wrong and should make things easier to debug, and it won't use up much disk space.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +84,112 @@ def to_json(self) -> str:
...
+
+    def _upload(self):
+        """Uploads the session store database file to the specified remote path on the cloud storage."""
+        # Connect to the remote file system
+        protocol, _ = get_protocol_and_path(self._remote_path)
+        db_name = _get_dbname()
+        fs = fsspec.filesystem(protocol)
+        # Open the local database file and upload it to the remote path
+        with open(self.location, "rb") as file:
+            with fs.open(f"{self._remote_path}/{db_name}", "wb") as s3f:
```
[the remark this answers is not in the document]
antonymilne: P.S. you can also do two context managers in one like `with ... as file, with ... as remote_file` but I don't think we should need to do any manual file writing here and instead just rely on some built-in fsspec method.

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: If there's a connection already open, it closes it. Just to make sure we don't have too many open connections

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +84,112 @@ def to_json(self) -> str:
...
+        """Save the session store info on db and uploads it to the cloud if a remote cloud path is provided ."""
         engine, session_class = create_db_engine(self.location)
         Base.metadata.create_all(bind=engine)
         database = next(get_db(session_class))
-        session_store_data = RunModel(id=self._session_id, blob=self.to_json())
+
+        session_store_data = RunModel(id=self._session_id, blob=self._to_json())
         database.add(session_store_data)
         database.commit()
+        self.sync()
```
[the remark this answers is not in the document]
rashidakanchwala: > This is coming on well, great work! But I think there's two big things to consider still:
> 
> 1. what happens if my internet is broken? Everything should still work ok and not crash, but e.g. the `upload` method I think will break.
> 2. where is the logic to avoid duplicating rows in the database? i.e. the thing that says "only insert rows into my database if I don't already have them". Otherwise we'll just be repeatedly adding more and more rows to the database that we already have?

Not sure about point 1. it's a bit confusing. let me have a think and revert.

for your point 2. there's a unique constraint on run_ids in runs table so it won't add a run with the same timestamp. It throws an error which is why I have to do session.rollback()

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +84,112 @@ def to_json(self) -> str:
...
+                        data = database_conn.execute(table_obj.select()).fetchall()
+                        for row in data:
+                            row_dict = row._asdict()
+                            session_store_data = RunModel(**row_dict)
+                            try:
+                                database.add(session_store_data)
+                                database.commit()
+                            except Exception as e:
+                                database.rollback()
+                                logging.exception(f"Failed to add runs to the {database}: {e}")
```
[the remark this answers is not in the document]
rashidakanchwala: I tried, but it wasn't working properly. I guess because the exception gets thrown here. I will retry

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: This was just to keep the user's project directory clean. I mean these were only temp databases.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -24,6 +30,11 @@ def get_db(session_class: sessionmaker) -> Generator:
         database.close()
 
 
+def _get_dbname():
```
rashidakanchwala, line 33: I have to do it this way otherwise unit tests fail

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -24,6 +30,11 @@ def get_db(session_class: sessionmaker) -> Generator:
...
+    username = os.environ.get("KEDRO_SQLITE_STORE_USERNAME") or getpass.getuser()
+    return username + ".db"
```
rashidakanchwala, line 35: I have to do it this way rather than a single return statement because if do that the related unit test fails (gives me None.db)

antonymilne: Let's talk about this!

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +85,109 @@ def to_json(self) -> str:
...
+            database_location:  A list of local filepath in string format for all the databases
+
+        """
+        # Connect to the user's local session_store.db
+        engine, session_class = create_db_engine(self.location)
+        Base.metadata.create_all(bind=engine)
+        database = next(get_db(session_class))
+
+        all_runs_data = []
+        existing_run_ids = [id[0] for id in database.query(RunModel.id).all()]
```
antonymilne, line 141: I think there must be some way to do this without the list comprehension.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +85,109 @@ def to_json(self) -> str:
...
+            with temp_engine.connect() as database_conn:
+                db_metadata = MetaData()
+                db_metadata.reflect(bind=temp_engine)
+                # Merge data from the 'runs' table
+                for table_name, table_obj in db_metadata.tables.items():
+                    if table_name == "runs":
+                        data = database_conn.execute(table_obj.select()).fetchall()
+                        for row in data:
+                            if row.id not in existing_run_ids:
+                                existing_run_ids.append(row.id)
```
antonymilne, line 163: I'm sure there's some way of doing this in one query also, e.g. `query(RunModel).filter(RunModel.id.not_in(existing_run_ids)).all()`. See https://docs.sqlalchemy.org/en/14/orm/internals.html#sqlalchemy.orm.PropComparator.not_in.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -60,10 +85,109 @@ def to_json(self) -> str:
...
+                                all_runs_data.append((row._asdict()))
+            temp_engine.dispose()
+
+        database.bulk_insert_mappings(RunModel, all_runs_data)
```
antonymilne, line 167: Apparently this is a legacy method so probably we shouldn't use it. https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.bulk_insert_mappings

antonymilne (commented):
I think we're nearly there! There's one thing that I don't really like still: the style of sqlalchemy used in `_merge` is very different e.g. from what we're doing in `RunsRepository`, which I think is the style we should go for. I'm pretty sure it would be possible to write things that way - it just means using the sqlalchemy a bit differently.

I recommend taking a look through the SQLAlchemy docs and understanding the difference between ORM and core and the different APIs a bit better. I should do this too because I'm pretty new to it also and forgot most of what I learnt when I read it before. It's probably worth talking to Ravi about this also since I'm pretty sure he knows sqlalchemy better than me...

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: done

package/kedro_viz/server.py
```diff
@@ -32,7 +31,7 @@ def populate_data(
     data_access_manager: DataAccessManager,
     catalog: DataCatalog,
     pipelines: Dict[str, Pipeline],
-    session_store: BaseSessionStore,
+    session_store: Optional[BaseSessionStore],
```
antonymilne, line 34: Sounds like a minor comment, but it's actually important... why is this `Optional`?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -120,16 +124,21 @@ def _download(self):
...
-        - In the version 1.0 - we only merge the runs table which contains all the experiments.
-        - The downloaded database files are deleted after it's runs are merged with the user's local session_store.db
+        - In the version 1.0 - we only merge the runs table which
+        contains all the experiments.
+        - The downloaded database files are deleted after it's runs are merged
+        with the user's local session_store.db
 
         Args:
-            database_location:  A list of local filepath in string format for all the databases
+            database_location:  A list of local filepath in string
```
antonymilne, line 140: This argument no longer exists.

package/kedro_viz/server.py
rashidakanchwala: My bad I was getting a weird test_cli.error earlier so I thought this might be the issue but I think the issue finally was a session-scoped pytest fixture.

package/test_requirements.txt
```diff
@@ -4,11 +4,13 @@ kedro-telemetry>=0.1.1  # for testing telemetry integration
 bandit~=1.7
 behave~=1.2
 black~=23.3
+boto3~=1.26
 flake8~=5.0
 fastapi[all]>=0.73.0, <0.96.0
 isort~=5.11
 matplotlib~=3.5
-mypy~=0.990
+mypy==0.990
```
rashidakanchwala, line 12: I have pinned this for now due to some pydantic issues that we are resolving in another ticket.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +79,95 @@ def to_json(self) -> str:
...
+        }
+
+        logger.debug(f"Checking {len(downloaded_db_locations)} downloaded session stores for new runs...")
+        for downloaded_db_location in downloaded_db_locations:
+            engine = create_engine(f"sqlite:///{downloaded_db_location}")
+            with Session(engine) as session:
+                new_runs = (
+                    session.query(RunModel)
+                    .filter(RunModel.id.not_in(existing_run_ids))
+                    .all()
```
antonymilne, line 145: I don't like using `session.query` here (compared to `session.execute` above), but I know it matches what we have in runs.py so it's fine for now. We should update it as part of https://github.com/kedro-org/kedro-viz/issues/1360.

demo-project/.gitignore
```diff
@@ -23,6 +23,9 @@ logs/**
 # also keep all .gitkeep files
 !.gitkeep
 
+# logging.yml 
+conf/base/logging.yml
+
```
antonymilne, line 28: I don't think this should be git ignored. It's actually probably a good idea to commit the debug level change.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -105,10 +103,14 @@ def _download(self):
             # In theory we should be able to do this as a single operation:
             # self._remote_fs.get(f"{self.remote_location}/*.db", str(Path(self.location).parent))
             # but this does not seem to work correctly - maybe a bug in fsspec. So instead
-            # we do it in two steps.
+            # we do it in two steps. Also need to add os.sep so it works with older s3fs version.
+            # This is a known bug in s3fs
```
antonymilne, line 107: Would be nice to link to the issue here :)

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -105,10 +103,14 @@ def _download(self):
...
             remote_dbs = self._remote_fs.glob(f"{self.remote_location}/*.db")
-            logger.debug(f"Downloading {len(remote_dbs)} remote session stores to local...")
-            self._remote_fs.get(remote_dbs, str(Path(self.location).parent))
+            logger.debug(
+                f"Downloading {len(remote_dbs)} remote session stores to local..."
+            )
+            for remote_db in remote_dbs:
+                self._remote_fs.get(remote_db, str(Path(self.location).parent) + os.sep)
```
antonymilne, line 113: Just to double check: this works on the latest s3fs?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,102 @@ def to_json(self) -> str:
...
+            # we do it in two steps. Also need to add os.sep so it works with older s3fs version.
+            # This is a known bug in s3fs
+            remote_dbs = self._remote_fs.glob(f"{self.remote_location}/*.db")
+            logger.debug(
+                f"Downloading {len(remote_dbs)} remote session stores to local..."
+            )
+            for remote_db in remote_dbs:
+                self._remote_fs.get(remote_db, str(Path(self.location).parent) + os.sep)
+        except Exception as exc:
+            logger.exception(exc)
```
merelcht, line 115: Similar comment as for the upload method: I think it would be useful to mention that this exception happens in `download` so it's easier to trace.

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,102 @@ def to_json(self) -> str:
...
+        user's session_store.db and to all the other downloaded dbs.
+        - It is assumed that all the databases share the same schema.
+        - In the version 1.0 - we only merge the runs table which
+        contains all the experiments.
+        """
+
+        all_new_runs = []
+
+        with self._db_session_class() as session:
+            existing_run_ids = session.execute(select(RunModel.id)).scalars().all()
```
merelcht, line 132: what does `.scalars()` do?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,102 @@ def to_json(self) -> str:
...
+        Synchronizes the user's local session_store.db with
+        remote session_store.db stored on a cloud storage service.
+        """
+
+        if self.remote_location:
+            self._download()
+            # We don't want a failed merge to stop the whole kedro-viz process.
+            try:
+                self._merge()
+            except Exception as exc:
```
merelcht, line 176: Again I'd be more specific about where this exception is thrown, so in this case on `sync`

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -29,13 +29,166 @@ def db_session_class(store_path):
...
+def mock_db4(tmp_path):
+    database_loc = str(tmp_path / "db4.db")
+    engine = create_engine(f"sqlite:///{database_loc}")
+    session_class = sessionmaker(bind=engine)
+    Base.metadata.create_all(bind=engine)
+    session = session_class()
+    session.add(RunModel(id="3", blob="blob3"))
+    session.commit()
+    session.close()
+    yield Path(database_loc)
```
merelcht, line 103: That's a lot of very similar looking fixtures. I think it would be helpful to add a comment to each to make it clear what purpose it solves. It took me a while to figure out that db4 has the same data as db3, so if you add something about that being a duplicate it would help understanding the use quicker.

merelcht: And do you actually need 4? Wouldn't 3 be enough? 1 and 2 as unique dbs and then 3 as a duplicate of 2?

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -67,9 +237,146 @@ def test_update_git_branch(self, store_path, mocker):
...
+        sqlite_store._remote_fs.glob.return_value = mocked_db_in_s3
+        sqlite_store._download()
+        sqlite_store._merge()
+        db_session = sqlite_store._db_session_class
+        downloaded_dbs = set(Path(sqlite_store.location).parent.glob("*.db")) - {
+            Path(sqlite_store.location)
+        }
+        assert len(downloaded_dbs) == 3
+        with db_session() as session:
+            assert session.query(RunModel).count() == 3
```
merelcht, line 318: How does this check that the `merge` was successful? In `test_download_from_s3_success` you also get three paths in `downloaded_dbs`. I think this test should almost be the same as `test_download_from_s3_success` but there should be a very clear difference when merge was called vs when it wasn't.

merelcht (commented):
This looks really good! I've left some comments, mainly about the exception messages and the tests. Other than that I think it's almost ready 👍

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: Basically scalars() just returns the data as a list instead of tuples so we don't have to do a list conversion.

package/tests/test_integrations/test_sqlite_store.py
rashidakanchwala: makes sense. I will change

package/tests/test_integrations/test_sqlite_store.py
merelcht: Ahh okay that makes sense. It's just that I couldn't figure out the proper difference from this test and `test_download_from_s3_success`. Would it perhaps make sense to show that in the other one each db has a row?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -26,16 +37,27 @@ def _is_json_serializable(obj: Any):
...
         self._db_session_class = make_db_session_factory(self.location)
+        self._remote_path = remote_path
+
+        if self.remote_location:
+            protocol, _ = get_protocol_and_path(self.remote_location)
+            self._remote_fs = fsspec.filesystem(protocol)
 
     @property
-    def location(self) -> Path:
+    def location(self) -> str:
```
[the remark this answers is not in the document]
rashidakanchwala: Good question - I will let @antonymilne comment on this change.

rashidakanchwala:
> This looks really good! I've left some comments, mainly about the exception messages and the tests. Other than that I think it's almost ready 👍

Thanks Merel for the thorough review <3 I have made the changes as suggested.

package/tests/conftest.py
```diff
@@ -23,14 +23,31 @@
...
+    # Create a MagicMock object for SQLiteStore
+    session_store_mock = mock.MagicMock(spec=SQLiteStore)
+
+    # Create a PropertyMock object for the location attribute
+    location_mock = mock.PropertyMock(return_value=Path(tmp_path / "session_store.db"))
+    type(session_store_mock).location = location_mock
+
+    session_store_mock.sync = mock.MagicMock()
+
+    yield session_store_mock
```
antonymilne, line 48: Why do you need to mock this all out? Why not use a real `BaseSessionStore("dummy_path", "dummy_session_id")`?

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -29,13 +29,145 @@ def db_session_class(store_path):
...
+    mocker.patch("getpass.getuser", return_value="computer_user_name")
+    dbname = _get_dbname()
+    assert dbname == "env_user_name.db"
+
+
+def test_get_dbname_without_env_var(mocker):
+    mocker.patch(
+        "os.environ",
+        cast(Dict[str, str], {**os.environ, "KEDRO_SQLITE_STORE_USERNAME": None}),
+    )
```
antonymilne, line 158:
```suggestion
```

I guess there's a good reason for this line to exist but I don't see why?

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -59,6 +191,23 @@ def test_save_multiple_runs(self, store_path, db_session_class):
         with sqlite_store2._db_session_class() as db_session:
             assert db_session.query(RunModel).count() == 2
 
+    def test_save_run_with_remote_path(self, mocker, store_path, remote_path):
```
antonymilne, line 194: These two are very good tests 👍

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -67,9 +216,138 @@ def test_update_git_branch(self, store_path, mocker):
...
+            store_path, next(session_id()), remote_path=remote_path
+        )
+        sqlite_store._remote_fs.glob.side_effect = ConnectionError("Connection error")
+        mock_log = mocker.patch.object(logging.Logger, "exception")
+        sqlite_store._download()
+        downloaded_dbs = set(Path(sqlite_store.location).parent.glob("*.db")) - {
+            Path(sqlite_store.location)
+        }
+        # Assert that the number of databases downloaded is 0
+        assert len(downloaded_dbs) == 0
```
antonymilne, line 270: Similar comment here to above test about a better `assert` that doesn't copy and paste `downloaded_dbs` from the implementation of SQLitestore.

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -67,9 +216,138 @@ def test_update_git_branch(self, store_path, mocker):
...
+            store_path, next(session_id()), remote_path=remote_path
+        )
+        sqlite_store._remote_fs.glob.return_value = mocked_db_in_s3_repeated_runs
+        sqlite_store._download()
+        sqlite_store._merge()
+        db_session = sqlite_store._db_session_class
+        downloaded_dbs = set(Path(sqlite_store.location).parent.glob("*.db")) - {
+            Path(sqlite_store.location)
+        }
+        assert len(downloaded_dbs) == 3
```
antonymilne, line 309:
```suggestion
```

This is covered in the other tests already, no need to check it again here.

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -67,9 +216,138 @@ def test_update_git_branch(self, store_path, mocker):
...
+        mock_upload.assert_called_once()
+
+    def test_sync_without_remote_path(self, mocker, store_path):
+        mocker.patch("fsspec.filesystem")
+        sqlite_store = SQLiteStore(store_path, next(session_id()))
+        mock_download = mocker.patch.object(sqlite_store, "_download")
+        mock_merge = mocker.patch.object(sqlite_store, "_merge")
+        mock_upload = mocker.patch.object(sqlite_store, "_upload")
+        sqlite_store.sync()
+        assert not mock_download.called
```
antonymilne, line 334:
```suggestion
        mock_download.assert_not_called()
```

And similarly for the others.

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -67,9 +216,138 @@ def test_update_git_branch(self, store_path, mocker):
...
+        sqlite_store._remote_fs.glob.return_value = mocked_db_in_s3_repeated_runs
+        sqlite_store._download()
+        sqlite_store._merge()
+        db_session = sqlite_store._db_session_class
+        downloaded_dbs = set(Path(sqlite_store.location).parent.glob("*.db")) - {
+            Path(sqlite_store.location)
+        }
+        assert len(downloaded_dbs) == 3
+        with db_session() as session:
+            assert session.query(RunModel).count() == 2
```
antonymilne, line 311: Same suggestion again here about asserting you have the right run ids.

antonymilne (approved):
Thank you for all the work on this! I know it's taken a lot of time and many iterations but it looks great now ⭐ Nice work on the tests especially, because those weren't easy I know.

I've left some minor comments - happy to discuss if any don't make sense. Otherwise all good to go from me with two additional bits:
* would be nice to have some sort of PR description that says what's here
* please can you double check kedro-viz still works as normal if you don't supply a `remote_path` and if you don't use `SQLiteStore` at all? I already checked both of these last week, but just to be absolutely sure

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,105 @@ def to_json(self) -> str:
...
+            # we do it in two steps. Also need to add os.sep so it works with older s3fs version.
+            # This is a known bug in s3fs
+            remote_dbs = self._remote_fs.glob(f"{self.remote_location}/*.db")
+            logger.debug(
+                f"Downloading {len(remote_dbs)} remote session stores to local..."
+            )
+            for remote_db in remote_dbs:
+                self._remote_fs.get(remote_db, str(Path(self.location).parent) + os.sep)
+        except Exception as exc:
+            logger.exception(f"Download failed: {exc}")
```
idanov, line 118:
```suggestion
            logger.exception("Download failed:", exc)
```

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -26,16 +37,27 @@ def _is_json_serializable(obj: Any):
 class SQLiteStore(BaseSessionStore):
     """Stores the session data on the sqlite db."""
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, remote_path: Optional[str] = None, **kwargs):
+        """Initializes the SQLiteStore object."""
         super().__init__(*args, **kwargs)
         self._db_session_class = make_db_session_factory(self.location)
```
idanov, line 43: Not your code, ,but this seems to be session object and not session class. Could you rename it to simply `_db_session`?

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,105 @@ def to_json(self) -> str:
         return json.dumps(session_dict)
 
     def save(self):
-        """Save the session store info on db ."""
-
+        """Save the session store info on db and uploads it
+        to the cloud if a remote cloud path is provided ."""
         with self._db_session_class.begin() as session:
```
idanov, line 82: Here the `as session` should be `as tx` (from transaction).

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,105 @@ def to_json(self) -> str:
...
+        remote session_store.db stored on a cloud storage service.
+        """
+
+        if self.remote_location:
+            self._download()
+            # We don't want a failed merge to stop the whole kedro-viz process.
+            try:
+                self._merge()
+            except Exception as exc:
+                logger.exception(f"Merge failed on sync: {exc}")
```
idanov, line 180:
```suggestion
                logger.exception("Merge failed on sync:", exc)
```

idanov (approved):
Pretty good PR overall 🧨

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,105 @@ def to_json(self) -> str:
...
+        remote path on the cloud storage."""
+        db_name = _get_dbname()
+        logger.debug(
+            f"""Uploading local session store to remote with name
+              {db_name} to {self.remote_location}..."""
+        )
+        try:
+            self._remote_fs.put(self.location, f"{self.remote_location}/{db_name}")
+        except Exception as exc:
+            logger.exception(f"Upload failed: {exc}")
```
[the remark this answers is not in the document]
antonymilne: You're missing the `%s` in your suggestion.
```suggestion
            logger.exception("Upload failed: %s", exc)
```

package/kedro_viz/integrations/kedro/sqlite_store.py
antonymilne: This is actually a class: https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.sessionmaker. The docs here aren't very clear because they call the returned `Session` an object, but if you read more it is definitely a class.
```
Session = sessionmaker(engine)

with Session() as session:
```

package/kedro_viz/integrations/kedro/sqlite_store.py
antonymilne: Since `db_session_class` is indeed a class, `db_session-class.b egin()` does produce a `session` and not a transaction. It's basically shorthand for `Session().begin()`. See https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.sessionmaker.begin

demo-project/conf/base/globals.yml
```diff
@@ -1 +1,2 @@
-base_location: data/
\ No newline at end of file
+base_location: data/
+s3_location: s3://544538232364-kedro-vizstest/Master/kedro-proj-2/
```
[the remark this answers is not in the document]
rashidakanchwala: reverting it to just base_location since we know it works on s3.

demo-project/conf/base/logging.yml
```diff
@@ -0,0 +1,36 @@
+version: 1
```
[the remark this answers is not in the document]
rashidakanchwala: we originally didn't have logging.yml on demo-project. But it's useful for logging output and debugging so @antonymilne suggested we can let it be part of demo project now.

package/kedro_viz/integrations/kedro/sqlite_store.py
rashidakanchwala: There are chances you have already setup an env variable on your machine and if so the related unit test will fail. So we have to explicitly mock the environment variable for the tests -- see below. 

image

Now for second test when we mock env variable to None, weirdly with the one-liner, it returns None.db instead of doing getpass.getUser() and returning the computer_name.db 

Hence I reverted to this just so it passes the unit tests.

package/tests/test_integrations/test_sqlite_store.py
rashidakanchwala: So if I don't mock the environment variable explicitly to None it takes the previous tests environment variable. Rearranging tests also did not help because it still took the environment variable incase your environment had one.

Mocking it without cast threw one of the two (i can't remember): mypy or pylint error so with this the linting passes. :\

package/tests/conftest.py
rashidakanchwala: works great!

package/tests/test_integrations/test_sqlite_store.py
antonymilne: I think mocking like this should fix it: https://stackoverflow.com/a/31583528. We should definitely have this set up so that the mocking of environment variables doesn't leak from one test into another.

If you still have a linting error after trying that then I'd just put in an `ignore` rather than `cast`.

astrojuanlu:
Note to self: I managed to make this work locally on MinIO by setting the appropriate endpoint and credentials:

```
FSSPEC_S3_ENDPOINT_URL=http://127.0.0.1:9000 FSSPEC_S3_KEY=minioadmin FSSPEC_S3_SECRET=minioadmin kedro run
```

(inspiration https://github.com/fsspec/s3fs/issues/482#issuecomment-1445270169, https://s3fs.readthedocs.io/en/latest/#s3-compatible-storage)

antonymilne:
@astrojuanlu nice! I meant to try it out with MinIO but didn't get around to it. You can actually test it out even more easily (but less thoroughly) just by specifying `"remote_path": "path/to/local/folder/`.

demo-project/conf/base/catalog_09_tracking.yml
```diff
@@ -1,19 +1,19 @@
 train_evaluation.linear_regression.r2_score:
   type: tracking.MetricsDataSet
-  filepath: ${base_location}/09_tracking/linear_score.json
+  filepath: ${s3_location}/tracking_data/linear_score.json
```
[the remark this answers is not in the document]
rashidakanchwala: done.

demo-project/conf/base/globals.yml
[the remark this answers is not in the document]
rashidakanchwala: done

package/kedro_viz/integrations/kedro/sqlite_store.py
```diff
@@ -55,7 +77,105 @@ def to_json(self) -> str:
...
+        if self.remote_location:
+            self._upload()
+
+    def _upload(self):
+        """Uploads the session store database file to the specified
+        remote path on the cloud storage."""
+        db_name = _get_dbname()
+        logger.debug(
+            f"""Uploading local session store to remote with name
+              {db_name} to {self.remote_location}..."""
```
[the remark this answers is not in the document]
rashidakanchwala: done

demo-project/conf/base/globals.yml
```diff
@@ -1 +1,2 @@
-base_location: data/
\ No newline at end of file
+base_location: data/
```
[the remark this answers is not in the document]
rashidakanchwala: done

tynandebold:
Seconding and raising again what @antonymilne wrote above:

> would be nice to have some sort of PR description that says what's here.

For future reference, let's make sure there's something written in the PR description.

package/tests/test_integrations/test_sqlite_store.py
rashidakanchwala: it works as you suggested. also makes it easier to revert to a one line return in the sqlite_store

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -67,9 +216,138 @@ def test_update_git_branch(self, store_path, mocker):
...
+        mocker.patch("fsspec.filesystem")
+        sqlite_store = SQLiteStore(
+            store_path, next(session_id()), remote_path=remote_path
+        )
+        sqlite_store._remote_fs.glob.return_value = mocked_db_in_s3
+        sqlite_store._download()
+        sqlite_store._merge()
+        db_session = sqlite_store._db_session_class
+        with db_session() as session:
+            assert session.query(RunModel).count() == 2
```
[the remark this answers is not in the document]
rashidakanchwala: done

package/tests/test_integrations/test_sqlite_store.py
rashidakanchwala: done

package/tests/test_integrations/test_sqlite_store.py
rashidakanchwala: done

package/tests/conftest.py
rashidakanchwala: BaseSessionStore("dummy_path", "dummy_session_id")
this works but i get code coverage error in server.py (lines 40-42 where it checks for if(sessionStore?) 

should I revert it to back to SQLiteStore mock or shld i just do #pragma:ignore. 
I wouldn't want to do ignore.

NeroOkwa:
@antonymilne I tried out the setup with MinIO and it worked perfectly.

PR #1288, opened 2023-03-15
Preview dataset

Huongg:
## Description
Fixes for #907. For now, every time the user clicks on CSV, Excel or Parquet Dataset in viz, it will load the first 40 rows in the metadata panel

Design:
[Link to the journey and behaviour description](https://www.figma.com/file/akJ7QasxvLcgqc8Hxud7QB/Kedro-Design%3A-OUT?node-id=2021%3A86784)
[Link to the metadata side panel](https://www.figma.com/file/akJ7QasxvLcgqc8Hxud7QB/Kedro-Design%3A-OUT?node-id=539%3A28859)

## Development notes

To test this locally, ensure you clone and pull the latest changes from this branch `https://github.com/kedro-org/kedro-plugins/tree/preview-csv-dataset`

Then cd to location of `kedro-plugins/kedro-datasets`, then run `pip install -e .` To check, run ` pip list` to see if `kedro-dataset` and `kedro-viz` are pointed to your local machine
Once done, go back to `kedro-viz/demo-project` run `kedro run` then `kedro viz`. It should show you all the changes from both repos. 

WIP: tests to be added to cover both changes, here and in kedro-plugins

https://user-images.githubusercontent.com/32060364/225598177-50d571b9-7493-4f1e-b412-3a9053081d07.mov

## Checklist

- [ ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #1282, opened 2023-03-06
Upgrade to React v17 and MUI v5

tynandebold:
## Description

Resolves #1265.

## Development notes

I've gone up one major version, to React 17. I also bumped Google's Material UI component library to v5 from v4, since we'll need to do that when we upgrade to React 18.

## QA notes

Please delete both the `node_modules` folder and the `package-lock.json` file. Then run `npm install` and make sure everything installs and runs without issue.

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

Issue #1165, opened 2022-11-14
Experiment Tracking - Link plot/metrics/json dataset names to the flowchart


NeroOkwa:
## Context
This feature request is an output from the experiment tracking User Testing Synthesis [#1627](https://github.com/kedro-org/kedro/issues/1627#top)

**Pain point:** User wishes to know where the dataset displayed in the Plots tab is coming from and how it is connected to what is seen in the flow chart view. Overall, it seems out of context and may be confusing when presenting to stakeholders.

**Proposed Solution**: User proposes a description for the data shown to be able to understand and easily present the data to stakeholders from Kedro-Viz. 

## Supporting quotes
“So basically I would like to have a description or something that I can click so that I can understand the data that is involved here. Because if I'm going to show this to a client for example I don't want to keep switching from documentation to here”.

NeroOkwa:
@rashidakanchwala I have updated ticket with more info on user's pain point.

tynandebold:
@NeroOkwa can you please go back to the user and see what specific information about that dataset they want to show and what they were switching back and fourth between? We still need som additional information here in order to decide where to go with this.

It sounds like this could be a case of where the plot data is derived from, and that's something we should show more clearly.

NeroOkwa:
@tynandebold yes, this is the case. The user wants to know where the plot data is derived from, and how it is connected to the flow chart. 

The current way plots are displayed without a reference or link to the flow chart seems out of context and confusing especially when the user is presenting to a client.

tynandebold:
It seems to me that the user is misunderstanding the view.

The plot data is connected to a `kedro run`. Unless the plot is from the user's most-recent run, there isn't any connection with the flowchart.

Further, the proposed solution doesn't seem like it would help. All we'd really be able to do would be to surface the location of the file that drives the plot (e.g. `code/data/08_reporting/price_histogram.json`). Does that help with presentation to stakeholders?

There is merit in connecting the flowchart with a plot in experiment tracking, as we've talked about before. That seems outside the scope of this ticket.

tynandebold:
After discussion with the team, let's aim to link the dataset name (e.g. `reporting.feature_importance`) to a node in the flowchart (`https://demo.kedro.org/?pipeline_id=__default__&selected_id=eb7d6d28`). 

That way, if someone asks, "Where did that plot/metric/data/etc come from?" all the user has to do is click the dataset name and they'll be taken to that part of the flowchart.

[...]

PR #1289, opened 2023-03-16
Build out new Viz version panel 

tynandebold:
## Description

Resolves #1281.

## Development notes

I've augmented what we already had and built it out to the new design specs. 

## QA notes

Look at the branch as-is and you'll see the new version tag in the top left. That's the up-to-date state. To change the version of Viz in order to see the out-of-date state, change the version number in the `package/kedro_viz/__init__.py` to something before `5.3.0`.

We should probably think of a better color combo for the up-to-date state for the light theme. This is why: 

image

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [x] Added tests to cover my changes

PR #1297, opened 2023-03-24
Release v6.0.0

tynandebold:
## Description

Release v6.0.0.

It's a **major-version release** because we've bumped the minimum version of React to a new major version, which is a breaking change for users using Viz as a React package.

Here's what's going in:

#### Major features and improvements

- Remove metrics plots from metadata panel and add link to the plots on experiment tracking. (#1268)
- Link plot and JSON dataset names from experiment tracking to the flowchart. (#1165)
- Bump minimum version of React from 16.8.6 to 17.0.2. (#1282)
- Show preview of data in metadata panel. (#1288)

#### Bug fixes and other changes

- Remove pandas and plotly dependencies from Kedro-viz. (#1268)

## Checklist


RELEASE.md
```diff
@@ -6,11 +6,11 @@ Please follow the established format:
...
-# Release 6.x.x
+# Release 6.0.0
 
 ## Major features and improvements
 
-- Remove metrics plots from metadata panel and add link to the plots on Experiment tracking. (#1268)
+- Remove metrics plots from metadata panel and add link to the plots on experiment tracking. (#1268)
 - Link plot and JSON dataset names from experiment tracking to the flowchart. (#1165)
 - Bump minimum version of React from 16.8.6 to 17.0.2. (#1282)
 - Show preview of data in metadata panel. (#1288)
```
AntonyMilneQB, line 16: I'd consider reordering these so they're more in order of what's interesting to users (e.g. removing plots isn't very interesting, and nor is the react version for most people; adding the preview in is more exciting).

AntonyMilneQB (approved):
Looks good. I think we should clearly explain what the bump to version 6 means though (in the announcement and maybe also in the release notes), since as I understand it for those who `pip install kedro-viz` (which I guess is 99% of kedro-viz users), this is not actually a breaking release. We don't want those people to be put off upgrading because they think it will break stuff.

I don't know what conventions and expectations are for react packages, but if it's not too offensive/risky to that 1% of users who use it as a react component then I'd even be tempted to just do this as a minor release.

NeroOkwa (changes requested):
Agree with @AntonyMilneQB . These can be reordered with Link plots and Show preview of data at the top.

RELEASE.md
tynandebold: The convention is to go from least to greatest PR number. I think we can leave it like this here and reorder by excitement level in the announcement note.

AntonyMilneQB:
You're right - the rules we use are semver, not React/Python. But we're in a tricky situation here that we have the same version number for the React package as the Python one, and what constitutes a breaking changes is different for the React users compared to the Python ones.

Given that Python users are (I presume) the vast majority of consumers, it might make sense to version according to whether it's a breaking release for them, which would make this a non-breaking release. The problem then is that there might be React consumers who would be tripped up since it's a breaking release for them. I don't know what convention and expectation is in the React universe so don't know how common that would be though (like if people are pinning to `5.x` then it would be a problem, but if they tend to pin to `5.x.y` then it would not).

Undoubtedly it's safest to do as you do and err on the side of "if it's breaking for React or Python users then better bump the major release". This will trip no one up, and so long as we make it clear in the release announcement that it's actually non-breaking for Python kedro-viz consumers then that's fine by me. I think the risk isn't actually so much that users will be put off upgrading but that they won't receive the update automatically. Most people in Python universe will have their version pinned to `5.x` ([as we do in the starters](https://github.com/kedro-org/kedro-starters/blob/main/spaceflights/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/requirements.txt#L11)) and so won't receive the new kedro-viz version until they manually choose to do so.

So it's a balance of "do we want the majority of users to automatically receive this upgrade automatically" vs. "do we want to not break the minority of users packages". Usually I'd go for the latter of those two option, but given that the majority vs. minority ratio here is so extreme, maybe the balance shifts to the former. I don't mind, just want to point out the pros/cons either way 🙂

AntonyMilneQB:
Also where is this "What's New" panel? 😮 I've never seen that before.

tynandebold:
> Also where is this "What's New" panel? 😮 I've never seen that before.

It's [this](https://github.com/kedro-org/kedro-viz/pull/1289)!

src/components/update-reminder/update-reminder-content.js
```diff
@@ -9,10 +9,10 @@ export const updateContent = {
...
+      copy: "You're now able to preview certain datasets in the Metadata panel and expand the preview for greater inspection. If you don't want this enabled, you can turn it off in Settings by selecting the cog icon in the lower left of the screen and toggling the option. In the future, enablement of this feature will be done at a dataset level in Kedro's catalog.yml file.",
```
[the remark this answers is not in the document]
stichbury: "In the future, enablement of this feature will be done at a dataset level in Kedro's catalog.yml file." isn't very clear what this means. Does it mean, at some point in the future, we will adjust where you set the feature on/off? What is this future point, is it a separate release or ?

I removed the sentence for now because it seems to create more confusion than help (in my mind) but if you want to add it back in, please do so --  just clarify what/when/how

stichbury:
Would it also be possible to change the copy in "New update panel" to replace "...will provide more context..." with "...provides more context..." -- if it's not too much hassle? I couldn't see that copy in the file to make the adjustment myself.

PR #1301, opened 2023-03-28
Fix SQLAlchemy Connection Error

rashidakanchwala:
## Description

In the runs repository, we open sqlalchemy sessions for a lot of read operations but we don't close them using session.close(). Hence we eventually get this error on our terminal 

Screenshot 2023-03-28 at 14 26 58

More details on this error is here - https://docs.sqlalchemy.org/en/20/errors.html#error-3o7r) 

In this PR, i fixed this by doing `with self._db_session_class() as session:` for each time we connect to a db for reading.

Note - In my collaborative experiment tracking, I have added additional try/catch blocks every time we do an add which I haven't added a part of this ticket because anytime an exception is thrown, the session doesn't close even with `with ....as begin`. @AntonyMilneQB -- should I add it here too? wdyt? 

## Checklist


noklam (approved):
Thanks, I think closing the session is the right thing to do. 

One improvement we can do (not a must) is something like this. Instead of creating all these context manager in the class method, we can create a decorator - which handle the session creation/close, we can then add this `@require_session` decorator on this method. The `session` will be available for the function scope and destroy automatically.
```python
from functools import wraps

def require_session(func):
    @wraps(func)
    def wrapper(self, *args, **kwargs):
        #  session = self._db_session_class() # acquire the session
        print("Create the session")
        session = None

        try:
            return func(self, session, *args, **kwargs) # pass the resource to the function
        finally:
            print("session.close()")
#                 session.close() # close the resource after the function call
    return wrapper

class FakeSession:
    def _db_session_class(self):
        print("getting session")
    
    @require_session
    def do_something(self, session):
        print("session.do_something")
            
FakeSession().do_something()
```

We can further split it to `read` `write` if it requires different way of handling the session.

AntonyMilneQB (commented):
This is definitely a good fix, thank you!

Please could we actually change all of the `self._db_session_class.begin()` to just `self._db_session_class()` throughout this file. [There's no need to use `begin` now that we're on sqlalchemy 1.4](https://docs.sqlalchemy.org/en/14/orm/session_basics.html#session-autobegin) and this way it will be consistent everywhere.

> Note - In my collaborative experiment tracking, I have added additional try/catch blocks every time we do an add which I haven't added a part of this ticket because anytime an exception is thrown, the session doesn't close even with with ....as begin. @AntonyMilneQB -- should I add it here too? wdyt?

Let's leave this out of her ejust to keep it as a clean PR and fix only one thing.

rashidakanchwala:
> This is definitely a good fix, thank you!
> 
> Please could we actually change all of the `self._db_session_class.begin()` to just `self._db_session_class()` throughout this file. [There's no need to use `begin` now that we're on sqlalchemy 1.4](https://docs.sqlalchemy.org/en/14/orm/session_basics.html#session-autobegin) and this way it will be consistent everywhere.
> 
> > Note - In my collaborative experiment tracking, I have added additional try/catch blocks every time we do an add which I haven't added a part of this ticket because anytime an exception is thrown, the session doesn't close even with with ....as begin. @AntonyMilneQB -- should I add it here too? wdyt?
> 
> Let's leave this out of here just to keep it as a clean PR and fix only one thing.

hi @AntonyMilneQB, without having begin() -- it doesn't commit implicitly (at least at my end) 

Try -- running viz, changing a name of a run, click refresh. It's not committing to the database

I am on sqlalchemy 2.0

AntonyMilneQB:
Thanks for trying it out @rashidakanchwala! I think it must be because of the settings we have for `sessionmaker` in database.py:
```
session_class = sessionmaker(autocommit=False, autoflush=False, bind=engine)
```
These options have been set since the beginning in #575 but I don't see any explanation of why anywhere. Do you remember?

For sqlalchemy 2 `autocommit` is deprecated and just set to `False` anyway, so that shouldn't make a difference. Please could you try removing `autoflush=False` and see if that works?

rashidakanchwala:
> Thanks for trying it out @rashidakanchwala! I think it must be because of the settings we have for `sessionmaker` in database.py:
> 
> ```
> session_class = sessionmaker(autocommit=False, autoflush=False, bind=engine)
> ```
> 
> These options have been set since the beginning in #575 but I don't see any explanation of why anywhere. Do you remember?
> 
> For sqlalchemy 2 `autocommit` is deprecated and just set to `False` anyway, so that shouldn't make a difference. Please could you try removing `autoflush=False` and see if that works?
> 
> Edit: actually I'm pretty sure that still won't do the SQL commit. I think I see why, but would be good if you could double check.

Yup, doesn't work and autocommit=True is not supported.

AntonyMilneQB:
OK, I think the code is correct as you have it (i.e. some have `.begin()` and others don't) but we should remove `autocommit=False, autoflush=False` from `sessionmaker` because we don't need them and they just make it confusing.

I'm like 99% sure that will all be good but please could you give that a go and double check it works for sqlalchemy 1.4  also? 🙏

rashidakanchwala:
> OK, I think the code is correct as you have it (i.e. some have `.begin()` and others don't) but we should remove `autocommit=False, autoflush=False` from `sessionmaker` because we don't need them and they just make it confusing.
> 
> I'm like 99% sure that will all be good but please could you give that a go and double check it works for sqlalchemy 1.4 also? 🙏

Yup, it works on 1.4 :) thankss!!!.. i have removed it. Could you please approve?

AntonyMilneQB (approved):
Thank you for this change! 🎉

PR #1303, opened 2023-03-29
Append params to Run Command if some were used for the run

MaximeSteinmetz:
## Description
It is not currently possible to easily reproduce a run that used CLI params (`--params=key:value`) because the params are not shown in Kedro Viz. 

This change appends the params in CLI form to Run Command if the run used some. By using the same git commit and CLI params, runs should be reproducible.

[Slack thread](https://kedro-org.slack.com/archives/C03RKP2LW64/p1679922405309659) for context

## Development notes
- Modified `serializers.py` so that it appends run params information (already present in session data) to run command
- Added a test case in `test_queries.py` as well as fixtures for it in `conftest.py`

image

## Checklist

### Additional remarks
- The change assumes all `command_path` accept `--params` option. If not, additional checks must be added to ensure the resulting `run_command` makes sense
- I didn't find pre-existing logic to convert _params dict_ --> _CLI params string_, so added [custom logic](https://stackoverflow.com/a/6027615) which looks non-ideal. We could add logic in `kedro.framework.cli.utils` and reuse it `serializers.py`?


AntonyMilneQB:
Hello @MaximeSteinmetz and thank you very much for your contribution! Unfortunately I think the implementation here isn't quite right though for a few reasons:
* as you found, it requires some quite fiddly logic to convert from the `params` dictionary back to the `--params` string
* with the introduction of omegaconf the syntax for `--params` is changing slightly, so this is a bit fragile and will not necessarily represent exactly what the user inputs
* for better reproducibility the run command should really contain all the flags (e.g. `--pipeline` as well) and not just `--params`

What would be the correct implementation here would be to modify what kedro saves to the session store in the first place. The original run command is available there and so can be saved to the store in full with no need to re-create it later in kedro-viz from the dictionary representation.

The change should be quite simple: we want to change the [`"command_path"` key](https://github.com/kedro-org/kedro/blob/7acff044fc9b446b6d69e18e0e7620960eebdad7/kedro/framework/session/session.py#L63) to give something like this:
```
"command_path" : " ".join(["kedro"] + sys.argv[1:])
```

Let me know if this makes sense and if you'd be able to open the PR on the kedro repo instead. I'd be happy to review it and help if required.

MaximeSteinmetz:
Hey @AntonyMilneQB thanks for the review! I completely agree with your assessment. The solution you propose sounds much better! I'll have a closer look at it and raise a PR to Kedro as you suggest if all goes well 🙂

PR #1313, opened 2023-04-05
Feature/select-metrics

### LINKED ISSUE #1294: Select Metrics for Visualization in Parallel Coordinates
## Description
Currently, in Kedro, all metrics are displayed in parallel coordinates, which can make the visualization cluttered and difficult to interpret. This feature request is aimed at allowing users to select which metrics to display in the visualization, making it easier to focus on the most relevant metrics and gain deeper insights into the data.

## Context
As a Kedro user, I work with complex datasets and often need to visualize high-dimensional data using parallel coordinates. However, when there are many metrics, the visualization becomes difficult to interpret. With this feature, I could select the most relevant metrics to display and create clearer, more informative visualizations. Additionally, this feature would benefit other Kedro users who work with complex data and need to create clear and informative visualizations.

Huongg:
## Description

Fixes #1294 

Enable users to hide/show different metrics on both parallel and coordinate. We store those selected metrics values in our localStorage so that it will remember even when users move to different pages. The same selected metrics values will be applied on all the runs 

[Design link](https://www.figma.com/file/akJ7QasxvLcgqc8Hxud7QB/Kedro-Design%3A-OUT?node-id=3106-100354&t=h4q8fpJ0u73YSdhz-0)
## Development notes

https://user-images.githubusercontent.com/32060364/230265332-620da781-2595-4d94-a3d0-43d421c6563c.mov

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1313"></a>


tynandebold (approved):
Looks and works great! Approving straight away with a couple of comments:

I think the light theme needs a slight style update. It looks a little off. The select button should probably have a background color so it easy to differentiate. Also, the hover is still the dark theme color:

image

And then when there are no metrics selected we should add one line of text telling the user what to do:

image

Use this copy: `No data to display. Select a metric to view a visualisation.`

You can see how we do it when there's no plot to view [here](http://localhost:4141/experiment-tracking?run_ids=2021-09-01T09.45.55.269Z&view=Plots&comparison=false).

stephkaiser:
looks good to me! agree with Tynan on having a message when there are no metrics selected as its just blank right now. 

QQ - do we have designs on what that dropdown looks like when we have lots of metrics + scrolling? or do users normally only have a short list of metrics?

stephkaiser:
blank message looks great! on scrolling - would it be possible to fix the buttons on the bottom so we are only scrolling the list of metrics? :)

stephkaiser (approved):

Mackay031:
Hi thanks for this. I agree with the above comments. I would just add a couple more:

- Agreeing with the buttons not scrolling, if we make a rule, I would say that metrics (no buttons) start to scroll within the component if there are more than 7. 
- The initial tooltip that appears on FIRST hover could be sped up a bit. The speed after the first one is good.
- The space under the "apply and close' button is a bit large, could we make it 24px.
- For the message when no metrics are selected, I think we should stick to the 'nothing to display' design we use when there are no experiments 

Metric_no_data

WDYT?

NeroOkwa:
Looks really good! I agree with the above comments.

PR #1315, opened 2023-04-11
Fix gitpod

AntonyMilneQB:
On gitpod the Python dependencies were taking aaaaages to resolve and we were on a very old version of `pip`. I've updated the Docker image used and added `pip install -U pip` to keep things up to date. It should all work more smoothly again 🤞

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -14,7 +14,7 @@ tasks:
   - name: backend
     before: echo PIP_USER=no >> ~/.bashrc && export PIP_USER=no
     init: |
-      pip install -U pip --user
+      pip install "pip==23.0.1" --user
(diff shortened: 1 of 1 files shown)


.gitpod.yml
```diff
@@ -10,29 +10,32 @@ tasks:
     command: |
       gp sync-done build_complete
       DANGEROUSLY_DISABLE_HOST_CHECK=true npm run start
+
   - name: backend
     before: echo PIP_USER=no >> ~/.bashrc && export PIP_USER=no
     init: |
+      pip install -U pip --user
```
astrojuanlu, line 17: Maybe install a specific version of pip to avoid accidental breakage? Not a big deal, but can prevent some headaches:

```suggestion
      pip install "pip==23.0.1" --user
```

rashidakanchwala (approved):

PR #1373, opened 2023-05-30
Make Kedro-viz compatible with new layer conf format under metadata. 

rashidakanchwala:
Prior to version 1.3.0 of kedro-datasets, layers were defined directly within the catalog as follows:

```yaml
type: pandas.CSVDataSet
filepath: /filepath/to/dataset
layer: raw
```

However, starting with kedro-datasets 1.3.0, layers are now defined within the 'metadata' section under the 'kedro-viz' plugin:

```yaml
type: pandas.CSVDataSet
filepath: /filepath/to/dataset
metadata:
  kedro-viz:
    layer: raw
```

Previously, the layers' information was computed in the DataCatalog and provided to kedro-viz. With the new format, the layer information under 'metadata' -> 'kedro-viz' needs to be computed by kedro-viz itself.

This ticket aims to update kedro-viz to handle this change in layer configuration, while still ensuring compatibility with both the new and the old ways of configuring layers within a kedro project.

Please note: From Kedro version 0.19.x and Kedro-viz version 7.x onwards, support for the old way of configuring layers will be discontinued.

ADDITIONALLY - I have removed kedro.extras from the tests and added kedro-dataset as a test_requirement. So now we always do unit testing against kedro-datasets. 

## QA notes

I have commented the old way of doing things in the catalog.yml files so you can test both using layers as a top-level attribute or inside metadata. I will remove the comments before merge. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1373"></a>


demo-project/conf/base/catalog_01_raw.yml
```diff
@@ -1,14 +1,38 @@
...
   filepath: ${base_location}/01_raw/shuttles.xlsx
-  layer: raw
+  metadata:
+    kedro-viz:
+      layer: raw
+
+
+
+
+# companies:
```
rashidakanchwala, line 25: I have commented the old way of doing things in the catalog.yml files so you can test both using layers as a top-level attribute or inside metadata. I will remove the comments before merge.

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -20,24 +21,67 @@ def get_catalog(self) -> DataCatalog:
...
+        for dataset_name in self._catalog._data_sets:
+            dataset = self._catalog._get_dataset(dataset_name)
+
+            if hasattr(dataset, "metadata"):
+                metadata = dataset.metadata or {}
+                dataset_layer = metadata.get("kedro-viz", {}).get("layer")
+                if dataset_layer:
+                    layers[dataset_layer].add(dataset_name)
+        return layers or None
+
```
rashidakanchwala, line 63: This is exactly what was happening in framework side - https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L272-L277

package/test_requirements.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements.txt
-kedro[pandas.ParquetDataSet, pandas.CSVDataSet, plotly.JSONDataSet]>=0.17.0  
+kedro >=0.17.0  
+kedro-datasets~=1.0
```
merelcht, line 3: I guess you still only need these datasets `[pandas.ParquetDataSet, pandas.CSVDataSet, plotly.JSONDataSet]` right? So you can still specify that in the import.

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -20,24 +21,65 @@ def get_catalog(self) -> DataCatalog:
...
+        It's recommended to follow the newest format for defining layers in the catalog.
+        """
+        if self._catalog.layers:
+            return self._catalog.layers
+
+        layers: Dict[str, Set[str]] = defaultdict(set)
+
+        for dataset_name in self._catalog._data_sets:
+            dataset = self._catalog._get_dataset(dataset_name)
+            metadata = getattr(dataset, "metadata", {}) or {}
```
merelcht, line 56: What is the `or` condition for? The third argument in `getattr` is already the default return if `metadata` isn't found.

merelcht (commented):
I haven't tried running viz with this, but in general it looks good. Left two small comments!

package/kedro_viz/data_access/repositories/catalog.py
rashidakanchwala: it's to handle a case where metadata = None. When I was working with preview function, this error got thrown multiple times so just handling it here. 
I guess this is because we define metadata as None by default here - https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/kedro_datasets/pandas/csv_dataset.py#L80

rashidakanchwala:
> Very nice and clear PR! 👍
> 
> I think this sort of implementation is ok, but there's some big things to consider already which might change it quite a bit:
> 
> 1. what is the correct way to handle different kedro versions and kedro-viz versions? Restricting kedro-viz 7.x to kedro 0.19 sounds a bit harsh to me. And what is the correct way to handle catalogs which have both the old style and new style `layer` syntax? At the moment your logic is: so long as there's a single dataset with the old style specification, assume we use that for every dataset. What I was imagining originally is that it would work for a catalog with some new and some old style entries and we look at each dataset on a case by case basis. The logic you have is probably fine tbh, but worth thinking about what the user experience is here.
> 2. should we simplify the flow using the new style syntax? Note that `_layers_mapping` is `dataset_full_name -> layer`, and the input catalog is exactly that format already. The munging that we currently do is necessary because `self._catalog.layers` is not already in a format that resembles that. What you have now is basically going from something that looks like `dataset_full_name -> layer` to the `layers` dictionary that maps `layer -> datasets` and then back again to `dataset_full_name -> layer`. Can we avoid doing all this manipulation for the new style syntax? The current solution should work ok but once we remove the logic for the old style syntax will become very roundabout. So we could either do it now or save it for a later refactoring if you prefer.
> 3. Moving over https://github.com/kedro-org/kedro/blob/main/kedro/framework/context/context.py#L122 to kedro-viz, but let's save that for a different PR probably. Good to anticipate what difference it will make in advance though.
> 4. How to handle deprecation warnings? Not sure exactly where these belong actually - needs some careful thought.

Hi Antony, thanks I did think of the above. my thoughts below. 

1.  It's true the way I have handled it right now is that either you use the old way of doing things for all the datasets or the new way, a mix would break. I didn't anticipate someone doing a mix, but you are right, let's say it's an existing project with the old way of doing things, and now the new developers on the project use the new way of defining layers then it would break. But I think if they use the new recommended way anyway for a project, they should also update the existing. I think a mix could cause confusion to the user as well. @yetudada what do you think? 

Regarding the restrictions, @merelcht and @AhdraMeraliQB had decided that Kedro 19 would be a breaking change and this is when they would stop the support so deprecation warnings for layers change would go out in Kedro 18 (@merelcht ?) whenever the user does 'kedro run' but we could also add these warnings when they do 'kedro viz'. Also since Kedro 19 won't support it , it won't work with any version of Kedro-viz that use >= Kedro 19 (unless we handle it) so decided that for Kedro-viz 7.x this would be discontinued as well. 

2. Yes, the way it's done right now is a bit going back on forth with the mapping. Once we remove the support layers processing part in data catalog, on framework side, I was thinking of doing the refactor. Even if we refactor it now, we still need to accommodate the old way and will have to do additional refactoring later. I need to add a TODO note in the code though for future refactoring.

3. Yes, we need to tackle this. I will create a new ticket to do this in Kedro-viz. Thanks

noklam (commented):
I expect there are follow-up tickets to change the documentation on `kedro` as well? or maybe it's there already?

yetudada:
It seems weird that we'd need to accommodate the case where `catalog.yml` could have both ways of writing `layer`. I don't have evidence to suggest this would happen. However, it might be wise to think of a way to solve for this. Let me describe what would be an ideal user experience and you let me know if this is possible. 

1. If the user writes `layer` as a top-level attribute (the old way) for any dataset value in `catalog.yml` 
2. Then they can still see their layers on Kedro-Viz 
3. But perhaps when they do `kedro run`, they are notified that this will be changing and which datasets are affected 
4. **Nice to have**: It would be awesome if we could create a script to convert their catalog.yml for them which we would include in the docs

rashidakanchwala (commented):
@antonymilne , @merelcht , @noklam , @yetudada 

I have added support for both configs now even if they are in the same project. Basically, we first read all the layers defined in DataCatalog (top-level layers) and map them as dataset_name : layer. 
And then we check for any layers defined under metadata , and map them as dataset_name: layer

package/tests/test_data_access/test_repositories/test_data_catalog.py
```diff
@@ -17,3 +17,20 @@ def test_get_layer_mapping_for_transcoded_dataset(self):
         repo.set_catalog(catalog)
         assert repo.get_layer_for_dataset("cars") == "raw"
         assert repo.get_layer_for_dataset("cars@pandas") == "raw"
+
+    def test_get_layer_mapping_from_metadata(self):
```
merelcht, line 21: I'd suggest adding another test to check that indeed a mix of `layer` definitions is supported.

noklam (commented):
Almost there! Agree with Merel should have a test for testing the mixed `layers` catalog. I am happy to approve once the test is in place.

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -26,24 +26,59 @@ def strip_encoding(dataset_name: str) -> str:
...
                     }
                 )
+
+        # Maps layers according to the new format
+        for dataset_name in self._catalog._data_sets:
+            dataset = self._catalog._get_dataset(dataset_name)
+            metadata = getattr(dataset, "metadata", {})
+            if not metadata:
+                continue
+            dataset_layer = metadata.get("kedro-viz", {}).get("layer")
```
antonymilne, line 77: I'd definitely go for [EAFP](https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python) for the clearest way to write a nested dictionary lookup. This way you avoid many `get`s and `continue` e.g. https://stackoverflow.com/questions/43491287/elegant-way-to-check-if-a-nested-key-exists-in-a-dict
```
try:
    dataset_layer = dataset.metadata["kedro-viz"]["layer"]
except AttributeError, KeyError:
    pass
else:
    # update self._layers_mapping
```

antonymilne (commented):
This looks much nicer now without all the dictionary reshaping being needed 👍 

I think we're nearly there but I'm really not sure about the decision for kedro-viz 7.x to only support kedro 0.19.x. This is inconsistent with previous policies where we aimed to maintain support for two major versions of kedro and seems unnecessarily harsh to me since in this case it's actually very easy to maintain compatibility with 0.18.x just by keeping in a few lines of legacy code to handle the old layers syntax. We could put this in a `KEDRO_VERSION.match` conditional like we do in other places so it doesn't even get run for kedro 0.19.x when it's irrelevant.

antonymilne (commented):
This looks much nicer now without all the dictionary reshaping being needed 👍 

I think we're nearly there but I'm really not sure about the decision for kedro-viz 7.x to only support kedro 0.19.x. This is inconsistent with previous policies where we aimed to maintain support for two major versions of kedro and seems unnecessarily harsh to me since in this case it's actually very easy to maintain compatibility with 0.18.x just by keeping in a few lines of legacy code to handle the old layers syntax. We could put this in a `KEDRO_VERSION.match` conditional like we do in other places so it doesn't even get run for kedro 0.19.x when it's irrelevant.

rashidakanchwala (commented):
That's true. Ideally viz should support 2 previous versions of Kedro. So it makes sense to leave it there even if Kedro stops supporting the old format of layers. I will bring up above in tomorrow's standup for others input . It also makes sense to add the KEDRO_VERSION == also for code clarity. 
QQ - should I add that now, with the latest version of 18 this feature was released in, or shld i wait till 19? @antonymilne

rashidakanchwala (commented):
That's true. Ideally viz should support 2 previous versions of Kedro. So it makes sense to leave it there even if Kedro stops supporting the old format of layers. I will bring up above in tomorrow's standup for others input  It also makes sense to add the KEDRO_VERSION == for code clarity. 
QQ - should I add that now, with the latest version of 18 this feature was released in, or shld i wait till 19? @antonymilne

rashidakanchwala (commented):
Ignore comment

antonymilne:
> That's true. Ideally viz should support 2 previous versions of Kedro. So it makes sense to leave it there even if Kedro stops supporting the old format of layers. I will bring up above in tomorrow's standup for others input . It also makes sense to add the KEDRO_VERSION == also for code clarity. QQ - should I add that now, with the latest version of 18 this feature was released in, or shld i wait till 19? @antonymilne

I would do it now with a `if KEDRO_VERSION.match("<=0.19.0")` on the legacy method.

noklam (approved):
LGTM

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -26,24 +26,60 @@ def strip_encoding(dataset_name: str) -> str:
...
+        # Maps layers according to the new format
+        for dataset_name in self._catalog._data_sets:
+            dataset = self._catalog._get_dataset(dataset_name)
+            metadata = getattr(dataset, "metadata", None)
+            if not metadata:
+                continue
+            try:
+                dataset_layer = dataset.metadata["kedro-viz"]["layer"]
+            except (AttributeError, KeyError):  # pragma: no cover
+                pass
```
merelcht, line 80: I think it would be helpful to add an error message for debugging.

merelcht (approved):
One minor comment but otherwise this looks good! 🙂 Also don't forget to add this to the release notes!

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -26,24 +29,63 @@ def strip_encoding(dataset_name: str) -> str:
...
-            for layer, dataset_names in self._catalog.layers.items():
-                self._layers_mapping.update(
-                    {
-                        self.strip_encoding(dataset_name): layer
-                        for dataset_name in dataset_names
-                    }
+        self._layers_mapping = {}
+
+        # Maps layers according to the old format
+        if KEDRO_VERSION.match("<=0.19.0"):
```
antonymilne, line 59:
```suggestion
        if KEDRO_VERSION.match("<0.19.0"):
```
I think I made a typo before, sorry... Please do check I got the `match` syntax right also because this was just from memory.

antonymilne (approved):
Just one thing to check but otherwise looks good, nice work! 👍

package/kedro_viz/data_access/repositories/catalog.py
rashidakanchwala: it's correct. i checked :)

Issue #662, opened 2021-12-01
Visualize size of processed datasets


lukaszdz:
## Description
I'm always frustrated when I'm running daily or weekly sets of modular pipelines and my final output does not make complete sense. This indicates that there was an issue when running the pipeline but I'm not sure, at a glance, what step didn't provide output.

One example problem: one initial dataset had the mapping of market IDs. One day, the market ID for our second biggest market was omitted from the first step, causing all subsequent downstream analysis to be off by a nontrivial amount. 

## Context
This change is important to me because it would help me, at a glance, identify changes across runs through visual cues, so I know where to begin.

## Possible Implementation
Visualize the total size of each dataset that has been processed via kedro viz:
The day that things ran correctly:
![image](https://user-images.githubusercontent.com/588971/144320322-197f2345-8aa9-4918-8f7e-fcef38937ab7.png)

The day that things failed:
![image](https://user-images.githubusercontent.com/588971/144320409-c53ba724-0aaf-4798-b5ab-0d7ed32a48d1.png)
Would be nice to also visualize the nodes that had been attempted to run, but failed
![image](https://user-images.githubusercontent.com/588971/144320725-bf58a897-7d4a-4d5c-abbb-fda7d596eec0.png)

In this example, by visualizing the size of each step that had been run, you would immediately see that the data set with the biggest difference was the companies set. Even though the pipeline strictly failed a step later, you would immediately know where to start debugging.

AntonyMilneQB:
This is more of a kedro-viz issue so I've moved it 🙂  This is a great suggestion @lukaszdz and is also something I've pondered before so let me add some thoughts here... Related: https://github.com/quantumblacklabs/kedro/issues/1076 https://github.com/quantumblacklabs/private-kedro/issues/1148

### Current methods for tracking dataset size
For an immediate solution, outside viz there are actually a couple of different ways you might be able to achieve what you're looking for already:
* great expectations via the [kedro-great plugin](https://github.com/tamsanh/kedro-great). I'm not at all familiar with this myself but I imagine you should be able to write some rule that validates the number of rows in a dataframe 
* use a hook that emits a log message (to the console or a file) giving the number of rows in the dataframe [like this](https://github.com/AntonyMilneQB/kedro-intermediate-training/blob/master/completed-spaceflights/src/spaceflights/hooks.py#L78-L81)

And one which will show you something, though not exactly what you want, in kedro viz:
[...]

PR #1374, opened 2023-05-30
Add preview to datasets as specified in the Kedro catalog under metadata

rashidakanchwala:
## Description

This ticket is part 2 of the preview dataset feature development. In the first part, we enabled preview functionality for CSV and Excel datasets in the Kedro Framework. In Kedro-viz, we showed previews for all CSV and Excel datasets when the user turned on the preview option under the settings panel.

In this ticket, we are removing the preview option from the settings panel. Now, if the user wants to preview a dataset, they need to specify it via the catalog.yml file.

```yaml
type: pandas.CSVDataSet
filepath: /filepath/to/dataset
metadata:
  kedro-viz:
    layer: raw
    preview_args:
          nrows: 3
```

In the above the user wants to see the first 3 rows of the dataset. Previously we always showed the first 40 rows, but now we show the number of rows specified by the user in the catalog.yml file.

## QA notes

The tests and linter is currently failing but the changes to this are already part of the first PR related to layers so it should be fixed once that's merged. So we can ignore that in this PR. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1374"></a>


src/components/metadata-modal/metadata-modal.js
```diff
@@ -46,7 +46,7 @@ const MetadataModal = ({ metadata, onToggle, visible }) => {
         </div>
         {hasPreview && (
           <div className="pipeline-metadata-modal__preview-text">
-            Previewing first 40 rows only
+            Previewing first {metadata.preview.data.length} rows only
```
Huongg, line 49: I think this copy line might not make sense anymore. We use it originally because we were only going to show the first 40 rows. But now since users can define it themself, do we need it anymore?

demo-project/conf/base/catalog_01_raw.yml
```diff
@@ -1,14 +1,41 @@
...
-  layer: raw
+  metadata:
+    kedro-viz:
+      layer: raw
+      preview: 15 
+
+
+
+
+# companies:
```
Huongg, line 28: are we not going to remove these completely? if not then it might be helpful to have a comment to say why we just commented them out here

antonymilne:
My suggested syntax for this is:
```
type: pandas.CSVDataSet
filepath: /filepath/to/dataset
metadata:
  kedro-viz:
    layer: raw
    preview_args:
      nrows: 5
```

This would match `load_args` and `save_args` and mean that future preview functions can take an arbitrary set of arguments. Otherwise you're hardcoding it here to only accept `nrows` as a parameter.

You'd call the preview function as `dataset._preview(**preview_args)`.

There's a question about what the default behaviour should be though. If I don't specify `preview_args`, should the preview show at all?
* if yes, then should a user be able to turn it off if they like and how would they do this?
* if no, then you always need to specify `nrows` explicitly, which is annoying.

Huongg (commented):
hey @rashidakanchwala all looks good to me, just left 2 small comments but nothing seems to be an issue to me.

merelcht:
I agree with @antonymilne 's comment above. `preview_args` would be much more flexible and allow for other things than just number of rows to be shown in the preview. 

> There's a question about what the default behaviour should be though. If I don't specify preview_args, should the preview show at all?

IMO if the user doesn't specify `preview_args` on the dataset it shouldn't preview at all. It seems confusing if it would show without adding anything in the catalog. Also, if we add more options and not just number of rows, what would you show by default?

rashidakanchwala:
> My suggested syntax for this is:
> 
> ```
> type: pandas.CSVDataSet
> filepath: /filepath/to/dataset
> metadata:
>   kedro-viz:
>     layer: raw
>     preview_args:
>       nrows: 5
> ```
> 
> This would match `load_args` and `save_args` and mean that future preview functions can take an arbitrary set of arguments. Otherwise you're hardcoding it here to only accept `nrows` as a parameter.
> 
> You'd call the preview function as `dataset._preview(**preview_args)`.
> 
> There's a question about what the default behaviour should be though. If I don't specify `preview_args`, should the preview show at all?
> 
> * if yes, then should a user be able to turn it off if they like and how would they do this?
> * if no, then you always need to specify `nrows` explicitly, which is annoying.

The preview shouldn't show at all. Basically, if you want to show preview then you need to have the preview_args in your paml and nrows would follow. If people don't specify nrows but want to still preview what would it look like. I am a bit confused. 

> ```
> type: pandas.CSVDataSet
> filepath: /filepath/to/dataset
> metadata:
>   kedro-viz:
>     layer: raw
>     preview_args: ?
> ```

Currently, we don't have any other options. But I guess if the user doesn't add nrows, it would default to 40, which is what is specified already in kedro-datasets. I will still test it

package/kedro_viz/models/flowchart.py
```diff
@@ -488,7 +488,12 @@ def is_tracking_node(self):
 
     def is_preview_node(self):
         """Checks if the current node has a preview"""
-        return hasattr(self.kedro_obj, "_preview")
+        metadata = getattr(self.kedro_obj, "metadata", {}) or {}
+        return bool(metadata.get("kedro-viz", {}).get("preview"))
```
noklam, line 492: It feels like it shouldn't be that complicated. 
Many questions here:
* Does `preview: 0` count as `True` or `False`? It will be `False` now, then the `get("preview", 0) path will never be executed.

The same pattern accessing the nested dict appears in multiple places, maybe refactor it with a small util function

antonymilne:
@merelcht 
> > There's a question about what the default behaviour should be though. If I don't specify preview_args, should the preview show at all?
> 
> IMO if the user doesn't specify `preview_args` on the dataset it shouldn't preview at all. It seems confusing if it would show without adding anything in the catalog. Also, if we add more options and not just number of rows, what would you show by default?

I think the problem with this approach is that the user always has to specify `nrows`. Currently this has a default value in kedro-datasets:
```
_preview(self, nrows: int = 40)
```
... but as it stands there's no way to actually use that default value, since as soon as you write `preview_args` then you need to specify `nrows`. In theory it might be nice for a user to just switch `preview` on but without needing to supply any more information.

For the case of pandas datasets it's not too much effort for the user, but in future maybe we'll have `preview` methods for lots of different datasets. In that case it would be nice for a user to just be able to say `preview: True` without needing to specify all the `preview_args` which might well be different for different dataset tyes.

It's not a big thing so I'm ok with this if it's what we want to do, but I'd probably make `nrows` compulsory in kedro-datasets in this case.

The alternative solution is maybe we have two keys: `preview` and `preview_args`, like this:
```
preview: true
preview_args:
  nrows: ...
```
where `preview_args` is optional and only for "advanced" users who want to fine-tune the preview shown.

Or another option is that `preview` is enabled by default and they need to specify `preview: false` to disable it.

merelcht:
@antonymilne We discussed this in stand up last week and agreed that the user needs to be specific about what needs to be previewed. So they will have to specify `preview_args` or nothing will show. There just won't be a default at all. We can always iterate on that solution once we get user feedback on their experience, but at the moment we don't even know yet what potential other preview functions we or other people will add.

src/components/metadata-modal/metadata-modal.js
merelcht: I think it's still useful to have the message but maybe rephrase it to: "Previewing first ... rows"

tynandebold: Agree with Merel here. Let's still give them the number.

package/kedro_viz/models/flowchart.py
```diff
@@ -486,14 +486,28 @@ def is_tracking_node(self):
...
+    def get_viz_metadata(self):
+        """Gets the kedro-viz metadata config specified in the catalog"""
+        metadata = getattr(self.kedro_obj, "metadata", None)
+        if not metadata:
+            return None
+        try:
+            viz_metadata = metadata["kedro-viz"]
+        except (AttributeError, KeyError):
+            return None
+        return viz_metadata
```
noklam, line 498: Can this be replaced with the `getattr` method since it's not a nested key?

noklam: additional comment: Would it be better to just set this as a property `viz_metadata` instead of a get method?

rashidakanchwala: i did try this but I got some weird errors, so changed it to get_ . but i agree, it's better as a property. I will ping u again, if u have a couple of min to help me with the errors that would be appreciated. let me try doing it again.

demo-project/conf/base/catalog_01_raw.yml
```diff
@@ -1,14 +1,29 @@
 companies:
   type: pandas.CSVDataSet
   filepath: ${base_location}/01_raw/companies.csv
-  layer: raw
+  metadata:
+    kedro-viz:
+      layer: raw
+      preview_args:
```
merelcht, line 7: I think there's a value missing here?

demo-project/conf/base/catalog_04_feature.yml
```diff
@@ -29,6 +29,9 @@
 feature_importance_output: 
   type: pandas.CSVDataSet
   filepath: ${base_location}/04_feature/feature_importance_output.csv
-  layer: feature 
+  metadata:
+    kedro-viz:
+      layer: feature 
+      preview: 2
```
merelcht, line 35: I thought we would just do the `preview_args` and not the `preview`?

package/kedro_viz/models/flowchart.py
```diff
@@ -458,6 +460,12 @@ def __post_init__(self):
         self.modular_pipelines = self._expand_namespaces(
             self._get_namespace(self.full_name)
         )
+        metadata = getattr(self.kedro_obj, "metadata", None)
+        if metadata:
+            try:
+                self.viz_metadata = metadata["kedro-viz"]
+            except (AttributeError, KeyError):
+                pass
```
merelcht, line 468: Would it be helpful here to emit an error message for debugging purposes?

merelcht (commented):
@rashidakanchwala I'll message you on Slack as well because I'm a bit confused now about the implementation: are we using `preview_args` or `preview` with a numeric value? It seems mixed in the changes you've made..

demo-project/conf/base/catalog_04_feature.yml
rashidakanchwala: this is just old code. i removed it. thanks for flagging

package/kedro_viz/models/flowchart.py
```diff
@@ -488,7 +497,15 @@ def is_tracking_node(self):
...
         """Checks if the current node has a preview"""
-        return hasattr(self.kedro_obj, "_preview")
+        try:
+            is_preview = bool(self.viz_metadata["preview_args"])
+        except (AttributeError, KeyError):
+            return False
+        return is_preview
+
+    def get_preview_args(self):
+        """Gets the number of rows for the preview dataset"""
```
merelcht, line 507:
```suggestion
        """Gets the preview arguments for a dataset"""
```

package/tests/test_models/test_flowchart.py
```diff
@@ -368,6 +369,17 @@ def test_data_node_metadata(self):
         assert data_node_metadata.filepath == "/tmp/dataset.csv"
         assert data_node_metadata.run_command == "kedro run --to-outputs=dataset"
 
+    def test_get_preview_nrows(self):
+        metadata = {"kedro-viz": {"preview": 3}}
```
merelcht, line 373: I'm guessing this test needs to be updated to use `preview_args` right?

merelcht (commented):
Almost there!

package/tests/test_models/test_flowchart.py
rashidakanchwala: thanks for this. i missed it. the tests are still failing but that's because of the kedro.extras and it should go once the other PR is merged.

merelcht (approved):
Nice work! ⭐ Don't forget to add the change to the release notes.

package/tests/test_models/test_flowchart.py
```diff
@@ -368,6 +369,18 @@ def test_data_node_metadata(self):
...
+    def test_get_preview_args(self):
+        metadata = {"kedro-viz": {"preview_args": {"nrows": 3}}}
+        dataset = CSVDataSet(filepath="test.csv", metadata=metadata)
+        data_node = GraphNode.create_data_node(
+            full_name="dataset",
+            tags=set(),
+            layer=None,
+            dataset=dataset,
+        )
+        assert data_node.is_preview_node() is True
```
antonymilne, line 381:
```suggestion
        assert data_node.is_preview_node()
```

antonymilne (approved):
A few small suggestions but generally this looks great ⭐

package/kedro_viz/models/flowchart.py
```diff
@@ -488,7 +496,15 @@ def is_tracking_node(self):
 
     def is_preview_node(self):
         """Checks if the current node has a preview"""
-        return hasattr(self.kedro_obj, "_preview")
+        try:
+            is_preview = bool(self.viz_metadata["preview_args"])
+        except (AttributeError, KeyError):
+            return False
+        return is_preview
```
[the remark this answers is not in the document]
antonymilne: btw I see this is probably the opposite of what @noklam was suggesting, so feel free to ignore and just do whatever you think is best 😁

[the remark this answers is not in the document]
rashidakanchwala: I think viz_metadata might be eventually used for more than preview. for e.g. looking at this (https://github.com/kedro-org/kedro-viz/issues/662) ticket. if we were to also show table statistics we could probably have 

metadata :
   kedro-viz: 
       preview_args : 
             nrows=4 
      profiler_args:
              show = true.. 
              
something like it

PR #1388, opened 2023-06-08
Fix/linking flowchart error

### LINKED ISSUE #1320: Kedro-viz flowchart linking breaks when you are on a pipeline that is not default
## Description

Kedro-viz flowchart linking breaks when you are on a pipeline that is not default

Screenshot 2023-04-18 at 11 19 36

## Steps to Reproduce

1. Select a pipeline 'Data Ingestion' or any other from the dropdown on Viz Flowchart
2. Now go to Experiments and click on linking to flowchart 
3. You get the above error saying 'No Url found' , note - you are still on the 'Data Ingestion' pipeline but on the URL it says pipeline_id=default. 

## Expected Result
It must take you to the Default Pipeline and show you the selected node on the flowchart. 

```
```

```
```

## Your Environment

## Checklist

Huongg:
## Description

Fixes #1320. 

## Development notes

The issue that causing this is when switching from ET to FC, and especially when pipeline is not on the default view, it would not have enough time to load the correct set of data, the pass it to redirectLocation hook. Hence the error is thrown. 

This fix includes remove the `useRedirectLocationInFlowChart` from the hooks, and moved it to be inside `FlowchartWrapper`. Also there is no more `reload` to check, instead we used `isGraphLoading` which is handling in redux . 

https://github.com/kedro-org/kedro-viz/assets/32060364/a05c7951-135e-41f7-8b48-dcc6fce3d213

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1388"></a>


Huongg:
> Can we do it so the app respects which pipeline the user has selected? In your video you have a different pipeline than the default, but when you go from experiment tracking to the flowchart via the link the pipeline is back to the default.

hey we would need to switch back to the default pipeline as the selected pipeline (eg: feature_engineering in this case) doesn't have the node that user wants to see

tynandebold:
It's working almost all the time but I've found an instance in which it doesn't:

1. Open the app
2. Select the `Data Ingestion` pipeline
3. Go to Exp. Tracking and click on `Plots`
4. Click on either of the dataset plot links
5. Observe that the URL is invalid

Huongg:
> It's working almost all the time but I've found an instance in which it doesn't:
> 
> 1. Open the app
> 2. Select the `Data Ingestion` pipeline
> 3. Go to Exp. Tracking and click on `Plots`
> 4. Click on either of the dataset plot links
> 5. Observe that the URL is invalid

hmm it seems to load the page from my side ...thats strange. Does it happen again if you re-try it? 

https://github.com/kedro-org/kedro-viz/assets/32060364/adbe3987-e134-46cd-865d-88a2328e2175

tynandebold:
Nice, not sure why it's like that on my end.

@rashidakanchwala and @ravi-kumar-pilla can you please try to reproduce my tests above and see what you get?

ravi-kumar-pilla:
fix_linking_flowchart_test_1

RELEASE.md
```diff
@@ -11,6 +11,10 @@ Please follow the established format:
 
 - Show original node input and output names in metadata panel. (#1381)
 
+## Bug fixes and other changes
+
+- Fix broken URL link between Experiment Tracking and Flowchart when the pipeline is not the default one. (#1388)
```
tynandebold, line 16: Think we historically have these lowercase in the release notes.

```suggestion
- Fix broken URL link between experiment tracking and flowchart when the pipeline is not the default one. (#1388)
```

tynandebold (approved):
Looks great. If it works for both you and Ravi then I'm fine with it!

One other thing while you're in here Huong, can you tweak the z-index of the back button to prevent this problem: 

image

Looks like `.pipeline-wrapper__go-back-btn` needs a `z-index: 3;`.

ravi-kumar-pilla:
break_flowchart_link

I tried multiple times, but it breaks randomly sometimes

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,147 @@ const linkToFlowchartInitialVal = {
...
+   */
   useEffect(() => {
-    const timer = setTimeout(() => {
-      setReload(false);
-    }, 200);
+    const isGraphEmpty = Object.keys(graph).length === 0;
 
-    return () => clearTimeout(timer);
-  }, []);
+    if ((graphState.current === null || usedNavigationBtn) && !isGraphEmpty) {
```
rashidakanchwala, line 150: maybe we change this to graphRef

Huongg: done 👍

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
...
 }) => {
   const history = useHistory();
+  const { pathname, search } = useLocation();
+  const searchParams = new URLSearchParams(search);
 
-  // Reload state is to ensure it will call redirectLocation
-  // only when the page is reloaded.
-  const [reload, setReload] = useState(false);
+  const [errorMessage, setErrorMessage] = useState({});
+  const [invalidUrl, setInvalidUrl] = useState(false);
```
tynandebold, line 65: Thinking we can use our usual `is` convention for boolean state values.

```suggestion
  const [isInvalidUrl, setIsInvalidUrl] = useState(false);
```

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
...
+    window.addEventListener('popstate', () => setUsedNavigationBtn(true));
 
+    return () => {
+      window.removeEventListener('popstate', () => setUsedNavigationBtn(false));
+    };
+  }, []);
+
+  useEffect(() => {
     const linkToFlowchart = loadLocalStorage(localStorageFlowchartLink);
     setGoBackToExperimentTracking(linkToFlowchart);
```
tynandebold, line 145:
```suggestion
    setGoBackToExperimentTracking(loadLocalStorage(localStorageFlowchartLink));
```

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
...
   }, []);
 
+  /**
+   * To handle redirecting to different location via URL, eg: selectedNode, focusNode, etc
+   * We only need to call these matchPath actions either when:
+   * graphRef.current is null means the page first loads
+   * or when user navigating through the back and forward button
+   * or when the invalidUrl is true, that means the user enters some thing wrong in the URL
+   * and we should allow them to reset it by clicking on a different node on the list
+   */
```
tynandebold, line 155:
```suggestion
  /**
   * To handle redirecting to a different location via the URL (e.g. selectedNode,
   * focusNode, etc.) we only need to call the matchPath actions when:
   * 1. graphRef.current is null, meaning the page has just loaded
   * 2. or when the user navigates using the back and forward buttons
   * 3. or when invalidUrl is true, meaning the user entered something wrong in
   * the URL and we should allow them to reset by clicking on a different node.
   */
```

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
...
+
+  const redirectToSelectedPipeline = () => {
+    const pipelineId = searchParams.get(params.pipeline);
+    const foundPipeline = pipelines.find((id) => id === pipelineId);
+
+    if (!foundPipeline) {
+      setErrorMessage(errorMessages.pipeline);
+      setInvalidUrl(true);
+    }
+  };
```
tynandebold, line 119: I'm confused by this function's name. It doesn't actually redirect to anything, does it? It sets a couple state values if a pipeline value **_isn't_** found. There's no logic in the function for when a pipeline value **_is found_**. Where does that happen? If it's not in this function I think we should rename it.

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
...
+      if (matchedSelectedPipeline) {
+        redirectToSelectedPipeline();
+      }
+
+      if (matchedFocusedNode) {
+        redirectToFocusedNode();
+      }
+
+      // Once all the matchPath check is finished
+      // ensure the local states are reset
```
tynandebold, line 183:
```suggestion
      // Once all the matchPath checks are finished
      // ensure the local states are reset
```

src/components/flowchart/flowchart.js
```diff
@@ -458,8 +458,8 @@ export class FlowChart extends Component {
    */
   handleChartClick = () => {
     this.props.onLoadNodeData(null);
-    // To reset URL to '/' when click outside of a node on flowchart
-    this.props.toFlowchartPage();
+    // To reset URL to current active pipeline when click outside of a node on flowchart
+    this.props.toSelectedPipeline();
```
tynandebold, line 462: I think this is having an unintended consequence. When I load Viz in a fresh time and click on the flowchart background, this is what I see in my browser's history:

image

I don't think we should be pushing to history so often.

src/utils/match-path.js
```diff
@@ -0,0 +1,38 @@
+import { matchPath } from 'react-router-dom';
+import { routes } from '../config';
+
+export const findMatchedPath = (pathname, search) => {
+  // find match path
```
tynandebold, line 5: I don't think we need this comment since the function name is very clear :)

src/components/flowchart-wrapper/flowchart-wrapper.js
tynandebold, line 1: It's fantastic how clean the code is here! Well done.

tynandebold (commented):
Great job both of you! This is so clean and reads very well. Fantastic work.

I'm leaving an initial round of comments after a pretty extensive review. Two larger points:

1. I think we need to tighten up how often we're pushing to the browser's history
2. We're too often displaying the `Return to Experiment Tracking` button. I can find a couple ways where that button is on the page and shouldn't be.

Maybe that second point is outside the scope of this PR though. You let me know. I think we should fix the large amount of history events we're storing before merging though.

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
   showGoBackBtn: false,
 };
 
+const getKeyByValue = (object, value) => {
+  return Object.keys(object).find((key) => object[key] === value);
+};
```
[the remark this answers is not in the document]
ravi-kumar-pilla: +1

src/components/flowchart-wrapper/flowchart-wrapper.js
rashidakanchwala: yeah, i agree... i think it should just be handleSelectedPipeline, handleFocusMode etc

src/components/flowchart/flowchart.js
rashidakanchwala: good catch!

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
...
+
+    if (nodeId) {
+      const modularPipeline = nodes[nodeId];
+      const hasModularPipeline = modularPipeline?.length > 0;
+
+      if (hasModularPipeline) {
+        onToggleModularPipelineExpanded(modularPipeline);
+      }
+      onToggleNodeSelected(nodeId);
+      resetErrorMessage();
```
ravi-kumar-pilla, line 104: should we resetErrorMessage everytime ? Can we call resetErrorMessage function only when there is already an error message displayed ?

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -35,63 +35,158 @@ const linkToFlowchartInitialVal = {
...
+        redirectToSelectedNode();
+      }
+
+      if (matchedSelectedPipeline) {
+        redirectToSelectedPipeline();
+      }
+
+      if (matchedFocusedNode) {
+        redirectToFocusedNode();
+      }
```
ravi-kumar-pilla, line 180: [Nit] Do we follow any heirarchy in ordering the if checks for readability ? 
FlowchartMainPage -> selectedPipeline -> selectedNode -> focusedNode ?

ravi-kumar-pilla (changes requested):
Great work !! 
I have added few comments and apart from the history issue mentioned by Tynan, I observed that the node in focus or clicked state does not appear clicked (css) when returning back from ET (though the flowchart and metadata panel show correct data). Added a gif for reference. Thank you

clicked_node_not_having_clicked_css

src/utils/match-path.js
Huongg: ah yeah sorry it wasnt meant to be there, will remove it

src/components/flowchart/flowchart.js
Huongg: hey thanks for spotting this @tynandebold, I've limited the amount of time we push to history by checking if the node has been clicked or not. This also helps with `onLoadNodeData`. It works pretty well

```
if (this.props.clickedNode) {
      // To reset URL to current active pipeline when click outside of a node on flowchart
      this.props.toSelectedPipeline();
    }
```

src/components/flowchart-wrapper/flowchart-wrapper.js
Huongg: hmm i don't think we follow that rule inside the component. We need it for CSS as all the lines are next to each other so it helps with readability.

Huongg:
> Great work !! I have added few comments and apart from the history issue mentioned by Tynan, I observed that the node in focus or clicked state does not appear clicked (css) when returning back from ET (though the flowchart and metadata panel show correct data). Added a gif for reference. Thank you
> 
> clicked_node_not_having_clicked_css

Hey @ravi-kumar-pilla thanks for taking a look. I can't seem to reproduce that issue at all. I tried even in incognito mode several time but it seems to be working fine for me.

Huongg:
> 2\. We're too often displaying the `Return to Experiment Tracking` button. I can find a couple ways where that button is on the page and shouldn't be.

hey @tynandebold , thanks for your review. Please see my replies as below:

- I've mentioned in the comment that I limited the time we push to history when clicking outside node, by checking if any node has been clicked yet. And other cases where we need to do history.push include. I went through them all and see if we can simplify or limit any action but they are all independently and not happening very frequently. 

  1. onCloseClick metadata
  2. click outside a node to close the metadata panel
  3. onItemClick(when you select a node, or when you select or focus on one)
  4. reset the URL to the current active pipeline when switching between different view
  5. resetView in pipeline 
  6. To reset URL to current active pipeline when click outside of a node on flowchart 


- For this point, I'd suggest not to fix in this PR, but in a separate one, perhaps on when we're working on the refactor for ET instead as it links to the ET more. I've created the issue here https://github.com/kedro-org/kedro-viz/issues/1402

Let me know your though on this 😄

tynandebold (approved):
Looking good, thank you for addressing my comments! I'm approving with one last observation that I found while testing which I think we still need to address. Try this out:

1. Open Viz
2. Go to the exp. tracking page
3. Click on a metadata link to be taken to the flowchart
4. Click on the `Return to Exp. Tracking` button
5. Go back to the flowchart page by clicking on the flowchart icon

When I do the above the flowchart loads with the preselected node from step 3 but the URL is only showing `http://localhost:4141/`. I don't think this should happen. Either we should show the selected information in the URL **_or_** clear everything and show the initial flowchart view, when nothing is selected and there are no params in the URL.

tynandebold:
> 

I can't reproduce this either.

tynandebold:
> > Great work !! I have added few comments and apart from the history issue mentioned by Tynan, I observed that the node in focus or clicked state does not appear clicked (css) when returning back from ET (though the flowchart and metadata panel show correct data). Added a gif for reference. Thank you
> > clicked_node_not_having_clicked_css
> >       
> >         [
> >           
> >         
> >             clicked_node_not_having_clicked_css
> >           ](https://user-images.githubusercontent.com/87735534/246467014-11e4eea7-3689-4ee6-8e1f-2313988e7c82.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/87735534/246467014-11e4eea7-3689-4ee6-8e1f-2313988e7c82.gif)
> 
> Hey @ravi-kumar-pilla thanks for taking a look. I can't seem to reproduce that issue at all. I tried even in incognito mode several time but it seems to be working fine for me.

I can't reproduce this either.

tynandebold:
> > 2. We're too often displaying the `Return to Experiment Tracking` button. I can find a couple ways where that button is on the page and shouldn't be.
> 
> hey @tynandebold , thanks for your review. Please see my replies as below:
> 
> * I've mentioned in the comment that I limited the time we push to history when clicking outside node, by checking if any node has been clicked yet. And other cases where we need to do history.push include. I went through them all and see if we can simplify or limit any action but they are all independently and not happening very frequently.
>   
>   1. onCloseClick metadata
>   2. click outside a node to close the metadata panel
>   3. onItemClick(when you select a node, or when you select or focus on one)
>   4. reset the URL to the current active pipeline when switching between different view
>   5. resetView in pipeline
>   6. To reset URL to current active pipeline when click outside of a node on flowchart
> * For this point, I'd suggest not to fix in this PR, but in a separate one, perhaps on when we're working on the refactor for ET instead as it links to the ET more. I've created the issue here [Refactor redirect hook for Experiment Tracking #1402](https://github.com/kedro-org/kedro-viz/issues/1402)
> 
> Let me know your though on this 😄

Yes, agreed! Let's do it this way.

Huongg:
> Looking good, thank you for addressing my comments! I'm approving with one last observation that I found while testing which I think we still need to address. Try this out:
> 
> 1. Open Viz
> 2. Go to the exp. tracking page
> 3. Click on a metadata link to be taken to the flowchart
> 4. Click on the `Return to Exp. Tracking` button
> 5. Go back to the flowchart page by clicking on the flowchart icon
> 
> When I do the above the flowchart loads with the preselected node from step 3 but the URL is only showing `http://localhost:4141/`. I don't think this should happen. Either we should show the selected information in the URL **_or_** clear everything and show the initial flowchart view, when nothing is selected and there are no params in the URL.

Hey good catch, I forgot to update one of the config field for the default main pipeline to be `"/"`. Its now updated and the issue is fixed

ravi-kumar-pilla (approved):

PR #1381, opened 2023-06-06
Show unpretty input/output names in node details

ravi-kumar-pilla:
## Description

Resolves #1312 

## Development notes

- Modified the [GET] node metadata endpoint to accept a query param boolean isPrettyName which defaults to True
- Modified the response of [GET] node metadata endpoint inputs/outputs based on the boolean value isPrettyName

## QA notes

- Modified existing unit test for TestNodeMetadataEndpoint class - test_task_node_metadata to test_task_node_metadata_with_pretty_name
- Added new unit test for TestNodeMetadataEndpoint - test_task_node_metadata_without_pretty_name

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added new entries to the `RELEASE.md` file
- [x] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1381"></a>

PR #1394, opened 2023-06-12
Refactor Pretty Logic from backend to frontend

### LINKED ISSUE #1384: Refactor pretty logic from backend to frontend
## Description
Refactor the _pretty_name logic to frontend for maintaining API design consistency.

## Context
The context for this request is to support #1381 (Show unpretty input/output names in node details) and also adhere to the API design of not sending query parameters.

ravi-kumar-pilla:
## Description

Resolves #1384 

## Development notes

- Moved pretty and stripNamespace Logic from backend to frontend
- Removed name attribute from node schema. Every node should contain an attribute full_name.
- Modified the test case which falls back on name attribute when full_name is not present to verify all nodes should contain full_name. 

## QA notes

- Unit Tests are included in #1381 
- The metadata panel displays node name, inputs/outputs according to the pretty name flag set by the user in the settings modal

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1394"></a>


package/tests/test_api/example_pipelines.json
```diff
@@ -129,18 +119,18 @@
...
-    { "id": "split", "name": "Split" },
-    { "id": "train", "name": "Train" }
+    { "id": "split", "name": "split" },
+    { "id": "train", "name": "train" }
   ],
   "pipelines": [
-    { "id": "__default__", "name": "Default" },
-    { "id": "data_science", "name": "Data Science" },
-    { "id": "data_processing", "name": "Data Processing" }
+    { "id": "__default__", "name": "__default__" },
```
rashidakanchwala, line 126: nit - maybe for consistency we can replace full_name to name everywhere since we just only send 'name' now.

package/tests/test_api/test_rest/test_reponses.py
```diff
@@ -66,6 +66,7 @@ def assert_modular_pipelines_tree_equal(response: Dict, expected: Dict):
 
 
 def assert_example_data(response_data):
+    print(response_data)
```
rashidakanchwala, line 69: we can remove this

package/tests/test_models/test_flowchart.py
```diff
@@ -243,13 +237,13 @@ def test_create_non_existing_parameter_node_empty_dataset(self, patched_warning)
 
 
 class TestGraphNodePipelines:
-    def test_registered_pipeline_pretty_name(self):
+    def test_registered_pipeline_name(self):
```
rashidakanchwala, line 240: maybe these tests are no longer required

rashidakanchwala (changes requested):
hey Ravi, thanks for this. While I am still reviewing, just had one nit comment around naming, i feel we could keep it to `name` in the BE instead of `full_name`, i think everywhere we use `full_name`, but in pipelines there's just `name` which then gets a bit confusing. Plus since we only send one name now maybe we don't need to distinguish whether it's full_name or name. 

besides that, i found one issue when i toggled 'Pretty name' off 
Screenshot 2023-06-14 at 14 28 03

PR #1406, opened 2023-06-20
Validate layers for transcoding datasets 

### LINKED ISSUE #1382: Validate layers for transcoding datasets
## Description

Previously, the layers were computed in DataCatalog on Kedro framework. However, with the introduction of the new layer configuration under metadata -> kedro-viz, this computation now happens in kedro-viz.

Similarly, when dealing with transcoded datasets that point to the same dataset, the layer information should also be the same. In Kedro framework, this validation was performed in the DataCatalog (see https://github.com/kedro-org/kedro/blob/main/kedro/framework/context/context.py#L121C1-L142), but now it needs to be carried out in Kedro-viz.

This change aligns with the fact that layers are specific to Kedro-viz and not to Kedro framework. So, if users provide incorrect layer information when running 'kedro run', it should not result in failure. Failure should only occur when users are attempting to view this layer information on the flowchart using 'kedro-viz'.

## Possible Implementation
Move the layer validation for transcoded datasets to Kedro-viz. 

## Checklist

rashidakanchwala:
## Description

Resolves #1382 

Moved layer validation to Kedro-viz now. This supports both old and new way of defining layers. 

The Kedro Framework part of this ticket is [here](https://github.com/kedro-org/kedro/pull/2705)

## Development notes

The `strip_encoding` function in the Catalog Repository now performs the same task as `_strip_transcoding()`, which is to remove the '@dataset' from the transcoding dataset name. Therefore, I have removed it.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1406"></a>


package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -23,9 +24,14 @@ def get_catalog(self) -> DataCatalog:
...
-    def strip_encoding(dataset_name: str) -> str:
-        return dataset_name.split("@")[0]
+    def _validate_layers_for_transcoding(self, dataset_name, layer):
+        if TRANSCODING_SEPARATOR in dataset_name:
+            dataset_name = _strip_transcoding(dataset_name)
+            existing_layer = self._layers_mapping.get(dataset_name)
+            if existing_layer is not None and existing_layer != layer:
+                raise ValueError(f"Transcoded datasets should have the same layer. Mismatch found for: {dataset_name}")
+        return dataset_name, layer
+
```
noklam, line 34: I know it's copied from the `kedro` codebase, but I found it weird that the validation function actually return something.

Is it easy to move the `strip_transcoding` outside and the validation only validate but not return anything?

noklam (commented):
https://github.com/kedro-org/kedro/pull/2705

The example you use here may be a good addition to the unit tests?

package/kedro_viz/data_access/repositories/catalog.py
rashidakanchwala: done !

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -59,17 +64,16 @@ def layers_mapping(self):
...
             else:
                 for layer, dataset_names in self._catalog.layers.items():
-                    self._layers_mapping.update(
-                        {
-                            self.strip_encoding(dataset_name): layer
-                            for dataset_name in dataset_names
-                        }
-                    )
+                    for dataset_name in dataset_names:
+                        if TRANSCODING_SEPARATOR in dataset_name:
```
AhdraMeraliQB, line 73: Why check for the transcoding separator here and not above (line 67)? In either case the check is not necessary as `_strip_transcoding()` handles dataset names that do not contain the transcoding separator, returning `('dataset_name', '')`

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -23,9 +24,13 @@ def get_catalog(self) -> DataCatalog:
     def set_catalog(self, value: DataCatalog):
         self._catalog = value
 
-    @staticmethod
-    def strip_encoding(dataset_name: str) -> str:
-        return dataset_name.split("@")[0]
+    def _validate_layers_for_transcoding(self, dataset_name, layer):
+        existing_layer = self._layers_mapping.get(dataset_name)
+        if existing_layer is not None and existing_layer != layer:
```
AhdraMeraliQB, line 29:
```suggestion
        if existing_layer != layer:
```
If we use 
`existing_layer = self._layers_mapping.setdefault(dataset_name, layer)`
in the line above, we can shorten the if condition, and drop the `self._layers_mapping[dataset_name] = layer` after the if conditions on lines 76 and 95

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -78,14 +82,18 @@ def layers_mapping(self):
...
-                dataset_layer = dataset.metadata["kedro-viz"]["layer"]
+                layer = dataset.metadata["kedro-viz"]["layer"]
             except (AttributeError, KeyError):  # pragma: no cover
                 logger.debug(
                     "No layer info provided under metadata in the catalog for %s",
                     dataset_name,
                 )
             else:
-                self._layers_mapping[self.strip_encoding(dataset_name)] = dataset_layer
+                if TRANSCODING_SEPARATOR in dataset_name:
```
AhdraMeraliQB, line 92: Same as above comment, we don't need to check for the transcoding separator here

AhdraMeraliQB:
Just a note - with this approach, errors would only be thrown on datasets one by one as the transcoding is evaluated, whereas previously the error message would include all transcoded datasets with ill defined layers. From a user perspective I would find the latter less frustrating, as it eliminates the need to fix one error, only for the next dataset to throw one, over and over again.

If possible, could we mimic the procedure in the framework-side implementation?

package/kedro_viz/data_access/repositories/catalog.py
rashidakanchwala: as discussed, we will keep the transcoding_seperator if statement for now so that validation doesn't happen for all datasets just the transcoding ones also because transcoding is not very common.

package/kedro_viz/data_access/repositories/catalog.py
rashidakanchwala: as discussed, we will keep the function as is so that's it does what it says i.e. purely validates if the layer for transcoding datasets and raises an error if two datasets have different layers.

rashidakanchwala:
> [kedro-org/kedro#2705](https://github.com/kedro-org/kedro/pull/2705)
> 
> The example you use here may be a good addition to the unit tests?

@noklam - out of scope for this ticket. i think it's already handled in kedro-framework here - https://github.com/kedro-org/kedro/blob/main/tests/pipeline/test_pipeline.py#L25C41-L27

noklam (approved):

AhdraMeraliQB (commented):
Thanks for walking through everything, happy to move ahead with these changes once tests are passing 👍

PR #1422, opened 2023-06-29
Fix/switching-pipeline

### LINKED ISSUE #1397: Go back button does not work when switching between different pipeline in Flowchart
## Description
This bug is caused by the refactor work for the redirectUrl [here](https://github.com/kedro-org/kedro-viz/pull/1388). 

After switching between different pipelines, if user uses the back button to go back the previous pipeline, it does not load the page or any data at all, unless user refreshes the page. 

https://github.com/kedro-org/kedro-viz/assets/32060364/d87ed623-4a43-4371-816c-f57e44421a2d

## Steps to Reproduce

1. Go to any pipeline
2. Select a different pipeline
3. Click on the go back button in the browser

## Expected Result
It should load the previously selected pipeline

## Actual Result
The page gets stuck on the current pipeline, even the URL already updated the previous pipeline 

## Checklist

Huongg:
## Description

Fixes #1397

Changes include: (I've also put specific comments in where things are changed for easier discussion)
- move `state.pipeline.active` outside `applyFixes` in `initialState`
- call the following inside `flowchart-wrapper` as we already track the back navigation button there
     
```
onUpdateActivePipeline(foundPipeline);
onToggleNodeSelected(null);
onToggleFocusMode(null)
```
- on the pipeline dropdown list, when click or select a different one, make sure it only gets called once, rather than 3 times. 
- remove `history.go(0)` on reset button

Known issue: 

1. If you select any pipeline, eg `modelling stage`, then click on any other 2 nodes
2. Then click on a different pipeline, eg `data-ingestion`
3. Navigate back with the back button

https://github.com/kedro-org/kedro-viz/assets/32060364/579093a3-3f8e-454b-a323-8c69cd619757

- Issue: it will switch to the previous pipeline which is `modelling stage` but it wont show any node yet, even though the URL is updated. But the one after it will be okay It is because when we use the back button to switch pipeline, at that point it doesnt have time to load the correct set of data yet. 
- Solution: I dont think this will be a big issue as not many user will use the back button that often, and if they do, and notice it, once they refresh the page it will load again normally. Perhaps we can just ignore this issue and see if there's any users will notice and raise an issue about it? 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1422"></a>

### DIFF
--- a/src/components/app/app.test.js
+++ b/src/components/app/app.test.js
@@ -101,7 +101,7 @@ describe('App', () => {
     const activePipeline = spaceflights.pipelines.find(
       (pipeline) => !demo.pipelines.map((d) => d.id).includes(pipeline.id)
     );
-    const { container, rerender } = render(<App data={spaceflights} />);
+    const { container } = render(<App data={spaceflights} />);
     const pipelineDropdown = container.querySelector('.pipeline-list');
     const menuOption = within(pipelineDropdown).getByText(
       prettifyName(activePipeline.name)
@@ -115,8 +115,5 @@ describe('App', () => {
     expect(pipelineDropdownLabel.innerHTML).toBe(
       prettifyName(activePipeline.name)
     );
-    rerender(<App data={demo} />);
-    // the default dropdown placeholder is 'Please select...' which is not returned right after a rerender
-    // expect(pipelineDropdownLabel.innerHTML).toBe('Please select...');
   });
 });
--- a/src/components/app/app.test.js
+++ b/src/components/app/app.test.js
@@ -117,6 +117,6 @@ describe('App', () => {
     );
     rerender(<App data={demo} />);
     // the default dropdown placeholder is 'Please select...' which is not returned right after a rerender
-    expect(pipelineDropdownLabel.innerHTML).toBe('Please select...');
+    // expect(pipelineDropdownLabel.innerHTML).toBe('Please select...');
   });
 });
--- a/src/components/ui/dropdown/dropdown.js
+++ b/src/components/ui/dropdown/dropdown.js
@@ -144,7 +144,8 @@ const Dropdown = (props) => {
         onClosed();
       }
     }
-  });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedObject, selectedObjRef]);
 
   // Event to be fired on componentWillUnmount
   useEffect(() => {
--- a/src/store/initial-state.js
+++ b/src/store/initial-state.js
@@ -74,14 +74,14 @@ export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
   const state = mergeLocalStorage(normalizeData(data, expandAllPipelines));
 
   const search = new URLSearchParams(window.location.search);
-  const pipelineFromURL = search.get(params.pipeline);
+  const pipelineIdFromURL = search.get(params.pipeline);
 
-  if (pipelineFromURL) {
+  if (pipelineIdFromURL) {
     // Use main pipeline if pipeline from URL isn't recognised
-    if (!state.pipeline.ids.includes(pipelineFromURL)) {
+    if (!state.pipeline.ids.includes(pipelineIdFromURL)) {
       state.pipeline.active = state.pipeline.main;
     } else {
-      state.pipeline.active = pipelineFromURL;
+      state.pipeline.active = pipelineIdFromURL;
     }
   }
 
--- a/src/store/initial-state.js
+++ b/src/store/initial-state.js
@@ -72,23 +72,25 @@ export const mergeLocalStorage = (state) => {
  */
 export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
   const state = mergeLocalStorage(normalizeData(data, expandAllPipelines));
-  if (applyFixes) {
-    const search = new URLSearchParams(window.location.search);
-    const pipelineFromURL = search.get(params.pipeline);
 
-    // Use main pipeline if active pipeline from localStorage isn't recognised
-    if (!state.pipeline.ids.includes(state.pipeline.active)) {
+  const search = new URLSearchParams(window.location.search);
+  const pipelineFromURL = search.get(params.pipeline);
+
+  if (pipelineFromURL) {
+    // Use main pipeline if pipeline from URL isn't recognised
+    if (!state.pipeline.ids.includes(pipelineFromURL)) {
       state.pipeline.active = state.pipeline.main;
-    } else if (pipelineFromURL) {
-      // Use main pipeline if pipeline from URL isn't recognised
-      if (!state.pipeline.ids.includes(pipelineFromURL)) {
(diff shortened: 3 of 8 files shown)


src/components/ui/dropdown/dropdown.js
```diff
@@ -144,7 +144,8 @@ const Dropdown = (props) => {
         onClosed();
       }
     }
-  });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedObject, selectedObjRef]);
```
Huongg, line 148: if we don't have this to limit the amount of time this useEffect gets called, everything here will get called around 3 times every time you switch a pipeline. And as a result, it will push to the history 3 times.

Huongg: but it does affect the test from line https://github.com/kedro-org/kedro-viz/blob/8d330b075ec34865fc0bd0bfe8a2d37fd5d94330/src/components/app/app.test.js#L120

I'm not sure if we need this test as it basically checks if the "Please select..." renders when the component renders. Should we check this after we select a different pipeline from the dropdown?

src/components/app/app.test.js
```diff
@@ -117,6 +117,6 @@ describe('App', () => {
     );
     rerender(<App data={demo} />);
     // the default dropdown placeholder is 'Please select...' which is not returned right after a rerender
-    expect(pipelineDropdownLabel.innerHTML).toBe('Please select...');
+    // expect(pipelineDropdownLabel.innerHTML).toBe('Please select...');
```
tynandebold, line 120: I think we can remove this test, as I don't feel it adds much value.

src/components/ui/dropdown/dropdown.js
tynandebold: Like I said above, I think we can remove this.

src/utils/hooks/use-generate-pathname.js
```diff
@@ -18,20 +18,24 @@ export const useGeneratePathname = () => {
...
-
-    // FIX: this is just a temporary solution for a known issue from https://github.com/kedro-org/kedro-viz/issues/1397
-    history.go(url);
   }, [history]);
 
-  const toSelectedPipeline = useCallback(() => {
-    const activePipeline = getCurrentActivePipeline();
+  const toSelectedPipeline = useCallback(
+    (pipelineValue) => {
+      // Get the value from param if existed first
```
tynandebold, line 25:
```suggestion
      // Get the value from param if it exists first
```

tynandebold (approved):
Looks and works well! I'm approving with two comments:

- When I use the back button and land on our root URL (http://localhost:4141/), the flowchart doesn't update. I feel like it probably should. What do you think?
- Paste this URL into a new browser window: http://localhost:4141/?pipeline_id=__default__&selected_id=72baf5c6. It'll crash the app. It seems to only happen if the node is of type `parameter`. You didn't introduce this in your PR, as it's on the demo site, too. I'm wondering if you can have a very quick look and see if you can solve it in this PR? And by quick, I mean like 15 minutes, max. Otherwise I'll create an issue for it.

Huongg:
> Looks and works well! I'm approving with two comments:
> 
> * When I use the back button and land on our root URL (http://localhost:4141/), the flowchart doesn't update. I feel like it probably should. What do you think?
> * Paste this URL into a new browser window: http://localhost:4141/?pipeline_id=__default__&selected_id=72baf5c6. It'll crash the app. It seems to only happen if the node is of type `parameter`. You didn't introduce this in your PR, as it's on the demo site, too. I'm wondering if you can have a very quick look and see if you can solve it in this PR? And by quick, I mean like 15 minutes, max. Otherwise I'll create an issue for it.

hey @tynandebold thanks for taking a look. 

For point 1, when you say the flowchart doesn't update. do you mean the view doesnt set back to be in central? As I can see the data does update back to default and no nodes are selected when navigating back through the button

https://github.com/kedro-org/kedro-viz/assets/32060364/41a67463-1257-4524-96be-e8d197fcb13c


For point 2, I have a quick call with Rashida after lunch to confirm on this approach, it can be a quick fix but I just want to double check with her. Will let you know after our call.

tynandebold:
Not exactly what I meant for the first point. Hopefully this is more clear:

https://github.com/kedro-org/kedro-viz/assets/16869061/f7574cf3-20f7-4410-93d2-9bd0711a3a6f

Huongg:
> Not exactly what I meant for the first point. Hopefully this is more clear:
> 
>  Screen.Recording.2023-07-06.at.12.08.13.mov

ah I see...that's an interesting point. As we still have the pipeline stored in localStorage, so unless you go from `http://localhost:4141/?pipeline_id=__default__` to `http://localhost:4141/?pipeline_id=Modelling%20stage`, it will know to direct back to default pipeline when go back. But if its just `http://localhost:4141/` it will just check what currently in localStorage. Maybe we should discuss whether we should store and track pipeline from localStorage anymore, now that we have it in the URL anyway? 


For the second point, best if we can have a separate ticket for this, as it's related to the `parameters` is always off by default. (previous [PR ](https://github.com/kedro-org/kedro-viz/pull/416)for that)  . So I assume we will need to check if the node is parameters, then will need to turn parameter back on, it wont be a fix in 15 mins I'm afraid. prob 3 points.

tynandebold:
Ok, noted for point two. I'll make a ticket. 

And for the first point, I do think that when we go from `http://localhost:4141/?pipeline_id=Modelling%20stage` back to `http://localhost:4141` with the back button we should update the flowchart to show the default pipeline. We can also do that in a separate ticket if it's too much to do here. What you've done is a great improvement, so we can get it in.

Huongg:
> Ok, noted for point two. I'll make a ticket.
> 
> And for the first point, I do think that when we go from `http://localhost:4141/?pipeline_id=Modelling%20stage` back to `http://localhost:4141` with the back button we should update the flowchart to show the default pipeline. We can also do that in a separate ticket if it's too much to do here. What you've done is a great improvement, so we can get it in.

hey thanks for making the ticket for point 2. I think for point 1, if we do it that way, we basically say `http://localhost:4141` will always be the default pipeline? If so then we don't really need to read from local storage anymore. But I agree lets discuss it and do it in a separate ticket

src/store/initial-state.js
```diff
@@ -72,23 +72,25 @@ export const mergeLocalStorage = (state) => {
  */
 export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
   const state = mergeLocalStorage(normalizeData(data, expandAllPipelines));
-  if (applyFixes) {
-    const search = new URLSearchParams(window.location.search);
-    const pipelineFromURL = search.get(params.pipeline);
 
+  const search = new URLSearchParams(window.location.search);
+  const pipelineFromURL = search.get(params.pipeline);
```
ravi-kumar-pilla, line 77: [nit] can we rename this to pipelineIdFromURL ?

ravi-kumar-pilla (approved):

PR #875, opened 2022-05-30
Feature/better apply and close ux

tynandebold:
## Description

Resolves #860.

## Development notes

The heft of this is in the `Modal` files. I added some state to be able to track what's been clicked, and then the different button text and styles act on that.

I thought about refactoring the actual `Modal` component and decided against it. I think it would be a premature optimization.

## QA notes

Here's what it looks like:

https://user-images.githubusercontent.com/16869061/171032346-e1adfb49-4926-4b6d-a5b1-b8445167c867.mov

Take a look at the branch via Gitpod and play around with it.

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [x] Added tests to cover my changes

PR #1424, opened 2023-06-29
Fix button for large pipeline warning

rashidakanchwala:
## Description

This is to fix this issue 

Screenshot 2023-06-29 at 17 41 31

## Development notes

This bug was introduced due to changes on this ticket - #875 where display:flex is being added [here](https://github.com/kedro-org/kedro-viz/blob/main/src/components/ui/button/button.scss#L38). I have removed it for now. Removing this hasn't affected the Apply/Close buttons but @tynandebold can confirm.


Huongg (approved):
thanks for the fix, LGTM 👍

PR #1428, opened 2023-06-30
Fix button for large pipeline warning

rashidakanchwala:
## Description

This is to fix this issue 

Screenshot 2023-06-29 at 17 41 31

## Development notes

This bug was introduced due to changes on this ticket - #875 where display:flex is being added [here](https://github.com/kedro-org/kedro-viz/blob/main/src/components/ui/button/button.scss#L38). I have removed it for now. Removing this hasn't affected the Apply/Close buttons but @tynandebold can confirm. 

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1426"></a>

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1428"></a>

### DIFF
--- a/src/components/pipeline-warning/pipeline-warning.js
+++ b/src/components/pipeline-warning/pipeline-warning.js
@@ -47,10 +47,12 @@ export const PipelineWarning = ({
             a while to render. You can use the sidebar controls to select a
             smaller graph.
           </p>
-          <Button onClick={onHide}>Render it anyway</Button>
-          <Button mode="secondary" onClick={onDisable} size="small">
-            Don't show this again
-          </Button>
+          <div className="pipeline-warning__button-wrapper">
+            <Button onClick={onHide}>Render it anyway</Button>
+            <Button mode="secondary" onClick={onDisable} size="small">
+              Don't show this again
+            </Button>
+          </div>
         </div>
       )}
       {isEmptyPipeline && componentLoaded && (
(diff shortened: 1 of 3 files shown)


src/components/ui/button/button.scss
```diff
@@ -35,7 +35,6 @@ $secondary-underline-offset-hover: 4px;
   box-shadow: none;
   color: var(--color-button__text);
   cursor: pointer;
-  display: flex;
```
tynandebold, line 38: We need this, otherwise it alters the `Apply changes and close` button:

Screenshot 2023-06-30 at 11 45 58

tynandebold (commented):
Left a comment about how removing the `display: flex` changes something we don't want it to. Also, I think the buttons should be next to each other, like they are in our modals:

image

tynandebold (approved):
LGTM!

src/components/pipeline-warning/pipeline-warning.js
```diff
@@ -47,10 +47,12 @@ export const PipelineWarning = ({
             a while to render. You can use the sidebar controls to select a
             smaller graph.
           </p>
-          <Button onClick={onHide}>Render it anyway</Button>
-          <Button mode="secondary" onClick={onDisable} size="small">
-            Don't show this again
-          </Button>
+          <div className="pipeline-warning__button-wrapper">
```
Huongg, line 50: hey @rashidakanchwala  Do we not need any styling for this new classname ?

rashidakanchwala: nope, this div basically ensures that the buttons are not next to each other ... just followed this way (https://github.com/kedro-org/kedro-viz/blob/main/src/components/settings-modal/settings-modal.js#L149)

Huongg (approved):
LGTM 👍

PR #1443, opened 2023-07-11
Clean e2e test setup

### LINKED ISSUE #805: Tidy e2e test Python environment

rashidakanchwala:
## Description

Resolves #805. 

## Development notes

Ok so the main things done in this PR:-
- Added support for the earliest Kedro version that Kedro-viz supports, which is Kedro 0.17.5. The end-to-end (e2e) tests for Kedro 0.17.5 will skip if the Python version is greater than 3.8, as Kedro 18 only works with Python 3.9 and above.
- Removed old code related to Kedro 0.16.0 in environment.py, which involved unnecessary package installations.
- Implemented a suggestion by @tynandebold to capture Kedro-viz breaking due to old dependencies in our end-to-end (e2e) tests. I created a lower-bound requirements.txt file for testing Kedro-viz functionality. If the tests fail, we know that we need to increase the lower-bound on our dependencies.
- Added an additional assert statement to check if kedro-viz is running, as the viz flowchart for Kedro 0.17.5 and Kedro 0.18 are different.
- Removed win-e2e-test for Python 3.7 since it was failing. Since we plan to drop support for Python 3.7 soon, we decided not to spend time investigating the issue.
also thanks @astrojuanlu for help on this ticket :) 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1443"></a>


package/features/viz.feature
```diff
@@ -2,9 +2,25 @@ Feature: Viz plugin in new project
...
         Given I have prepared a config file with example code
 
+    Scenario: Execute viz with the earliest Kedro version that it supports 
+        Given I have installed kedro version "0.17.5"
+        And I have run a non-interactive kedro new with pandas-iris starter
+        And I have installed the project's requirements
+        When I execute the kedro viz command
+        Then kedro-viz should start successfully
+
+    Scenario: Execute viz with latest Kedro with lower-bound viz requirements
```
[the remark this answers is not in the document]
rashidakanchwala: @merelcht 
this is because in our last release we had an issue where strawberry-graphql below < 0.190 wouldn't work with Kedro-viz and we were not able to capture that in our manual testing. 

So we decided to have an e2e-test that would check if kedro-viz works with the lower-bound of all our dependencies (this would mean some maintenance work but we don't have so many dependencies) and if it didn't we know we need to update these dependencies.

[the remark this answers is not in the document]
merelcht: Okay cool, that's very useful 👍

.circleci/continue_config.yml
```diff
@@ -346,11 +340,6 @@ workflows:
           matrix:
             parameters:
               python_version: ['3.7', '3.8', '3.9', '3.10']
-          filters:
```
ravi-kumar-pilla, line 349: Why are we removing branch filters from win_unit_tests ? I saw that for e2e tests we skip the behave scenario in package/features/environment.py , does that apply to unit tests as well ?

package/features/environment.py
```diff
@@ -63,20 +90,7 @@ def _setup_context_with_venv(context, venv_dir):
...
-            "-U",
-            "pip>=21.2",
-            "setuptools>=38.0",
-            "cookiecutter>=1.7.2",
-            "wheel",
-            "botocore",
-            "PyYAML>=4.2, <6.0",
-            "click<9.0",
-        ],
+        [context.python, "-m", "pip", "install", "-U", "pip>=21.2", "setuptools>=38.0"],
```
ravi-kumar-pilla, line 93: are the removed requirements no longer needed ?

package/features/steps/cli_steps.py
```diff
@@ -86,6 +87,19 @@ def install_project_requirements(context):
         assert False
 
 
+@given("I have installed the lower-bound Kedro-viz requirements")
+def install_lower_bound_requirements(context):
+    cwd = Path(__file__).resolve().parent
+    requirements_path = cwd / "lower_requirements.txt"
```
ravi-kumar-pilla, line 93: should we be updating the lower_requirements.txt everytime manually ? If so, will it be helpful to add somewhere in the docs when to do so ?

ravi-kumar-pilla (approved):
Awesome work !! 
I had few questions and added comments, but neither of them is a blocker to merge. So I am approving it :)

.circleci/continue_config.yml
rashidakanchwala: ok good question -  so this is just for test purpose. 
win-e2e and win-unit tests are only done when you merge to main, and during demo. it doesn't happen when you are doing commits in your PR. The reason is because win tests take too long 20min or so to execute and it kind of becomes a blocker. 

I have only temporarily removed this because I want to check how e2e test work on windows. I will remove it before I merge.

rashidakanchwala: we skip testing for python>=3.9 and kedro 17.5 because kedro 17.5 does not support python > 3.8 , so it's no point testing that scenario

package/features/steps/cli_steps.py
rashidakanchwala: good idea.

package/features/environment.py
rashidakanchwala: yes, they were there because of kedro 16 i think. we stopped support for kedro 16 but had this bit of chore leftover to do

.circleci/continue_config.yml
```diff
@@ -331,8 +331,8 @@ workflows:
       - win_e2e_tests:
           matrix:
             parameters:
-              #3.10 fails with win-build, hence reverting it to 3.10
-              python_version: ['3.7', '3.8', '3.9', '3.10']
+              #removing 3.7 for now since it's failing circleCI and we are planning to soon stop support for 3.7
```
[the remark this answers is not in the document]
rashidakanchwala: @merelcht  -- they were WindowsPath failures related to subprocess and stuff.
 I also realised they were only failing for lower-bound requirements. and that's like super niche and not at all important thing to spend time on. 
however I agree we shouldn't remove win-37 completely because of it so thanks for the nudge.  what I have done now in the code is if you are on windows and using python 3.7, then that scenario which checks with lower-bounds just gets skipped.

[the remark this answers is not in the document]
merelcht: Thanks for explaining! I think that's a good solution 👍

merelcht (approved):
Great work @rashidakanchwala ! ⭐ ⭐ Does it make sense to mention this in the release notes?

package/features/environment.py
```diff
@@ -26,6 +30,29 @@ def before_scenario(context, scenario):
     Installs kedro by running pip in the top level directory.
     """
 
+    kedro_version = Version.parse("1.0.0")
```
astrojuanlu, line 33: I'm confused, shouldn't this be 

```
import kedro

kedro_version = Version.parse(kedro.__version__)
```

?

rashidakanchwala: kedro_version was just a placeholder but yes, can replace it with kedro.__version__ 

I will make the semver change to packaging in another PR. As we will need to update it in multiple places.

CONTRIBUTING.md
```diff
@@ -311,6 +311,12 @@ make e2e-tests
 make lint
 ```
 
+### Managing python dependencies on Kedro-viz 
+
+Python dependencies in Kedro-viz are usually updated automatically through tools like Dependabot or triggered by deprecation warnings. When updating dependencies, especially the lower-bound version, it's essential to also update the 'lower-requirements.txt' file manually. This ensures that our e2e tests reflect the latest dependencies, maintaining project stability and compatibility.
+
+If the lower-bound e2e test fails, it indicates that some dependencies may not work correctly with Kedro-viz. To resolve this, update the problematic dependency in both 'requirements.txt' and 'lower-requirements.txt'.
```
rashidakanchwala, line 318: @tynandebold @ravi-kumar-pilla  --as Ravi mentioned I added some info in contributing docs.  let me know if this is fine here?

package/features/environment.py
astrojuanlu: Thanks for the clarification! Maybe adding a code comment would make it more clear

CONTRIBUTING.md
```diff
@@ -311,6 +311,12 @@ make e2e-tests
 make lint
 ```
 
+### Managing python dependencies on Kedro-viz
```
tynandebold, line 314: Nit: we capitalize this.

```suggestion
### Managing Python dependencies on Kedro-viz 
```

CONTRIBUTING.md
```diff
@@ -311,6 +311,12 @@ make e2e-tests
 make lint
 ```
 
+### Managing python dependencies on Kedro-viz 
+
+Python dependencies in Kedro-viz are usually updated automatically through tools like Dependabot or triggered by deprecation warnings. When updating dependencies, especially the lower-bound version, it's essential to also update the 'lower-requirements.txt' file manually. This ensures that our e2e tests reflect the latest dependencies, maintaining project stability and compatibility.
```
tynandebold, line 316:
```suggestion
Python dependencies in Kedro-Viz are usually updated automatically through tools like Dependabot or triggered by deprecation warnings. When updating dependencies, especially the lower-bound version, it's essential to also update the `lower-requirements.txt` file manually. This ensures that our e2e tests reflect the latest dependencies, maintaining project stability and compatibility.
```

CONTRIBUTING.md
tynandebold, line 318: `Kedro-Viz` and not `Kedro-viz`.

Might also put the file paths in code formatting.

```suggestion
If the lower-bound e2e test fails, it indicates that some dependencies may not work correctly with Kedro-Viz. To resolve this, update the problematic dependency in both `requirements.txt` and `lower-requirements.txt`.
```

tynandebold (approved):
Looks great. Approved with some nits in the new comment you added.

PR #1449, opened 2023-07-12
Fix/url-for-parameter

### LINKED ISSUE #1440: Ensure URL parameters and parameter nodes are in sync
## Description

If you try to paste this URL (http://localhost:4141/?pipeline_id=__default__&selected_id=72baf5c6) into a browser and open it, Kerdo-Viz will crash.

More context [here](https://github.com/kedro-org/kedro-viz/pull/1422#pullrequestreview-1516139380) and in subsequent comments. 

## Expected Result

Viz should open fine.

## Actual Result

The app doesn't load.

## Checklist

Huongg:
## Description
Fixes #1440

## Development notes

There is one failed test which I'm currently working on to address it. But I think the PR is ready for review in the meantime anyway. Let me know your thoughts on the changes or any other issue you can spot. 

- In `initial-state`, set disabled.parameter to false if the nodeId from URL is parameter type
- in `initial-state`, only set disable.parameter to true if the localStorage is empty, otherwise always read from the localStorage
- in `flowchart-wrapper`, extra check for selected node nested in a modular-pipeline, only expanding it if its not a parameter type

https://github.com/kedro-org/kedro-viz/assets/32060364/21c301cf-07a2-4572-b105-29fb2161e635

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1449"></a>


src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -119,7 +119,13 @@ export const FlowChartWrapper = ({
...
       const hasModularPipeline = modularPipeline?.length > 0;
 
-      if (hasModularPipeline) {
+      const isParameterType =
+        graph.nodes &&
+        graph.nodes.find(
+          (node) => node.id === nodeId && node.type === 'parameters'
+        );
+
+      if (hasModularPipeline && !isParameterType) {
```
tynandebold, line 128: Could you explain why this is needed again? You mentioned it earlier though now I've forgot.

src/store/initial-state.js
```diff
@@ -85,12 +89,25 @@ export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
     }
   }
 
+  // Set the nodeType.disable to false depends on what type of data it is, eg: parameters, or data, etc
```
tynandebold, line 92:
```suggestion
  // Set the nodeType.disable to false depending on what type of data it is, e.g. parameters, data, etc.
```

src/store/initial-state.js
```diff
@@ -85,12 +89,25 @@ export const preparePipelineState = (data, applyFixes, expandAllPipelines) => {
...
+  if (nodeTypes.includes(state.node.type[nodeIdFromUrl])) {
+    state.nodeType.disabled[state.node.type[nodeIdFromUrl]] = false;
+  }
+
+  // If there is a "selected_name" in the URL
+  // we need to ensure task and data are on so it can redirect back to the selected node
```
tynandebold, line 98:
```suggestion
  // If there is a "selected_name" in the URL we need to ensure task and
  // data tags are on so the app can redirect back to the selected node
```

src/store/initial-state.js
```diff
@@ -131,12 +148,18 @@ export const prepareNonPipelineState = (props) => {
  */
 const getInitialState = (props = {}) => {
   const nonPipelineState = prepareNonPipelineState(props);
-  saveLocalStorage(localStorageName, {
-    nodeType: {
-      // Default to disabled parameters and other types enabled
-      disabled: { parameters: true, task: false, data: false },
-    },
-  });
+
```
tynandebold, line 151:
```suggestion
```

src/store/initial-state.js
```diff
@@ -131,12 +148,18 @@ export const prepareNonPipelineState = (props) => {
...
+  const localStorageState = loadLocalStorage(localStorageName);
+
+  // If the localStorage is empty
+  // Default to disabled parameters and other types enabled
```
tynandebold, line 155:
```suggestion
  // If localStorage is empty, default to disabled parameters and enable the
  // other element types in the filters panel
```

tynandebold (approved):
This works well and solves the problem nicely. I'm approving with some nits about the comments and one question. Thank you!

Oh, and don't forget the failing test :)

src/components/flowchart-wrapper/flowchart-wrapper.js
Huongg: hey because both parameters and some nested nodes can both have a modular pipeline as a parent. therefore it needs to check that the node is not a parameter type before it tries to expand the modelarPipeline. 

I'll include a comment here to explain this 👍

rashidakanchwala (changes requested):
hey @Huongg , i am a little confused  about the what the initial-state.js (getInitialState) changes are for. 

Yesterday, you had suggested that we should not keep the resetting the filters to parameters: hidden, task: show, and data: show. This way, if a user chooses to display parameters and refreshes the browser, the settings will remain the same for parameters, data, and tasks.

If that's what we're aiming for, I believe there's a simpler way to achieve it. Happy to get on a call on Monday and discuss :)

src/store/initial-state.js
```diff
@@ -1,9 +1,11 @@
 import deepmerge from 'deepmerge';
-import { loadLocalStorage, saveLocalStorage } from './helpers';
+import { loadLocalStorage } from './helpers';
 import normalizeData from './normalize-data';
 import { getFlagsFromUrl, Flags } from '../utils/flags';
 import { settings, sidebarWidth, localStorageName, params } from '../config';
 
+const nodeTypes = ['parameters', 'task', 'data'];
```
rashidakanchwala, line 7: does this need to be global. Can it be local to the func.

rashidakanchwala (approved):
Just one small comment. Looks good. thanks @Huongg

RELEASE.md
```diff
@@ -6,6 +6,12 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+# Release x.x.x
+
+## Bug fixes and other changes
+
+- Ensure URL parameters and parameter nodes are in sync. (#1449)
```
rashidakanchwala, line 13: should we add another release note that says something like 

Ensure filter states for task, data, and parameter nodes as set by the user, even after browser refresh.

src/store/initial-state.js
Huongg: ah yeah i think we can have it inside a func 👍

PR #1465, opened 2023-07-27
[DON'T MERGE] PoC of recording stats during kedro run

noklam:
## Description

#662

If you run

```py
cd demo_project
kedro run
```

You should see a `stats.json` created.
## Development notes
Pros:
- No extra cost to read data when users fire up viz to see the statistics.
- Need better way to manage the version, or maybe it is ok because the current pipeline view also only display the latest pipeline. If viz saw this file then it can optionally skip the read.

Cons:
- It can only show the stats for the last run, if it's the first time user fire up Viz it won't work.
- Requires an upgrade of kedro-viz & a new kedro run to generate this stats.json

Notes:
It's a PoC, if we actually want to implement this we need to be more careful about the type checking, maybe we can check the type of the `dataset` instead of `data`. We may also need to take care of the remote storage (or it could be a next step)

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1465"></a>

### DIFF
--- a/demo-project/src/demo_project/settings.py
+++ b/demo-project/src/demo_project/settings.py
@@ -4,14 +4,49 @@
 
 from pathlib import Path
 
+from traitlets import default
+
 # Define where to store data from a KedroSession. Defaults to BaseSessionStore.
 # from kedro.framework.session.store import ShelveStore
 from kedro_viz.integrations.kedro.sqlite_store import SQLiteStore
 
+from kedro.framework.hooks import hook_impl
+from collections import defaultdict
+
+
+class DatasetStatsHook:
+    def __init__(self):
+        self._stats = defaultdict(dict)
+
+    @hook_impl
+    def after_context_created(self, context):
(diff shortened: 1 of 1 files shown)


demo-project/src/demo_project/settings.py
```diff
@@ -4,14 +4,49 @@
...
+    def __init__(self):
+        self._stats = defaultdict(dict)
+
+    @hook_impl
+    def after_context_created(self, context):
+        self._catalog = context.catalog
+
+    @hook_impl
+    def after_dataset_loaded(self, dataset_name, data):
+        import pandas as pd
```
datajoely, line 27: Should this be in a `try/except` if the user doesn't have pandas installed?

noklam: This is written in 20 minutes gap between meeting, definitely not ready to be merged 😂

noklam:
Closed in favor of #1472

PR #1472, opened 2023-08-02
Visualize Dataset statistics in metadata panel

### LINKED ISSUE #662: Visualize size of processed datasets

ravi-kumar-pilla:
## Description

Resolves #662 

## Development notes

Dataset_stats_compressed

**Screenshots:**

**All stats -**

Screenshot 2023-08-02 at 8 18 07 AM

**File size only -**

Screenshot 2023-08-02 at 8 22 02 AM

**File size not configured due to no file path -**

Screenshot 2023-08-02 at 8 23 24 AM

**Overflow Dataset stats -**

Screenshot 2023-08-02 at 8 25 01 AM

## QA notes

**Steps to QA :**
1. Checkout feature/viz-size-datasets
2. Pip install -e package
3. cd demo-project
4. Execute kedro run
5. A file stats.json will be created in the demo-project along with changes in the data folder due to new kedro run
6. cd .. (go into kedro-viz folder)
7. Start the backend server executing make run
8. Start frontend by executing npm start
9. Click on datasets (covered only instances of pd.Dataframe). You should see the file statistics (rows, columns and file size). If file size is not available, you will see it as NA

**Note:**

1. There needs to be further discussion on where to extract file size as it is not available at the hook level (or I do not know how to get file path at hook level). As of now, I am using fsspec inside the flowchart model to get the file size. 
2. Transcoded data name is stored as ingestion.int_typed_shuttles@pandas2

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1472"></a>


amandakys:
For the overflow case, I thought 29000 rows would be displayed as 29K? Or is that just a forced example to show the overflow behaviour? 

Would you be able to provide a screenshot of the skeleton loader in action?

ravi-kumar-pilla:
> For the overflow case, I thought 29000 rows would be displayed as 29K? Or is that just a forced example to show the overflow behaviour?

I wanted to show you how the overflow looks. If you want it to be in a single row, I will implement the row and column shortforms (like 29K). During our call we thought this will not show exact number (unless we have tooltip or something like that). If the 2 rows are not bad, we can have it this way for now.

> Would you be able to provide a screenshot of the skeleton loader in action?

Skeleton loader is not needed for this implementation as the dataset statistics are pre-calculated. However, we might need it as a general improvement for the api call which is not related to this specific ticket. I will create a new ticket to track the skeleton loader implementation.

merelcht (commented):
I've done a first review and left initial comments. It's a lot of changes, so I'll do a second review asap!

In terms of using hooks for getting the file size. You could potentially do something like: 
```
@hook_impl
    def after_catalog_created(self, catalog: DataCatalog):
        datasets = catalog.datasets
        for ds_name, ds in datasets.items():
            try:
                fp = ds._filepath
            except Exception:
                logger.error("Something went wrong trying to get the filepath")
```

This is super rough and I haven't tried it properly, but it's a way to get the filepath for each dataset in the catalog after the catalog has been created.

src/components/metadata/styles/metadata.scss
```diff
@@ -315,3 +315,12 @@ $list-inline-spacing: 0.2em;
   bottom: 0;
   height: 40px;
 }
+
+.pipeline-metadata-label__profiler {
+  margin-left: 8px;
+  margin-right: 24px;
```
tynandebold, line 321: Nit:

```suggestion
  margin: 0 24px 0 8px;
```

tynandebold (approved):
I'll approve this PR from the JS side and will let Merel, Nok, and others weight in on the Python part.

I've left a couple of smaller comments that I hope you can get to before merging.

Well done! Looking forward to this being released.

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -54,11 +54,27 @@ def _bootstrap(project_path: Path):
         return
 
 
+def get_dataset_stats(project_path: Path):
+    """Return the stats saved at stats.json"""
+    import json as json_lib
```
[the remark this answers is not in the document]
ravi-kumar-pilla: I saw this type of imports in other part of the code and thought to separate Kedro json dataset from json we need to import inside a function scope. I have moved it to top level now

package/kedro_viz/models/flowchart.py
```diff
@@ -557,12 +558,28 @@ class DataNodeMetadata(GraphNodeMetadata):
...
+    profiler: Optional[Dict] = field(init=False, default=None)
+
     # TODO: improve this scheme.
-    def __post_init__(self, data_node: DataNode):
+    def __post_init__(self, data_node: DataNode, dataset_stats: Dict):
         self.type = data_node.dataset_type
         dataset = cast(AbstractDataset, data_node.kedro_obj)
         dataset_description = dataset._describe()
         self.filepath = _parse_filepath(dataset_description)
+        self.profiler = dataset_stats
```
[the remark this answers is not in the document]
ravi-kumar-pilla: It was initially termed as profiler. I have reverted back to stats and now it is consistent with the package.

ravi-kumar-pilla:
Thank you @merelcht , @tynandebold for the suggestions. I have made changes as suggested. Please review when you have some time.

NeroOkwa (approved):

src/components/metadata/metadata.js
```diff
@@ -107,6 +109,27 @@ const MetaData = ({
       : value?.split('.').pop();
   };
 
+  const statsContent = (statValue, statLabel) => {
```
vladimir-mck, line 112: Optional: move this to its own component so this function component doesn't get created on every time it metadata components changed or renders.

vladimir-mck (approved):

stephkaiser:
Hey @ravi-kumar-pilla - Amanda and I had a catch up about this yesterday and not sure if you both have discussed since then. If not, these were some thoughts from a design perspective:

- I'd push for showing the exact number, rather than rounding up/down (eg. show 29,365 rather than 29k) - If we go back to the original user pain point we are trying to solve here - its to help users identify where to start debugging when a run has failed, by allowing them to inspect dataset metrics after a run. So in my opinion, showing the exact number here is quite important for that debugging use case. I'd also avoid hiding the exact number behind a tooltip since the user already has to click on the node and open the metadata panel in order to view the metrics, adding an additional barrier to viewing that exact number isn't great. 

- In this case, the 3 metrics probably won't fit nicely on a single row and we'd have to move to 3 rows instead - we can provide the designs for you. When values are unavailable to be displayed, we can keep them on a single row showing N/A, so not to take up unused empty space in the panel. 

- Lastly, a very minor thing but could we show large values with a comma (29365 as 29,365) and NA as N/A? 

Let me know if you have any questions :)

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -0,0 +1,106 @@
+# pylint: disable=broad-exception-caught
+# pylint: disable=protected-access
+"""`kedro_viz.integrations.kedro.hooks` defines hooks to add additional
+functionalities for a kedro run."""
+
+import json as json_lib
```
noklam, line 6: Can we keep this as `json`?

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -0,0 +1,106 @@
...
+
+        Args:
+            catalog: The catalog that was created.
+        """
+        try:
+            datasets = catalog._data_sets
+
+            for dataset_name, dataset in datasets.items():
+                if not isinstance(dataset, MemoryDataset):
+                    file_path = dataset._filepath  # noqa: no-member
```
noklam, line 50: This is quite fragile, `_filepath` is not a API of dataset. In addition do we track DataFrame only or we are tracking more dataset?

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -0,0 +1,106 @@
...
+    def after_pipeline_run(self):
+        """Hook to be invoked after a pipeline runs.
+        Once the pipeline run completes, write the dataset
+        statistics to stats.json file
+
+        """
+        try:
+            with open("stats.json", "w", encoding="utf8") as file:
+                sorted_stats_data = {
+                    dataset_name: stats_order(stats)
```
noklam, line 97: Is it necessary to order the JSON file?

package/kedro_viz/models/flowchart.py
```diff
@@ -13,7 +13,7 @@
 from kedro.pipeline.node import Node as KedroNode
 from kedro.pipeline.pipeline import TRANSCODING_SEPARATOR, _strip_transcoding
 
-from .utils import get_dataset_type
+from .utils import get_dataset_type, get_file_size
```
noklam, line 16: Can we use absolute import?

package/kedro_viz/models/utils.py
```diff
@@ -39,3 +45,45 @@ def get_dataset_type(dataset: "AbstractDataset") -> str:
...
+def get_file_size(file_path: Union[str, None]) -> int:
+    """Get the dataset file size using fsspec. If the file_path is a directory,
+    get the latest file created (this corresponds to the latest run)
+
+    Args:
+        file_path: The file path for the dataset
+    """
+    try:
+        if not file_path:
+            return 0
```
noklam, line 59: Should it be None instead? How would front end handle this?

noklam:
Btw, I think this PR is too big to be reviewed. Is it possible to breakdown to 2-3 smaller PRs?

1. Implementation of Hooks and Plugins that produce the JSON
2. Backend Python viz code that consume the JSON
3. Front end JS code + Design

2 & 3 may be needed to be reviewed together ( I don't know if it's possible to be reviewed separately), as long as the interface of JSON is set, it should be easy to breakdown the PR into smaller PRs.

tynandebold:
> Btw, I think this PR is too big to be reviewed. Is it possible to breakdown to 2-3 smaller PRs?
> 
> 1. Implementation of Hooks and Plugins that produce the JSON
> 2. Backend Python viz code that consume the JSON
> 3. Front end JS code + Design
> 
> 2 & 3 may be needed to be reviewed together ( I don't know if it's possible to be reviewed separately), as long as the interface of JSON is set, it should be easy to breakdown the PR into smaller PRs.

@noklam can we push through this one and get it reviewed with a promise to break something like this out better in the future?

ravi-kumar-pilla:
> Hey @ravi-kumar-pilla - Amanda and I had a catch up about this yesterday and not sure if you both have discussed since then. If not, these were some thoughts from a design perspective:
> 
> * I'd push for showing the exact number, rather than rounding up/down (eg. show 29,365 rather than 29k) - If we go back to the original user pain point we are trying to solve here - its to help users identify where to start debugging when a run has failed, by allowing them to inspect dataset metrics after a run. So in my opinion, showing the exact number here is quite important for that debugging use case. I'd also avoid hiding the exact number behind a tooltip since the user already has to click on the node and open the metadata panel in order to view the metrics, adding an additional barrier to viewing that exact number isn't great.

Yes, exact numbers will be shown

> * In this case, the 3 metrics probably won't fit nicely on a single row and we'd have to move to 3 rows instead - we can provide the designs for you. When values are unavailable to be displayed, we can keep them on a single row showing N/A, so not to take up unused empty space in the panel.

Overflow example -
image

N/A example -
image



> * Lastly, a very minor thing but could we show large values with a comma (29365 as 29,365) and NA as N/A?

Done !

> Let me know if you have any questions :)

When you say **values are unavailable**, do you mean for every dataset we should show - rows, columns, file size as default labels and values as N/A if we do not have the information (stats row something like - **N/A rows N/A columns N/A file size**) ? Or for example if we only have file size, can we show just the file size label (stats row something like **N/A file size**)

package/kedro_viz/integrations/kedro/hooks.py
ravi-kumar-pilla: I think for now we will track only DataFrame

package/kedro_viz/integrations/kedro/hooks.py
ravi-kumar-pilla: It will help when reading the json in the frontend. Also, will be helpful to structure the data

package/kedro_viz/models/utils.py
ravi-kumar-pilla: In the frontend, if the file size is returned as 0 bytes, we will be showing N/A

demo-project/stats.json
```diff
@@ -0,0 +1,64 @@
...
+  "train_evaluation.linear_regression.r2_score": { "file_size": 37 },
+  "train_evaluation.random_forest.r2_score": { "file_size": 36 },
+  "train_evaluation.linear_regression.experiment_params": { "file_size": 102 },
+  "train_evaluation.random_forest.experiment_params": { "file_size": 338 },
+  "params:ingestion.typing.reviews.columns_as_floats": {
+    "rows": 1,
+    "file_size": 88
+  },
+  "ingestion.prm_agg_companies": { "rows": 50098, "columns": 5 },
+  "params:feature_engineering.feature.derived": { "rows": 3, "file_size": 88 },
```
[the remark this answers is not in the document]
ravi-kumar-pilla: I think this needs further discussion, on how to show parameter info. I added the rows as size of the list or dict. But, I think it does not make sense when it is termed as rows. Removing this for the POC

package/kedro_viz/data_access/managers.py
```diff
@@ -91,6 +92,25 @@ def add_pipelines(self, pipelines: Dict[str, KedroPipeline]):
...
+        self.dataset_stats = stats_dict
+
+    def get_dataset_stats(self, data_node: Union[DataNode, TranscodedDataNode]) -> Dict:
+        """Returns the dataset statistics for the data node if found else returns None
+
+        Args:
+            The data node for which we need the statistics
+        """
+
+        return self.dataset_stats.get(data_node.name, {})
```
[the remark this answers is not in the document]
ravi-kumar-pilla: It should be an empty dict. This method is called in the router.py where I was checking for None and initializing the stats to empty dict. Instead this way, we always get a dict for a data node

package/kedro_viz/integrations/kedro/hooks.py
ravi-kumar-pilla: in some files we have json from Kedro datasets. To distinguish the library import from kedro datasets, I used this convention through out.

package/kedro_viz/integrations/kedro/hooks.py
ravi-kumar-pilla: Yes even Circle CI build fails as the `AbstractDataset` does not contain _filepath field. Do you think it is better to read the file size in `DataNodeMetadata` class ? I am already doing that - please check line 571-573 in `package/kedro_viz/models/flowchart.py`

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -54,11 +58,35 @@ def _bootstrap(project_path: Path):
...
+        if not stats_file_path.exists():
+            return {}
+
+        with open(stats_file_path, encoding="utf8") as stats_file:
+            stats = json_lib.load(stats_file)
+            return stats
+
+    except Exception as exc:  # pylint: disable=broad-exception-caught
+        logger.warning(
+            "Unable to get dataset stats from project path %s : %s", project_path, exc
```
merelcht, line 80: Since this error can appear in logs for users I think it's better to not use abbreviations. 

```suggestion
            "Unable to get dataset statistics from project path %s : %s", project_path, exc
```

package/kedro_viz/integrations/kedro/hooks.py
merelcht: If we have json from a dataset doesn't it make more sense to call that `json_dataset` and use `json` here since this is the actual json library we're calling?

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        try:
+            stats_dataset_name = get_stats_dataset_name(dataset_name)
+            if isinstance(data, pd.DataFrame):
+                self._stats[stats_dataset_name]["rows"] = int(data.shape[0])
+                self._stats[stats_dataset_name]["columns"] = int(data.shape[1])
+
+        except Exception as exc:  # pragma: no cover
+            logger.warning(
+                "Error creating the stats for the dataset %s : %s", dataset_name, exc
```
merelcht, line 45:
```suggestion
                "Error creating the statistics for the dataset %s : %s", dataset_name, exc
```

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -0,0 +1,67 @@
...
+        try:
+            with open("stats.json", "w", encoding="utf8") as file:
+                sorted_stats_data = {
+                    dataset_name: stats_order(stats)
+                    for dataset_name, stats in self._stats.items()
+                }
+                json_lib.dump(sorted_stats_data, file)
+
+        except Exception as exc:  # pragma: no cover
+            logger.warning("Error writing the stats for the pipeline: %s", exc)
```
merelcht, line 64:
```suggestion
            logger.warning("Error writing the dataset statistics for the pipeline: %s", exc)
```

package/tests/test_integrations/test_hooks.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+    # Test for Transcoded data node
+    transcoded_dataset_name = "raw_data@pandas1"
+    assert (
+        dataset_stats_hook.after_dataset_loaded(
+            transcoded_dataset_name, example_data_frame
+        )
+        is None
+    )
+
```
merelcht, line 20: What are you testing here?

package/tests/test_models/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    from kedro.io import MemoryDataset
+except ImportError:
+    # older versions
+    # older versions
+    from kedro.io import MemoryDataSet as MemoryDataset
+
+
+def test_get_dataset_type(dataset=None):
+    assert get_dataset_type(dataset) == ""
+    assert get_dataset_type(MemoryDataset()) == "io.memory_dataset.MemoryDataset"
```
merelcht, line 14: What is this test for?

package/tests/test_models/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    mock_file.write_text("This is a mock file content.")
+
+    assert get_file_size(mock_file) == 28
+
+    mock_file = tmp_path / "mock_file.txt"
+    mock_file.write_text("This is a mock file content.")
+    assert get_file_size(tmp_path) == 0
+
+    # Testing no file found exception
+    assert get_file_size("raw.txt") == 0
```
merelcht, line 31: In python you would either write four different tests to check all these cases, or you can parametrise your test: https://docs.pytest.org/en/7.3.x/how-to/parametrize.html

merelcht (changes requested):
The code looks pretty good to me, but the python tests for the hooks and to check the file size should be changed. I don't think the ones for hooks are checking anything useful and the ones for testing the file size should be refactored to be more pythonic. Have a look at https://docs.pytest.org/en/7.3.x/how-to/parametrize.html. There's also plenty of good test examples in the kedro framework repo 🙂

stephkaiser:
> When you say values are unavailable, do you mean for every dataset we should show - rows, columns, file size as default labels and values as N/A if we do not have the information (stats row something like - N/A rows N/A columns N/A file size) ? Or for example if we only have file size, can we show just the file size label (stats row something like N/A file size)

Yes exactly, I mean the first option you mentioned - where we show all labels even if we don't have the information. Example:
N/A rows N/A columns N/A file size (on a single row) 
N/A rows N/A columns 156MB file size (on a single row)

demo-project/stats.json
```diff
@@ -0,0 +1,64 @@
...
+  "companies": { "rows": 77096, "columns": 5, "file_size": 1810602 },
+  "reviews": { "rows": 77096, "columns": 10, "file_size": 2937144 },
+  "shuttles": { "rows": 77096, "columns": 13, "file_size": 4195290 },
+  "ingestion.int_typed_companies": {
+    "rows": 77096,
+    "columns": 5,
+    "file_size": 550616
+  },
+  "ingestion.int_typed_shuttles@pandas1": { "file_size": 1235685 },
+  "ingestion.int_typed_shuttles@pandas2": {
```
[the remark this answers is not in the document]
ravi-kumar-pilla: stripped the transcoding now

package/tests/test_integrations/test_hooks.py
ravi-kumar-pilla: This was primarily to get the code coverage and it covers the utility function where we strip transcoded name. As you said the test cases are really not adding any value here. I need to rewrite them.

package/kedro_viz/integrations/kedro/hooks.py
noklam: May need to catch up on this later. Top of my head I think we should limit the scope to just DataFrame and don't try to address the most generic case now. `_filepath` is a problem because a `SQLTableDataSet` is a pandas dataframe but it doesn't have a filepath (worse if it has one it is the path of the `.sql` file instead of the dataframe).

amandakys:
@ravi-kumar-pilla 

to your questions: 
in the N/A case - then no stats are available
image
in the N/A case - when 1+ stats are unavailable 
image
overflow state: when it won't all fit in one row, split into 3 rows 
@stephkaiser and i are still discussing this one, but i didn't want to block you :) will update when we've decided

package/kedro_viz/models/utils.py
noklam: Could it be return `None` here? then in the frontend use if not filesize since 0 or None would both be treated as `False`

package/kedro_viz/integrations/kedro/hooks.py
ravi-kumar-pilla: okay, if that is the case, getting the file size without checking for extension here `(line 569 in package/kedro_viz/models/flowchart.py)` will be a problem ? So should we check for file extensions ?

package/tests/test_models/test_utils.py
ravi-kumar-pilla: Testing utility function get_dataset_type which returns the dataset type as a string based on the dataset input

ravi-kumar-pilla:
Hi @merelcht , I have modified the test cases as suggested. 

@noklam I need some help on the circle ci e2e tests failing. It is failing at `line 166 in kedro_viz/integrations/kedro/data_loader.py file` though I haven't modified anything related to this. Error - 

`ModuleNotFoundError: No module named 'matplotlib'
    Then kedro-viz should start successfully                            # features/steps/cli_steps.py:128 30.047s
        File "/home/circleci/.pyenv/versions/3.10.12/lib/python3.10/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/home/circleci/.pyenv/versions/3.10.12/lib/python3.10/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "features/steps/cli_steps.py", line 144, in check_kedroviz_up
          assert context.result.poll() is None
      AssertionError`

ravi-kumar-pilla:
> @ravi-kumar-pilla
> 
> to your questions: in the N/A case - when no stats are available image in the N/A case - when 1+ stats are unavailable image overflow state: when it won't all fit in one row, split into 3 rows image

Modified the design to account for overflow - @amandakys @stephkaiser 

image

Normal -
image

All N/A -
image

Single N/A -
image

ravi-kumar-pilla:
Hi @tynandebold , @vladimir-mck 
I have modified the frontend code to account for design changes due to overflow. Please review

Thank you

merelcht (approved):
Thanks @ravi-kumar-pilla , the python tests & code look good now! 👍 ⭐

tynandebold:
> Hi @tynandebold , @vladimir-mck I have modified the frontend code to account for design changes due to overflow. Please review
> 
> Thank you

Where can we view that in action? I clicked on several nodes but didn't see anything overflowing.

ravi-kumar-pilla:
> > Hi @tynandebold , @vladimir-mck I have modified the frontend code to account for design changes due to overflow. Please review
> > Thank you
> 
> Where can we view that in action? I clicked on several nodes but didn't see anything overflowing.

You can check for datasets - ` int typed companies, Experiment params, Regressor, R2 score, Regressor, Model Input Table, Prm Spine Table`

package/kedro_viz/integrations/kedro/hooks.py
noklam: I found the name `stats_order` slightly confusing. Maybe `format_json` or `prettify_json`?

noklam: And can it be just a helper method stay within the class of the `Hook`? I don't see why we want to put it in `utils.py` since this is very specific to format the JSON produce by the Hook itself.

noklam: Same apply to the other functions. Can we not have the `utils.py`?

package/kedro_viz/models/utils.py
```diff
@@ -39,3 +45,45 @@ def get_dataset_type(dataset: "AbstractDataset") -> str:
     abbreviated_module_name = ".".join(dataset.__class__.__module__.split(".")[-2:])
     class_name = f"{dataset.__class__.__qualname__}"
     return f"{abbreviated_module_name}.{class_name}"
+
+
+def get_file_size(file_path: Union[str, None]) -> Union[int, None]:
+    """Get the dataset file size using fsspec. If the file_path is a directory,
+    get the latest file created (this corresponds to the latest run)
+
```
noklam, line 53: Why can't we calculate this from the hook? It should be much simpler to do this if you have the `Dataset` object. That way you don't need to re-create all the `fsspec` logic.

ravi-kumar-pilla: I tried doing that using _filepath in after_context_created hook. But since AbstractDataset does not have an api supporting that, you mentioned it to be fragile and so thought it would be a good idea to move it here

ravi-kumar-pilla:
Hey Nok,

For your question - Same apply to the other functions. Can we not have the utils.py?

So I want the hooks.py to contain only hook impl and any helper functions to be moved to utils. This way in future if we need more helper functions, we can manage code better.

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -0,0 +1,68 @@
+# pylint: disable=broad-exception-caught
+"""`kedro_viz.integrations.kedro.hooks` defines hooks to add additional
+functionalities for a kedro run."""
+
+import json
+import logging
+from collections import defaultdict
+from typing import Any
+
+import pandas as pd
```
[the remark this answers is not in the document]
ravi-kumar-pilla: yea makes sense, so we need to inform user that to have this feature, pandas need to be installed. @tynandebold

ravi-kumar-pilla:
@noklam I have modified the hooks to include file_size. Though I am calculating the file_size only if the dataset instance is pd.Dataframe. Earlier, I calculated it for every datanode and transcoded datanode which contains file_path. Please have a look and let me know if there needs to be any changes. Thank you !

noklam (approved):
Approved with non-blocking comments

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -0,0 +1,174 @@
...
+            if isinstance(data, pd.DataFrame):
+                self._stats[stats_dataset_name]["rows"] = int(data.shape[0])
+                self._stats[stats_dataset_name]["columns"] = int(data.shape[1])
+
+                current_dataset = self.datasets.get(dataset_name, None)
+
+                if current_dataset:
+                    self._stats[stats_dataset_name]["file_size"] = self.get_file_size(
+                        current_dataset
+                    )
```
oulianov, line 105: Good start ! However it only works with dataframes.

There are 2 additional cases that can be easily implemented :
- In the case of a PartitionedDataSet that loads pd.DataFrame s
- In the case of loading multiple sheets of an Excel file (load_args: sheet_name: None )

I think it's not too difficult to account for those cases by adding additional logic on data, and it would avoid just throwing N/A in the UI. 

Do you see how to implement that ?

lukaszdz:
I'm not sure that this fully resolves the request (https://github.com/kedro-org/kedro-viz/issues/662), though please correct me in case I didn't fully understand the update.  Added more details: https://github.com/kedro-org/kedro-viz/issues/662#issuecomment-1699317699

ravi-kumar-pilla:
> I'm not sure that this fully resolves the request (#662), Ideally, we would want to see the dataset sizes directly in the graph view, so we can view any issues with the pipeline without having to click through each node in the graph. Even better if we had some way to set up some rules to color the nodes (if N=0, then color the node red).
> 
> Please correct me in case I didn't fully understand the update. Added more details: [#662 (comment)](https://github.com/kedro-org/kedro-viz/issues/662#issuecomment-1699317699)

Hi @lukaszdz , we are planning on integrating the statistics into the graph which would resolve this issue completely. We are exploring options on how to approach this without cluttering the flowchart view. This is the first step towards visualizing the statistics in Kedro Viz. Thank you

package/kedro_viz/integrations/kedro/hooks.py
oulianov: Nice ! thank you

PR #1485, opened 2023-08-14
Fix for sidebar search result based on Pretty name flag

jitu5:
## Description

This resolves case below for the Pretty name feature [1252](https://github.com/kedro-org/kedro-viz/issues/1252) 

When the feature is turned on, the sidebar search will only allow the user to search for pretty-named things. For example, if one searches for **data_processing.preprocessed_shuttles** nothing will be returned, even though **preprocessed shuttles** exists and is the same thing. The same is true in the opposite direction.
 

## Development notes

Now it works for both the type of search terms Pretty name and non Pretty name if it exist.

Screenshot 2023-08-14 at 7 21 26 p m
Screenshot 2023-08-14 at 7 22 07 p m

## Checklist

- [X] Added new entries to the `RELEASE.md` file

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1485"></a>


src/selectors/nodes.js
```diff
@@ -113,6 +113,15 @@ export const getNodeLabel = createSelector(
     isPrettyName ? nodeName : nodeFullName
 );
 
+/**
```
ravi-kumar-pilla, line 116: Can we use getNodeName and getNodeFullName selectors for searching instead of creating a new selector ?

src/selectors/modular-pipelines.js
```diff
@@ -44,7 +44,12 @@ export const searchTree = (
       // if the child node is a leaf, simply search the leaf's name
       // and add to the search result if there is a match.
       const found = searchString(childNode.data.name, searchValue);
-      if (found) {
+      const foundOpposite = searchString(
+        childNode.data.oppositeForPrettyName,
+        searchValue
+      );
+
+      if (found || foundOpposite) {
```
ravi-kumar-pilla, line 52: This works in searching for both pretty and non-pretty names but the node list displayed is changed.

Have a look at the gif - The display names in the list gets changed based on search value i.e., I disabled pretty name flag and searched for a pretty name "Feat" and the node list shows Feature Engineering instead of feature_engineering. Though the search results return a value, we should push names based on the pretty flag I guess

Display_Name changes

ravi-kumar-pilla (changes requested):
I have added few comments. Please have a look. Thank you

src/selectors/modular-pipelines.js
jitu5: I checked current version on https://demo.kedro.org and this is happening before my changes. We can create a separate ticket for this if this behaviour is not as expected.

src/selectors/nodes.js
ravi-kumar-pilla: I think we have 2 selectors at line 13. 14 in selectors -> nodes.js 

```
export const getNodeName = (state) => state.node.name;
export const getNodeFullName = (state) => state.node.fullName)
```
These do not depend on pretty flag. Can you please check once.

src/selectors/modular-pipelines.js
```diff
@@ -44,7 +44,12 @@ export const searchTree = (
       // if the child node is a leaf, simply search the leaf's name
       // and add to the search result if there is a match.
       const found = searchString(childNode.data.name, searchValue);
-      if (found) {
+      const foundOpposite = searchString(
```
ravi-kumar-pilla, line 47: [Nit]  I feel opposite terminology not fitting but at the same time, I cannot come up with a suitable name here. @tynandebold any suggestions ?

ravi-kumar-pilla (approved):
I have left few comments and none of them is a blocking change, so approving it now

src/selectors/modular-pipelines.js
tynandebold: No, that isn't expected. @jitu5 is right though, this isn't due to his change.

In the screen capture, what's returned when you search for `feat` when `prettyName` is **off** should be `feature_engineering` and not `Feature Engineering`. 

Jitendra, can you please create a separate ticket to fix that?

tynandebold:
Looking really good so far! The results are returning the way they should, and that's the main point of the PR, so well done there.

I am noticing one style bit that's off. On your branch, the search results aren't always highlighted like they should be. Notice the results aren't highlighted here:

image

It should look something like this: 

image

This could be tricky to solve though, because what's being highlighted is the same exact text string that the user has entered. Have a look though and see if you can get it.

jitu5:
@tynandebold Now searched/matched text is being highlighted in both the cases. please let me know if anythings is missing or broken.

RELEASE.md
```diff
@@ -18,6 +18,7 @@ Please follow the established format:
 - Fix incorrect rendering of datasets in modular pipelines. (#1439)
 - Fix broken SVG/PNG exports in light theme. (#1463)
 - Fix dataset and global toolbar error with standalone React component (#1351)
+- Fix Sidebar search result based on Pretty name setting (#1252)
```
tynandebold, line 21: These should be PR numbers.

```suggestion
- Fix Sidebar search result based on Pretty name setting (#1485)
```

src/selectors/modular-pipelines.js
tynandebold: I'm fine with it.

tynandebold (approved):
Thanks for making that style change. It looks good to me now, and I'll approve.

I do think there's a little bit of odd UX wherein there can be a mismatch between what's returned in the sidebar and what's show in the flowchart. E.g. if `prettyName` is off and you search for `Create mat` you'll get back `Create Matplot Chart`, but in the flowchart the text of that node is `create_matplotlib_chart`.

I think this is ok for now. Let's see if anyone complains. The benefit is now you can search for both pretty and normal names. Nicely done!

jitu5:
@tynandebold For this E.g. if prettyName is off and you search for `Create mat ` you'll get back `Create Matplot Chart`, but in the flowchart the text of that node is `create_matplotlib_chart`.

Before the changes I did just now, it was returning correct text same as flowchart but it want highlighting because the enter text is pretty name (`Create mat `) and returned text is non pretty name (`create_matplotlib_chart`). 
But as you said we will see if anyone complains.

Issue #1460, opened 2023-07-23
Replace semver with packaging 


rashidakanchwala:
## Description

Should we replace semver.Version with packaging.version.Version as PEP 440 versions are not exactly compatible with SemVer ?

https://peps.python.org/pep-0440/#semantic-versioning, https://packaging.pypa.io/en/stable/version.html

@astrojuanlu  to provide more info. 

## Checklist

- [x] Include labels so that we can categorise your feature request

astrojuanlu:
More info on the specific features of Python versioning https://sethmlarson.dev/pep-440?date=2023-08-03

rashidakanchwala:
Opening this because we are using packaging now in backend so we might as well replace semver with packaging
cc - @noklam

PR #1498, opened 2023-08-21
Shareable Kedro-viz backend implementation

### LINKED ISSUE #1520: Shareable Kedro-viz backend implementation
## Description

This ticket is related to parent ticket #1116 

Scope of this ticket includes 
- creating an endpoint that packages Kedro-viz in a static html app and deploys it on AWS 
- writing related tests 
- manual testing in multiple environments for robustness. 
- writing backend setup docs 

## Checklist

rashidakanchwala:
## Description

Resolves #1520 

This PR involves the backend implementation of Shareable URLs on Kedro-viz. In this PR, we have used fsspec to facilitate the uploading of the Kedro-viz build and API responses (in JSON format) to an S3 storage location. Kedro-viz, when deployed in a hosted environment, functions as a serverless application, serving static HTML content and retrieving Kedro project data from the API JSON files.

This PR solves two many issues 

- Hosting Kedro-viz.
- Fixing the bug with the kedro viz --save-file and --load-file methods. These methods were previously not saving and loading the metadata associated with each node as well as registered pipelines flowchart. They were only saving flowchart for __default__ pipeline

#### Router.py:

This file defines two routers in this PR:
- api/deploy: A POST request that sends the bucket_name and region from the frontend (FE) and forwards this information to S3Deployer.
- api/package_compatibilities: A GET request sent by the FE to the BE to check if the fsspec version is greater than or equal to 2023.9.0, which is a requirement for Shareable Kedro-viz. If the user has an older version, the FE will prompt the user to update the dependency to the correct version before enabling this feature.

#### apps.py:

The main change in this file is the `create_api_app_from_file` function, which is invoked when the user runs kedro-viz --load-file xyz. Now, this function loads everything, including metadata for nodes and other registered pipeline information.

#### responses.py:

This file has undergone some refactoring, and one of the main changes is that it contains functions that enable the saving of backend responses to a specified filepath (either local or remote).

#### integrations/deployment/s3-deployer:
This class, S3Deployer, is the first of many deployer classes. As we expand to more deployer classes, we will introduce a baseDeployer class. The S3Deployer class is self-explanatory and handles the task of uploading content to S3.

##### server.py:
In this file, we make a change to the kedro-viz --save-file method. It now saves everything, including the flowchart, metadata, and other pipeline information, to the specified filepath.

## QA notes

If you have an s3 bucket or a Minio. You can simply do a post request using curl or something else to upload your kedro-viz to s3.

```bash

curl -X POST http://localhost:4142/api/deploy \
-H "Content-Type: application/json" \
-d '{"region": "us-east-1", "bucket_name": "s3://kedroviz2"}'
```

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1498"></a>


package/kedro_viz/api/rest/responses.py
```diff
@@ -19,6 +28,11 @@ class Config:
         orm_mode = True
 
 
+class UserCredentials(BaseModel):
```
ravi-kumar-pilla, line 31: can we have this more specific like AWSCredentials (since we might have different deployers)

package/kedro_viz/utils/api_tools.py
```diff
@@ -0,0 +1,79 @@
...
+    get_default_response,
+    get_node_metadata_response,
+    get_selected_pipeline_response,
+)
+from kedro_viz.data_access import data_access_manager
+
+logger = logging.getLogger(__name__)
+
+
+def save_api_responses_to_fs(filepath: str):
```
ravi-kumar-pilla, line 18: can we have save_api_responses_to_fs inside -> `package/kedro_viz/api/rest/responses.py ` as this is also creating some kind of response ?

package/kedro_viz/api/rest/router.py
```diff
@@ -38,53 +38,48 @@ async def main():
...
-        layers=data_access_manager.get_sorted_layers_for_registered_pipeline(
-            registered_pipeline_id
-        ),
-        pipelines=data_access_manager.registered_pipelines.as_list(),
-        selected_pipeline=registered_pipeline_id,
-        modular_pipelines=modular_pipelines_tree,  # type: ignore
-    )
+
+@router.get(
+    "/package_compatibilities",
```
ravi-kumar-pilla, line 72: In general, REST API routes should not contain special characters, the only exception being hyphen. It would be better if we call this as /package-compatibilities. 

However, I feel we should tackle this route differently if we check for compatibility in the backend. Something like a [POST] /feature-compatibility route which accepts a json body like 

```json
   {
       "feature": "DEPLOY_VIZ"
   }
```
and have some kind of dictionary in the backend for feature_package_compatibility like -

```python
feature_package_compatibility = {
  "DEPLOY_VIZ": {
     "fsspec": "2023.9.0" 
  }
}
```
This way we can check for feature compatibility and extend this if we need to check for multiple features in future. Since the frontend needs information on why some feature is not compatible, we can send specific message (containing the missing package dependency) or a generic message to the frontend like -

```json
{
"feature": "DEPLOY_VIZ",
"is_compatible": true,
"compatible_packages": {
        "fsspec": "2023.9.0" 
    }
}
```

ravi-kumar-pilla (commented):
Looks good to see all tests passing on Circle CI. Awesome !! 

Most of the code looks good to me but I have a suggestion for the feature compatibility API design. Please have a look. Thank you !

package/kedro_viz/api/rest/router.py
rashidakanchwala: yup, already updated it to` package-compatibilities` in the FE PR and will also do the same in this PR. 

So for now it's only one package compatibility that we return but yes in future there might be more. In that case I was thinking we would return a List[PackageCompatibilityAPIResponse]. 

Also, at this point, I am not so sure if we want to make it a POST request having said that we can revise this in the future if the need arises, unless others feel strongly about this approach.

tynandebold: Agree with both of you: this would be something we can scale going forward, though right now, without a use case to do so, I don't think we need to implement it at this stage.

package/kedro_viz/api/rest/responses.py
```diff
@@ -273,19 +330,154 @@ def get_default_response() -> GraphAPIResponse:
...
+        layers=data_access_manager.get_sorted_layers_for_registered_pipeline(
+            registered_pipeline_id
+        ),
+        pipelines=data_access_manager.registered_pipelines.as_list(),
+        selected_pipeline=registered_pipeline_id,
+        modular_pipelines=modular_pipelines_tree,  # type: ignore
+    )
+
+
+def get_package_version(package_name: str):
```
merelcht, line 384: I'm not quite sure if all the methods her belong in this file. As stated at the top this `responses.py` is meant for: `kedro_viz.api.rest.responses defines REST response types.` Again, I haven't looked at the architecture for a while, but this definitely seems like it's not supposed to be here.

package/kedro_viz/integrations/deployment/s3_deployer.py
```diff
@@ -0,0 +1,71 @@
...
+            logger.exception("Upload failed: %s ", exc)
+            raise exc
+
+    def _deploy(self):
+        self._upload_api_responses()
+        self._upload_static_files()
+        self._upload_deploy_viz_metadata_file()
+
+    def get_deployed_url(self):
+        """Returns an S3 URL where Kedro viz is deployed"""
```
merelcht, line 69: This also calls `_deploy()` so that's more than just returning the S3 URL, I'd make that clearer in the function name and doc string.

merelcht (commented):
I've done a first review and left some comments. I think my main points are around the "architecture" of the viz backend and where some of this new could should actually be. I'm sure this all works, but I'd be cautious about just breaking the intended architecture design that is in place now.

package/kedro_viz/api/rest/responses.py
```diff
@@ -1,13 +1,37 @@
...
 from kedro_viz.data_access import data_access_manager
+from kedro_viz.models.flowchart import (
+    DataNode,
+    DataNodeMetadata,
+    ParametersNodeMetadata,
+    TaskNode,
+    TaskNodeMetadata,
+    TranscodedDataNode,
+    TranscodedDataNodeMetadata,
+)
```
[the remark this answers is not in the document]
rashidakanchwala: Previously, in rest/router.py, we directly called models that were data classes, which is a common approach in [FastAPI](https://fastapi.tiangolo.com/advanced/dataclasses/).

Now, we've refactored this by moving the logic to two functions: `get_node_metadata_response` and `get_selected_pipeline_response`, which are in responses.py. This change is not new; it aligns with our existing approach for `get_default_response`, where the logic for /api/main resided in `responses.py`.

The reason we keep `get_default_response` in `responses.py` is because it's used in two places: in `router.py` when the user hits the endpoint `/api/main` and also during `kedro viz --save-file` in `server.py`, where we save these responses as JSON files. We've extended this approach to other API responses, such as node_metadata and selected_pipeline. This also fixed the previously broken` kedro viz --save-file` functionality, as we now save all the required files.

So yeah the above aligns with the same approach already used for  `get_default_response`. Happy to get on a call and discuss in more detail.

package/kedro_viz/api/rest/responses.py
```diff
@@ -1,13 +1,37 @@
...
+
+logger = logging.getLogger(__name__)
+
+_FSSPEC_PACKAGE_NAME = "fsspec"
+_FSSPEC_COMPATIBLE_VERSION = "2023.9.0"
```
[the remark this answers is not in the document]
rashidakanchwala: I didn't want to hard code this in the function. Is there a better way to do this?

package/kedro_viz/api/rest/responses.py
```diff
@@ -273,19 +330,154 @@ def get_default_response() -> GraphAPIResponse:
...
+        modular_pipelines=modular_pipelines_tree,  # type: ignore
+    )
+
+
+def get_package_version(package_name: str):
+    """Returns the version of the given package."""
+    return version(package_name)  # pragma: no cover
+
+
+def get_package_compatibilities_response():
```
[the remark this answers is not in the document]
rashidakanchwala: true, the reason I have kept this generic is incase we have more package compatibilties to test in the future.

package/kedro_viz/api/rest/responses.py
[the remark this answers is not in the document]
merelcht: We discussed this in person. Bottomline is that we should refactor Viz to use `dataclasses` in a better way and not do these response "conversions".

package/kedro_viz/api/rest/responses.py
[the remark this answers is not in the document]
merelcht: Also discussed in person :)

package/kedro_viz/api/rest/responses.py
```diff
@@ -273,19 +326,149 @@ def get_default_response() -> GraphAPIResponse:
...
+    jsonable_response = jsonable_encoder(response)
+    encoded_response = EnhancedORJSONResponse.encode_to_human_readable(
+        jsonable_response
+    )
+
+    with remote_fs.open(file_path, "wb") as file:
+        file.write(encoded_response)
+
+
+def save_api_main_response_to_fs(main_loc: str, remote_fs: Any):
```
merelcht, line 405: `main_loc` is a slightly vague variable name, I'd name this differently and also add some doc string to describe what this is.

merelcht: Perhaps `main_path`?

package/kedro_viz/api/rest/responses.py
```diff
@@ -273,19 +326,149 @@ def get_default_response() -> GraphAPIResponse:
...
+    """Saves API /main response to a file."""
+    try:
+        write_api_response_to_fs(main_loc, get_default_response(), remote_fs)
+    except Exception as exc:  # pragma: no cover
+        logger.exception("Failed to save default response. Error: %s", str(exc))
+        raise exc
+
+
+def save_api_node_response_to_fs(nodes_loc: str, remote_fs: Any):
```
merelcht, line 414: I'd change `loc` to `path` here as well if it makes sense to you.

package/kedro_viz/integrations/deployment/s3_deployer.py
```diff
@@ -0,0 +1,71 @@
...
+        self._remote_fs = fsspec.filesystem(self._protocol)
+
+    def _upload_api_responses(self):
+        save_api_responses_to_fs(self._bucket_name)
+
+    def _upload_static_files(self):
+        logger.debug("""Uploading static html files to %s.""", self._bucket_name)
+        try:
+            self._remote_fs.put(
+                f"{str(_HTML_DIR)}/*", self._bucket_name, recursive=True
```
merelcht, line 38: Instead of hardcoding this `global` value here, I'd add an argument for `html_dir` and then when you call the method do `_upload_static_files(_HTML_DIR)` That way upload static files is a more generic method and can be called with any HTML path in the future.

package/tests/test_api/test_rest/test_responses.py
```diff
@@ -794,15 +806,220 @@ def test_get_non_existing_pipeline(self, client):
...
+    def test_get_package_compatibilities_response_incompatible(self, mocker):
+        expected_version = "2023.8.1"
+        mocker.patch(
+            "kedro_viz.api.rest.responses.get_package_version",
+            return_value=expected_version,
+        )
+        response = get_package_compatibilities_response()
+        assert response.package_name == "fsspec"
+        assert response.package_version == expected_version
+        assert response.is_compatible is False
```
merelcht, line 844: Minor: You could parametrise these two tests into one.

package/tests/test_api/test_rest/test_responses.py
```diff
@@ -794,15 +806,220 @@ def test_get_non_existing_pipeline(self, client):
...
+            write_api_response_to_fs(file_path, response, mock_remote_fs)
+            mock_remote_fs.open.assert_called_once_with(file_path, "wb")
+            mock_encode_to_human_readable.assert_called_once()
+
+    @pytest.mark.parametrize(
+        "main_loc, expected_default_response",
+        [
+            ("/main", {"test": "json"}),
+        ],
+    )
```
merelcht, line 891: You've parameterised this test and several of the below, but as far as I can tell you're only passing one set of values right? In that case it might actually be more readable if you don't parameterise the tests where you don't have multiple test values.

merelcht (commented):
I had another look after our discussion about the architecture. I left some minor comments, but I think it's nearly ready 🙂

package/kedro_viz/api/rest/router.py
rashidakanchwala: So i read this earlier today, and thought it seemed relevant here. Hence sharing - https://www.techtarget.com/whatis/definition/You-arent-gonna-need-it#:~:text=YAGNI%20principle%20(%22You%20Aren',desired%20increased%20frequency%20of%20releases.

package/kedro_viz/api/apps.py
```diff
@@ -131,6 +131,24 @@ async def index():
 
     @app.get("/api/main", response_class=JSONResponse)
     async def main():
-        return json.loads(Path(filepath).read_text(encoding="utf8"))
+        return json.loads((Path(filepath) / "main").read_text(encoding="utf8"))
```
noklam, line 134: Maybe we could update the variable name `filepath`? I understand it used to be 1 file and we use `--save-file`, this is now a folder and we can reflect this internally at least.

package/kedro_viz/api/rest/responses.py
```diff
@@ -273,19 +326,151 @@ def get_default_response() -> GraphAPIResponse:
...
+            logger.exception(
+                "Failed to save pipeline data for pipeline ID %s. Error: %s",
+                pipelineId,
+                str(exc),
+            )
+            raise exc
+
+
+def save_api_responses_to_fs(filepath: str):
+    """Saves all Kedro Viz API responses to a file."""
```
noklam, line 449: Is this a single file or a direcotory?

package/kedro_viz/api/rest/router.py
```diff
@@ -38,53 +38,48 @@ async def main():
...
+    try:
+        return get_package_compatibilities_response()
+    except Exception as exc:
+        logger.exception(
+            "An exception occured while getting package compatibility info : %s", exc
+        )
+        return JSONResponse(
+            status_code=500,
+            content={"message": "Failed to get package compatibility info"},
+        )
```
noklam, line 85: When does this endpoint get called?

package/kedro_viz/integrations/deployment/s3_deployer.py
```diff
@@ -0,0 +1,69 @@
...
+deployment class for S3"""
+
+import json
+import logging
+from datetime import datetime
+from pathlib import Path
+
+import fsspec
+from kedro.io.core import get_protocol_and_path
+from semver import VersionInfo
```
noklam, line 11: I see earlier you are using `packaging`, I think we should be consistent using either that of `semver`. And if I remember correctly I read it in some other issue that `packaging` is a better option?

package/kedro_viz/api/rest/responses.py
```diff
@@ -273,19 +326,151 @@ def get_default_response() -> GraphAPIResponse:
...
+def write_api_response_to_fs(file_path: str, response: Any, remote_fs: Any):
+    """Encodes, enhances responses and writes it to a file"""
+    jsonable_response = jsonable_encoder(response)
+    encoded_response = EnhancedORJSONResponse.encode_to_human_readable(
+        jsonable_response
+    )
+
+    with remote_fs.open(file_path, "wb") as file:
+        file.write(encoded_response)
+
```
noklam, line 405: I was a bit surprised it is using "wb" instead of "w". I expect the result of `encode_to_human_readable` should be text that is readable, but seems like it is returning `bytes`?

noklam (commented):
If we can't change the `--save-file` arguments, I think we could still update the help text to clarify it is saving multiple files instead of one.

image

It's confusing where I should provide a filename or a directory (maybe it should be an optional argument and have a default name?)

In addition, when I run `kedro viz --save-file` it try to open a new tab in my chrome browser, I think it's a bug. Should file be save as a `.json` suffix as seems like they are all JSON.

rashidakanchwala:
> If we can't change the `--save-file` arguments, I think we could still update the help text to clarify it is saving multiple files instead of one.
> 
> image
> 
> It's confusing where I should provide a filename or a directory (maybe it should be an optional argument and have a default name?)
> 
> In addition, when I run `kedro viz --save-file` it try to open a new tab in my chrome browser, I think it's a bug. Should file be save as a `.json` suffix as seems like they are all JSON.

@noklam -- definitely will fix the text... thanks for flagging! 

Good question -- I am not sure if it's a bug. 
So when you do `kedro viz --save-file` it does 2 things -- it opens kedro viz and it also saves your kedro-viz api files in the folder you specify. 

@stephkaiser , @tynandebold , @merelcht  -- what do you think of the above? To be frank; i was also initially confuse because kedro-viz would open when I did `kedro viz --save-file` but then I assumed this was the functionality and user's haven't highlighted this.

tynandebold:
> > If we can't change the `--save-file` arguments, I think we could still update the help text to clarify it is saving multiple files instead of one.
> > image
> > It's confusing where I should provide a filename or a directory (maybe it should be an optional argument and have a default name?)
> > In addition, when I run `kedro viz --save-file` it try to open a new tab in my chrome browser, I think it's a bug. Should file be save as a `.json` suffix as seems like they are all JSON.
> 
> @noklam -- definitely will fix the text... thanks for flagging!
> 
> Good question -- I am not sure if it's a bug. So when you do `kedro viz --save-file` it does 2 things -- it opens kedro viz and it also saves your kedro-viz api files in the folder you specify.
> 
> @stephkaiser , @tynandebold , @merelcht -- what do you think of the above? To be frank; i was also initially confuse because kedro-viz would open when I did `kedro viz --save-file` but then I assumed this was the functionality and user's haven't highlighted this.

Thanks for pointing that out, Nok! The action of Viz starting and opening when you run `kedro viz --save-file` has always been there and wasn't introduced in this PR. We can decide if we think it's a bug or not, though either way, we don't need to implement a solution in this PR, so it shouldn't block this work.

package/kedro_viz/integrations/deployment/s3_deployer.py
rashidakanchwala: Great point. I have reopened the issue and we should replace it. https://github.com/kedro-org/kedro-viz/issues/1460

package/kedro_viz/api/rest/router.py
rashidakanchwala: sure, so in the FE when the user clicks on `Deploy Kedro Viz` button -- we first make a call to this endpoint /api/package-compatibilities and get the fsspec version. If the version is compatible with Shareable viz, then they will get a form to enter thier bucket-name and region. If not, then they will get a message asking them to enable Shareable viz; they need to be on a particular fsspec version. See example from the screenshot

package/kedro_viz/api/rest/responses.py
rashidakanchwala: directory! updating :)

package/kedro_viz/api/apps.py
rashidakanchwala: renamed it to `api_dir`?

merelcht (approved):
I think this is in a good state now. Great work @rashidakanchwala ⭐ !

IMO the Viz backend should be refactored before adding any other new big features, this PR showed some of the strange stuff going on with the `pydantic` models and dataclasses. It might be a good time to do that after shareable viz is merged.

package/kedro_viz/api/rest/responses.py
rashidakanchwala: yes, it's currently returning bytes. I will have to create an issue to figure the best way to make this more clearer. But it would be out of scope for the current one.

noklam (approved):
Thanks! I agree with Merel that it would be great to spend some time after this to clean up the backend.

It's a nice feature to add and I am interested to make a GitHubPageDeployer after this is merged  :)

PR #1507, opened 2023-08-29
Fix for Kedro Viz Connection Error

### LINKED ISSUE #1277: Browser launches too soon, causing a connection error
## Description
As per title. After waiting for a couple of seconds, reloading the page just shows Kedro Viz as it should. But it might not be obvious to all users.

## Context
It doesn't affect me personally but I think it's bad UX. I'm wondering if there's a way to wait until the service is ready to accept connections to launch the browser. Or just give a longer timeout, but this isn't very robust.

## Steps to Reproduce

1. Run `kedro viz` on the terminal
2. See a connection error in the browser
3. Wait a couple of seconds, refresh, and see Kedro Viz

## Expected Result
100 % of the times, the browser should open and show Kedro Viz and not a connection error, so that the user doesn't have to refresh the page.

## Your Environment

* Web browser system and version: Firefox 110.0.1
* Operating system and version: macOS Ventura
* Kedro version used (if relevant): 0.18.5 and current `main`
* Python version used (if relevant): 3.10

## Checklist

ravi-kumar-pilla:
## Description

Resolves #1277 

## Development notes

- Start Fast API server using multiprocessing similar to jupyter launcher with/without --autoreload option
-  Wait for server to start before opening web browser 

Resolved_Connection_Error

## QA notes

- Checkout branch `fix/connection-error`
- Build frontend using `make build`
- Build backend using `pip install -e package`
- cd demo-project, execute command `kedro viz` or `kedro viz --autoreload`
- Browser window should open after FAST API server is started and the browser should not dispay a connection error

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1507"></a>


package/kedro_viz/launchers/jupyter.py
```diff
@@ -175,7 +116,7 @@ def run_viz(port: int = None, local_ns: Dict[str, Any] = None) -> None:
     viz_process.start()
     _VIZ_PROCESSES[port] = viz_process
 
-    _wait_for(func=_check_viz_up, host=host, port=port)
+    wait_for(func=check_viz_up, host=host, port=port)
```
noklam, line 119: What is the reason to change from private function to public one?

package/kedro_viz/launchers/cli.py
```diff
@@ -102,36 +105,47 @@ def viz(host, port, browser, load_file, save_file, pipeline, env, autoreload, pa
...
+
+        viz_process.start()
+        _VIZ_PROCESSES[port] = viz_process
+
+        wait_for(func=check_viz_up, host=host, port=port)
+
+        print("Kedro Viz Backend Server started successfully...")
+
+        if browser:
+            start_browser(host, port)
```
noklam, line 147: 👍🏼

noklam (commented):
Leave a small comment. So to check with my understanding, the main change here is moving the `start_browser` from `run_server` to the cli, and the run_server only handle the backend process.

I have a quick question, why `multiprocessing` is needed?

package/kedro_viz/launchers/jupyter.py
ravi-kumar-pilla: Earlier these functions were inside jupyter.py and used only by jupyter, now it is also used in cli launcher. so moved it to a utils file and these functions are now used in both places

ravi-kumar-pilla:
> Leave a small comment. So to check with my understanding, the main change here is moving the `start_browser` from `run_server` to the cli, and the run_server only handle the backend process.

Yes, I have moved the start_browser logic out of run_server as run_server should just focus on starting the FAST API server.

> I have a quick question, why `multiprocessing` is needed?

So, earlier we used to start the browser without waiting for backend server to start. Now if one process is used to start the server, there should be another process which can wait for the server to start and then start a browser (This can also be done using threading, but since we were using multiprocessing in jupyter already, I thought of using the same in cli launcher as well. )

package/kedro_viz/launchers/jupyter.py
noklam: I see they are now shared among two modules, but I think it should be an internal method `_wait_for` instead of `wait_for`.

For example, we have similar internal utils here:
https://github.com/kedro-org/kedro/blob/edbbcc7f7a863b7c980f9efcaa444e2fa01f45fa/kedro/config/common.py#L209

Changing from `_wait_for` -> `wait_for` singals this is a public function. (though in reality 99% people are using kedro-viz as an appllication but not a module). So this is just to keep things clean internally for ourselves.

https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

noklam (approved):
I've experienced this issue, definite an improvement for UX. Great fix, thanks!

NeroOkwa (approved):

rashidakanchwala (approved):
LGTM. thanks :)

Issue #1521, opened 2023-09-05
Preview SQL code in Kedro-Viz for SQL Datasets


delvillarjavier:
## Description
Currently Kedro-Viz only previews data from CSVs and Excel files, I would like to view the SQL code that produces that table, not the resulting table.

## Context
I use Kedro-Viz very often to review projects status, it would be nice if I can see all the datasources and the SQL code that produces them in one place.

## Possible Implementation
Add a toggle in SQL based datasets
![image](https://github.com/kedro-org/kedro-viz/assets/110201830/55ff0d69-baee-412d-b138-6a0b5a443d56)

yetudada:
This makes me wonder, do you want to also see the code for other datasets listed in `catalog.yml` too?

delvillarjavier:
Hello! 
I was thinking about that, I just want some idea of the data I would get or where is it, in the case there is no "File Path" attribute. :)

MehdiNV:
**Update:** Leaving an update here on progress, and potential next steps! 

We decided to split this issue into two parts:

1. Implement a 'Preview' for SQLQueryDataset-based data node
2. Extend the Preview mechanism for other 7 remaining datasets (all who don't have a file, and hence would benefit)

Reason for this is that we'd need some more time on finalising the design for those other 7 datasets - the SQL-dataset is still technically 'Code', and hence would fall in-line with a 'Show Code' / Code preview window. The other 7 in contrast may only have a URL (e.g. API) or source database, and our plan is to re-think where exactly they should go in the metadata window instead of adding it now and needing to refactor it later. 

Current PR implements the 1st task fully in the meantime! https://github.com/kedro-org/kedro-viz/pull/1564

astrojuanlu:
One question (maybe a bit too late, apologies): wouldn't it be easier to display the whole YAML definition of the dataset? So, instead of just the SQL query, the whole thing:

```
shuttle_id_dataset:
  type: pandas.SQLQueryDataset
  sql: "select shuttle, shuttle_id from spaceflights.shuttles;"
  credentials: db_credentials
```

that way (1) this works for every dataset, not only SQL ones, and (2) possibly addresses the coupling @merelcht flagged in https://github.com/kedro-org/kedro-viz/pull/1564#issuecomment-1771260062

MehdiNV:
**Update, had a discussion on feature:** 

[...]

PR #1564, opened 2023-10-09
Add 'preview' mechanism for SQL queries

MehdiNV:
## Description
If a data node relies on a file, then the user has the ability to see where that file is and have a preview into its data.

The same is not true for non-file datasets, like `SQLQueryDataset` or `APIDataset` who pull their data from non-file sources. In this instance, the user has a data node but it's unclear of its origin. 

This adds one in for those non-file sources. Essentially, a `filepath` attribute for datasets that have no files.

## TL:DR 
Resolves https://github.com/kedro-org/kedro-viz/issues/1521, adds the preview mechanism (already present for task nodes) to data nodes as well

## Development notes

Modifies DataNodeMetadata to have a `code` attribute. This is already present for TaskNodeMetadata, this is just making DataNode have the same capability.

This `code` attribute would store SQL command (whatever was used to generate data) or an address for API-sources. 

There are 8 non-file datasets available to users; this PR handles all 8 accordingly.

## QA notes

- All tests run normally.
- Added new test cases in `kedro_viz`-`test_models`-`test_utils.py` file to achieve 100% test coverage.
- To test your own data node's with 'preview', create a SQLQueryDataset-based data node. Exact reproduction is:
  - Create a new data node in `demo-project/conf/base/parameters/catalog_01_raw.yml` - make sure its type is SQLQueryDataset 

<p align="center">
image
</p>

  - Add that node into `.../src/pipelines/data_ingestion/pipeline.py` 
  - If necessary, create a 'dummy' function and link it to the code in `pipeline.py` if kedro complains about missing functions
  - See [commit](https://github.com/kedro-org/kedro-viz/pull/1564/commits/8ca2d32b2ff1920b87fde01d9fb071f0a85727cc) for example codes to help you with all of the above

## What's next?

- Modifying Node's `filepath` attribute to contain a glimpse of the `code` attribute value (e.g. if dataset is API, the code would be the full source details whilst the `filepath` would be a simplistic version of it like 'www.api_source.com'). 
- Aiming to address this extension in a separate ticket though (PR is already large)

## Image Example
_Where the data node has some 'code' value associated with it, a switch is displayed_
&nbsp;
image

_This displays a 'Source block' window; displays the underlying logic or source location that generated this data_
&nbsp;
image

_This is in-line with the same mechanism used for task / function nodes_
&nbsp;
image


MehdiNV:
I'll just aim to squash all these commits when merging ;-;

rashidakanchwala:
Amazing work Mehdi . This is really great. I like that this solves for not just SQLDatasets but also extends it for other non-file datasets. Also when we started building Preview Datasets; initially we added support for CSV, and Excel and eventually support the others. So this is great step in going in that direction. 

Before I delve deep into the code review, I want to ensure that design is on board with this approach. While the 'Show Code' function is typically used to display task node functions, most dataset information is on the metadata panel. I do see the value in having a 'Show Source' feature for dataset. Looping in @stephkaiser

rashidakanchwala:
A few more thoughts after discussing with Mehdi. 'Show Source' originally 'Show Code' is a code block to display code for each task node. This makes a lot of sense for SQLDatasets given that it's SQL Code and it can be lengthy. Hence it's nice to display it in the existing code panel. But that does not apply to other datasets like APIDataset where we display the URL which is not code or SparkHiveDataset which is a string 'Derived from'. 

I do like the idea of calling it 'Source' because now we are making it generic.

package/kedro_viz/models/utils.py
```diff
@@ -43,3 +44,41 @@ def get_dataset_type(dataset: "AbstractDataset") -> str:
...
+    elif format_type in {"SparkHiveDataSet", "SnowparkTableDataSet"}:
+        extracted_table = (
+            data_desc["table_name"]
+            if format_type == "SnowparkTableDataSet"
+            else data_desc["table"]
+        )
+        source = f'Derived from {extracted_table} table, in {data_desc["database"]} database.'
+        return source
+    elif format_type == "SparkJDBCDataSet":
+        return f'Derived from {data_desc["table"]} table, from {data_desc["url"]} url.'
```
ravi-kumar-pilla, line 80: Awesome !!

package/tests/test_integrations/test_sqlite_store.py
```diff
@@ -1,6 +1,6 @@
 # We need to disable pylint because of this issue -
 #  https://github.com/pylint-dev/pylint/issues/8138
-# pylint: disable=E1102
+# pylint: disable=E1102,useless-suppression
```
ravi-kumar-pilla, line 3: I do not see any file changes here, can you please comment on why this was added ?

ravi-kumar-pilla:
This looks awesome @MehdiNV . Great work !! I have left few comments on the PR and I agree with Rashida on having the design team signoff the UI changes. But the backend stuff is really good. Thank you

demo-project/src/demo_project/pipelines/data_ingestion/nodes.py
```diff
@@ -116,3 +116,9 @@ def combine_shuttle_level_information(
     working_table = combined_table.dropna(how="any")
     id_columns = [x for x in working_table.columns if x.endswith("id")]
     return working_table, working_table[id_columns]
+
+
+def apply_basic_logic_to_sql_example_data(
+    sqlExampleParam
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    return sqlExampleParam
```
tynandebold, line 124: Is this still needed?

tynandebold (commented):
How can we test the UI? Can you please provide some QA notes on how we can display a node that allows us to preview a SQL query?

demo-project/src/demo_project/pipelines/data_ingestion/nodes.py
MehdiNV: Ah crap! Missed that earlier, not needed yeah

MehdiNV: Re UI, good call! On:
- **QA Notes:** Added now
- **Testing UI on .js side:** Shouldn't be necessary; front-end side does not seem to distinguish between 'data' and 'task' nodes - it just treats them all as one type (node). 'Preview' is only enabled if back-end feeds it a 'code' value; this PR hijacks the backend to feed the same thing for its data nodes. So for front-end, the current UI testing it does for its 'preview' nodes will cover both task and data nodes from the get-go

package/kedro_viz/models/flowchart.py
```diff
@@ -577,6 +580,7 @@ def __post_init__(self, data_node: DataNode):
         self.type = data_node.dataset_type
         dataset = cast(AbstractDataset, data_node.kedro_obj)
         dataset_description = dataset._describe()
+        self.code = extract_data_source(self.type, dataset_description)
```
rashidakanchwala, line 583: can we not follow the approach we have for other datasets where we add dataset specific properties. 

if data_node.is_sql_node():
    self.query = extract_query(data_node)

MehdiNV: We can, but my concern is that it'll take too long - some refactoring would be needed, as currently we get:
image

due to line 609 below
image

What do you think of merging as it is? The solution seems to work reliably, and this solution can be implemented when we refactor the other datasets in?

package/kedro_viz/models/utils.py
```diff
@@ -43,3 +44,27 @@ def get_dataset_type(dataset: "AbstractDataset") -> str:
...
+    Returns:
+        String to display to user as the node source code
+    """
+    if node_type is None:
+        return None
+
+    format_type = node_type.split(".")[-1]
+
+    if format_type in {"SQLQueryDataSet", "GBQQueryDataSet"}:
+        return data_desc["sql"]
```
rashidakanchwala, line 68: @MehdiNV  - I'm not entirely confident in this approach, especially after discussing it with @merelcht. It seems this approach may not be robust, as changes to kedro-datasets could potentially break it. I would recommend collaborating with someone from the framework team to explore ways to improve our approach

ravi-kumar-pilla: May be we should stop hardcoding strings and rely on something like -
`if data_node.is_sql_node():
     self.code = dataset_description.get("sql")`
    as suggested by @rashidakanchwala 
    
 - Even accessing the sql key is not maintainable but we do similar logic fetching the file_path and other properties from the describe function
 - The better way of accessing any values would be having getters inside the kedro-datasets plugin something like `sqlDataset.getCode` - this way we wont rely on the underlying key name. But for now we can check the node type and get the sql code.

rashidakanchwala: @merelcht -- also let us know if you think the above approach makes sense. It's not ideal but the alternative as Ravi mentioned is to have getter functions in kedro-datasets ?

merelcht:
I have had a look at this PR and done some thinking about the approach and IMO a lot of this logic should live in `kedro-datasets` on the specific datasets instead of in Kedro Viz. 

If you hard code things like dataset names `if format_type in {"SQLQueryDataSet", "GBQQueryDataSet"}:` or fetch keys from the dataset `data_desc["sql"]` this will break as soon as the name or fields are changed on the datasets. The name is already an issue now, because we've renamed all datasets to end with `Dataset` instead of `DataSet` and all the classes with the old name will be removed from Kedro in Kedro `0.19.0` and kedro-datasets `2.0.0`. 

This might seem like a small feature and I understand you want to merge this in quickly, but the problem is that this is just adding to the already massive technical debt and coupling that exists between Kedro Viz and `kedro-datasets`. We moved datasets into its own separate package to make sure we could develop it faster, make breaking changes faster without the need of introducing breaking changes in the Kedro framework. This has been hugely successful so far, with the exception of Kedro-Viz. For example, the dataset rename was done in a backward compatible way so that currently using either syntax will work (e.g. `pandas.CSVDataSet` and `pandas.CSVDataset`). However, because of all the hardcoding of class names in Kedro Viz the rename broke experiment tracking. 

If features keep being developed in a way that tightens viz to datasets, you will be running into stuff breaking more and more often. We've come a long way with Kedro to make it more and more robust and I would urge Viz to stick to the same standards or you'll end up fighting fires every day.

package/kedro_viz/models/utils.py
MehdiNV: I've pushed commits for the first approach in the meantime, have just seen latest comments though so don't think they're viable; will chat with framework tomorrow to see if there's any way forward as I think this issue could be blocked for us for now.

MehdiNV:
> I have had a look at this PR and done some thinking about the approach and IMO a lot of this logic should live in `kedro-datasets` on the specific datasets instead of in Kedro Viz.
> 
> If you hard code things like dataset names `if format_type in {"SQLQueryDataSet", "GBQQueryDataSet"}:` or fetch keys from the dataset `data_desc["sql"]` this will break as soon as the name or fields are changed on the datasets. The name is already an issue now, because we've renamed all datasets to end with `Dataset` instead of `DataSet` and all the classes with the old name will be removed from Kedro in Kedro `0.19.0` and kedro-datasets `2.0.0`.
> 
> This might seem like a small feature and I understand you want to merge this in quickly, but the problem is that this is just adding to the already massive technical debt and coupling that exists between Kedro Viz and `kedro-datasets`. We moved datasets into its own separate package to make sure we could develop it faster, make breaking changes faster without the need of introducing breaking changes in the Kedro framework. This has been hugely successful so far, with the exception of Kedro-Viz. For example, the dataset rename was done in a backward compatible way so that currently using either syntax will work (e.g. `pandas.CSVDataSet` and `pandas.CSVDataset`). However, because of all the hardcoding of class names in Kedro Viz the rename broke experiment tracking.
> 
> If features keep being developed in a way that tightens viz to datasets, you will be running into stuff breaking more and more often. We've come a long way with Kedro to make it more and more robust and I would urge Viz to stick to the same standards or you'll end up fighting fires every day.

@merelcht Understandable yeah; I don't think there's anyway for us to resolve this issue though from viz side otherwise, and this may need changes on the framework side instead for the most optimal solution. Perhaps can just pair quickly tomorrow?

delvillarjavier:
wow! it's happening! 😍😍😍

MehdiNV:
Closing until further notice; see recent notes on the [issue thread](https://github.com/orgs/kedro-org/projects/1/views/4?pane=issue&itemId=38503256)

PR #1565, opened 2023-10-09
Refactor Flowchart models from dataclass to pydantic base models

### LINKED ISSUE #1538: [Spike] Clarifying Overlap: Pydantic Models vs. Data Classes in Backend
## Description

We recently had a look at our backend structure while reviewing the Shareable Viz PR, we realised that using both Pydantic Base Models and data classes in our backend might be unnecessary. This ticket is to investigate this issue and refactor the code accordingly.

## Context:

FastAPI's documentation recommends the use of Pydantic for [input validation and data serialization](https://fastapi.tiangolo.com/tutorial/body/). It mentions [data classes](https://fastapi.tiangolo.com/advanced/dataclasses/) offering a more advanced approach. Using both Pydantic models and data classes for similar purposes has lead to redundancy and has caused issues with mypy 1.0.

In this issue; we want to understand the below:-
- Are there performance implications between the two? If one is more efficient than the other for specific tasks, we should clarify their roles. Was the decision to use both related to our data access manager and repositories?
- Given that Pydantic models are classes as well, can't we achieve similar functionalities with them as with data classes?

## Checklist

ravi-kumar-pilla:
## Description

Partially Resolves #1538 

## Development notes

* Replaced all data classes from models/flowchart.py with pydantic models
* Used root-validators and field validators
* Modified pytests

## QA notes

* All features in Kedro Viz should run normally. 
* Start the app either using `kedro viz` command from demo-project or run the dev server using `make run`
* Run pytests using `make pytest`

## Checklist


package/kedro_viz/models/flowchart.py
```diff
@@ -337,10 +326,13 @@ class ModularPipelineNode(GraphNode):
     # when using the pipeline() factory function.
     # More information can be found here:
     # https://kedro.readthedocs.io/en/latest/06_nodes_and_pipelines/03_modular_pipelines.html#how-to-connect-existing-pipelines
-    internal_inputs: Set[str] = field(default_factory=set)
-    internal_outputs: Set[str] = field(default_factory=set)
-    external_inputs: Set[str] = field(default_factory=set)
-    external_outputs: Set[str] = field(default_factory=set)
+    internal_inputs: Set[str] = set()
```
rashidakanchwala, line 329: I think you can use default_factory here too.

package/kedro_viz/models/flowchart.py
```diff
@@ -36,14 +36,26 @@ def _parse_filepath(dataset_description: Dict[str, Any]) -> Optional[str]:
     return str(filepath) if filepath else None
 
 
-@dataclass
-class RegisteredPipeline:
+def _extract_wrapped_func(func: FunctionType) -> FunctionType:
```
merelcht, line 39: What is this used for?

package/kedro_viz/models/flowchart.py
```diff
@@ -93,29 +108,36 @@ class GraphNode(abc.ABC):
...
+        strict = True
+        arbitrary_types_allowed = True
+
+    @classmethod
+    def _hash(cls, value: str):
         return hashlib.sha1(value.encode("UTF-8")).hexdigest()[:8]
 
-    @staticmethod
-    def _get_namespace(dataset_name: str) -> Optional[str]:
+    @classmethod
```
merelcht, line 139: Why is this now a `classmethod` instead of a `staticmethod`? And same for the functions above and below.

merelcht (commented):
Sorry for the delay in review @ravi-kumar-pilla. I don't know too much about pydantic, but generally these changes look fine to me. I have a couple of questions.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: This is used to get the wrapped function code which we show in kedro viz source code block.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: my bad, there was an error when using staticmethod with pydantic and I thought classmethod is what we use in pydantic. No particular reason and it makes sense to have it as static as it is not relying on any class properties. I will revert this change. Thank you !

package/kedro_viz/models/flowchart.py
merelcht: Got it!

merelcht (approved):
LGTM! 👍 Might be good to also get @noklam or @astrojuanlu to look at this. I think they know more about `pydantic` than I do 🙂

package/kedro_viz/models/flowchart.py
```diff
@@ -356,39 +348,57 @@ def inputs(self) -> Set[str]:
...
+    type: str = GraphNodeType.TASK.value
+
+    def __init__(self, **data):
+        super().__init__(**data)
+        self.namespace = self.kedro_obj.namespace
+
+        # the modular pipelines that a task node belongs to are derived from its namespace.
+        self.modular_pipelines = self._expand_namespaces(self.kedro_obj.namespace)
+
+
```
noklam, line 377: Can you move this up? It makes it harder to see what are the actual changes.

package/kedro_viz/models/flowchart.py
```diff
@@ -93,22 +108,29 @@ class GraphNode(abc.ABC):
...
 
     # the set of registered pipeline IDs this node belongs to
-    pipelines: Set[str] = field(default_factory=set)
+    pipelines: Set[str] = set()
 
     # The original namespace on this node.
     # N.B.: in Kedro, modular pipeline is implemented by declaring namespace on a node.
     # For example, node(func, namespace="uk.de") means this node belongs
     # to the modular pipeline "uk" and "uk.de"
-    namespace: Optional[str] = field(init=False, default=None)
```
noklam, line 108: Noted that it used to be not possible to be initialised (it is possible now). I am not sure how this may affect downstream after many layers of propagation. Is it possible to maintain this behavior using the validation. see https://github.com/pydantic/pydantic/issues/2019

package/kedro_viz/models/flowchart.py
```diff
@@ -416,52 +429,54 @@ def __post_init__(self, task_node: TaskNode):
...
+        try:
+            metadata = getattr(self.kedro_obj, "metadata", None)
+
+            if isinstance(metadata, dict):
                 self.viz_metadata = metadata["kedro-viz"]
-            except (AttributeError, KeyError):  # pragma: no cover
-                logger.debug("Kedro-viz metadata not found for %s", self.name)
+
+        except (AttributeError, KeyError):  # pragma: no cover
+            logger.debug("Kedro-viz metadata not found for %s", self.name)
```
noklam, line 475: I don't find this refactoring is better than the previous form.  Original Line 455 is using `getattr` already and shouldn't be buried in a try-except block. 

In fact I found the original logic weird too (What could lead to `AttributeErorr`? and if metadata is not a dict you get `TypeError` which isn't caught). I will either use try-except to catch the whole metadata block or use `getattr`

Since this PR focus on dataclass refactoring I suggest leave it and don't make this change.

package/kedro_viz/models/flowchart.py
```diff
@@ -619,58 +590,95 @@ def __post_init__(self, data_node: DataNode):
                 )
 
 
-@dataclass
+class TranscodedDataNode(GraphNode):
```
noklam, line 593: Why is this moved down?

noklam (commented):
Have the 1st review.  
1. There are changes that are not 1:1 between Pydantic and dataclass. It's less restrictively i.e. frozen class is now mutable, non-initialisable attributes are now initialisable. It probably works the same for kedro-viz but it's better to keep it strict to reduce the chance that we fall into our own traps.
2. If re-ordering is needed, I suggest to make it in separate PR since it's quite hard to review when things get shuffled.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: I was trying to reorder node classes with metadata node class. I reverted the re-order as it is making PR reviews difficult

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: There was an issue with linting. If we try to access metadata["kedro-viz"] without the isinstance check. So I added that. I removed the getattr from try block now, since it has a default None value

noklam:
We had call discussed this last week and I haven't seen any change. @ravi-kumar-pilla is this ready for review already?

ravi-kumar-pilla:
> We had call discussed this last week and I haven't seen any change. @ravi-kumar-pilla is this ready for review already?

We did not made a decision yet. Since this is not a 1:1 conversion and there are no performance benefits, we are thinking of either keeping the data classes as is or going with the pydantic base class as per this PR (though I need to modify some code here). 

@rashidakanchwala can you please confirm when you get some time. Thank you

ravi-kumar-pilla:
I would like to close this ticket as it is not a 1:1 conversion possible with pydantic v1. However, with pydantic v2, we can have much closer conversion (retaining __post_init__) - https://github.com/pydantic/pydantic/issues/1729#issuecomment-1300576214

I will be doing the refactor on a separate ticket which supports pydantic v2 on kedro viz - https://github.com/kedro-org/kedro-viz/issues/1603

package/kedro_viz/models/flowchart.py
```diff
@@ -61,15 +61,19 @@ class GraphNodeType(str, Enum):
...
-@dataclass(frozen=True)
-class ModularPipelineChild:
+class ModularPipelineChild(BaseModel):
     """Represent a child of a modular pipeline."""
 
     id: str
     type: GraphNodeType
 
+    def __init__(self, **data):
+        super().__init__(**data)
```
antonymilne, line 71: This doesn't do anything at all.

package/kedro_viz/models/flowchart.py
```diff
@@ -267,31 +277,30 @@ def add_pipeline(self, pipeline_id: str):
...
     """Represent a graph node of type TASK"""
 
-    modular_pipelines: List[str] = field(init=False)
-    parameters: Dict = field(init=False, default_factory=dict)
+    modular_pipelines: List[str] = []
+    parameters: Dict = {}
     type: str = GraphNodeType.TASK.value
 
-    def __post_init__(self):
+    def __init__(self, **data):
```
antonymilne, line 302: This will work ok but is not really the right way to do things in pydantic. What I'd recommend is something like this:
```
@validator("namespace")
def set_namespace(cls, values):
    return values["kedro_obj"].namespace

@validator("modular_pipelines")
def set_modular_pipelines(cls, values):
    return self._expand_namespaces(values["kedro_obj"].namespace)
```

[Default factories](https://docs.pydantic.dev/1.10/usage/models/#field-with-dynamic-default-value) are also useful here and work like they do in dataclasses, but can't be used to reference other fields in the same model, hence the need for validators here.

ravi-kumar-pilla:
There are some mypy lint errors for strawberry graphql schema that does not have any relevance to this PR. I am checking the issue, but in the mean time I would like @noklam , @rashidakanchwala and @antonymilne (if you have some time) to please review the PR. To summarize, this PR only aims at converting all the dataclasses in flowchart.py to pydantic Base Models. 

Thank you

noklam:
Can you update the Issue title to reflect the current status since the scope have changed and most discussion are obsoleted.

package/kedro_viz/models/flowchart.py
```diff
@@ -61,54 +69,72 @@ class GraphNodeType(str, Enum):
...
+        id (str): Id of the modular pipeline child
+        type (GraphNodeType): Type of modular pipeline child
+    """
 
     id: str
     type: GraphNodeType
 
+    def __hash__(self) -> int:
+        return hash(self.id)
+
```
noklam, line 85: Why this is needed now? Is this a bug fix or part of the refactoring

package/kedro_viz/models/flowchart.py
```diff
@@ -278,24 +304,36 @@ def has_metadata(self) -> bool:
         return self.kedro_obj is not None
 
 
-@dataclass
-class GraphNodeMetadata(abc.ABC):
+class GraphNodeMetadata(BaseModel, abc.ABC):
```
noklam, line 307: Why do you need to inherit from `abc`?

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: Dataclasses provide default methods for example (__hash__), but for pydantic base class, we need to add these methods manually

package/kedro_viz/models/flowchart.py
```diff
@@ -278,24 +304,36 @@ def has_metadata(self) -> bool:
...
+
+    # The type for Task node
     type: str = GraphNodeType.TASK.value
 
-    def __post_init__(self):
-        self.namespace = self.kedro_obj.namespace
+    @root_validator(pre=True)
+    def check_kedro_obj_exists(cls, values):
+        assert "kedro_obj" in values
+        return values
```
noklam, line 328: Can you use `@validator` instead? From my understanding `root_validator` is used for validating entire model data, here it seems that you are interested in a single field.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: I kept the original signature, but added BaseModel to it. Though we do not have any fields in this class, this is supposed to be an abstract base class.

package/kedro_viz/models/flowchart.py
```diff
@@ -337,10 +368,25 @@ class ModularPipelineNode(GraphNode):
...
+        pipeline node with other modular pipelines""",
+    )
+    external_outputs: Set[str] = Field(
+        set(),
+        description="""The dataset outputs connecting the modular
+        pipeline node with other modular pipelines""",
+    )
+
+    # The type for Modular Pipeline Node
+    type: str = GraphNodeType.MODULAR_PIPELINE.value
```
noklam, line 389: Why is this field shifted down?

package/kedro_viz/models/flowchart.py
```diff
@@ -496,189 +610,278 @@ def is_tracking_node(self):
...
+        if not (self.viz_metadata and self.viz_metadata.get("preview_args", None)):
             return False
-        return is_preview
+
+        return True
 
     def get_preview_args(self):
         """Gets the preview arguments for a dataset"""
-        return self.viz_metadata["preview_args"]
+        return self.viz_metadata.get("preview_args", None)
```
noklam, line 620: It seems that you assume self.viz_metadata is always exist, do you need the first condition?

package/kedro_viz/models/flowchart.py
```diff
@@ -496,189 +610,278 @@ def is_tracking_node(self):
...
 
 
-@dataclass
 class TranscodedDataNode(GraphNode):
-    """Represent a graph node of type DATA"""
+    """Represent a graph node of type DATA
```
noklam, line 624:
```suggestion
    """Represent a graph node of type data
```

package/kedro_viz/models/flowchart.py
```diff
@@ -496,189 +610,278 @@ def is_tracking_node(self):
...
 
-    # whether the data node is a free input
-    is_free_input: bool = field(default=False)
+    Args:
+        layer (Optional[str]): The layer that this transcoded data
+        node belongs to. Defaults to `None`.
+        is_free_input (bool): Determines whether the transcoded data
+        node is a free input. Defaults to `False`.
+        stats (Optional[Dict]): Statistics for the data node
```
noklam, line 631: Nit. If the docstring is multi-line it should be indented, it's a bit hard to read which one is the argument, there should be example in the codebase

package/kedro_viz/models/flowchart.py
```diff
@@ -496,189 +610,278 @@ def is_tracking_node(self):
...
             except Exception as exc:  # pylint: disable=broad-except # pragma: no cover
                 logger.warning(
                     "'%s' could not be previewed. Full exception: %s: %s",
-                    data_node.name,
+                    cls.data_node.name,
                     type(exc).__name__,
                     exc,
                 )
+                return None
+        return None
```
noklam, line 788: If you take a shortcut `if not cls.data_node.is_preview_node()` return `None` immediately, the logic will be less nested. (Github don't let me to add changes to the line you didn't edit)

package/kedro_viz/models/flowchart.py
```diff
@@ -717,25 +920,47 @@ def parameter_value(self) -> Any:
...
+    Args:
+        source (str): The id of the source node.
+        target (str): The id of the target node.
+    """
 
     source: str
     target: str
+
+    def __hash__(self) -> int:
+        return hash((self.source, self.target))
```
noklam, line 966: Does it actually changed the hash?

noklam (commented):
Left some comments, most are minor.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: no big reason for the shift, but following the order from other classes. In pydantic validations happen in the order we define fields. Though we are not setting any value through pydantic validator for this field, I thought of having the enum types at the end.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: I agree, I was thinking of doing the same but I thought of sticking to the pattern of checking the true case (i.e., set_tracking_data, set_image functions above this one)

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: my bad, This was added in the earlier refactor when I did not add frozen attribute which made the classes unhashable type when used in sets and dict. Now this is not required anymore. Thanks for pointing

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: this is not needed, will revert this

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: sure, we can add it but `get_preview_args` is only called after checking `is_preview_node`, which checks for self.viz_metadata. should we add it again ?

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: looks like we need the __hash__ function as pylint does not understand pydantic frozen class. Though it does not complain with frozen dataclass - [issue]( https://app.circleci.com/pipelines/github/kedro-org/kedro-viz/10998/workflows/21c09563-e6ce-4cf5-84da-1505b416a428/jobs/66399)

rashidakanchwala: is this only a pylint issue? if so, can we ignore it?

ravi-kumar-pilla: yes according to https://github.com/pydantic/pydantic/pull/1881 , we do not need to add __hash__ it is generated for frozen class. Will ignore pylint issue

noklam (approved):
Approve as I don't want to block the PR, there are minor clean up the unnecessary `hash` function.

ravi-kumar-pilla:
> Approve as I don't want to block the PR, there are minor clean up the unnecessary `hash` function.

fyi, I removed the hash function

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: removed hash function

package/kedro_viz/models/flowchart.py
```diff
@@ -61,54 +69,69 @@ class GraphNodeType(str, Enum):
...
+        id (str): Id of the modular pipeline child
+        type (GraphNodeType): Type of modular pipeline child
+    """
 
     id: str
     type: GraphNodeType
 
 
-@dataclass
 class Tag(RegisteredPipeline):
```
rashidakanchwala, line 84: Maybe not related to this PR exactly but I don't understand why Tag is a RegisteredPipeline. Both are different from one another and maybe tag should have it's own id and name, instead of inheriting from RegisteredPipeline

rashidakanchwala:
```
class Tag(BaseModel, frozen=True):
    """Represent a tag"""
    id: str
    name: str
```

package/kedro_viz/models/flowchart.py
```diff
@@ -364,100 +416,174 @@ def outputs(self) -> Set[str]:
...
-    run_command: Optional[str] = field(init=False, default=None)
+    @root_validator(pre=True)
+    def check_task_node_exists(cls, values):
+        assert "task_node" in values
+        cls.set_task_and_kedro_node(values["task_node"])
+        return values
 
-    inputs: List[str] = field(init=False)
+    @classmethod
+    def set_task_and_kedro_node(cls, task_node):
```
rashidakanchwala, line 456: nit - I got a bit confused with the function name. maybe we just call it set_task_node

package/kedro_viz/models/flowchart.py
```diff
@@ -496,189 +622,286 @@ def is_tracking_node(self):
...
     type: str = GraphNodeType.PARAMETERS.value
 
-    def __post_init__(self):
+    @root_validator(pre=True)
+    def check_kedro_obj_and_name_exists(cls, values):
+        assert "kedro_obj" in values
+        assert "name" in values
+        return values
+
+    def __init__(self, *args, **kwargs):
```
rashidakanchwala, line 902: do we need to replace this __init__ with validator?

rashidakanchwala (approved):

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: added a base class called NamedEntity (following the pattern in responses.py) for Tag/Registered Pipeline - https://github.com/kedro-org/kedro-viz/pull/1565/commits/6ce4e37e6d9cdd91cee7f2c4c17343acbcd29358

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: we are using @property to set few values in parameters node. @property values get initialized during __init__ and I am not sure how to use them inside validators. Using __init__ for now

Issue #1548, opened 2023-10-02
Missing Pipeline Name in kedro-viz `Run Command`


drlight0715:
## Description
Pipeline names are not displayed in visuals created by kedro-viz.

## Context
I aim to display a pipeline name in `Run Command` section.

## Steps to Reproduce

1. Set up a kedro pipeline
2. Run `kedro viz`

## Expected Result
A pipeline name should be visible in `Run Command`.

## Actual Result
No pipeline name is displayed in `Run Command`.

![image](https://github.com/kedro-org/kedro-viz/assets/48976896/0a3ee8d5-8788-4c55-804e-b057eb4c9c3f)

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version: Google Chrome
* Operating system and version: Mac OS Ventura 13.6
* NodeJS version used (if relevant):
* Kedro version used (if relevant):  0.18.13
* Python version used (if relevant): 3.9.13

## Checklist
- [x] Include labels so that we can categorise your issue

rashidakanchwala:
I was able to recreate this bug at my end using the demo_project code. 

### Steps to Reproduce
- in demo_project/pipelines/modelling/pipeline.py add 'name' to the split_data function so it will be like below

```python
  split_stage_pipeline = pipeline(
        [
            node(
                func=split_data,
                inputs=["model_input_table", "params:split_options"],
                outputs=test_train_refs,
                name='split_data'
            )
        ]
    )
```

next if you run kedro_viz and click on the split_data function - you should see this. 

![Screenshot 2023-10-02 at 10 26 21](https://github.com/kedro-org/kedro-viz/assets/37628668/c4519127-6197-4bd2-905b-f7054d77805c)


### Expected Result - There should be no None. It should just say the below
` kedro run --to-nodes=None.split_data`

PR #1569, opened 2023-10-11
Issue #1548: Fix improper display of 'run command' text

MehdiNV:
## Description
Resolves https://github.com/kedro-org/kedro-viz/issues/1548

Task nodes that have no namespace, but a name present, will display their run command as `...None.(Node_name)`.

The correct display should just be the `...(Node_name)` only

## Development notes

Purposely made changes very light to avoid unintentional bugs around the code (still unfamiliar with code base, so don't want to make large changes I can't predict the effects of).

Primary change for this PR is just to the text display; tells it to ignore 'None' values. 

## QA notes

- All tests run normally.
- Added new test case where a TaskNode is created w/ a name but no namespace, to check the namespace 'None' will not be included

## Image Example
_Before_
&nbsp;
image

_After_
&nbsp;
image

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -9,7 +9,7 @@ Please follow the established format:
 # Next Release
 
 ## Bug fixes and other changes
-- Fix improper display of 'run-command' inside the metadata panel. (#1548)
+- Fix improper display of 'run-command' inside the metadata panel. (#1569)
 
 # Release 6.6.0
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,11 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+# Next Release
+
+## Bug fixes and other changes
+- Fix improper display of 'run-command' inside the metadata panel. (#1548)
+
 # Release 6.6.0
 
 ## Major features and improvements
--- a/package/kedro_viz/models/flowchart.py
+++ b/package/kedro_viz/models/flowchart.py
@@ -411,6 +411,8 @@ class TaskNodeMetadata(GraphNodeMetadata):
         if kedro_node._name is not None:
             self.run_command = (
                 f"kedro run --to-nodes={task_node.namespace}.{kedro_node._name}"
+                if task_node.namespace is not None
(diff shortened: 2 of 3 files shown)


package/kedro_viz/models/flowchart.py
```diff
@@ -411,6 +411,8 @@ def __post_init__(self, task_node: TaskNode):
         if kedro_node._name is not None:
             self.run_command = (
                 f"kedro run --to-nodes={task_node.namespace}.{kedro_node._name}"
+                if task_node.namespace is not None
+                else f"kedro run --to-nodes={kedro_node._name}"
```
tynandebold, line 415: Is this the way we normally write `if/else` statements in the codebase?

tynandebold (approved):
Nice one! This fixes the problem.

I've left a small question about style and will let @rashidakanchwala or @ravi-kumar-pilla answer it.

package/kedro_viz/models/flowchart.py
merelcht: Yes we do use this style.

merelcht (approved):
LGTM! 👍 
Should this be added to the release notes?

tynandebold:
Yes, good call. Please add a line to the [release notes](https://github.com/kedro-org/kedro-viz/blob/main/RELEASE.md) after creating a new section titled `# Next Release`.

MehdiNV:
> ff0944f

Sounds good! Done just now, [ff0944f](https://github.com/kedro-org/kedro-viz/pull/1569/commits/ff0944f7001cbf06ed92bccf314fb84b4e6bc463)

RELEASE.md
```diff
@@ -6,6 +6,11 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+# Next Release
+
+## Bug fixes and other changes
+- Fix improper display of 'run-command' inside the metadata panel. (#1548)
```
tynandebold, line 12:
```suggestion
- Fix improper display of 'run-command' inside the metadata panel. (#1569)
```

We use the PR number, not the issue number.

MehdiNV: Ah slow day, makes sense!

PR #1572, opened 2023-10-12
Rename any mentions to data set

### LINKED ISSUE #1560: Support for Kedro Datasets 1.7.1 and above
## Description
This ticket aims at replacing all the kedro dataset instances (eg. CSVDataSet) with CSVDataset to support kedro datasets release [1.7.1](https://github.com/kedro-org/kedro-plugins/releases/tag/kedro-datasets-1.7.1)

## Context
This change will bring support for latest kedro datasets on kedro viz

## Possible Implementation

1. Identify all the dataset names that are modified
2. Rename the datasets in imports, requirements files
3. Fix pytests

## Checklist

merelcht:
## Description

Part of https://github.com/kedro-org/kedro/issues/2129

Resolves #1560 

## Development notes
This started out as just renaming any `DataSet` mentions to `Dataset`, but then I discovered "tracking" datasets weren't displayed properly in Kedro-Viz anymore. 

There's a lot of hardcoded references to datasets in kedro-viz which is problematic and the renaming of datasets has surfaced that again. The way the renaming was done is backwards compatible, but not if strings are hardcoded. 

Updating all dataset mentions to the new ending of `Dataset` seems to have fixed the problem. 

Note: I've left the "old" dataset names in place in the demo project to make sure that the old naming still works as well. 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ Once the above set-up is complete, tracking datasets can be used to track releva
 
 ```yaml
 train_evaluation.r2_score_linear_regression:
-  type: tracking.MetricsDataSet
+  type: tracking.MetricsDataset
   filepath: ${base_location}/09_tracking/linear_score.json
   versioned: true
 ```
--- a/cypress/fixtures/graphql/compareThreeRuns.json
+++ b/cypress/fixtures/graphql/compareThreeRuns.json
@@ -2968,7 +2968,7 @@
                     ]
                 },
                 "datasetName": "reporting.feature_importance",
-                "datasetType": "plotly.json_dataset.JSONDataSet",
+                "datasetType": "plotly.json_dataset.JSONDataset",
                 "runIds": [
                     "2022-12-24T21.05.59.296Z",
                     "2022-10-05T12.22.35.825Z",
@@ -3050,7 +3050,7 @@
                     ]
                 },
                 "datasetName": "train_evaluation.linear_regression.r2_score",
-                "datasetType": "tracking.metrics_dataset.MetricsDataSet",
+                "datasetType": "tracking.metrics_dataset.MetricsDataset",
                 "runIds": [
                     "2022-12-24T21.05.59.296Z",
                     "2022-10-05T12.22.35.825Z",
@@ -3104,7 +3104,7 @@
                     ]
                 },
                 "datasetName": "train_evaluation.random_forest.r2_score",
-                "datasetType": "tracking.metrics_dataset.MetricsDataSet",
+                "datasetType": "tracking.metrics_dataset.MetricsDataset",
                 "runIds": [
                     "2022-12-24T21.05.59.296Z",
                     "2022-10-05T12.22.35.825Z",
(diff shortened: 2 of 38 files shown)


ravi-kumar-pilla:
Attaching kedro viz story created - This PR also resolves #1560

tynandebold (commented):
Wow, so many changes. Thank you!

I still see a fair amount of `DataSet` mentions when I do a project-wide search:

image

Do we need to replace those? We should for at least the demo-project tracking datasets, otherwise the data doesn't show up in experiment tracking and instead is blank:

image

tynandebold:
> > We should for at least the demo-project tracking datasets, otherwise the data doesn't show up in experiment tracking and instead is blank:
> > image
> 
> I deliberately didn't change these, because it should actually work without the rename.. When I run my branch locally on the project I see:
> 
> Screenshot 2023-10-13 at 13 19 49

Got it. I had to update `kedro-datasets` to the latest version. It didn't work on my previous one, which was version `1.5.1`. I think that's the expected behavior?

tynandebold (approved):

ravi-kumar-pilla:
Hi @merelcht , Thank you for this. 

I see ImageDataSet has not changed its type to ImageDataset. Is there a list of left-over datasets that are yet to be changed ? 
image


I see the warnings mention - CSVDataSet, ExcelDataSet, ParquetDataSet, PickleDataSet, PlotlyDataSet, JSONDataSet, MetricsDataSet (which might not be the entire list of changed datasets)

merelcht:
> Got it. I had to update kedro-datasets to the latest version. It didn't work on my previous one, which was version 1.5.1. I think that's the expected behavior?

@tynandebold Yes that is the expected behaviour. We could also keep the references to both datasets with `DataSet` endings and `Dataset` ending, but `kedro-datasets` `1.7.1` is a non-breaking release so based on pins in our own requirements users should get this more up to date version. From the kedro/kedro-datasets side it doesn't matter, so I leave the choice up to you if you prefer to keep both references for now and only remove later.

merelcht:
> I see ImageDataSet has not changed its type to ImageDataset. Is there a list of left-over datasets that are yet to be changed ? image
> 
> I see the warnings mention - CSVDataSet, ExcelDataSet, ParquetDataSet, PickleDataSet, PlotlyDataSet, JSONDataSet, MetricsDataSet (which might not be the entire list of changed datasets)

@ravi-kumar-pilla I haven't renamed anything inside the `demo-project` yet as a check that the old naming, when it's used in a Kedro project, still works with `kedro-viz`. If the team prefers it, I can rename that now as well or do it later when the old names are actually fully removed.

ravi-kumar-pilla:
Also, there seems to be an issue with trackingDataset in experiment tracking when using kedro-datasets 1.7.1. I am yet to debug the issue but @merelcht do you think we should fix this in this PR or create a different issue for this ? Thank you

merelcht:
> Also, there seems to be an issue with trackingDataset in experiment tracking when using kedro-datasets 1.7.1. I am yet to debug the issue but @merelcht do you think we should fix this in this PR or create a different issue for this ? Thank you

Yes this fixes it. 

About the `ImageDataSet`, this is because this is a custom dataset in the demo project and I didn't want to rename anything in that project yet.

Issue #1582, opened 2023-10-16
Resolve Python dependency PRs


tynandebold:
## Description

We have a number of open dependabot PRs that we should resolve. They are:

- [x] #1483 
- [x] #1518 
- [x] #1546 
- [x] #1547 
- [x] #1552 
- [x] #1561 

This issue should only be closed when all of those PRs are closed, either by merging and accepting the update or closing them because they're not needed.

## Checklist
- [x] Include labels so that we can categorise your issue

astrojuanlu:
Comment from the peanut gallery: since Kedro-Viz dependabot is using the default configuration, `versioning-strategy` is set to `auto`, meaning 

> Try to differentiate between apps and libraries. Use `increase` for apps and `widen` for libraries.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

and since Kedro-Viz uses `requirements.txt` for its Python requirements, it's probably applying the `increase` strategy, which is arguably too aggressive for a component like Kedro-Viz. In fact, since Kedro switched to `pyproject.toml`, dependabot PRs have been using the `widen` strategy.

Viz can either

- Do nothing and keep aggressively upgrading dependencies, but this can lead to problems with people that have environments with lots of dependencies.
- Configure dependabot to use `widen`.
- Migrate dependencies to `pyproject.toml` #1527 and keep the `auto` config.

tynandebold:
Thanks for the input. I think we'll go with your third bullet point above (Migrate dependencies to pyproject.toml). I've pulled that issue into this sprint now, so when that's done I hope these upgrades quiet down a bit.

MehdiNV:
The `pyproject.toml` ticket may take a while due to its complexity, so going to progress this ticket forward so that it's not blocked

PR #1584, opened 2023-10-16
Migrate from '...requirements.txt' to 'pyproject.toml'

### LINKED ISSUE #1527: Move to pyproject.toml
## Description

Kedro-viz currently has `requirements.txt` file which the users use to install all the requirements. We want to now move to pyproject.toml which is a newer way of dependency management.

Pyproject.toml has some advance features that will also make our depandabot much more .... smarter? ;) 

## Checklist

MehdiNV:
## Description
Draft PR, TBA 

Resolves #1527 

## What's missing for the Draft PR? 
- Making sure that setup.py has a 'package_data' parameter value passed into it

## TL:DR 
TBC

## Development notes
TBC

## QA notes
TBC

## What's next?
TBC


package/setup.py
```diff
@@ -1,53 +1,47 @@
...
-here = path.abspath(path.dirname(__file__))
-
-
-jsbuild = [
-    os.path.join(dirpath, f)
-    for dirpath, dirnames, files in os.walk("kedro_viz/html/")
-    for f in files
+from setuptools import setup
+
+extras_require["test"] = [
```
astrojuanlu, line 4: You can have these under 

```toml
[project.optional-dependencies]
test = [
    ...
]
```

see https://github.com/kedro-org/kedro-plugins/blob/b946eec6a4e5979a3302e1a682c006a4411bddaa/kedro-airflow/pyproject.toml#L25-L40 for example

package/pyproject.toml
```diff
@@ -0,0 +1,114 @@
...
+
+[project]
+name = "kedro-viz"
+authors = [
+    {name = "Kedro"}
+]
+description = "Kedro-Viz helps visualise Kedro data and analytics pipelines"
+requires-python = ">=3.7"
+dependencies = [
+    "semver>=2.10",  # Needs to be at least 2.10.0 to make use of `VersionInfo.match`.
```
ravi-kumar-pilla, line 15: we have replaced semver with packaging, so you can remove this dependency

package/pyproject.toml
```diff
@@ -0,0 +1,119 @@
...
+
+[tool.setuptools]
+zip-safe = false
+
+[tool.setuptools.packages.find]
+include = ["kedro_viz*"]
+exclude=["tests*", "features*"]
+
+[tool.setuptools.dynamic]
+readme = {file = "README.md", content-type = "text/markdown"}
```
ravi-kumar-pilla, line 89: unable to read the README file as it is in a different directory. We need to consider following the directory structure of other kedro-plugins to be consistent since kedro-viz is also a plugin. 

@rashidakanchwala @noklam 

Our directory structure - `kedro-viz/package/kedro_viz`
Other plugins have (example) - `kedro-telemetry/kedro_telemetry` . 

Is it better we restructure our directory structure to `kedro-viz/kedro_viz` ?

package/setup.py
```diff
@@ -17,37 +16,6 @@
...
-    long_description=readme,
-    long_description_content_type="text/markdown",
-    license="Apache Software License (Apache 2.0)",
-    url="https://github.com/kedro-org/kedro-viz",
-    python_requires=">=3.7",
-    install_requires=requires,
-    keywords="pipelines, machine learning, data pipelines, data science, data engineering, visualisation",
-    author="Kedro",
-    packages=find_packages(exclude=["tests*", "features*"]),
     package_data={"kedro_viz": list(files)},
```
ravi-kumar-pilla, line 20: @MehdiNV I tested removing the package_data and also deleted the setup.py, setup.cfg completely. I was able to install editable package via `pip install -e package` and also was able to run `kedro viz`. I am not sure where the jsbuild is used tbh. 

@astrojuanlu - is the jsbuild used during publishing the package ? We execute the command `make package` and the below script runs -
```makefile
package:
	find . -regex ".*/__pycache__" -exec rm -rf {} +
	find . -regex ".*\.egg-info" -exec rm -rf {} +
	test -f package/kedro_viz/html/index.html || (echo "Built npm package not found; packaging process cancelled."; exit 1)
	cd package && python setup.py clean --all
	cd package && python setup.py sdist bdist_wheel
``` 
@MehdiNV we might need to change these steps as well if we get rid of setup.py completely

@astrojuanlu - how can we test if the migration is success, apart from installing the package locally ? We also need to test the `make package` command works and the package includes everything as before ?

Thank you

MehdiNV: @ravi-kumar-pilla I think setup.py is only needed as a fallback for users who cannot utilise pyproject.toml for one reason or another ([see stackoverflow post on it](https://stackoverflow.com/questions/73310069/should-i-be-using-only-pyproject-toml)), what do you think about keeping it as a plan B? Or perhaps it's not necessary as it's very unlikely a user will have incompatability w/ .toml?

ravi-kumar-pilla: I think we can migrate the requirements.txt, test_requirements.txt, setup.cfg and part of setup.py (excluding the readme and package data) to pyproject.toml. I want to discover more on this and see how can we test the setup locally and how it affects our release process of `make package ` I mentioned above. Will work on this today !

package/pyproject.toml
astrojuanlu: It's tricky in your case because there's a separation between the backend and the frontend 🤔 `../README.md` doesn't work?

package/setup.py
astrojuanlu: I think the way to test if the migration is successful is producing the final `.tar.gz` and `.whl` files that go to https://pypi.org/project/kedro-viz/#files, with and without this PR, and making sure that the relevant files are identical

package/pyproject.toml
ravi-kumar-pilla: yes it throws an error saying toml cannot access any files outside the directory it is in

astrojuanlu: you can use `setup.py` for reading the `README.md`, then setuptools will merge whatever is on `setup.py` with `pyproject.toml`. see what  `kedro-datasets` does for example

package/setup.py
ravi-kumar-pilla: Hi @astrojuanlu , 

Thank you for the input. I tested them and they seem to be the same. I have a question on installing dependencies. 

Previously when we used to install dependencies of kedro-viz via requirements.txt file, kedro-viz was not installed as a package. However, when installing the same dependencies via pyproject.toml, the list of installed dependencies contain kedro-viz. 

I checked this for kedro-datasets and even kedro-datasets get installed as dependency when we install its dependencies from pyproject.toml. Is this normal ?

Thank you

astrojuanlu: Yes, because there is no such a thing as `pip install . --only-deps` https://github.com/pypa/pip/issues/11440

If you want to still be able to install only the dependencies but _not_ kedro-viz, you should probably keep `requirements.txt` around and read it from `pyproject.toml`, declaring it as dynamic https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata 

```toml
dependencies = {file = "requirements.text"}}
```

(didn't test this syntax)

astrojuanlu: It's actually `dependencies = {file = "requirements.txt"}`

ravi-kumar-pilla: Hi @astrojuanlu , Thank you for the suggestion. 

I have a question regarding keeping the requirements file as is. Was the whole idea was to move away from the requirements file ? What is the advantage of using pyproject.toml and then loading the deps from requirements file. Thank you

astrojuanlu: This:

> If you want to still be able to install only the dependencies but not kedro-viz

astrojuanlu: Caveat: https://github.com/kedro-org/kedro-viz/issues/1582#issuecomment-1764451560

ravi-kumar-pilla: Since we are not moving to pyproject.toml completely, I do not see why this migration is needed. For handling the caveat mentioned, I feel we should go by the option - `Configure dependabot to use widen.` and migrate to pyproject.toml when the ticket [pypa/pip#11440](https://github.com/pypa/pip/issues/11440) is merged. What do you think of this approach ? @astrojuanlu @tynandebold

astrojuanlu: I'm supportive of that 👍🏽

MehdiNV: ^ This makes sense; the .toml file is now linked to a `requirements.txt` file containing the dependencies, however I've kept the `test_requirements.txt` file deleted (installing test dependencies is via .toml). 

In my head, a user should have no need for the test-dependencies unless they need to use / install kedro-viz as well (in which case, they'd use the .toml fully anyway). Is this a suitable approach?

ravi-kumar-pilla: @MehdiNV we will not go ahead with this draft PR. Instead, we will use the option of `Configuring dependabot to use widen.` to handle - https://github.com/kedro-org/kedro-viz/issues/1582#issuecomment-1764451560

MehdiNV: Ahhh makes sense; apologies, I hadn't reloaded the page before so hadn't seen the new comments in the last hour when I was replying. Will shift towards the widen option then - will close this PR in the meantime as it's redundant

astrojuanlu:
If I may, migrating from `setup.py` to `pyproject.toml`, even if it's partially, is still worth it. But you're right that if you want to keep `requirements.txt` around to be able to install only the dependencies, it's not as useful.

PR #1587, opened 2023-10-18
Add forward compatibility with renames `_dataset` arguments.

merelcht:
## Description
In https://github.com/kedro-org/kedro/issues/3041 we renamed any mentions of `data_set`/`_data_set` to `dataset` and `_dataset` respectively. This was merged to the `develop` branch, because it's a breaking change. For the new add_ons work we need to build new code inside the starters against the Kedro `develop` branch, but the starters also rely on Kedro-Viz and so the change in this PR makes sure that the starters can use Kedro `develop` as well as viz with a "forward compatible" mention of the renamed arguments. 

## Development notes
It's not possible yet to just rename the arguments inside of Kedro-Viz, because that requires a release of the Kedro `develop` branch. 

## Checklist

### DIFF
--- a/package/kedro_viz/data_access/repositories/catalog.py
+++ b/package/kedro_viz/data_access/repositories/catalog.py
@@ -78,12 +78,16 @@ class CatalogRepository:
 
         self._layers_mapping = {}
 
+        try:
+            datasets = self._catalog._data_sets
+        except:
+            datasets = self._catalog._datasets
         # Maps layers according to the old format
         if KEDRO_VERSION < parse("0.19.0"):
             if self._catalog.layers is None:
                 self._layers_mapping = {
                     _strip_transcoding(dataset_name): None
-                    for dataset_name in self._catalog._datasets
+                    for dataset_name in datasets
                 }
             else:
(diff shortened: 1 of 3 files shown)


SajidAlamQB (approved):
Looks great, Thank you! @merelcht

rashidakanchwala (approved):
I see you mentioned this temporary for now. But in future, given that Kedro-viz needs to be backward compatible with Kedro -- would we continue to have the try catch block with both _data_sets and _dataset

merelcht:
> I see you mentioned this temporary for now. But in future, given that Kedro-viz needs to be backward compatible with Kedro -- would we continue to have the try catch block with both _data_sets and _dataset

In `viz` you already have logic for pre-0.19 versions: https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/data_access/repositories/catalog.py#L82 which should be updated with the new naming,
but since our `develop` branch doesn't have the bumped version yet it doesn't solve the compatibility right now.

tynandebold (approved):

PR #1594, opened 2023-10-20
Add heap parameter to docs script for analytics

stichbury:
## Description

My attempt to set up Heap. Needs to be verified by someone who understands what they're doing.

## Checklist


docs/source/_templates/layout.html
```diff
@@ -32,6 +32,9 @@
 {%- block extrahead %}
 <script type="text/javascript">
   window.heap=window.heap||[],heap.load=function(e,t){window.heap.appid=e,window.heap.config=t=t||{};var r=document.createElement("script");r.type="text/javascript",r.async=!0,r.src="https://cdn.heapanalytics.com/js/heap-"+e+".js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(r,a);for(var n=function(e){return function(){heap.push([e].concat(Array.prototype.slice.call(arguments,0)))}},p=["addEventProperties","addUserProperties","clearEventProperties","identify","resetIdentity","removeEventProperty","setEventProperties","track","unsetEventProperty"],o=0;o<p.length;o++)heap[p[o]]=n(p[o])};
+
+  heap.load({{ "522942930" }});
```
astrojuanlu, line 36: I think to mirror what's happening on Kedro, we'd need to do something like

https://github.com/kedro-org/kedro/blob/c3c93cb4786b4fd38c62baa2ec7f6fc73a63791e/docs/source/_templates/layout.html#L6

and then 

https://github.com/kedro-org/kedro/blob/c3c93cb4786b4fd38c62baa2ec7f6fc73a63791e/docs/source/conf.py#L488-L518

I can take a stab at this but not today

stichbury: Thanks! I don't believe we need `env_override` though, do we? I think it's used to differentiate `stable` and `latest` but for these two subprojects I don't see that as particularly desirable or important.

stichbury:
@astrojuanlu @tynandebold Please can I check on the progress of this in my absence, if any?

tynandebold:
No progress from me. I've just added it to our new sprint, so we're tracking it now and will get to it soon.

docs/source/_templates/layout.html
tynandebold: I think it's used to differentiate between `development` and `production`, which is more important. I would argue we probably don't need that though. The number of us who'll be previewing the Viz docs and inflating the number of "real" users will be near zero.

tynandebold (approved):
This is working as it should be for me. I noted in a comment we're not switching between dev and prod environments, and I don't think we need to.

Here are the events coming through:

Screenshot 2023-11-01 at 15 34 06

PR #1602, opened 2023-10-24
Adjust Dependabot versioning strategy

MehdiNV:
## Description

Currently dependabot uses `auto` as its versioning strategy (which effectively becomes `increase` for us), which is a too aggressive for our liking for a component like Kedro-Viz. 

This PR adjusts it to use `increase-if-necessary` to reduce the aggressiveness and ensure dependancies are only increased if necessary. Makes dependency management a more relaxing task for us going forward.

> **All of the possible versioning strategies available, regardless of package ecosystem**

image

> **Ideally we would've used `widen`, however this strategy is unavailable for `pip`:**

image

> **Hence the decision to go forward with increase-if-necessary, as it behaves the closest and works with pip. Example of how each strategy works below**

image

## Development notes

Added `versioning strategy: increase-if-necessary` to dependabot's .yml configuration. Without this line, dependabot would have defaulted to `auto` if no versioning strategy was specified to it.

## QA notes

No major behavioural changes - however going forward, dependabot should try to not be as aggressive when it comes to upgrading dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,7 @@ updates:
     directory: "/package" # Location of package manifests
     schedule:
       interval: "monthly"
+    versioning-strategy: increase-if-necessary
     ignore:
       - dependency-name: "types-*"
     labels:


RELEASE.md
```diff
@@ -6,6 +6,12 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+# Release 6.6.2
+
+## Major features and improvements
+
+- Reduce aggressiveness of dependabot's versioning strategy (#1602)
```
tynandebold, line 13: We don't need this in the release notes. This information should only be user focused, and this wouldn't fall into that realm.

tynandebold (approved):
Looks good. One comment about the release notes, otherwise nice one.

RELEASE.md
MehdiNV: Ah that makes sense! Good call, will remove then yeah

PR #1355, opened 2023-05-12
Add improved security headers to FastAPI responses

tynandebold:
## Description

Resolves #1348.

## Development notes

I've added and am using the [secure](https://pypi.org/project/secure/) package to add the necessary security headers infosec requested, those headers being:

- `Strict-Transport-Security`
- `X-Frame-Options`

## QA notes

You need to run the app locally or with Gitpod and view the main server response.

This is before the change:

Screenshot 2023-05-12 at 09 07 33

This is after:

Screenshot 2023-05-12 at 09 06 28

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1355"></a>

Issue #1348, opened 2023-05-02
Add necessary security headers to demo.kedro.org


tynandebold:
## Description

- Return an appropriate Strict-Transport-Security HTTP header in application responses.
- Return appropriate X-Frame-Options and/or Content-Security-Policy HTTP headers in application responses.
- To prevent yourself from email spoofing, you should configure the SPF, DKIM, and DMARC records appropriately. These are the higher standards of authentication.

## Checklist

- [x] Include labels so that we can categorise your feature request

PR #1658, opened 2023-11-28
Fix Kedro-viz embedded as an IFrame

rashidakanchwala:
## Description

In response to issue #1348, which required the addition of security headers to demo.kedro.org, we implemented a solution in PR #1355. This solution involved adding security headers to the FastAPI application which results in all instances of kedro-viz, whether hosted or local, having these security headers. Having the security headers introduced a limitation where kedro-viz could not be used as an IFrame, affecting functionalities like %run_viz that embed kedro-viz in an iframe.

To address this, the current ticket introduces a conditional approach. We will add security headers only if the environment variable ADD_SECURITY_HEADER is set to true. This modification will be implemented in the Dockerfile when creating the docker image for the demo project. This image will then be uploaded to an EC2 instance and deployed using Lightsail.

## QA notes

I've conducted local tests and confirmed that security headers are no longer being added. Additionally, I tested this by creating a Docker image of the demo-project. In this scenario, the security headers are present and functioning as expected. 

## Checklist

### DIFF
--- a/demo-project/Dockerfile
+++ b/demo-project/Dockerfile
@@ -8,3 +8,5 @@ COPY . /code
 RUN pip install --no-cache-dir --upgrade -r /code/src/docker_requirements.txt
 
 CMD ["kedro", "viz", "--host", "0.0.0.0", "--port", "4141", "--no-browser"]
+
+ENV ADD_SECURITY_HEADERS=true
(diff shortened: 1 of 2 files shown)


ravi-kumar-pilla (approved):
Awesome !! Looks good to me and works well with local and jupyter notebook %run_viz line magic.

PR #1659, opened 2023-11-28
Fix for dataset existence check in factory pattern discovery

### LINKED ISSUE #1645: Disable dataset existence checks for unavailable sources
## Description
I'm having issues using version 6.7.0. I have a source defined in my catalog that I access through a VPN. I only use this source for data extraction, so I don't always have the VPN enabled.

When I try to start the Kedro Viz server without the VPN enabled, the following exception is raised:

```
  File "/usr/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/kedro/.venv/lib/python3.10/site-packages/kedro_viz/server.py", line 100, in run_server
    populate_data(
  File "/kedro/.venv/lib/python3.10/site-packages/kedro_viz/server.py", line 40, in populate_data
    data_access_manager.resolve_dataset_factory_patterns(catalog, pipelines)
  File "/kedro/.venv/lib/python3.10/site-packages/kedro_viz/data_access/managers.py", line 85, in resolve_dataset_factory_patterns
    catalog.exists(dataset_name)
  File "/kedro/.venv/lib/python3.10/site-packages/kedro/io/data_catalog.py", line 404, in exists
    return dataset.exists()
  File "/kedro/.venv/lib/python3.10/site-packages/kedro/io/core.py", line 313, in exists
    raise DatasetError(message) from exc
kedro.io.core.DatasetError: Failed during exists check for data set SQLTableDataSet(load_args={}, save_args={'index': False}, table_name=community2collection).
(psycopg2.OperationalError) connection to server at "postgres" (xxx.xxx.xxx.xxx), port xxxx failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?

```

## Context
I understand that the error is due to the feature introduced in this #1472 . More precisely, the error occurs here 

https://github.com/kedro-org/kedro-viz/blob/918997354704741237ced336833f8b54910d1fce/package/kedro_viz/server.py#L99-L102

It seems that the issue stems from Kedro attempting to check the existence of all data sources to display statistics. However, due to the lack of access, not only does it prevent viewing statistics, but it also hinders the ability to use the tool for visualizing pipelines. This occurs regardless of whether there is access to the data sources or not, and it was not an issue in versions prior to 6.7.0

Is there a way to disable this feature? I have no issues with version 6.6.1.

## Steps to Reproduce

1. Define a source of type pandas.SQLTableDataSet to which you do not have access (at least for testing purposes).
2. Start Kedro Viz

ravi-kumar-pilla:
## Description

Resolves #1645 

## Development notes

* Add exception handler to handle DatasetError caused during dataset_factory_pattern discovery

## QA notes

* All tests should pass

## Checklist

### DIFF
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -38,8 +38,6 @@ def populate_data(
         session_class = make_db_session_factory(session_store.location)
         data_access_manager.set_db_session(session_class)
 
-    # resolve dataset factory patterns, add catalog
-    # and add relevant tracking datasets
     data_access_manager.add_catalog(catalog, pipelines)
 
     # add dataset stats before adding pipelines as the data nodes
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -36,11 +36,9 @@ def populate_data(
         session_class = make_db_session_factory(session_store.location)
         data_access_manager.set_db_session(session_class)
 
-    # resolve the dataset factory patterns
-    data_access_manager.resolve_dataset_factory_patterns(catalog, pipelines)
-
-    # add catalog and relevant tracking datasets
-    data_access_manager.add_catalog(catalog)
+    # resolve dataset factory patterns, add catalog
+    # and add relevant tracking datasets
+    data_access_manager.add_catalog(catalog, pipelines)
 
     # add dataset stats before adding pipelines as the data nodes
     # need stats information and they are created during add_pipelines
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -5,6 +5,11 @@ Please follow the established format:
 - Use present tense (e.g. 'Add new feature')
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
+# Upcoming Release
+
+## Bug fixes and other changes
+
+- Fix for dataset existence check in factory pattern discovery (#1659)
 
 # Release 6.7.0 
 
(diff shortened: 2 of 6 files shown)


ankatiyar (approved):
LGTM!

rashidakanchwala (approved):
thanks!

rashidakanchwala (approved):
LGTM. thanks!

package/kedro_viz/server.py
```diff
@@ -36,11 +36,9 @@ def populate_data(
         session_class = make_db_session_factory(session_store.location)
         data_access_manager.set_db_session(session_class)
 
-    # resolve the dataset factory patterns
-    data_access_manager.resolve_dataset_factory_patterns(catalog, pipelines)
-
-    # add catalog and relevant tracking datasets
-    data_access_manager.add_catalog(catalog)
+    # resolve dataset factory patterns, add catalog
```
rashidakanchwala, line 39: we can remove this comment.

ravi-kumar-pilla: okay, but add_catalog method does all the 3 functions right ?

ravi-kumar-pilla: you want me to remove entire comment as we have doc strings ?

PR #1588, opened 2023-10-18
Fix dataset factory patterns in Experiment Tracking

ravi-kumar-pilla:
## Description

Resolves #1480 

## Development notes

* Discovering datasets defined via dataset factory pattern using datasets in pipelines before adding catalog and tracking datasets to kedro viz
* Inspired from the PR - #1491 

Example patterns -

Screenshot 2023-10-18 at 1 11 46 PM

## QA notes

* [Steps to Reproduce](https://github.com/kedro-org/kedro-viz/issues/1480)
* You can also follow the dataset factory patterns [guide](https://docs.kedro.org/en/stable/data/kedro_dataset_factories.html#kedro-dataset-factories) and modify the demo-project or create a spaceflights project
* After executing `kedro run`, you should see the factory pattern datasets in the experiment tracking run

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added new entries to the `RELEASE.md` file
- [x] Added tests to cover my changes

PR #1696, opened 2024-01-02
Set Kedro Viz start method to spawn process while launching it from Jupyter Notebook

### LINKED ISSUE #1656: SparkContext not reinitialized due to forked process start
## Description
In the latest release the following was added:
[`data_access_manager.resolve_dataset_factory_patterns(catalog, pipelines)`](https://github.com/kedro-org/kedro-viz/blob/86243d0f299cc2e718aceaf830ba7c5dd5ef3b91/package/kedro_viz/server.py#L40)

This forces a check of exists for all datasets, which is good but caused the following issue:

With SparkDatasets only when starting it through Jupyter Notebooks.
The issue being that the Spark Context gets initialised in this scenario on the Kedro context loading (`%load_ext kedro.ipython`) on a process.
Now when `%run_viz` gets called you are using multiprocessing to spin up a new forked process for unix only, since that's default. On Windows and MacOS a completely new process is default.

```python
    viz_process = multiprocessing.Process(
        target=run_server,
        daemon=True,
        kwargs={"project_path": project_path, "host": host, "port": port},
    )
```
[Link](https://github.com/kedro-org/kedro-viz/blob/86243d0f299cc2e718aceaf830ba7c5dd5ef3b91/package/kedro_viz/launchers/jupyter.py#L110)

What ends up happening is that due to it being a forked process, Spark still thinks its already initialised but can't access it since its on a different process.

The solution is to add the following, as it will force a fresh process instead of a forked one.
```python
    multiprocessing.set_start_method("spawn")
    viz_process = multiprocessing.Process(
        target=run_server,
        daemon=True,
        kwargs={"project_path": project_path, "host": host, "port": port},
    )
```

## Context
Can we add this change or will it break some other functionality you guys are using forked processes for?

ravi-kumar-pilla:
## Description

Resolves #1656 

## Development notes

* Add multiprocess start method to spawn
* Add mypy type ignore for `add_route` and `add_websocket_route` due to strict mypy check introduced [here](https://github.com/encode/starlette/pull/2180) in starlette

## QA notes

* All tests should pass
* Launch Kedro Viz via [jupyter notebook](https://docs.kedro.org/en/latest/notebooks_and_ipython/kedro_and_notebooks.html#)
* Run line magic %run_viz

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -5,6 +5,12 @@ Please follow the established format:
 - Use present tense (e.g. 'Add new feature')
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
+# Upcoming Release
+
+## Bug fixes and other changes
+
+- Set Kedro Viz start method to spawn process while launching it from Jupyter Notebook. (#1696)
+
 # Release 7.0.0
 
 ## Major features and improvements
--- a/package/kedro_viz/api/graphql/router.py
+++ b/package/kedro_viz/api/graphql/router.py
@@ -1,4 +1,5 @@
 """`kedro_viz.api.graphql.router` defines GraphQL routes."""
+# mypy: ignore-errors
 from fastapi import APIRouter
 from strawberry.asgi import GraphQL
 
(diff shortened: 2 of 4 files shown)


noklam (commented):
Looks fine to me, but my question is does it matter to viz to have the Spark context initialised? Is it being used to read data?

rashidakanchwala:
> Looks fine to me, but my question is does it matter to viz to have the Spark context initialised? Is it being used to read data?

Ravi would have more clarity on this but from what it looks like is that if Spark context is not intiatilised; the spark dataset is loaded as a MemoryDataSet on viz.

ravi-kumar-pilla:
> Looks fine to me, but my question is does it matter to viz to have the Spark context initialised? Is it being used to read data?

Hi Nok, It should not matter to viz but in Viz 6.7.0 we introduced a [dataset exists check](https://github.com/kedro-org/kedro-viz/pull/1588) which forced the discovery of all the datasets. Since on unix forked process is default, this check caused an issue as viz process is a copy. Starting Viz as a new process would not have this dependency. The issue is explained [here](https://github.com/kedro-org/kedro-viz/issues/1656) in detail.

Also, for your information the dataset exists check was removed in v7.0.0 as the check introduced other time out issues. So the Spark re-initialization issue should be solved in v7.0.0. However, this change will force viz to always start as a new process irrespective of the OS.

Thank you

merelcht (approved):
LGTM 👍

Issue #1615, opened 2023-10-30
Kedro Viz Build 


amandakys:
## Description
We want to provide the `kedro viz build` command to give users access to a build directory that they can then deploy using any static site hosting provider of their choice. 

The 'backend' of this work is already done to enable the AWS focused Hosted Viz feature. It works by 
1. assembling all the necessary build files and generating static data files 
2. sending this to a s3 bucket so it can be hosted as a static site.  

For this feature, we would interrupt after step 1, and return the build files in a /build directory inside the project. 

## Acceptance Criteria 
- [ ] working `kedro viz build` command that creates the build files in the `/build` directory 
- [ ] running `kedro viz build` again will **overwrite** the existing contents of `/build` 
- [ ] if a `/build` directory doesn't already exist, create one 
- [ ] update the `kedro viz` help text to include the `build` command 
 `build       Create build directory of local Kedro Viz instance with static data `

PR #1697, opened 2024-01-04
CLI command kedro viz build added

jitu5:
## Description

Resolves [#1615](https://github.com/kedro-org/kedro-viz/issues/1615). Add `kedro viz build` to build directory of local Kedro Viz instance with static data.

## Development notes

- Utilized a combination of inheritance and the factory pattern. 
- Introduced a base class called "BaseDeployer" which encapsulates the core deployment logic.
- Created two specialized child classes: "LocalDeployer" and "S3Deployer: 
  - "LocalDeployer" is responsible for copying the build and API folders to the kedro-project folder locally without cloud deployment.
  - "S3Deployer" handles the task of uploading the same content to an S3 bucket.

Adopted this design pattern with the flexibility to easily extend to other cloud platforms like Azure and GCP in the future.

## QA notes

- Install editable kedro viz via `pip install -e package`
- `cd demo-project`
- Run `kedro viz build`

## Checklist

- [X] Updated the documentation to reflect the code changes
- [X] Added new entries to the `RELEASE.md` file


package/kedro_viz/launchers/cli.py
```diff
@@ -270,3 +273,49 @@ def deploy(region, bucket_name):
         )
     finally:
         viz_deploy_timer.terminate()
+
+
+@viz.command(context_settings={"help_option_names": ["-h", "--help"]})
+def build():
```
ravi-kumar-pilla, line 279: I am not sure if we just want to copy the static build folder of kedro viz. Ideally we should create a build folder similar to what we copy to S3, something like [this](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/integrations/deployment/s3_deployer.py#L112) and use it as a base folder for all the cloud providers

```python
self._upload_api_responses()
self._upload_static_files(_HTML_DIR)
self._upload_deploy_viz_metadata_file()
 ```

File: README.md [hunk 2]
```diff
@@ -147,6 +147,12 @@ Options:
   -h, --help          Show this message and exit.
 ```
 
+To create build directory of local Kedro Viz instance with static data from the command line, use the following command from the root folder of your Kedro project:
```
tynandebold, line 150:
```suggestion
To create a build directory of your local Kedro-Viz instance with static data from the command line, use the following command from the root folder of your Kedro project:
```

package/kedro_viz/integrations/deployment/base_deployer.py
```diff
@@ -0,0 +1,108 @@
...
+        injected_head_content.append("</head>")
+        html_content = html_content.replace("</head>", "\n".join(injected_head_content))
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_file_path = f"{temp_dir}/index.html"
+
+            with open(temp_file_path, "w", encoding="utf-8") as temp_index_file:
+                temp_index_file.write(html_content)
+
+            self._fs.put(temp_file_path, f"{self._path}/")
```
SajidAlamQB, line 70: Minor Nit: could we not do this directly in the path instead of copying from temp for efficiency?

SajidAlamQB (approved):
This is awesome work @jitu5! 🌟 

The implementation looks great just left a minor comment.

package/kedro_viz/integrations/deployment/base_deployer.py
```diff
@@ -0,0 +1,108 @@
...
+                "timestamp": datetime.utcnow().strftime("%d.%m.%Y %H:%M:%S"),
+                "version": str(parse(__version__)),
+            }
+            with self._fs.open(f"{self._path}/{_METADATA_PATH}", "w") as metadata_file:
+                metadata_file.write(json.dumps(metadata))
+        except Exception as exc:  # pragma: no cover
+            logger.exception("Upload failed: %s ", exc)
+            raise exc
+
+    def _deploy(self):
```
merelcht, line 101: To me it seems more natural to not make this method "private" now it's called from the individual deployer classes.

package/kedro_viz/launchers/cli.py
```diff
@@ -270,3 +270,25 @@ def deploy(region, bucket_name):
...
+        click.echo(
+            click.style(
+                "\u2728 Success! Kedro-Viz build files have been successfully added to the "
+                f"{url} directory.",
+                fg="green",
+            )
+        )
+
+    except Exception as ex:
+        traceback.print_exc()
```
merelcht, line 293: What kind of information will show up in this traceback? Is that easy to read/understand for users?

merelcht (commented):
I left some small comments, but generally the implementation looks very neat to me 🙂

package/kedro_viz/launchers/cli.py
jitu5: Yes, It won't be easy to read and understand for users. I updated the Exception part, Thanks.

package/kedro_viz/integrations/deployment/base_deployer.py
jitu5: I updated the "_deploy" method, Thanks.

rashidakanchwala: actually we exposed the deploy_and_get_url() as public method. _deploy is private method called from deploy_and_get_url()

package/kedro_viz/integrations/deployment/base_deployer.py
rashidakanchwala: I can recollect, we had to do it this way because we had some issues getting this to work with s3fs

package/kedro_viz/integrations/deployment/base_deployer.py
merelcht: @rashidakanchwala Yes I remember this from the original implementation, but now `deploy_and_get_url` is an abstract method that must be overwritten in class implementations and I see these in turn call `deploy` so I think now it makes sense to not make it private anymore.

README.md
```diff
@@ -147,6 +147,12 @@ Options:
   -h, --help          Show this message and exit.
 ```
 
+To create a build directory of your local Kedro-Viz instance with static data from the command line, use the following command from the root folder of your Kedro project:
+
+```bash
+kedro viz build
+```
+
```
noklam, line 155: Should we have some docs explaining how to view this build?

noklam (approved):
Thanks! This is a really nice implementation and very close to what I had in mind. It should be relative straightforward to expand the deploy command and add other targets from here. 

Great work!

README.md
rashidakanchwala: yes, in another PR we will focus on the build docs

package/kedro_viz/integrations/deployment/base_deployer.py
jitu5: @merelcht Could you please review again, I renamed it to "deploy" method. Thanks.

package/kedro_viz/integrations/deployment/deployer_factory.py
```diff
@@ -0,0 +1,18 @@
...
+    """A class to handle creation of deployer class instances."""
+
+    @staticmethod
+    def create_deployer(platform, region=None, bucket_name=None):
+        """Instantiate Kedro-viz deployer classes"""
+        if platform == "s3":
+            return S3Deployer(region, bucket_name)
+        if platform == "local":
+            return LocalDeployer()
+        raise ValueError("Invalid platform specified")
```
merelcht, line 18: It could be useful to add what the value of "platform" given was, so it's easier to debug. E.g. "Invalid platform `bla` specified"

merelcht (approved):
I left one minor comment, but otherwise this looks great! 👍 ⭐

rashidakanchwala (approved):

package/kedro_viz/integrations/deployment/deployer_factory.py
jitu5: Updated, Thanks.

NeroOkwa (commented):
Amazing !

PR #1698, opened 2024-01-05
Refactor datasets

rashidakanchwala:
## Description
Related to: https://github.com/kedro-org/kedro-viz/issues/1700

`Kedro-Viz` and `Kedro-Datasets` are tightly coupled , leading to issues where modifications in `Kedro-Datasets` can disrupt the functionality of `Kedro-Viz`. This ticket aims to refactor the codebase to enhance modularity and stability. The goal is to decouple the two components, allowing for smoother integration in the future without excessive interdependencies.

## Development notes

We've moved the logic for generating previews for several datasets from `kedro-viz` to `kedro-datasets`, reducing the coupling between `kedro-viz` and `kedro-datasets`.

Previews are rendered differently in the front-end, so we've introduced aliasing using [NewType](https://docs.python.org/3/library/typing.html#newtype). Currently, we support four types of previews in the front-end: json, image [png], plotly, and dataframe [as tables]. Users can enable previews for custom datasets as long as they fall into one of these categories.

Previously, we utilized the `load` function and overrode it in kedro-viz to load dataset previews. However, this logic has been migrated to kedro-datasets within the `preview()` function. In the kedro-viz API, we no longer include `plot`, `image`, or `tracking_data` as fields of DataNodeMetadata. Instead, we now only send `preview` and `preview_type`. `Preview` can be a preview of any dataset that has the preview function, and `preview_type` informs the front-end on how to render the preview.

## Checklist


merelcht:
Can the PR description be updated to add some context? It will make it easier to review.

package/kedro_viz/models/flowchart.py
```diff
@@ -586,52 +586,17 @@ def set_viz_metadata(cls, _, values):
...
-        if not (self.viz_metadata and self.viz_metadata.get("preview_args", None)):
-            return False
-
-        return True
-
     def get_preview_args(self):
         """Gets the preview arguments for a dataset"""
         return self.viz_metadata.get("preview_args", None)
 
+    def disable_preview(self):
```
merelcht, line 593: I was a bit confused reading this method at first, because `disable_preview` sounds like this method will disable it, not just check. Since it returns a boolean I'd rename it to: `is_preview_disabled`.

package/kedro_viz/models/flowchart.py
```diff
@@ -769,31 +725,22 @@ def set_run_command(cls, _):
...
-        if cls.data_node.is_tracking_node():
-            return cls.data_node.kedro_obj.load()
-        return None
-
     @validator("preview", always=True)
     def set_preview(cls, _):
-        if not (cls.data_node.is_preview_node() and hasattr(cls.dataset, "_preview")):
+        if not (
+            not cls.data_node.disable_preview() and hasattr(cls.dataset, "preview")
+        ):
```
merelcht, line 732: The double `not` makes this a bit hard to read. Can we do:

```python
if cls.data_node.disable_preview() or not hasattr(cls.dataset, "preview"):
        return None
```

instead?

package/kedro_viz/models/flowchart.py
```diff
@@ -769,31 +725,22 @@ def set_run_command(cls, _):
...
         try:
-            return cls.dataset._preview(**cls.data_node.get_preview_args())
+            preview_args = None
+            if cls.data_node.viz_metadata:
+                preview_args = cls.data_node.get_preview_args()
+            return (
+                cls.dataset.preview(**preview_args)
+                if preview_args is not None
+                else cls.dataset.preview()
+            )
```
merelcht, line 743: This is personal preference, but I find this code quite hard to read. I would personally change it to:

```python
preview_args = cls.data_node.get_preview_args() if cls.data_node.viz_metadata else None
if preview_args is not None:
    return cls.dataset.preview(**preview_args)
else:
    return cls.dataset.preview()

```

package/kedro_viz/models/flowchart.py
```diff
@@ -804,6 +751,25 @@ def set_preview(cls, _):
             )
             return None
 
+    @validator("preview_type", always=True)
+    def set_preview_type(cls, _):
+        if not (
+            not cls.data_node.disable_preview() and hasattr(cls.dataset, "preview")
+        ):
+            return None
```
merelcht, line 759: Again the double `not` makes this hard to read, I'd suggest changing it as above.

package/kedro_viz/models/flowchart.py
```diff
@@ -804,6 +751,25 @@ def set_preview(cls, _):
             )
             return None
 
+    @validator("preview_type", always=True)
+    def set_preview_type(cls, _):
+        if not (
+            not cls.data_node.disable_preview() and hasattr(cls.dataset, "preview")
```
merelcht, line 757: shouldn't this check for `preview_type` instead of `preview`?
```suggestion
            not cls.data_node.disable_preview() and hasattr(cls.dataset, "preview_type")
```

package/kedro_viz/models/flowchart.py
```diff
@@ -804,6 +751,25 @@ def set_preview(cls, _):
...
+        ):
+            return None
+
+        try:
+            return inspect.signature(cls.dataset.preview).return_annotation.__name__
+
+        except Exception as exc:  # pylint: disable=broad-except # pragma: no cover
```
merelcht, line 764: Why can't this be covered by tests?

demo-project/conf/base/catalog_08_reporting.yml
```diff
@@ -37,4 +37,7 @@ reporting.cancellation_policy_grid:
 reporting.confusion_matrix:
   type: matplotlib.MatplotlibWriter
   filepath: ${_base_location}/08_reporting/confusion_matrix.png
+  metadata:
```
merelcht, line 40: Is this related to the preview changes?

merelcht (commented):
I left several suggestions and questions. I also wanted to ask if the docs will be updated before this is released?

demo-project/conf/base/catalog_08_reporting.yml
ravi-kumar-pilla: I am not sure of the exact reason why this was added in the PR. But, this seems not related to the preview changes.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: The dataset/datanode only has a preview function. `preview_type` is derived from the signature.

ravi-kumar-pilla:
Hi @SajidAlamQB, The reason tests are failing for python3.8 is `kedro-datasets >= 2` require `python >= 3.9` . I am not sure if we want to drop viz support for python3.8 or do a conditional check for the preview feature that is added to `kedro-datasets`.  If we support python3.8, we need to conditionally add the dataset type checks again.

@merelcht what do you suggest here ?

noklam:
@ravi-kumar-pilla I think it's fine to drop Python 3.8 support, with the caveat that is it easy for user find out which version to install?

If an user is using Python 3.8, which combination of kedro/kedro-datasets/kedro-viz would work for them?

It's a result of kedro-plugins (at least kedro-datasets) are moving faster and we are following https://endoflife.date/python.

ravi-kumar-pilla:
> @ravi-kumar-pilla I think it's fine to drop Python 3.8 support, with the caveat that is it easy for user find out which version to install?

> If an user is using Python 3.8, which combination of kedro/kedro-datasets/kedro-viz would work for them?

We need to add/update the compatibility matrix in the [ReadME](https://github.com/kedro-org/kedro-viz/blob/main/README.md#compatibility-with-kedro) which includes kedro-viz/kedro/kedro-datasets. 

> It's a result of kedro-plugins (at least kedro-datasets) are moving faster and we are following https://endoflife.date/python.

Got it !

package/kedro_viz/models/flowchart.py
```diff
@@ -586,52 +587,17 @@ def set_viz_metadata(cls, _, values):
...
     def get_preview_args(self):
         """Gets the preview arguments for a dataset"""
         return self.viz_metadata.get("preview_args", None)
 
+    def is_preview_disabled(self):
+        """Checks if the dataset has a preview disabled"""
+        if self.viz_metadata:
+            preview_value = self.viz_metadata.get("preview")
+            return preview_value is not None and not preview_value
+        return False
```
noklam, line 599:
```suggestion
    def is_preview_disabled(self):
        """Checks if the dataset has a preview disabled"""
        if self.viz_metadata:
            preview_value = self.viz_metadata.get("preview")
            return preview_value is not None and not preview_value
        return False
```
I found this complicated, can we assume it's True unless specified as False?

ravi-kumar-pilla: +1

SajidAlamQB: Maybe something like this:

```suggestion
    def is_preview_disabled(self):
        """Checks if the dataset has a preview disabled"""
        return self.viz_metadata and self.viz_metadata.get("preview") is False
```

package/kedro_viz/models/flowchart.py
rashidakanchwala: I've now  included a test for the preview validator. I've marked preview_type with #pragma: no cover since it's a more challenging issue to test. I'm uncertain if this exception might arise.

It would only occur if there's a preview function without a return type, which seems unlikely. However, implementing this error handling is crucial to prevent kedro-viz from breaking if an exception is thrown.

jitu5 (approved):
Front end part looks good me!

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: so, the preview feature is always enabled unless the user specifies `viz_metadata: preview: false` ? @rashidakanchwala

rashidakanchwala: yup, that's true. by default

rashidakanchwala:
> I left one more question and wanted to check again if the docs will also be updated before release?
> 
> Otherwise, this looks all good from a code point of view 👍

Yes, before the release the docs will be updated. I will create a PR tomm for the docs.

package/kedro_viz/models/flowchart.py
```diff
@@ -769,35 +723,39 @@ def set_run_command(cls, _):
...
-    def set_image(cls, _):
-        if cls.data_node.is_image_node():
-            return cls.data_node.kedro_obj.load()
-        return None
+        try:
+            preview_args = (
+                cls.data_node.get_preview_args() if cls.data_node.viz_metadata else None
+            )
+            if preview_args is None:
+                return cls.dataset.preview()
```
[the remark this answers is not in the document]
rashidakanchwala: ok so if the user does the below 

```
companies:
  type: pandas.CSVDataset
  filepath: ${_base_location}/01_raw/companies.csv
```

if self.viz_metadata doesn't exist, the preview is still enabled also because there are no preview_args which is required only for TablePreview, it defaults it to 5. 

since there is no `viz_metadata` and hence no `preview_args` ... get_preview_args() will throw an error because it tries to access self.viz_metadata which doesn't exist. hence only if the user has specified viz_metadata, we allow `preview_args` to load.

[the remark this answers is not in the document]
rashidakanchwala: ok so if the user does the below

companies:
  type: pandas.CSVDataset
  filepath: ${_base_location}/01_raw/companies.csv


Preview remains enabled by default, displaying only the first 5 rows. Preview is disabled only if the user explicitly sets preview: false

In the provided configuration, self.viz_metadata does not exist, the `is_preview_disabled()` returns false which means preview is enable. However, because there are no preview_args, it defaults to displaying 5 rows.


```
 preview_args = (
                cls.data_node.get_preview_args() if cls.data_node.viz_metadata else None
            )
```

In the above code, since there is no viz_metadata, and therefore no preview_args, calling get_preview_args() will throw an error because it attempts to access self.viz_metadata, which does not exist. Therefore, we only allow preview_args to load if the user has specified viz_metadata.

package/kedro_viz/models/flowchart.py
ravi-kumar-pilla: if self.viz_metadata is None, the function will return None. Can we have a return value of boolean in all cases ? Can we do something like - `return viz_metadata is not None and viz_metadata.get("preview") is False` . Thank you

rashidakanchwala: ok have more clarity on what you and @merelcht meant. It returns None 
so i have changed it to return bool(self.viz_metadata and self.viz_metadata.get("preview") is False) -- now it will always return False except when preview: False

ravi-kumar-pilla (approved):

merelcht (approved):
Thanks for clarifying my earlier question @rashidakanchwala. This looks good to be merged now ⭐ 👍

rashidakanchwala:
Thanks team. I am pushing this, I will revert the dependencies once `kedro-datasets` is released. I will start on the docs now.

Issue #1148, opened 2022-10-31
Allow customization of flowchart icons


tynandebold:
A more strict version based off of this idea here: https://github.com/kedro-org/kedro-viz/issues/480

tynandebold:
See @NeroOkwa's [comment](https://github.com/kedro-org/kedro-viz/issues/480#issuecomment-1327747457) in #480 for more context.

datajoely:
If we were to make the following open-source icon libraries available to users that would be incredible:

https://simpleicons.org/
https://pictogrammers.com/

- We could could do defaults for things like Pandas, CSV etc.
- But the exciting possibility would be if we could allow the user (via factories as well 🚀 ) to do this sort of thing

```yaml
my_dataset:
   type: my.custom.dataset
   path: ...
   kedro-viz:
      layer: primary
+     icon: mdi/Bank        #example 1
+     icon: simple/kedro    #example 2
```

References:
> [mdi/Bank](https://pictogrammers.com/library/mdi/icon/bank/)
> [simple/kedro](https://simpleicons.org/?q=kedro)

astrojuanlu:
or rather, something like

```
my_dataset:
  type: ...
  metadata:
    kedro-viz:
      style:
        icon: mdi/Bank
```

with `style` including other coloring properties (see #1706)

astrojuanlu:
Related: #1807

datajoely:
I also want to reiterate the immediate user value we get from changing the highlight colour of MemoryDatasets.

nicolasboisseau:
Choosing icons or changing colours would open a very creative way to pass visual messages in the flow. That would be great!

datajoely:
Another library that would make this easy 
https://icons.run/

rashidakanchwala:
Closing this issue as not much traction for this feature.

PR #1706, opened 2024-01-16
[Experiment] Distinctive MemoryDataset view on flowchart

rashidakanchwala:
## Description

Suggested for quick experimentation by @datajoely: MemoryDatasets are non-persistent datasets, meaning they are not saved after a run. Some flowcharts contain numerous MemoryDatasets, and Kedro-viz currently lacks a way to differentiate them from persistent datasets. This PR introduces opacity to MemoryDatasets, making them easily distinguishable on the flowchart. The transparency also signifies their non-persistent nature.

Currently this feature is under an experiment flag. 

## Development notes

MemoryDatasets

## Checklist

### DIFF
--- a/src/components/flowchart/draw.js
+++ b/src/components/flowchart/draw.js
@@ -135,6 +135,7 @@ export const drawNodes = function (changed) {
     nodes,
     focusMode,
     hoveredFocusMode,
+    flags,
   } = this.props;
 
   const isInputOutputNode = (nodeID) =>
@@ -171,6 +172,10 @@ export const drawNodes = function (changed) {
       )
       .classed('pipeline-node--data', (node) => node.type === 'data')
       .classed('pipeline-node--task', (node) => node.type === 'task')
+      .classed(
+        'pipeline-node--memory-data',
+        (node) => flags?.diffMemoryDatasets && node?.dataset?.includes('memory')
+      )
       .on('click', this.handleNodeClick)
       .on('mouseover', this.handleNodeMouseOver)
       .on('mouseout', this.handleNodeMouseOut)
--- a/src/components/flowchart/flowchart.js
+++ b/src/components/flowchart/flowchart.js
@@ -708,6 +708,7 @@ export const mapStateToProps = (state, ownProps) => ({
   visibleSidebar: state.visible.sidebar,
   visibleCode: state.visible.code,
   visibleMetaSidebar: getVisibleMetaSidebar(state),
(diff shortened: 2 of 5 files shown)


datajoely:
It's beautiful! I'd suggest this should be default behaviour already - but it's should be really quick to get some user validation of that

inigohidalgo:
I think this is very nice functionality, speaking as someone who has a loot of memorydatasets in my pipeline. In my case it's not relevant, but I know some users use "dummy" inputs/outputs to force node order execution, I'm not sure if they tend to use memorydatasets or another type.

Maybe a user-configurable toggle per dataset type to fade/not fade?

src/components/flowchart/draw.js
```diff
@@ -171,6 +172,10 @@ export const drawNodes = function (changed) {
       )
       .classed('pipeline-node--data', (node) => node.type === 'data')
       .classed('pipeline-node--task', (node) => node.type === 'task')
+      .classed(
+        'pipeline-node--memory-data',
+        (node) => flags?.diffMemoryDatasets && node?.dataset?.includes('memory')
```
ravi-kumar-pilla, line 177: This looks good. Can we have a DatasetType check instead of string matching ?

rashidakanchwala: yes, i think we will eventually refactor and do it properly. for now this is just a quick dirty experiment

astrojuanlu:
I really like this idea (more than #1707).

As I said there, I'm wondering if eventually we could provide a more generic customisation feature that would allow users to have more control over this, instead of adding a toggle.

rashidakanchwala:
@astrojuanlu -- I agree. After speaking to a few people, I do realise that while some people might like this, others might not find it so usable. They might want to differentiate other datasets instead. Perhaps we need to build some sort of customisability, as you mentioned in the other PR, to allow users to distinguish datasets rather than us doing it for them.

Ravi also mentioned color-coding datasets, and as we have too many datasets, giving each a color-code might end up being more confusing and less visually appealing?  well, this is subjective, isnt it?  

hence, providing them some sort of customisability where they can decide how they want to highlight or distinguish certain datasets over others would be better. 

I will try and get more feedback on this short experiment. But keep the comments coming :)

rashidakanchwala:
Closing this experiment PR for now -- allowing users to differentiating datasets on kedro-viz seems like a good idea. We are going to try and do #1148 first and see how that picks up with users.

Issue #480, opened 2021-06-14
Provide simple mechanism for adding icons to datasets


datajoely:
## Description
> Is your feature request related to a problem? A clear and concise description of what the problem is: "I'm always frustrated when ..."

Users can label their dataset in the catalog and provide layers - but there is very little they can do to differentiate datasets beyond this from a visual perspective. Adding the facility to apply an icon from an existing library of icons would be an effective mechanism for making the pipeline visualisation a clearer and more efficient story-telling tool.

## Context
> Why is this change important to you? How would you use it? How can it benefit other users?

A simple example for where this would be useful would be to allow users to mark Excel datasources vs SQL datasources at a glance, even more so in the collapsed label-less view.

## Possible Implementation
> (Optional) Suggest an idea for implementing the addition or change.

On the YAML catalog side there could be an extra key for `icon` like so:

```yaml

flight_times:
    type: pandas.CSVDataSet
    layer: raw
    load_args:
          sep: '|'
    icon: carbon-csv
```

This could pull in the following icon from the Carbon design system ([by IBM](https://www.carbondesignsystem.com/guidelines/icons/library/)): provided by the `iconfiy` framework which collects several open source icon libraries. 
https://iconify.design/icon-sets/carbon/csv.html

By using the `[iconfiy-react](https://github.com/iconify/iconify-react)` library this would hopefully be a low effort addition

## Checklist

- [x] Include labels so that we can categorise your feature request

datajoely:
I'm going to reopen this - since users on Dicsord have been asking for the same thing

nicolasboisseau:
Such a feature would be nice, not only at the dataset level, but also at the node level. 

Personally when I make pipelines with kedro i try to make nodes responsible for only one main activity (that can be clearly explained to others). It would be nice to represent each node in kedro viz with an icon summarizing the main task the node is doing.  e.g. clean, stack, join, filter,  train, predict..

And naturally as you said in the issue title, it can be also interesting to identify the dataset types based on their icon, giving a hint to a data source origin : csv, sql, excel file, as a pipeline is often a mix of various dataset types..

Possibilities are quite numerous if a good icon collection can be provided!

beatslikeahelix:
I would second this. Could help keep the pipeline maintainer right as well as improve comprehension among a non-technical audience. It should go without saying that it should not be compulsory, however. Perhaps could be toggled in Viz.

To play devil's advocate, I would also say that the wrong implementation could add unnecessary complexity.
[...]

PR #1707, opened 2024-01-16
[Experiment] Show/Hide Memory Datasets on the flowchart

rashidakanchwala:
## Description

Related to #1706. should we provide users with the option to display or hide Memory Datasets in the view? This feature can be particularly helpful for larger and more complex pipelines. This PR offers an experimental toggle to show/hide Memory Datasets, functioning similarly to the show/hide dataset.

This is currently not under experiment flag. 

## Development notes

MemoryDatasetsHidden

## Checklist


inigohidalgo:
What happens to nodes which only have memory datasets as inputs and outputs? just a line connecting them to the last dataset which isn't memory?

ravi-kumar-pilla:
This is a great feature to show/hide dataset types reducing the size of the flowchart. I have concerns over the placement of Memory Datasets under Element Types. It would be nice to have a hierarchy where these fall under Datasets.

merelcht:
Would this just be for `MemoryDataset` or any non persistent datasets?

datajoely:
For the purposes of this experiment I think `MemoryDataSet` gets us 80% of the way. There has already been questions in this PR whether we can visually differentiate datasets of different types.

DebanjanBanerjeeQB:
Unpopular opinion and this is already fantastic work but 2 thoughts.

I dont see why this feature will be super impactful. Would love to hear your thoughts. 

- i would treat `memorydatasets` as any other dataset in the logical flow , it should have any different treatment / handling
- i wouldn't hide it ever , i wouldn`t hide any dataset ever because that risks an incorrect data lineage or risk of discounting a dataset while demoing or explaining the pipelines to anyone. 

If the goal is to signal users that these datasets use cache or inmemory storage and not persisted storage , i would create a tooltip or a separate action to explain datasets (all datasets) . Something that on hovering gives basic information about the datasets like 

- Dataset Type
- Some stats (Optional)
- Path 

WDYT ?

rashidakanchwala:
> What happens to nodes which only have memory datasets as inputs and outputs? just a line connecting them to the last dataset which isn't memory?

Hey, there would be line connecting the two task/function nodes. Below is an example of Memory Dataset shown and then hidden. 


Screenshot 2024-01-17 at 21 56 03
Screenshot 2024-01-17 at 21 57 11

datajoely:
I think it might be nice to have more powerful, persistent exclusion functionality - but I agree that #1706 is better. 

Again pointing to dbt, they have an ability for the user to provide [custom colors](https://docs.getdbt.com/reference/resource-configs/docs#custom-node-colors) - a long time ago I pitched https://github.com/kedro-org/kedro-viz/issues/480 which was turned into https://github.com/kedro-org/kedro-viz/issues/1148 but hasn't been prioritised.

DebanjanBanerjeeQB:
One more solution could be 

- Identify all the different types of datasets in the pipeline
- put them as a list and checkboxes
- whatever user needs can select / deselect from the view to highlight. 

That covers the memory dataset but also makes the approach more open to other dataset types.

DebanjanBanerjeeQB:
Screenshot 2024-01-18 at 10 30 28

Something like this ?

rashidakanchwala:
Closing this experiment PR for now -- allowing users to differentiating datasets on kedro-viz seems like a good idea. We are going to try and do #1148 first and see how that picks up with users.

Issue #1699, opened 2024-01-05
 Enable %run_viz magic line to use the arguments that Kedro-Viz supports on the command line


anapaularg15:
Currently, Kedro Viz supports additional arguments on the command line (for example, --pipeline).

However, the %run_viz magic line doesn't allow to pass any argument, which limits the use of kedro viz in notebooks.

astrojuanlu:
Hi @anapaularg15, thanks for filing this feature request! The team will look at it and prioritise it soon.

rashidakanchwala:
Closing this issue as it's completed and released with Kedro-viz 8.0.0

PR #1733, opened 2024-02-01
Enable %run_viz magic line to use the arguments that Kedro-Viz supports on the command line

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/1699

make `%run_viz` magic line support to passing any arguments of `kedro viz run` in notebooks.

Follow up PRs should update the documentation on this: https://docs.kedro.org/en/stable/notebooks_and_ipython/kedro_and_notebooks.html#run-viz-line-magic

## Development notes
A lot of the logic is borrowed from `cli.py` I had initially tried to use the click context for line magic but it was becoming quite complicated. I think the easiest approach is to just manually do the logic for these arguments. 

Example flow:
image

image

## QA notes

I have't added a comprehensive test suite for most of this new logic as it already tested in the `cli.py` and didn't want to repeat them again. Let me know if you all think otherwise.

## Checklist


package/kedro_viz/launchers/jupyter.py
```diff
@@ -78,26 +80,48 @@ def _display_databricks_html(port: int):  # pragma: no cover
...
-
+    # Parse arguments
+    arg_dict = parse_args(args)
+
+    host = arg_dict.get("host", _DATABRICKS_HOST if _is_databricks() else DEFAULT_HOST)
+    port = int(arg_dict.get("port", DEFAULT_PORT))
+    load_file = arg_dict.get("load-file", None)
+    save_file = arg_dict.get("save-file", None)
+    pipeline = arg_dict.get("pipeline", None)
+    env = arg_dict.get("env", None)
```
ravi-kumar-pilla, line 116: looks good overall. Are we not allowing shorthand in jupyter like (-a for autoreload, -e for env) ?

ravi-kumar-pilla:
Hi @SajidAlamQB ,

Approach looks good to me. I tried testing on jupyter, the default pipeline works well i.e., when --pipeline arg is not passed. But if I pass the --pipeline arg, I see the below behavior. 
1. The link to open kedro viz appears instantly 
2. When clicked, loads a 'site can't be reached' page. 
3. If reloaded, works fine. 

It might not be an issue for this ticket rather a bug if this happens via cli as well (need to check)

line_magic_args_issue

package/kedro_viz/launchers/jupyter.py
SajidAlamQB: We could but would need to add more parsing logic to `parse_args` and this becomes quite complicated especially given the number of arguments we have. At that point I think we should just use the click context somehow with the line magic but this is also very convoluted approach.

Given the complexities, I think adding these might be outside the scope of this PR.

ravi-kumar-pilla:
> > Approach looks good to me. I tried testing on jupyter, the default pipeline works well i.e., when --pipeline arg is not passed. But if I pass the --pipeline arg, I see the below behavior.
> > 
> > 1. The link to open kedro viz appears instantly
> > 2. When clicked, loads a 'site can't be reached' page.
> 
> Thanks @ravi-kumar-pilla I tried the same thing and did not see the `site can't be reached` page it went directly to viz. Could it be a browser thing, I did mine on Safari?

Not sure. I will try on Safari, in the mean time can you try on Chrome as well ? Thank you

ankatiyar:
@SajidAlamQB Does `kedro viz run --pipeline="Reporting stage"` work? I believe we only allow valid python package names for pipelines? So the actual pipeline names wouldn't contain spaces ideally.

ravi-kumar-pilla:
> Does `kedro viz run --pipeline="Reporting stage"` work? I believe we only allow valid python package names for pipelines? So the actual pipeline names wouldn't contain spaces ideally.

Hi @SajidAlamQB , I feel we should stick to providing options with `=` like `kedro viz run --pipeline="Reporting stage"`. If @magic_arguments has an issue parsing this, we can have our own parsing as we have in this PR. If our users ask for shorthand in future, we will try to refactor this. 

@ankatiyar we do have spaces for pipeline names on viz. I am not aware that this is not allowed.

SajidAlamQB:
> Does `kedro viz run --pipeline="Reporting stage"` work?

I tired `kedro viz run --pipeline="Reporting stage"` directly in the terminal and it works there.

astrojuanlu:
I think `kedro viz run --pipeline` should be consistent with how `kedro run --pipeline` behaves. This behavior was surprising also for me.

Maybe this is a separate issue though.

SajidAlamQB:
Given what we have discussed I tend to agree with @ravi-kumar-pilla that due to the issues of `@magic_arguemnts` not parsing properly sticking with the manual parsing so only having longer form of arguments (`--pipeline="Reporting stage"`) is acceptable for now. We can see if there's strong demand for shorthand arguments and consider them for future enhancement.

SajidAlamQB:
> It might not be an issue for this ticket rather a bug if this happens via cli as well (need to check)

@ravi-kumar-pilla regarding the issue with the `--pipeline` argument leading to a "site can't be reached" page, it seems like a browser related issue that resolves with a refresh. Its not specifically caused by `--pipeline` I was able to recreate it with just `%run_viz --host=127.0.0.1` I think it might be linked to how quickly we're trying to access `Kedro-Viz` after it launches. It looks like a minor browser caching or connection reuse issue, not directly tied to the scope of this issue a simple refresh or stopping the cell before running solves the issue.

ravi-kumar-pilla:
Hi @SajidAlamQB , 

There are other errors I am facing when changing the pipeline name as below. Can you please try at your end. I will also try to use command line and see if this is a bug and not related to current PR. Thank you !

image

astrojuanlu:
Opened an issue about the pipeline naming. Other than that, it was not clear to me if this PR is still desired and waiting for review.

ravi-kumar-pilla:
> Opened an issue about the pipeline naming. Other than that, it was not clear to me if this PR is still desired and waiting for review.

Hi @astrojuanlu , Thank you for creating a new ticket. This PR is almost done but we still have the issue of `site cannot be reached` (https://github.com/kedro-org/kedro-viz/pull/1733#issuecomment-1930866694) sometimes. I will try to fix it today.

rashidakanchwala (approved):
happy to merge this for now and we can fix the issue with browser rendering in another PR

PR #1790, opened 2024-03-05
Make `kedro viz` also do what `kedro viz run` does

### LINKED ISSUE #1789: Make `kedro viz` also do what `kedro viz run` does
## Description
I find being forced to write `kedro viz run` slightly annoying, since I (personally) never use `kedro viz deploy`. Is there a reason (or strong objection) to also enabling `kedro viz` (e.g. if I were to try and raise a PR)?

## Possible Implementation
I think it's possible. I looked into this the other day, because I wanted to implement, and there is https://github.com/click-contrib/click-default-group that will enable it (only dependency is unbounded `click`). FYI I had found this through some investigation a while back, last thread was https://github.com/pallets/click/issues/430.

## Checklist

deepyaman:
## Description

Closes #1789 

## Checklist


RELEASE.md
```diff
@@ -11,6 +11,7 @@ Please follow the established format:
 
 - Fix a bug on metadata panel when preview data is unavailable. (#1794)
 - Update Compatibility/Support Matrix in README. (#1774)
+- Make `kedro viz` also do what `kedro viz run` does. (#1790)
```
rashidakanchwala, line 14:
```suggestion
- Update CLI command to run Kedro-viz to support both `kedro viz run` and `kedro viz`. (#1790)
```

package/requirements.txt
```diff
@@ -1,15 +1,16 @@
-packaging~=23.0
-kedro>=0.18.0
-ipython>=7.0.0, <9.0
+aiofiles>=22.1.0
+click-default-group
```
rashidakanchwala, line 2: should we add any version?

rashidakanchwala (approved):
LGTM! :) thanks @deepyaman

NeroOkwa (approved):

ravi-kumar-pilla:
Hi @deepyaman , Thanks for this. This looks awesome !

[Nit] I understand from the [docs](https://github.com/click-contrib/click-default-group) that --help will now show (*) infront of the default command. I am wondering if this confuses someone who does not have clear understanding of what this means. Should we implement a class which lists the commands ? 

image

deepyaman:
> Hi @deepyaman , Thanks for this. This looks awesome !
> 
> [Nit] I understand from the [docs](https://github.com/click-contrib/click-default-group) that --help will now show (*) infront of the default command. I am wondering if this confuses someone who does not have clear understanding of what this means. Should we implement a class which lists the commands ?
> 
> image

I noticed that. I think it's very likely it could be a non-issue, in which case I don't think it needs to be over engineered.

If we want to remove the mark, or use something more explicit like ` (default)` instead of `*`, we could extend `DefaultGroup` to handle a `default_cmd_mark` parameter, which would get passed to https://github.com/click-contrib/click-default-group/blob/main/click_default_group.py#L100.

deepyaman:
> If we want to remove the mark, or use something more explicit like ` (default)` instead of `*`, we could extend `DefaultGroup` to handle a `default_cmd_mark` parameter, which would get passed to https://github.com/click-contrib/click-default-group/blob/main/click_default_group.py#L100.

Merged, but just let me know if want to do this; happy to slot it in quickly!

ravi-kumar-pilla:
> > If we want to remove the mark, or use something more explicit like ` (default)` instead of `*`, we could extend `DefaultGroup` to handle a `default_cmd_mark` parameter, which would get passed to https://github.com/click-contrib/click-default-group/blob/main/click_default_group.py#L100.
> 
> Merged, but just let me know if want to do this; happy to slot it in quickly!

I don't see as a blocker for now. If it is a quick fix, we can say (default) which is much clearer. But I am happy to hear from @rashidakanchwala . Thank you

PR #1792, opened 2024-03-05
Implementation of Directory Selection for Server Startup [Santana's PR]

rashidakanchwala:
## Checklist


package/kedro_viz/server.py
```diff
@@ -130,7 +131,7 @@ def run_server(
     from kedro.framework.startup import bootstrap_project
 
     parser = argparse.ArgumentParser(description="Launch a development viz server")
-    parser.add_argument("project_path", help="Path to a Kedro project")
+    parser.add_argument("--project-path", help="Path to a Kedro project")
```
ravi-kumar-pilla, line 134: For dev mode, earlier we had project_path as a required argument. This is good to make it consistent with cli, but can we add a default similar to cli ?

package/kedro_viz/server.py
```diff
@@ -139,16 +140,18 @@ def run_server(
...
 
     run_process_kwargs = {
-        "path": project_path,
+        "path": args.project_path,
         "target": run_server,
         "kwargs": {
             "host": args.host,
             "port": args.port,
-            "project_path": str(project_path),
+            "project_path": str(args.project_path),
```
ravi-kumar-pilla, line 154: since we are removing `path = Path(project_path) if project_path else Path.cwd()` from run_server, we can just pass the resolved `project_path` from line 143 here

Issue #1736, opened 2024-02-05
Extending stateful URLS 


rashidakanchwala:
## Description

This issue is to solve for this problem #1673 

The aims to enhance the stateful URLs to include additional functionalities such as 
- node type filtering, 
- tag filtering, and 
- expanding/collapsing all modular pipelines. 

We understand that this may result in longer URLs. Shortening our existing URL states is a priority, even if the resulting URLs are not human-readable.

Below is the current code 
```javascript 
export const params = {
  focused: 'focused_id',  (instead we use **fid**) 
  selected: 'selected_id', (instead we use **sid**) 
  selectedName: 'selected_name', (instead we use **sn**) 
  pipeline: 'pipeline_id', (instead we use **pid**) 
  run: 'run_ids', (instead we use **rid**) 
  view: 'view', (instead we use **v**) 
  comparisonMode: 'comparison', (instead we use **cmp**) 
};

```

Here are some suggestions for the new parameters. For Node Filtering by node types, or tags -- maybe we only mention what we want to hide in the params otherwise params will get very long. 
- To filter by types: type=parameters,dataset (this means only task nodes will be visible) 
- To filter by tags: tag=tagid1,tagid2,tagid3 (this means all nodes with thse tagsIds will be hidden) 
- To expand all: expAll=true/false 

- [ ] Include labels so that we can categorise your feature request

rashidakanchwala:
@NeroOkwa @astrojuanlu

stephkaiser:
suggestion to also include ability to save pretty name on/off

Issue #1764, opened 2024-02-26
Development installation instructions don't work


astrojuanlu:
## Description
As per title

## Context
I was trying to send a PR

## Steps to Reproduce

1. `micromamba create -n kedro-viz39-dev python=3.9 nodejs=18 -c conda-forge && micromamba activate kedro-viz39-dev`
2. `npm install`
3. `npm start`

(Steps 2 and 3 as per https://github.com/kedro-org/kedro-viz/blob/main/CONTRIBUTING.md#development-guidelines)

## Expected Result
> This will serve the app at [localhost:4141](http://localhost:4141/),

## Actual Result

```

> @quantumblack/kedro-viz@7.1.0 start
> REACT_APP_DATA_SOURCE=$DATA npm-run-all -p start:app start:lib


> @quantumblack/kedro-viz@7.1.0 start:lib
> rm -rf lib && babel src --out-dir lib --copy-files --watch


> @quantumblack/kedro-viz@7.1.0 start:app
> PORT=4141 react-scripts start


There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "jest": "26.6.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of jest was detected higher up in the tree:

  /Users/juan_cano/Projects/QuantumBlackLabs/Kedro/kedro-viz/node_modules/jest (version: 29.7.0) 

Manually installing incompatible versions is known to cause hard-to-debug issues.

If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
That will permanently disable this message but you might encounter other issues.

To fix the dependency tree, try following the steps below in the exact order:

  1. Delete package-lock.json (not package.json!) and/or yarn.lock in your project folder.
  2. Delete node_modules in your project folder.
  3. Remove "jest" from dependencies and/or devDependencies in the package.json file in your project folder.
  4. Run npm install or yarn, depending on the package manager you use.

In most cases, this should be enough to fix the problem.
If this has not helped, there are a few other things you can try:

  5. If you used npm, install yarn (http://yarnpkg.com/) and repeat the above steps with it instead.
     This may help because npm has known issues with package hoisting which may get resolved in future versions.

  6. Check if /Users/juan_cano/Projects/QuantumBlackLabs/Kedro/kedro-viz/node_modules/jest is outside your project directory.
[...]

PR #1542, opened 2023-09-27
Inherit tags from nodes for modular pipelines

rashidakanchwala:
## Description

Resolves #795 

## Development notes

We tried solving this issue in the backend but the way modular pipelines is computed in the backend is very complex. I thought maybe we can try solving it in the FE using selectors. 

In this PR, we have a new selector called getTagsforNodesAndModularPipelines.This selector identifies nodes associated with specific tags and then propagates these tags to the corresponding modular pipeline(s) (nested as well) for each node.

## QA notes

I have also added some tags to the demo project so we can test this. 

## Checklist

- [ ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1542"></a>

PR #1799, opened 2024-03-11
Extending stateful URLS with node filters and Expanding/collapsing modular pipelines flag

jitu5:
## Description

Resolves [#1736](https://github.com/kedro-org/kedro-viz/issues/1736) 

In this PR the stateful URLs to include additional functionalities such as

- Node type filtering (`type=task,data`)
- Tag filtering (`tags=companies,evaluate`)
- Expanding/collapsing all modular pipelines (`expandAllPipelines=true`)
- Reset filter button

Below are four scenarios where the possible values for new query parameters could be:

|   | Empty url & New browser | Empty url & Old browser | Stateful url & New browser | Stateful url & Old browser |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| types  | task,data (default)  | From local storage  | From stateful url  | If available, then from stateful URL, else from local storage.  |
| tags  |   | From local storage  | From stateful url  | If available, then from stateful URL, else from local storage.  |
| expandAllPipelines  | false (default)  | From local storage  | From stateful url  | From stateful url  |
| pipeline_id  | \__default__ (default)  | From local storage  | From stateful url  | From stateful url  |

https://github.com/kedro-org/kedro-viz/assets/38945204/7884e73e-27b6-460d-a7f6-a5bbd59eb1e8

## Development notes

- On first load query params `pipeline_id`,  `expandAllPipelines` and `types` will always be part of url.
- For path matching `URLSearchParams` used instead of  `matchPath`.
- Shortening our existing URL query params: 
   'focused_id' to **fid**
   'selected_id' to **sid**
   'selected_name' to **sn**
   'pipeline_id' to **pid**

## QA notes

- All tests should pass

## Checklist


RELEASE.md
```diff
@@ -5,6 +5,12 @@ Please follow the established format:
 - Use present tense (e.g. 'Add new feature')
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
+# Release 8.1.0
+
+## Major features and improvements
+
+- Extending stateful URLs with node types and tags. (#1799)
```
ravi-kumar-pilla, line 12: [Nit] I think the PR also includes expand/collapse modular pipeline. So the text can be something like - `Extending stateful URLs with node filters and expand/collapse modular pipelines`

ravi-kumar-pilla:
A minor inconsistency during Reset View. As mentioned in the ticket that the default qParams are - pid, types and expandAllPipelines. But when I click on Reset View, the pid is not present in the URL as shown below, but it does appear back if I open a new window. 

Also, I am not sure of the feasibility part but can we have the query params follow some order while we append to the URL ?

resetView_inconsistency

Thank you

src/config.js
```diff
@@ -178,3 +181,15 @@ export const KEDRO_VIZ_PREVIEW_DATASETS_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}preview
 export const KEDRO_VIZ_PUBLISH_AWS_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}publish_and_share_kedro_viz_on_aws.html#set-up-endpoint`;
 export const KEDRO_VIZ_PUBLISH_AZURE_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}publish_and_share_kedro_viz_on_azure.html#set-up-endpoint`;
 export const KEDRO_VIZ_PUBLISH_GCP_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}publish_and_share_kedro_viz_on_gcp.html#set-up-endpoint`;
+
+export const queryParamsToRetain = [
+  params.types,
+  params.tags,
```
ravi-kumar-pilla, line 187: Is this array to retain the default query params ? If so we should have `params.pipeline` and should not have `params.tags` right ?

jitu5: I missed `params.pipeline` and added but `params.tags` will also be part of query params to retain when user clicks on reset view. Also in existing behaviour, on reset view click we do not clear any tag if already applied.

jitu5:
> A minor inconsistency during Reset View. As mentioned in the ticket that the default qParams are - pid, types and expandAllPipelines. But when I click on Reset View, the pid is not present in the URL as shown below, but it does appear back if I open a new window.
> 
> Also, I am not sure of the feasibility part but can we have the query params follow some order while we append to the URL ?
> 
> 
> Thank you

I changed to retain pid after click on reset view. for query params to follow some order, let me check the feasibility.

Huongg (commented):
great work @jitu5, after a look through my first notice is `tags=...` doesnt get update immediately unless i refresh the page. It might actually a different bug too as if you notice, once you filtered the tag and refresh the page, the previous tags that you filtered are reset

https://github.com/kedro-org/kedro-viz/assets/32060364/c91a4b1f-29ae-4572-9e2c-61bf2206c5a1

src/config.js
ravi-kumar-pilla: I guess that might be a design question. From what I understand **Reset view** should take me back to an initial load. Our initial load (which does not have prior session) should only have - `types=task,data&pid=__default__&expandAllPipelines=false`. But I would like to hear from @rashidakanchwala and @Huongg

ravi-kumar-pilla:
Hi @jitu5 ,

There seems to be an issue with tag filtering (this is present in demo site as well, so it might be a bug) but it would be nice to resolve it here. 

Steps: 
1. Start viz
2. Click on tag filter Companies. This will show both Companies and Ingestion nodes
3. Refresh the page and the flowchart appears to only have Companies node. 
4. I tried similar filtering with other tags and it behaves differently. 

tag_filtering

jitu5:
> Hi @jitu5 ,
> 
> There seems to be an issue with tag filtering (this is present in demo site as well, so it might be a bug) but it would be nice to resolve it here.
> 
> Steps:
> 
> 1. Start viz
> 2. Click on tag filter Companies. This will show both Companies and Ingestion nodes
> 3. Refresh the page and the flowchart appears to only have Companies node.
> 4. I tried similar filtering with other tags and it behaves differently.
> 


@ravi-kumar-pilla @rashidakanchwala 
Here I see two diffreant behaviour when `expandAllPipelines` is true and false.

**When `expandAllPipelines` is false:**
1. Start viz
2. Click on tag filter Companies. This will show both Companies and Ingestion nodes(not its child node which contains Companies tag because Ingestion nodes collapsed on left side)
3. Refresh the page and the flowchart appears to only have Companies node.

https://github.com/kedro-org/kedro-viz/assets/38945204/8895f5fb-3b5a-41fe-9868-d081903634ee




**When `expandAllPipelines` is true:**
1. Start viz
2. Click on tag filter Companies. This will show both Companies, Apply Types To Companies & Int Typed Companies which contains Companies tag.
3. Refresh the page and the flowchart appears same as before refresh.

https://github.com/kedro-org/kedro-viz/assets/38945204/2ee1e3dc-cb9e-4fa7-87c6-f3fa3f3b4729



This is my understanding not completely sure is this a right behaviour.

ravi-kumar-pilla:
> > Hi @jitu5 ,
> > There seems to be an issue with tag filtering (this is present in demo site as well, so it might be a bug) but it would be nice to resolve it here.
> > Steps:
> > 
> > 1. Start viz
> > 2. Click on tag filter Companies. This will show both Companies and Ingestion nodes
> > 3. Refresh the page and the flowchart appears to only have Companies node.
> > 4. I tried similar filtering with other tags and it behaves differently.
> 
> @ravi-kumar-pilla @rashidakanchwala Here I see two diffreant behaviour when `expandAllPipelines` is true and false.
> 
> **When `expandAllPipelines` is false:**
> 
> 1. Start viz
> 2. Click on tag filter Companies. This will show both Companies and Ingestion nodes(not its child node which contains Companies tag because Ingestion nodes collapsed on left side)
> 3. Refresh the page and the flowchart appears to only have Companies node.
> 
>  Screen.Recording.2024-03-22.at.10.21.26.a.m.mov 
> **When `expandAllPipelines` is true:**
> 
> 1. Start viz
> 2. Click on tag filter Companies. This will show both Companies, Apply Types To Companies & Int Typed Companies which contains Companies tag.
> 3. Refresh the page and the flowchart appears same as before refresh.
> 
>  Screen.Recording.2024-03-22.at.10.17.06.a.m.mov 
> This is my understanding not completely sure is this a right behaviour.

There seems to be some bug/inconsistency here. I am not sure if that is stemming from the backend response or filtering happening on the frontend. Let me take some time and debug the issue. I feel the output should honor all the filtering criteria i.e., 

1. If there is a tag filter, it should show all the nodes containing the tag. 
2. Incase of expandModularPipelines = false and the filtered nodes happened to be inside a modular pipeline, it should show the modular pipeline node (collapsed view) which seems to be missing now.
3. In case of  expandModularPipelines = true and the filtered nodes happened to be inside a modular pipeline, it should show the expanded view (as it is happening now)

Thank you

ravi-kumar-pilla:
> > > Hi @jitu5 ,
> > > There seems to be an issue with tag filtering (this is present in demo site as well, so it might be a bug) but it would be nice to resolve it here.
> > > Steps:
> > > 
> > > 1. Start viz
> > > 2. Click on tag filter Companies. This will show both Companies and Ingestion nodes
> > > 3. Refresh the page and the flowchart appears to only have Companies node.
> > > 4. I tried similar filtering with other tags and it behaves differently.
> > 
> > 
> > @ravi-kumar-pilla @rashidakanchwala Here I see two diffreant behaviour when `expandAllPipelines` is true and false.
> > **When `expandAllPipelines` is false:**
> > 
> > 1. Start viz
> > 2. Click on tag filter Companies. This will show both Companies and Ingestion nodes(not its child node which contains Companies tag because Ingestion nodes collapsed on left side)
> > 3. Refresh the page and the flowchart appears to only have Companies node.
> > 
> > Screen.Recording.2024-03-22.at.10.21.26.a.m.mov
> > **When `expandAllPipelines` is true:**
> > 
> > 1. Start viz
> > 2. Click on tag filter Companies. This will show both Companies, Apply Types To Companies & Int Typed Companies which contains Companies tag.
> > 3. Refresh the page and the flowchart appears same as before refresh.
> > 
> > Screen.Recording.2024-03-22.at.10.17.06.a.m.mov
> > This is my understanding not completely sure is this a right behaviour.
> 
> There seems to be some bug/inconsistency here. I am not sure if that is stemming from the backend response or filtering happening on the frontend. Let me take some time and debug the issue. I feel the output should honor all the filtering criteria (even for the refresh) i.e.,
> 
> 1. If there is a tag filter, it should show all the nodes containing the tag.
> 2. Incase of expandModularPipelines = false and the filtered nodes happened to be inside a modular pipeline, it should show the modular pipeline node (collapsed view) which seems to be missing now on refresh.
> 3. In case of  expandModularPipelines = true and the filtered nodes happened to be inside a modular pipeline, it should show the expanded view (as it is happening now)
> 
> Thank you

I tried debugging the issue but it seems to be missing some part of the logic in the frontend in handling the below case -

`Incase of expandModularPipelines = false and the filtered nodes happened to be inside a modular pipeline, it should show the modular pipeline node (collapsed view) which seems to be missing now on refresh.`

The bug is not completely related to the current feature but it would be good to resolve it here. If the team feels to tackle this in a separate ticket, I am happy to create one. Please let me know what you think of this @rashidakanchwala @jitu5

rashidakanchwala:
> > > > Hi @jitu5 ,
> > > > There seems to be an issue with tag filtering (this is present in demo site as well, so it might be a bug) but it would be nice to resolve it here.
> > > > Steps:
> > > > 
> > > > 1. Start viz
> > > > 2. Click on tag filter Companies. This will show both Companies and Ingestion nodes
> > > > 3. Refresh the page and the flowchart appears to only have Companies node.
> > > > 4. I tried similar filtering with other tags and it behaves differently.
> > > 
> > > 
> > > @ravi-kumar-pilla @rashidakanchwala Here I see two diffreant behaviour when `expandAllPipelines` is true and false.
> > > **When `expandAllPipelines` is false:**
> > > 
> > > 1. Start viz
> > > 2. Click on tag filter Companies. This will show both Companies and Ingestion nodes(not its child node which contains Companies tag because Ingestion nodes collapsed on left side)
> > > 3. Refresh the page and the flowchart appears to only have Companies node.
> > > 
> > > Screen.Recording.2024-03-22.at.10.21.26.a.m.mov
> > > **When `expandAllPipelines` is true:**
> > > 
> > > 1. Start viz
> > > 2. Click on tag filter Companies. This will show both Companies, Apply Types To Companies & Int Typed Companies which contains Companies tag.
> > > 3. Refresh the page and the flowchart appears same as before refresh.
> > > 
> > > Screen.Recording.2024-03-22.at.10.17.06.a.m.mov
> > > This is my understanding not completely sure is this a right behaviour.
> > 
> > 
> > There seems to be some bug/inconsistency here. I am not sure if that is stemming from the backend response or filtering happening on the frontend. Let me take some time and debug the issue. I feel the output should honor all the filtering criteria (even for the refresh) i.e.,
> > 
> > 1. If there is a tag filter, it should show all the nodes containing the tag.
> > 2. Incase of expandModularPipelines = false and the filtered nodes happened to be inside a modular pipeline, it should show the modular pipeline node (collapsed view) which seems to be missing now on refresh.
> > 3. In case of  expandModularPipelines = true and the filtered nodes happened to be inside a modular pipeline, it should show the expanded view (as it is happening now)
> > 
> > Thank you
> 
> I tried debugging the issue but it seems to be missing some part of the logic in the frontend in handling the below case -
> 
> `Incase of expandModularPipelines = false and the filtered nodes happened to be inside a modular pipeline, it should show the modular pipeline node (collapsed view) which seems to be missing now on refresh.`
> 
> The bug is not completely related to the current feature but it would be good to resolve it here. If the team feels to tackle this in a separate ticket, I am happy to create one. Please let me know what you think of this @rashidakanchwala @jitu5

Thanks for flagging this, I had attempted to address this issue in the frontend (#1542) -- but I reckon it's not the best way. After some investigation, I think we need to fix this issue in the backend. Let's create a new pull request.

src/config.js
ravi-kumar-pilla: any comment on this ? @rashidakanchwala

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -78,6 +83,51 @@ export const FlowChartWrapper = ({
...
+            enabledKeys && toSetQueryParam(params.tags, enabledKeys);
+          }
+        },
+        nodeType: (value) => {
+          if (!searchParams.has(params.types)) {
+            const disabledKeys = getKeysByValue(value.disabled, false);
+            disabledKeys && toSetQueryParam(params.types, disabledKeys);
+          }
+        },
+        flags: (value) => {
```
ravi-kumar-pilla, line 111: is there any particular reason this is called flags ? other keys like pipeline, tag, nodeType mimick what the query is. Can we rename this to expandAllPipeline ?

src/components/flowchart-wrapper/flowchart-wrapper.js
```diff
@@ -78,6 +83,51 @@ export const FlowChartWrapper = ({
     matchedFocusedNode,
   } = findMatchedPath(pathname, search);
 
+  /**
+   * Sets the query params from local storage based on NodeType, tag, expandAllPipelines and active pipeline.
+   * @param {string} activePipeline - The active pipeline.
+   */
+  const setParamsFromLocalStorage = (activePipeline) => {
```
ravi-kumar-pilla, line 90: does this function only used on initial load ? like when the user is not modifying filters ? Based on the not conditions like - `if (!searchParams.has(params.tags))`, I think that is the case. If so, can we add that in the comment on line 87, just to be clear ?

src/utils/match-path.js
```diff
@@ -1,31 +1,27 @@
 import { matchPath } from 'react-router-dom';
-import { routes } from '../config';
+import { params, routes } from '../config';
 
 export const findMatchedPath = (pathname, search) => {
   const matchedFlowchartMainPage = matchPath(pathname + search, {
-    exact: true,
+    exact: false,
```
ravi-kumar-pilla, line 6: is this changed because of the default query params introduced ?

src/utils/hooks/use-generate-pathname.js
```diff
@@ -15,59 +34,83 @@ const getCurrentActivePipeline = () => {
...
   const toSelectedNode = useCallback(
     (item) => {
-      const activePipeline = getCurrentActivePipeline();
-
-      const url = generatePath(routes.flowchart.selectedNode, {
-        pipelineId: activePipeline,
-        id: item.id,
+      updateURLWithSearchParams((searchParams) => {
+        const activePipeline = getCurrentActivePipeline();
+        searchParams.set(params.pipeline, activePipeline);
```
ravi-kumar-pilla, line 77: what is the need to set active pipeline (params.pipeline) every time when the hook is called to set either selectedNode or focusedModularPipeline ? can't we just append/update the selectednode or focused mod pipeline to the existing searchParams (which should already have activePipeline), like we do in toSetQueryParam ?

src/utils/hooks/use-generate-pathname.js
```diff
@@ -1,12 +1,31 @@
...
 const getCurrentActivePipeline = () => {
   const localStorage = window.localStorage.getItem(localStorageName);
   return JSON.parse(localStorage)?.pipeline?.active;
 };
 
+/**
+ * Retains only the specified query parameters and removes all others from the given searchParams object.
+ * @param {URLSearchParams} searchParams - The searchParams object to modify.
+ */
+const retainOtherQueryParams = (searchParams) => {
```
ravi-kumar-pilla, line 19: [Nit] if I understand correctly, this function is used to retain the default queryParams and remove others, can we mention that in the comments (on line 16) and rename this to retainDefaultQueryParams ?

src/config.js
```diff
@@ -178,3 +181,16 @@ export const KEDRO_VIZ_PREVIEW_DATASETS_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}preview
 export const KEDRO_VIZ_PUBLISH_AWS_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}publish_and_share_kedro_viz_on_aws.html#set-up-endpoint`;
 export const KEDRO_VIZ_PUBLISH_AZURE_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}publish_and_share_kedro_viz_on_azure.html#set-up-endpoint`;
 export const KEDRO_VIZ_PUBLISH_GCP_DOCS_URL = `${KEDRO_VIZ_DOCS_URL}publish_and_share_kedro_viz_on_gcp.html#set-up-endpoint`;
+
+export const queryParamsToRetain = [
```
ravi-kumar-pilla, line 185: can we rename this to defaultQueryParams ?

src/utils/get-keys-by-value.js
```diff
@@ -0,0 +1,5 @@
+export const getKeysByValue = (object, value) => {
```
ravi-kumar-pilla, line 1: I find there is also another util file which says getKeyByValue, since this is a list and the previous one returns a single key, can we -
1. Try to use getKeysByValue[0] for the earlier version
2. If we need to keep both functions (I do feel its fine to keep both), can we move both to object-utils.js file ?
3. Looking at both the files, they seem related but we have 2 files for 2 functions which seems to be not needed. Moving both the functions to object-utils.js file would help in easy management in future

src/store/normalize-data.js
```diff
@@ -256,7 +326,8 @@ const normalizeData = (data, expandAllPipelines) => {
     data.layers.forEach(addLayer(state));
   }
 
-  return state;
+  const updatedState = getNodeFiltersFromUrl(state, data.tags);
+  return { ...state, ...updatedState };
```
ravi-kumar-pilla, line 330: is there a way where we just return state as before and have modifications to the state done within, instead of patching like this ? Like making getNodeFiltersFromUrl function (does not actually getNodeFilters but updates state with filters) as updateStateWithFilters which internally modifies the state. Also data.tags is a bit confusing. Can we do -

updateStateWithFilters(state, data.filters) - which takes care of modifying the state and data.filters contains both tags and nodeType info ? After this function is called we just return state (in the current implementation we are anyway returning - `return { ...state, ...updatedStateWithTags, ...updatedStateWithTypes };`

src/components/node-list/index.js
```diff
@@ -207,6 +253,52 @@ const NodeListProvider = ({
...
+    onToggleTagFilter(
+      tags.map((item) => item.id),
+      false
+    );
+
+    updateUrlParamsOnResetFilter();
+  };
+
+  // Updates the reset filter button status based on the node types and tags.
+  useEffect(() => {
```
rashidakanchwala, line 282: I was thinking maybe we can simplify this 
setIsResetFilterActive(false) --> whenever the 'Reset' button is clicked 

setIsResetFilterActive(true) --> whenever any tag, or filter is clicked. 

So we avoid the complex logic below.

src/components/node-list/index.js
```diff
@@ -207,6 +253,52 @@ const NodeListProvider = ({
       onToggleNodeSelected(null);
     }
   };
+
+  // Updates the URL parameters when the filter is reset.
+  const updateUrlParamsOnResetFilter = () => {
```
rashidakanchwala, line 258: I think this logic can go in use-generate-pathname.js

src/components/node-list/index.js
```diff
@@ -105,9 +114,42 @@ const NodeListProvider = ({
...
+  };
+
+  // To get existing values from search parameters
+  const getExistingValues = (paramName, searchParams) => {
+    const paramValues = searchParams.get(paramName);
+    return new Set(paramValues ? paramValues.split(',') : []);
+  };
+
+  // To update the URL parameters
+  const updateUrlParams = (item, paramName, existingValues) => {
```
rashidakanchwala, line 129: this logic can go also go in use-generate-pathname.js

rashidakanchwala (changes requested):
I am reviewing part by part. Already reviewed and shared changes for initial-state.js 
For nodelist/index.js -- we can refactor it so the URL Params handling can go in user-generated-pathname.js -- it feels more natural to me there , let me know what you think.

src/components/node-list/index.js
ravi-kumar-pilla: I think doing that will work if someone uses the UI to do filtering, but if someone directly changes the url, the reset button status should also watch out for the url.

rashidakanchwala: another suggestion is to follow the KISS principal and keep the `Reset` button always on but since this is also design suggestion looping @stephkaiser

jitu5: @ravi-kumar-pilla @rashidakanchwala @stephkaiser  I understand that the logic to set correct reset button status is complex, but are we keeping it simple like always ON or keeping what I did ?

rashidakanchwala: My suggestion is keeping it permanently enabled. Even when both the 'Reset' and reset filter are active, it doesn't alter the outcome. Many UI designs follow this convention, where the clear button remains visible even when the form is already empty. That's just my perspective, but would like to hear what others think.

src/components/node-list/index.js
jitu5: done.

src/components/node-list/index.js
jitu5: done.

src/components/flowchart-wrapper/flowchart-wrapper.js
jitu5: @ravi-kumar-pilla The `flags` key in the state object(loadLocalStorage) is used to store flags(expandAllPipelines, sizewarning).
If I use expandAllPipeline instead of flags I need to make extra checks and mapping here

```
      for (const [key, value] of Object.entries(localStorageParams)) {
        if (paramActions[key]) {
          paramActions[key](value);
        }
      }
```

Screenshot 2024-04-02 at 11 20 56 a m

src/components/flowchart-wrapper/flowchart-wrapper.js
jitu5: @ravi-kumar-pilla Yes this function used on initial load and when there will be any changes in active pipeline. same updated in function comment.

src/utils/match-path.js
jitu5: Yes, to match flowchart main page with new query params introduced.

src/utils/hooks/use-generate-pathname.js
jitu5: @ravi-kumar-pilla Good catch!
If you see existing behaviour on demo.kedro.org, we do not have `pipeline_id` params in url on initial load, it only added to url when you either click on focus button or on any node. So this was needed before stateful url changes. 
now as we already have `pipeline_id` params in url on initial load, so we don't need to set it along with either selectedNode or focusedModularPipeline ?

src/utils/hooks/use-generate-pathname.js
jitu5: Updated.

src/config.js
jitu5: Updated.

src/utils/get-keys-by-value.js
jitu5: Moved both the utils to object-utils.js

src/store/normalize-data.js
jitu5: @ravi-kumar-pilla modified and renamed to updateStateWithFilters  and removed patching. here `data.tags` is pass just for tag where `data.tags` contains all the tags.

src/components/node-list/index.js
ravi-kumar-pilla: I think this is more of a design choice. It might confuse some users if nothing happens on reset, but considering the complexity of checks needed, it should be fine to have it active always.

jitu5: In case if we keep always ON,
If the user clicks the reset button while the filter is already in the default state, the state will still be set to the same values. In this scenario, if the user is zoomed in on the flowchart, it will also be reset.

stephkaiser: Hi team, thanks for flagging to me. I think its confusing to users if the Reset button is always there even after they press it, theres no way for uses to tell they are on the default/reset view or the customised filtered view unless they look at the url. I think the convention is that when the reset/clear button is visible that implies that changes have been made and it can be reset to default (like when you've applied a filter on a website, a clear button will appear). 

I wanted to ask what the default view is? because in my opinion it should be without any filters applied (nodes, datasets, parameters). Otherwise it's a bit confusing as it implies to users that a filtered view is the default, unless we want to make a filtered view the default then the Reset button will not be visible when those filters are applied. 

What do you all think?

rashidakanchwala: The default view is set to 'Nodes', with 'Datasets' enabled and 'Parameters' disabled. Hence, we use 'Reset' instead of 'Clear' filters. 

Since majority including design agree having 'Reset' enabled all the time might confuse users, I think maintaining the way as Jitendra had previously done seems ok. We can see how we can make it less complex.

ravi-kumar-pilla: sounds good. For me this is similar to a subscriber pattern. The Reset button status should subscribe to any changes in the query params. I am not sure of the technical feasibility in this ticket, but yes we can modify in future if need be.

stephkaiser: Another thought - we also shouldn't rely on a colour change in the Reset text to indicate that a change has been applied as this doesn't work with colourblind users. Design fixes for the Reset button below:

Design fixes for the Reset button below:
- font size 13 (same as the Types and Tags label under Filters)
- font colour default FFFFFF 55%, and on hover FFFFFF 85% (similar pattern for Search)
- no underline as it's not a link
- after user clicks on Reset, button should disappear and only reappear if user changes have been made

src/store/normalize-data.js
```diff
@@ -196,6 +197,75 @@ const addLayer = (state) => (layer) => {
...
+  const queryParamsTagsArray = splitQueryParams(tagsQueryParam);
+
+  if (queryParamsTagsArray.length !== 0) {
+    const queryParamsTagsSet = new Set(queryParamsTagsArray);
+    const enabledTags = allNodeTags.reduce((result, tag) => {
+      result[tag.id] = queryParamsTagsSet.has(tag.id);
+      return result;
+    }, {});
+
+    state.tag.enabled = enabledTags;
```
rashidakanchwala, line 224:
```suggestion
    return {
  ...state,
  tag: {
    ...state.tag,
    enabled: enabledTags,
  },
```

rashidakanchwala (approved):
Hi LGTM thank @jitu5 ..awesome work :D

ravi-kumar-pilla:
Hi @jitu5 , Nice work. 

I am trying out the scenario below -
Lets say someone would like to share a URL - `http://127.0.0.1:4141/?pid=__default__&expandAllPipelines=false`. This does not have nodeTypes (i.e., I want all types including parameters). But if I open this in incognito or new browser, the url gets appended with types=task,data. Is this desired, if so how can someone ever share a url which has all 3 nodeTypes ? 

Thank you

jitu5:
> Hi @jitu5 , Nice work.
> 
> I am trying out the scenario below - Lets say someone would like to share a URL - `http://127.0.0.1:4141/?pid=__default__&expandAllPipelines=false`. This does not have nodeTypes (i.e., I want all types including parameters). But if I open this in incognito or new browser, the url gets appended with types=task,data. Is this desired, if so how can someone ever share a url which has all 3 nodeTypes ?
> 
> Thank you

@ravi-kumar-pilla
if you see the table above for nodeType with new browser & empty url(no nodeType params), the default ( types=task,data ) will apply. this is desired. If user wanted to to share with all nodeTypes, all should be included in urls.

ravi-kumar-pilla:
> > Hi @jitu5 , Nice work.
> > I am trying out the scenario below - Lets say someone would like to share a URL - `http://127.0.0.1:4141/?pid=__default__&expandAllPipelines=false`. This does not have nodeTypes (i.e., I want all types including parameters). But if I open this in incognito or new browser, the url gets appended with types=task,data. Is this desired, if so how can someone ever share a url which has all 3 nodeTypes ?
> > Thank you
> 
> @ravi-kumar-pilla if you see the table above for nodeType with new browser & empty url(no nodeType params), the default ( types=task,data ) will apply. this is desired. If user wanted to to share with all nodeTypes, all should be included in urls.

okay got it. I think it is a bit of design question as well. Should we make allSelected nodeTypes == noneSelected nodeTypes ? That is how the UI behaves at the moment. What do you think ? @stephkaiser @jitu5 @rashidakanchwala

ravi-kumar-pilla:
Hi @jitu5 , Appreciate your patience 💯 

I am testing the below scenario -
1. I have a tag filter with companies selected
2. I go to incognito and paste the url - it works fine
3. I try to remove the tags in the url and refresh the page, the tag re-appears as shown in the gif
4. I tested the scenario on normal browser and it has the same issue

Please let me know if you need more information. Thank you

stateful_test1

jitu5:
> Hi @jitu5 , Appreciate your patience 💯
> 
> I am testing the below scenario -
> 
> 1. I have a tag filter with companies selected
> 2. I go to incognito and paste the url - it works fine
> 3. I try to remove the tags in the url and refresh the page, the tag re-appears as shown in the gif
> 4. I tested the scenario on normal browser and it has the same issue
> 
> Please let me know if you need more information. Thank you
> 

@rashidakanchwala
if see below table 
Screenshot 2024-04-09 at 5 05 23 p m

- If tags is present in url it will apply on UI
- If tags is not part of url and you have new browser(no local-storage) nothing will apply
- If tags is not part of url and and if your browser's local storage contains any tag (in your case 'companies') it will applies from your browser's local storage.

this is expected behaviour.

stephkaiser:
> > > Hi @jitu5 , Nice work.
> > > I am trying out the scenario below - Lets say someone would like to share a URL - `http://127.0.0.1:4141/?pid=__default__&expandAllPipelines=false`. This does not have nodeTypes (i.e., I want all types including parameters). But if I open this in incognito or new browser, the url gets appended with types=task,data. Is this desired, if so how can someone ever share a url which has all 3 nodeTypes ?
> > > Thank you
> > 
> > 
> > @ravi-kumar-pilla if you see the table above for nodeType with new browser & empty url(no nodeType params), the default ( types=task,data ) will apply. this is desired. If user wanted to to share with all nodeTypes, all should be included in urls.
> 
> okay got it. I think it is a bit of design question as well. Should we make allSelected nodeTypes == noneSelected nodeTypes ? That is how the UI behaves at the moment. What do you think ? @stephkaiser @jitu5 @rashidakanchwala

Thanks Ravi, do you mind rephrasing or explaining in a less-technical way? Im struggling to follow :) are we discussing if the way all/none node types are displayed in the URL is intuitive to users?

I also had a question regarding the types - what are the task and data types? From the UI i see we have Nodes, Datasets, or Parameters so i'm a bit confused. 

Another question - shouldn't all pipelines be expanded by default?

stephkaiser:
> Hi @jitu5 , Appreciate your patience 💯
> 
> I am testing the below scenario -
> 
> 1. I have a tag filter with companies selected
> 2. I go to incognito and paste the url - it works fine
> 3. I try to remove the tags in the url and refresh the page, the tag re-appears as shown in the gif
> 4. I tested the scenario on normal browser and it has the same issue
> 
> Please let me know if you need more information. Thank you
> 
> stateful_test1 [ stateful_test1 ](https://private-user-images.githubusercontent.com/87735534/320918972-3527cf61-8813-4208-af23-2b3b85543b87.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTI2Nzg4NjIsIm5iZiI6MTcxMjY3ODU2MiwicGF0aCI6Ii84NzczNTUzNC8zMjA5MTg5NzItMzUyN2NmNjEtODgxMy00MjA4LWFmMjMtMmIzYjg1NTQzYjg3LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MDklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDA5VDE2MDI0MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTU1ZmZlZGE5YzAxMmJhNzQxMTY5YjIzYjVlOGI0NWRkNDEyZTQxOGY0Y2Q2MWIzZmM0OTNhZWQ5OGMwZWI0Y2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.1EHhJ9fXU64gtnbEkcyVPvKMFtaNldbbPfpckv33jfY) [ ](https://private-user-images.githubusercontent.com/87735534/320918972-3527cf61-8813-4208-af23-2b3b85543b87.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTI2Nzg4NjIsIm5iZiI6MTcxMjY3ODU2MiwicGF0aCI6Ii84NzczNTUzNC8zMjA5MTg5NzItMzUyN2NmNjEtODgxMy00MjA4LWFmMjMtMmIzYjg1NTQzYjg3LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MDklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDA5VDE2MDI0MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTU1ZmZlZGE5YzAxMmJhNzQxMTY5YjIzYjVlOGI0NWRkNDEyZTQxOGY0Y2Q2MWIzZmM0OTNhZWQ5OGMwZWI0Y2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.1EHhJ9fXU64gtnbEkcyVPvKMFtaNldbbPfpckv33jfY)

it does seem a bit strange that users wont be able to tweak the URL themselves. What happens if the user clicks on the Companies tag to remove it, will the URL update?

ravi-kumar-pilla:
> > Hi @jitu5 , Appreciate your patience 💯
> > I am testing the below scenario -
> > 
> > 1. I have a tag filter with companies selected
> > 2. I go to incognito and paste the url - it works fine
> > 3. I try to remove the tags in the url and refresh the page, the tag re-appears as shown in the gif
> > 4. I tested the scenario on normal browser and it has the same issue
> > 
> > Please let me know if you need more information. Thank you
> 
> @rashidakanchwala if see below table Screenshot 2024-04-09 at 5 05 23 p m
> 
> * If tags is present in url it will apply on UI
> * If tags is not part of url and you have new browser(no local-storage) nothing will apply
> * If tags is not part of url and and if your browser's local storage contains any tag (in your case 'companies') it will applies from your browser's local storage.
> 
> this is expected behaviour.

okay I understand that if there is any data in local storage, we try to get that and there is no way to check if the user removed all tags. For nodeTypes we do have some default but for tags we do not have such thing and that makes this test case confusing. Basically if there is local storage, the user can only disable a tag via UI and not via URL. May be we need to think about this approach. It is a bit confusing if I want to change the URL and it keeps appending tags.

rashidakanchwala:
> http://127.0.0.1:4141/?pid=__default__&expandAllPipelines=false

Historically, parameters in the flowchart are hidden by default based on user research indicating that most users do not need to see them. This approach has not drawn complaints, suggesting it is generally acceptable.I think the way it currently behaves should be fine, however if we see a lot of users complaining, maybe we can consider revising our approach. 
I think, when node types are unspecified, the URL should automatically revert to the default view, which includes both task and data types.

rashidakanchwala:
@jitu5 - I faced an issue related to tags that I got in the beginning but couldn't do it again whilst testing that case Ravi mentioned above. 

- Open the browser 
- Select `companies` tag. Updates URL to `http://localhost:4141/?types=task,data&pid=__default__&expandAllPipelines=false&tags=companies`
- Select `evaluate` tag. Updates URL to `http://localhost:4141/?types=task,data&pid=__default__&expandAllPipelines=false&tags=companies,evaluate`
- Now click on the `tag group` deselect. Updates URL to `http://localhost:4141/?types=task,data&pid=__default__&expandAllPipelines=false&tags=shuttles,train` 


team, lets connect tomm and discuss the tags approach. I think with us sa

rashidakanchwala:
> Team, I propose we meet tomorrow after the standup to discuss our approach to tags. Given our current way of saving, reading to local storage and reading from the URL, some of design decisions have been made that were thought most optimal while Steph was away. Let's see what we can release soon and identify areas for future improvement.

task = node,
 but  from user perspective, I agree we should keep the language consistent. so we should say type = data,node,parameters.

rashidakanchwala:
> I also had a question regarding the types - what are the task and data types? From the UI i see we have Nodes, Datasets, or Parameters so i'm a bit confused.

task = node,
but from user perspective, I agree we should keep the language consistent. so we should say type = data,node,parameters.

stephkaiser:
I think a meeting to discuss this work would be great, i wasn't involved in this so would be great to get a walkthrough and discuss tags and the default view. Also agree we should always keep language consistent, I would even suggest type = datasets,nodes,parameters.

astrojuanlu:
Please sign me up to the meeting!

jitu5:
> @jitu5 - I faced an issue related to tags that I got in the beginning but couldn't do it again whilst testing that case Ravi mentioned above.
> 
> * Open the browser
> * Select `companies` tag. Updates URL to `http://localhost:4141/?types=task,data&pid=__default__&expandAllPipelines=false&tags=companies`
> * Select `evaluate` tag. Updates URL to `http://localhost:4141/?types=task,data&pid=__default__&expandAllPipelines=false&tags=companies,evaluate`
> * Now click on the `tag group` deselect. Updates URL to `http://localhost:4141/?types=task,data&pid=__default__&expandAllPipelines=false&tags=shuttles,train`
> 
> I think it should revert it back to no tags.
> 

@rashidakanchwala I was able to reproduce this issue with group tag filter, yesterday I fix for this and now you can test it again.

jitu5:
> I think a meeting to discuss this work would be great, i wasn't involved in this so would be great to get a walkthrough and discuss tags and the default view. Also agree we should always keep language consistent, I would even suggest type = datasets,nodes,parameters.

@stephkaiser, Now types in URL matches with UI labels `type=datasets,nodes,parameters` thanks.

ravi-kumar-pilla (approved):
Great work @jitu5 !! Looks good to me

astrojuanlu:
I'm not qualified to comment on the code and also unfortunately don't have a lot of time to do manual QA of this in the face of #1764. Please don't wait on my review 🙏🏼 Thanks @jitu5 !

Issue #1159, opened 2022-11-02
Kedro-viz needs too much to run properly


foxale:
## Description

So according to this https://github.com/kedro-org/kedro-viz/issues/1125 kedro-viz requires 
- all plugins to be working (incl. mlflow connection to remote tracking server if leveraged)
- all python packages to be installed

For me it just doesn't make sense. Why would I need to have the python env ready to visualize project flow? The original issue seems dead, but kedro-viz is still in some way a core kedro feature, so..

merelcht:
Transferred this issue to the Kedro-Viz repo, because it's not framework related.

noklam:
> all plugins to be working (incl. mlflow connection to remote tracking server if leveraged)
> all python packages to be installed



@foxale Thank you for your question. The answer to that is pretty much covered. tl;dr
* Viz needs to load up the pipeline/nodes object, so the files would be loaded and the dependencies need to be installed. I don't see any way to get around it since you need to execute the code to get the Python representation of the pipeline.
* In most case - plugins aren't relevant, but as @deepyaman mentioned, the hook can actually change the DAG, so viz need to execute that. Otherwise, we can skip all the hooks and it would avoid the mlflow connection problem.

In theory, you could hide the dependencies so it won't be loaded, what viz need is just the pipeline & nodes objects.

foxale:
Well, it still looks like a fixable problem on the `kedro-viz` side. One solution would be to add a parameter to kedro viz that would skip all import errors and `after_catalog_created` hooks and reder the "raw" DAG.

noklam:
This is certainly doable if the team thinks we should do this. It's not the most elegant solution you can find, but I think simply

do something like
```python
if SOME_FLAG:
    session._hook_manager = _NullPluginManager()
# Before this line
context = session.load_context() 
```

This will remove all the hooks and should avoid this problem.

https://github.com/kedro-org/kedro-viz/blob/cc11edbec0777329eb66c875529ed35cfc5d7256/package/kedro_viz/integrations/kedro/data_loader.py#L72-L91

astrojuanlu:
Potential solution for this: https://github.com/kedro-org/kedro-viz/issues/1459#issuecomment-1662165936

rashidakanchwala:
Reopening this issue, as there's some issues around working with Spark that `--ignore--plugins` does no resolve

ravi-kumar-pilla:
[...]

PR #1818, opened 2024-03-26
Introduce --include-hooks option and remove --ignore-plugins in Kedro-Viz

### LINKED ISSUE #1808: Introducing `--include-hooks` parameter and deprecating `--ignore-plugins` in Kedro-Viz
## Description

In Kedro-Viz, hooks automatically run, but you can choose to skip them by using `kedro viz run --ignore-plugins`. However, since most of Kedro-Viz's features don't need hook information to work, and many users don't know about the `--ignore-plugins` option, we should consider a new approach. This issue about making hook execution opt-in by default, which means hooks would be disabled unless you choose to enable them.

## Possible Implementation

This ticket propose the following changes:

- Disable Hooks by default: Hooks will no longer be executed automatically. This change aims to streamline the operation of Kedro-Viz, reducing unnecessary processing and speeding up the visualization rendering for users.

- Introduce `--include-hooks` parameter: For users who require hook execution to generate runtime data for dynamic pipelines or other specific use cases, we will introduce a new `--include-hooks` parameter. This parameter allows users to explicitly opt-in to execute their hooks as needed, offering flexibility without sacrificing the tool's efficiency for the majority.

- Deprecate `--ignore-plugins` parameter: In light of introducing the `--include-hooks` option, the existing `--ignore-plugin`s parameter will be phased out. 

We want to enable this for all cli commands of kedro-viz including `run` , `build` and `deploy` 

Before implementing these changes, it would be nice to understand the usage of `--ignore-plugins` parameter. @NeroOkwa 

## Checklist

ravi-kumar-pilla:
## Description

Resolves #1808 

## Development notes

* Added `--include-hooks` option to all the viz cli commands (run, build, deploy)
* Added `--include-hooks` option to `%run_viz` Jupyter line magic
* Removed `--ignore-plugins` option from viz cli commands
* Updated pytests

## QA notes

* All tests should pass
* ` --include-hooks` should appear as an option with -h (eg., `kedro viz run -h`)

## Checklist


ravi-kumar-pilla:
CircleCI Build fix - https://github.com/kedro-org/kedro-viz/pull/1819

README.md
```diff
@@ -126,36 +126,31 @@ Usage: kedro viz run [OPTIONS]
...
                             host is localhost. Defaults to True.
-
-  --load-file FILE          Path to load kedro-viz data from a directory
-  --save-file FILE          Path to save kedro-viz data to a directory 
-  --pipeline TEXT           Name of the registered pipeline to visualise. If not
-                            set, the default pipeline is visualised
-
+  --load-file TEXT          Load Kedro-Viz using JSON files from the specified
+                            directory.
+  --save-file FILE          Save all API responses from the backend as JSON
```
rashidakanchwala, line 136: since this is more user facing I was thinking maybe we stick to the old 
'Save Kedro-Viz data as JSON files in a specified directory'

rashidakanchwala (approved):
Looks great to me !! thanks @ravi-kumar-pilla

SajidAlamQB (approved):
LGTM, thank you @ravi-kumar-pilla!

RELEASE.md
```diff
@@ -9,6 +9,7 @@ Please follow the established format:
 
 ## Major features and improvements
 
+- Introduce `--include-hooks` option and deprecate `--ignore-plugins` from cli commands. (#1818)
```
merelcht, line 12: It says `--ignore-plugins` is deprecated, but from looking at the code it looks like it's removed?

merelcht: And is this going to be included in a breaking release?

merelcht (changes requested):
The code itself looks good, but I just wanted to clarify if `--ignore-plugins` is meant to be deprecated, like it says in the release notes, or removed. And if it's removed will it be included in a breaking release?

RELEASE.md
merelcht: Deprecating and then removing would be the "right" process, but if no-one is using this flag I'm fine with removing straight away in a non-breaking release. Just make sure it's clear in the release notes 🙂

merelcht (approved):

brau0300:
This (as expected it seems) was a breaking change for any project using hooks or plugins and was non-trivial to chase down.

It also makes it less convenient for especially new users on a project needing this flag to start kedro viz; would we be open to considering ways to pass this option (now required for certain projects) by default (e.g. via project configuration) and/or more succinctly?

ravi-kumar-pilla:
> This (as expected it seems) was a breaking change for any project using hooks or plugins and was non-trivial to chase down.
> 
> It also makes it less convenient for especially new users on a project needing this flag to start kedro viz; would we be open to considering ways to pass this option (now required for certain projects) by default (e.g. via project configuration) and/or more succinctly?

Hi @brau0300 , 

Thank you for the comment. Ideally Kedro-Viz does not need any hook to start. Users should be able to do `kedro viz run` without any issue. Having said that, could you please let us know the use case where you need a hook to be enabled for kedro viz to start ? 

At this moment, we do not have any discussions on making this available via configuration of some sort. But we would be happy to hear from you the pain points and the issue this would solve for Kedro-Viz. Thank you

ravi-kumar-pilla:
Hi @brau0300 , 

Thank you for the response. The idea behind ignoring hooks by default was to make Kedro-Viz faster and work towards https://github.com/kedro-org/kedro-viz/issues/1159 . I agree it is a breaking change, but the team decided to load Kedro-Viz with bare minimum of what it needs and users can opt in hooks if they want any additional functionality. 

For the issue of `dropping toposort`, this could have been avoided if Kedro-Viz refrained from using transitive dependencies. We have created a [ticket](https://github.com/kedro-org/kedro-viz/issues/1821) to make Kedro-Viz better and avoid any such instances in future. 

Thank you

Issue #1786, opened 2024-03-04
Release notification doesn't parse markdown  ` backticks


datajoely:
Backticks should be parsed and turned into `<code>` blocks

![image](https://github.com/kedro-org/kedro-viz/assets/35801847/548b72a2-c2eb-4904-b7ae-9e7df329b0e6)

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro-viz/pull/1826

PR #1826, opened 2024-03-27
Add markdown support for version reminders

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/1786

The reminder pop-up notification doesn't parse markdown ` backticks this PR will attempt to add support for it using [markdown-it](https://www.npmjs.com/package/markdown-it).

## Development notes
image

## Checklist


rashidakanchwala (commented):
I was considering that, given the simplicity of our markdown needs—primarily replacing backticks with code-blocks —we could directly use < code > tags for any backticks we encounter. This approach would eliminate the need for any additional dependencies. What do you think?

SajidAlamQB:
> I was considering that, given the simplicity of our markdown needs—primarily replacing backticks with code-blocks —we could directly use < code > tags for any backticks we encounter. This approach would eliminate the need for any additional dependencies. What do you think?

That makes sense, I thought we wanted to support more markdown overall, happy to go with that instead.

jitu5:
@SajidAlamQB If we are not planing to add more style to `<code>` block then everything else looks good to me.

jitu5 (approved):
LGTM!

Issue #1797, opened 2024-03-08
Warning when parsing filepath in pandas.SQLQueryDataset even though pipeline works fine.


galenseilis:
## Description

https://kedro-org.slack.com/archives/C03RKP2LW64/p1709912797979079

## Context
How has this bug affected you? What were you trying to accomplish?

The impact is relatively minor. I just see a warning.

## Steps to Reproduce

1. Put a SQL script into an accessible path.
2. Put a SQL query dataset in the catalog with a filepath assigned to the path of the SQL script.
3. Run kedro.

## Expected Result

I expect that a warning should not be raised if nothing is wrong or situationally risky.

## Actual Result

I get a warning even though the pipeline works as desired. Purportedly this might be a Kedro-viz issue.

See [here](https://kedro-org.slack.com/archives/C03RKP2LW64/p1709912797979079).

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version: N/A
* Operating system and version: Windows 10
* NodeJS version used (if relevant): N/A
* Kedro version used (if relevant): See [here](https://kedro-org.slack.com/archives/C03RKP2LW64/p1709915286914139?thread_ts=1709912797.979079&cid=C03RKP2LW64).
* Python version used (if relevant): 3.10

PR #1853, opened 2024-04-08
Testing Path Fix

ravi-kumar-pilla:
## Description

Testing windows build [failure](https://app.circleci.com/pipelines/github/kedro-org/kedro-viz/11881/workflows/004e2ecc-5df6-419a-8ef1-0f43c6a21f28/jobs/75923) due to StatsHook on main branch.

```bash
   WARNING  Unable to get file size for the       hooks.py:142
                             dataset                                           
                             CSVDataset(filepath=C:/Users/circleci             
                             /AppData/Local/Temp/pytest-of-circlec             
                             i/pytest-0/test_get_file_size_dataset             
                             0_0/model_inputs.csv, load_args={},               
                             protocol=file, save_args={'index':                
                             False}): [WinError 3] The system                  
                             cannot find the path specified:                   
                             'C:Users/circleci/AppData/Local/Temp/             
                             pytest-of-circleci/pytest-0/test_get_             
                             file_size_dataset0_0/model_inputs.csv             
                             '
```

**NOTE:** While I tested on windows machine locally, it was working fine. However on CircleCI, filepath has `C:/ `, but the system is trying to find `C: ` (**missing the forward slash**). After looking into plugins kedro-datasets further, some dataset._filepath returns **PurePosixPath** while some return string. The fix for https://github.com/kedro-org/kedro-viz/issues/1797 should be made on the datasets side to be consistent. 

## Development notes

The pandas CSV, Excel, Feather, Generic, HDF, json, Parquet, xml has a code block -

```python
 super().__init__(
            filepath=PurePosixPath(path),
            version=version,
            exists_function=self._fs.exists,
            glob_function=self._fs.glob,
        )
```
while others like deltatable, gbq, sql keep filepath as `str`. I am not completely aware of the reason for this difference. @SajidAlamQB please add if you have any information on this. 

If this is not intentional, we should be fixing the [issue](https://github.com/kedro-org/kedro-viz/issues/1797) on kedro-datasets.

* As a platform-independent fix we can do as below in DataStatsHook if not on kedro-datasets

```python
file_path = get_filepath_str(PurePosixPath(dataset._filepath), dataset._protocol)
```            

**NOTE:** `continue_config.yml` changes are for testing **windows** build and they will be discarded before merging to main

## QA notes

* All tests should pass

## Checklist


rashidakanchwala:
also looping in @merelcht , does it make sense to fix this issue on kedro-datasets side? are we strict on return type validation for file_path? pls let us know.

ravi-kumar-pilla:
> also looping in @merelcht , does it make sense to fix this issue on kedro-datasets side? are we strict on return type validation for file_path? pls let us know.

Had a discussion on [slack](https://kedro-org.slack.com/archives/C0694PW10NS/p1712674911897309) and Nok mentioned that filepath is not a well defined API, means there may be entries of datasets which do not have filepath. Though I still think if there is filepath, the type should be consistent. From Viz perspective, we can have the conversion to be foolproof.

Thank you

package/kedro_viz/integrations/kedro/hooks.py
```diff
@@ -135,7 +135,9 @@ def get_file_size(self, dataset: Any) -> Union[int, None]:
             return None
 
         try:
-            file_path = get_filepath_str(Path(dataset._filepath), dataset._protocol)
+            file_path = get_filepath_str(
+                PurePosixPath(dataset._filepath), dataset._protocol
```
rashidakanchwala, line 139: I am confused, `Path` usually is the platform-agnostic way and works on both Windows and Linux. But PurePosixPath wouldn't work on Windows, so why are we changing this. Happy to get on a call to understand better.

rashidakanchwala (approved):
LGTM :) thanks Ravi -- great debugging

PR #1651, opened 2023-11-20
Fix modular pipelines breaking when collapsed. 

rashidakanchwala:
## Description

Resolves #1105 

## Development notes

In modular pipelines, a bug was identified when the pipelines were in collapsed mode. This issue was related to the incorrect handling of datasets due to a flawed logic in defining inputs and outputs. Here's the explanation:

### Inputs of a Modular Pipeline:
- An input can either be an external source or an internal one, provided it's not also used internally as an output.
If an internal input is simultaneously an internal output, it implies it's part of a sub-pipeline within the modular pipeline. Therefore, it should remain hidden in the collapsed view of the modular pipeline and not be displayed as an input.

### Outputs of a Modular Pipeline:
- An output is typically any external dataset used by other pipelines.
- In the Kedro framework, an external output must be explicitly defined as an output of the modular pipeline. It doesn't have a namespace.
- Internal outputs that also function as internal inputs are treated as components of the modular pipeline. These are not visible as outputs in the collapsed view but become visible when the pipeline is expanded.

This improved handling ensures a more accurate representation and functionality of modular pipelines, particularly in their collapsed state.

## QA notes

Tested this solution on below 4 edge cases :- 

1.  ### When a modular pipeline (external) output is used as an input to another pipeline and as an (internal) input to another function of the same modular pipeline. 
```python 
def create_pipeline(**kwargs) -> Pipeline:
    new_pipeline = pipeline(
        [
            node(lambda x: x,
                 inputs="dataset_in",
                 outputs="dataset_1",
                 name="step1"),
            node(lambda x: x,
                 inputs="dataset_1",
                 outputs="dataset_2",
                 name="step2"),
            node(lambda x: x,
                 inputs="dataset_2",
                 outputs="dataset_3",
                 name="step3"),
            node(lambda x: x,
                 inputs="dataset_3",
                 outputs="dataset_out",
                 name="step4"
            )
        ],
            namespace="main_pipeline",
        inputs=None,
        outputs={"dataset_out", "dataset_3"}
    )
    return new_pipeline
```
### Before 
Screenshot 2023-11-21 at 11 38 09

### After 
Screenshot 2023-11-21 at 11 32 48

2. ### When a nested modular pipeline output is used as an input to the outer modular pipeline and also used as an input to another external modular pipeline 
```python 
def create_pipeline(**kwargs) -> Pipeline:

    sub_pipeline = pipeline(
        [
            node(lambda x: x,
                 inputs="dataset_1",
                 outputs="dataset_2",
[...]

PR #1863, opened 2024-04-16
Fix Flowchart bug with datasets within a nested modular pipeline. 

### LINKED ISSUE #1814: While nesting namespace pipelines, intermediary datasets get exposed to top level of the `viz`
## Description
I found out that there might be a potential bug in how `kedro viz` visualizes **nested namespace pipelines**. In short, if there is an outer namespace (I will use `processing` in my example going forward) that has a single input and a single free output, instead of visually **collapsing everything** in between this input and output in the namespace, `kedro viz` **also exposes datasets shared by the inner namespaced pipelines** to the top level of the `viz`.

## Context
I encountered this issue in my project and, as discussed with @rashidakanchwala , opening an issue so the team can have a more detailed look. Also while writing this issue, as you'll see below, I created a very compact example of how you can reproduce this situation.

## Steps to Reproduce

Create a new kedro project with viz installed, and make the following pipeline:

```python
from kedro.pipeline import Pipeline, node
from kedro.pipeline.modular_pipeline import pipeline


def _get_generic_pipe() -> Pipeline:
    return Pipeline([
        node(
            func=lambda x: x,
            inputs="input_df",
            outputs="output_df",
        ),
    ])


def create_pipeline(**kwargs) -> Pipeline:
    pipe = Pipeline([
        pipeline(
            pipe=_get_generic_pipe(),
            inputs={"input_df": "input_to_processing"},
            outputs={"output_df": "post_first_pipe"},
            namespace="first_processing_step",
        ),
        pipeline(
            pipe=_get_generic_pipe(),
            inputs={"input_df": "post_first_pipe"},
            outputs={"output_df": "output_from_processing"},
            namespace="second_processing_step",
        ),
    ])
    return pipeline(
        pipe=pipe,
        inputs="input_to_processing",
        outputs="output_from_processing",
        namespace="processing",
    )
```

Then `kedro viz run` and see that `post_first_pipe` dataset, which should be fully encapsulated within `processing` namespace, gets exposed to the top level of `viz`.

## Expected Result
Since `post_first_pipe` dataset is fully internal to `processing` namespace, it should be visually encapsulated there and not exposed to the top level of the viz.

## Actual Result
What I actually see in the `viz` is this:

Screenshot 2024-03-19 at 6 08 48 PM

Let me highlight a few things here:
* So the problem is that `Post First Pipe` is visualized as if it is a "free output" of the `processing` namespace, while it is not. It should be fully encapsulated.
* An argument in favor of that is actually on the left: it is not mentioned under the "Search" bar.
* The other argument is: if I hover over this `Post First Pipe`, the tooltip is `processing.post_first_pipe`, so a logic that generates this name is aware that it's internal to `processing`
* The other argument is: if I do `kedro run` and see the logs, it actually captures those nested namespaces accurately. Meaning that prefixes are added where needed and not added where not needed.
* Another argument why that shouldn't be the case: [this page](https://docs.kedro.org/en/stable/nodes_and_pipelines/modular_pipelines.html) of the docs mentions that namespaces can be nested an arbitrary number of times. Which is definetely true from the `kedro run` perspective, but then to support it in the `viz`, this exposure of internal datasets should probably not be happening.

## Your Environment

* Web browser system and version: Google Chrome
* Operating system and version: Tried on both Win and Mac
* NodeJS version used (if relevant): No idea
* Kedro version used (if relevant): ~=0.19.2
* Kedro viz version used (if relevant): Tried both 7 and 8
* Python version used (if relevant): ~=3.10.0

## Checklist

rashidakanchwala:
## Description

Resolves #1814 

## Development notes

There was a bug in the logic of modular pipelines. Essentially, datasets that served as inputs and outputs to nested modular pipelines (internal_inputs/internal_outputs to modular pipeline) were mistakenly treated as external_inputs/external_outputs to the modular pipeline. This occurred because we were only checking if datasets were internal by comparing them against the nested modular pipeline, neglecting to verify against the parent modular pipeline. Now, we check against both the nested modular pipeline and the parent modular pipeline to determine whether the dataset is either an internal input/output to the modular popular or external input/output. 

## QA notes

You can verify that the issue is resolved by comparing it to the pipeline example shared in issue #1814. 

Additionally, if you examine example number 2 in issue #1651, which includes a nested pipeline, you'll notice that the problem with 'main_pipeline.dataset_1' is fixed. Now, it's hidden inside the 'main_pipeline' node when viewed in collapsed mode.

## Checklist

### DIFF
--- a/package/kedro_viz/data_access/repositories/modular_pipelines.py
+++ b/package/kedro_viz/data_access/repositories/modular_pipelines.py
@@ -14,21 +14,23 @@ from kedro_viz.models.flowchart import (
 )
 
 
-def _check_is_internal_input_output(
-    modular_pipeline_id: str, input_node_modular_pipeline: List
+def _is_internal_to_modular_pipeline(
+    input_node_modular_pipeline: List,
+    modular_pipeline_id: str,
 ) -> Optional[str]:
-    """Extracts the ID of the outer modular pipeline containing a nested modular pipeline.
+    """Checks if an input node's modular pipeline is part of
+    the given modular pipelines or it's parents.
 
     Args:
-        modular_pipeline_id: The ID of the nested modular pipeline.
+        modular_pipeline_id: The ID of the modular pipeline.
+        The modular pipeline can be a nested modular pipeline.
 
     Returns:
-        The ID of the outer modular pipeline containing the nested modular pipeline, if available.
+        True if the input node's modular pipeline ID
+        matches the given modular pipeline or it's parents, otherwise False.
 
-    Example:
-        >>> _get_outer_modular_pipeline_id("namespace_modular_pipeline.modular_pipeline")
-        'namespace_modular_pipeline'
     """
+
     all_modular_pipeline_ids = modular_pipeline_id.split(".")
     for pipeline_id in all_modular_pipeline_ids:
         if pipeline_id is not None and pipeline_id in input_node_modular_pipeline:
@@ -144,8 +146,8 @@ class ModularPipelinesRepository:
                 f"Attempt to add a non-data node as input to modular pipeline {modular_pipeline_id}"
             )
 
-        is_internal_input = _check_is_internal_input_output(
-            modular_pipeline_id, input_node.modular_pipelines
+        is_internal_input = _is_internal_to_modular_pipeline(
+            input_node.modular_pipelines, modular_pipeline_id
         )
 
         if is_internal_input:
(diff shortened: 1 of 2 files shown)


package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -14,6 +14,28 @@
 )
 
 
+def _check_is_internal_input_output(
+    modular_pipeline_id: str, input_node_modular_pipeline: List
+) -> Optional[str]:
```
ravi-kumar-pilla, line 19:
```suggestion
def _is_internal_node(
    modular_pipeline_id: str, internal_modular_pipelines: List
) -> Optional[str]:

```

ravi-kumar-pilla, line 19: The definition is a bit confusing to me, can we have -

```python
def _is_internal_node(
    modular_pipeline_id: str, internal_modular_pipelines: List
) -> Optional[str]:
```

yuryfedotov-mck:
Thanks for addressing this issue! Hope the example I provided helped localize it. I think it's a great enabler for kedro viz adoption in large projects.

Issue #1822, opened 2024-03-26
Flowchart view inconsistency with tag filters


ravi-kumar-pilla:
## Description

Incase of expandModularPipelines = false and the filtered nodes happened to be inside a modular pipeline, it should show the modular pipeline node (collapsed view) which seems to be missing on refresh. 

https://github.com/kedro-org/kedro-viz/pull/1799#issuecomment-2013933157

## Context

When I use tag filtering and if I do not have modular pipelines expanded, on page refresh I am missing few nodes which do have the tags. This causes an incorrect DAG rendering. This may be due to the modular pipelines not inheriting the tags from the children nodes. The changes needs to be propagated from the backend.

## Steps to Reproduce

Explained [here](https://github.com/kedro-org/kedro-viz/pull/1799#issuecomment-2013933157)

## Expected Result

When I use tag filtering and if I do not have modular pipelines expanded, on page refresh I should see the collapsed view of the modular pipeline as I see when I do not hit refresh

## Actual Result

When I use tag filtering and if I do not have modular pipelines expanded, on page refresh I am missing the entire modular pipeline which has nodes that have the tags.

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version: Chrome
* Operating system and version: macOS
* NodeJS version used (if relevant):
* Kedro version used (if relevant):
* Python version used (if relevant):

## Checklist
- [x] Include labels so that we can categorise your issue

rashidakanchwala:
@ravi-kumar-pilla to update this ticket with more information.

PR #1878, opened 2024-04-25
Inherit tags from nodes for modular pipelines

jitu5:
## Description

Related to #1822  

Incase of expandModularPipelines = false and the filtered nodes happened to be inside a modular pipeline, it should show the modular pipeline node (collapsed view) which seems to be missing on refresh. This PR aims to resolve this issue.

## Development notes

https://github.com/kedro-org/kedro-viz/assets/38945204/a336b39c-1d04-471c-a921-89b8a48587ce

## QA notes

- Added tests to cover changes.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro-viz/pull/1525"></a>


package/kedro_viz/models/flowchart.py
```diff
@@ -424,6 +424,8 @@ class ModularPipelineNode(GraphNode):
     # The type for Modular Pipeline Node
     type: str = GraphNodeType.MODULAR_PIPELINE.value
 
+    tags: Set[str] = Field(default_factory=set)
```
ravi-kumar-pilla, line 427: tags is already part of the parent (i.e., GraphNode). I don't think we need this here.

jitu5: Done.

rashidakanchwala:
Tried slighly different way of doing this. Let me know what you think - https://github.com/kedro-org/kedro-viz/compare/main...inherit-tags

rashidakanchwala:
Did a slightly different approach here https://github.com/kedro-org/kedro-viz/compare/main...inherit-tags--  .let me know what you think

ravi-kumar-pilla:
> Did a slightly different approach here [main...inherit-tags](https://github.com/kedro-org/kedro-viz/compare/main...inherit-tags) .let me know what you think

I think this updates the tags at the same time we add inputs/outputs to the immediate parent. We are anyway updating the grandparents here - `expanded_tree[parent_id].tags.update(modular_pipeline_node.tags)`. Looks good to me. Thank you

package/kedro_viz/data_access/managers.py
```diff
@@ -199,6 +199,9 @@ def add_pipeline(self, registered_pipeline_id: str, pipeline: KedroPipeline):
                     modular_pipelines.add_output(
                         current_modular_pipeline_id, output_node
                     )
+            # add tags
```
rashidakanchwala, line 202: we can remove this comment. as the code is obvious.

rashidakanchwala (approved):
LGTM. thanks!!

package/kedro_viz/data_access/managers.py
jitu5: Done.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -161,6 +162,27 @@ def add_output(self, modular_pipeline_id: str, output_node: GraphNode):
         else:
             self.tree[modular_pipeline_id].external_outputs.add(output_node.id)
 
+    def add_tags(self, modular_pipeline_id: str, node_tags: set):
```
ravi-kumar-pilla, line 165: I was going through the modular pipelines code and found that we update the modular_pipeline `pipelines` set with `node.pipelines`. May be we can update the tags at that point too. 

Like at line 239 in this file inside `extract_from_node` function having something like - `modular_pipeline.tags.update(node.tags)`. In that case we do not need this function and also do not need a call at line 203 in `managers.py`. What do you think ? @rashidakanchwala

rashidakanchwala: That's how I believe @jitu5 approached it previously. 
He had some code in the `extract_from_node` function, which I found very confusing because the function was originally defined to extract namespaces, but now it's getting tags. 
I think it's important to have clean functions that handle one task each, rather than one function doing too many things.

I'm also personally confused by the `extract_from_node` function being called three times here - [link to GitHub code](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/data_access/managers.py#L156-L202).
This is the exactly code refactor/simplification that Ivan mentioned to us the other day.

ravi-kumar-pilla (approved):
LGTM 👍 !!

Issue #1870, opened 2024-04-22
Enable users to disable preview for all when publishing Kedro-Viz


ravi-kumar-pilla:
## Description

This would help users to enable/disable preview for all the datasets.

## Context

* Users would like to have preview enabled in the catalog or for `kedro viz run` locally, but would like to disable the preview for all the datasets when deploying to cloud providers

https://kedro-org.slack.com/archives/C03RKP2LW64/p1713469330441069

## Possible Implementation

* Introduce a `--include-preview` option for all kedro viz cli commands
* Default the preview feature to False (need to be discussed)

## Checklist

- [x] Include labels so that we can categorise your feature request

rashidakanchwala:
had a Design PS session with Ivan, Yetu, and Steph. The following came out of discussion 

instead of `include_preview` or `exclude_preview` we simply have preview=true/false 

- Instead of using --include-preview or --exclude-preview, we'll switch to --preview=false/true.
- By default, kedro viz deploy and kedro viz build will set --preview=false.
- On the other hand, kedro viz and kedro viz run will set --preview=true by default.
- In the UI, users will specify whether they want to Enable or Disable preview before publishing. The backend will then create a JSON API both with and without preview based on the user's selection.

stephkaiser:
Designs can be found here - https://www.figma.com/file/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?type=design&node-id=1411-87299&mode=design&t=HEcOS0koRJ2SVfbX-4

Issue #1705, opened 2024-01-16
Add CLI reference


astrojuanlu:
## Description
It would be nice to have a CLI reference for `kedro viz` subcommands.

## Context
I was wondering what the output of `kedro viz run --save-file` should be, so I went to the documentation. But the docs explain usage, there is no reference docs (related: https://github.com/kedro-org/kedro/issues/2604)

## Possible Implementation
https://sphinx-click.readthedocs.io/

<img width="713" alt="image" src="https://github.com/kedro-org/kedro/assets/316517/53382509-b960-4fc5-9de5-8dde30683341">

## Possible Alternatives
...

## Checklist

- [ ] Include labels so that we can categorise your feature request

astrojuanlu:
To note, the fact that this is manually generated means that now we have to remember to document new commands or flags manually, see https://github.com/kedro-org/kedro-viz/pull/2077 cc @rashidakanchwala

PR #1894, opened 2024-05-08
Enable/disable preview for all the datasets when publishing Kedro-Viz from CLI

jitu5:
## Description

Resolves [1870](https://github.com/kedro-org/kedro-viz/issues/1870) 

## Development notes

* `--include-preview` flag added for Viz CLI:
* By default, kedro viz deploy and kedro viz build will have preview disabled for all the datasets
* Introduced a ClassVar `is_all_previews_enabled` for `DataNodeMetadata` class which controls the enable/disable functionality for all the datasets

## QA notes

* Run `kedro viz build --include-preview` to enable preview for all datasets
* Run `kedro viz deploy --include-preview` to enable preview for all datasets in the deployed version

**NOTE:** As discussed with @rashidakanchwala and @stephkaiser , `kedro viz run` will not have a preview flag and preview will be set to True by default. If the users want to change this behavior, they need to do this from UI. Designs will be updated soon (cc: @stephkaiser )

## Checklist


stephkaiser (approved):

rashidakanchwala:
There's a ticket here #1705  , we should add this to next sprint

package/kedro_viz/api/rest/responses.py
```diff
@@ -452,7 +457,7 @@ def save_api_pipeline_response_to_fs(pipelines_path: str, remote_fs: Any):
             raise exc
 
 
-def save_api_responses_to_fs(path: str, remote_fs: Any):
+def save_api_responses_to_fs(path: str, remote_fs: Any, are_datasets_previewable: bool):
```
merelcht, line 460: This is personal preference, but I would change `are_datasets_previewable` to `preview_datasets`.

To me `are_datasets_previewable` sounds like it's checking whether it's possible to preview these datasets or not, where when reading the code it should mean whether preview has been enabled or not. 

In my proposed change `preview_datasets=True` means preview is enabled and `preview_datasets=False` means it's disabled.

package/kedro_viz/models/flowchart.py
```diff
@@ -806,7 +812,11 @@ def set_run_command(cls, _):
     @field_validator("preview")
     @classmethod
     def set_preview(cls, _):
-        if cls.data_node.is_preview_disabled() or not hasattr(cls.dataset, "preview"):
+        if (
+            cls.data_node.is_preview_disabled()
+            or not hasattr(cls.dataset, "preview")
+            or not cls.are_datasets_previewable
```
merelcht, line 818: What's the difference here between `cls.data_node.is_preview_disabled()` and `cls.are_datasets_previewable`? Semantically, it sounds very similar, so I think we should think carefully about how these flags are named so it's easy to understand what's happening for future team members. 

Also in terms of functionality, what does it mean for preview to be disabled on a node level vs class level?

merelcht (commented):
The code looks good, but I have some reservations about the naming of the new `are_datasets_previewable` boolean. It's getting confusing with the `is_preview_disabled` functionality we have.

merelcht:
One thought is perhaps to name the new boolean after the new `--include-preview` flag, so it's clear those two are connected?

package/kedro_viz/models/flowchart.py
rashidakanchwala: Agree with @merelcht. 

i am also thinking for clarity we could possibly have :- 
- is_preview_disabled on node level (and explain this in the docstring, that this disabled via node level happens when the user specifies 'preview=false' in the DataCatalog)
- is_all_previews_disabled on class level (and once again explain in docstring this is to disable previews for all nodes and can be done via CLI and UI)

package/kedro_viz/api/rest/responses.py
ravi-kumar-pilla: okay, I too had a hard time renaming this :) .... so I want to make it differentiable to the `is_preview_disabled` which corresponds to each instance of DataNodeMetadata class, where-as the plural form `are_datasets_previewable` is a class variable. I personally prefer having is/are before a boolean variable/method to be clear of the return/data type.

package/kedro_viz/models/flowchart.py
jitu5: @merelcht As discussed with @rashidakanchwala the existing `is_preview_disabled` will rename to `is_preview_enabled` and new boolean variable `are_datasets_previewable` to `is_all_previews_enabled` .

rashidakanchwala (approved):
LGTM. Can we please also create a a ticket/PR to change  `is_preview_disabled` to `is_preview_enabled` so it is consistent. 

thanks both @jitu5 and @ravi-kumar-pilla :D

jitu5:
@rashidakanchwala Sure I will create a separate PR for change of `is_preview_disabled` to `is_preview_enabled`

ravi-kumar-pilla (approved):
LGTM. Thank you

jitu5:
@stephkaiser Requesting you approval on disable preview from CLI, few days back, I demoed it you. please let me know if you have any questions.

stephkaiser:
> @stephkaiser Requesting you approval on disable preview from CLI, few days back, I demoed it you. please let me know if you have any questions.

no questions from me! looks good, thanks Jitendra!

stephkaiser (approved):

PR #1896, opened 2024-05-09
Include expandAllPipelines flag in initial state

### LINKED ISSUE #1883: Undefined Value for `expandAllPipelines` Flag after Removal from `config.js`
## Description

See slack thread: https://kedro-org.slack.com/archives/C04EHA3K7C2/p1714426656093459

Following, https://github.com/kedro-org/kedro-viz/pull/1858, the `expandAllPipelines` flag from `config.js` was removed which is causing an undefined value, as it is no longer has a default value. This flag was previously used to control the state of expanding all modular pipelines on the initial load of the application in the `settings-modal.js`.  

The code within [settings-modal.js](https://github.com/kedro-org/kedro-viz/blob/be7236e11022d0bb44a416ca7e08a139b96bd5d8/src/components/settings-modal/settings-modal.js#L133-L148) uses the `getFlagsState` function to pull flag data, which included `expandAllPipelines`. The assumption was that removing it from `config.js` would suffice for this UI adjustment. However, this led to the flag being undefined elsewhere in the application.

We need to think about how the `expandAllPipelines` flag should be removed from just from the settings modal.

Suggestions:

- We reintroduce the `expandAllPipelines` flag into `config.js` and change settings-modal to not include.
-  implement a new way to initialise this flag across viz, maybe outside of config.js?

## Steps to Reproduce

Load the application and see that the flag's value is undefined.

## Expected Result
`expandAllPipelines` should have a default value.

## Actual Result
`expandAllPipelines` is undefined.

ravi-kumar-pilla:
## Description

Resolves #1883 

## Development notes

* Add `expandAllPipelines` to Initial state object
* Modify applyingURLParams to pipeline and non-pipeline state
* Validate the url param `expandAllPipelines` without breaking the UI using `isValidBoolean` helper function
* Update tests

expandAll_initial_state_fix

## QA notes

* On initial load of Kedro-Viz, expandAllPipelines query param in the URL should not be undefined
* All tests should pass

## Checklist


src/store/initial-state.js
```diff
@@ -18,7 +18,7 @@ import {
  */
 export const createInitialState = () => ({
   chartSize: {},
-  flags: Flags.defaults(),
+  flags: { ...Flags.defaults(), expandAllPipelines: false },
```
rashidakanchwala, line 21: since it's no more a flag, do we need to add it in flag. Can it be it's own key like themes?

ravi-kumar-pilla: I do agree its not a flag anymore (atleast not in settings) but we are updating the flags object when we click the expandAllPipelines icon. The issue happened as this property was missing in the flags object in initial state. If we want to move this out, that will be a refactor which includes the changes of the icon click as well.

I would rather have this inside an object. I was about to propose making the state object more structured :D .... like having all the nonPipelineSettings (theme, featureHintStep etc) into an object ?

rashidakanchwala (approved):
LGTM. Thanks @ravi-kumar-pilla

SajidAlamQB (approved):
Thank you @ravi-kumar-pilla this looks great! 🔥

PR #1903, opened 2024-05-14
Fix for schema change in strawberry-graphql JSON scalar

ravi-kumar-pilla:
## Description

* Update GraphQL schema which is breaking tests due to [strawberry-graphql release](https://github.com/strawberry-graphql/strawberry/releases/tag/0.228.0)

## Development notes

* Updated GraphQL schema JSON scalar specifiedBy location to fix build

## QA notes

* All tests should pass

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #1904, opened 2024-05-15
Fix messaging level when package compatibility is not satisfied

### LINKED ISSUE #1877: Fix messaging level when package compatibility is not satisfied
## Description

We check for 2 package compatibilities when running `kedro viz run` and if the satisified versions are not available we display an error message on the console (based on users debug level). We should make the error message as a warning instead and display more useful message

## Context

Image

## Steps to Reproduce

* Try executing `kedro viz run` on a python environment which does not have `kedro-datasets`

## Expected Result

* As these packages do not stop kedro viz from loading but hinder only the shareableviz feature, we should let the users know of this

## Actual Result

* It shows an error log but does not affect `kedro viz run`

## Your Environment

* Web browser system and version: Chrome
* Operating system and version: mac

## Checklist

ravi-kumar-pilla:
## Description

Resolves #1877 

## Development notes

* Update log level to `warning`
* Modify `PACKAGE_REQUIREMENTS` to have `warning_message` and `min_compatible_version` dict
* Update tests

**NOTE:** Lint will be fixed with the PR - https://github.com/kedro-org/kedro-viz/pull/1903

## QA notes

* Start `kedro viz run` on a kedro project which does not use `kedro-datasets`
* Users should see a warning message instead of error
* All tests should pass

## Checklist


ravi-kumar-pilla:
Hi @ankatiyar , 

Not related to this ticket but I was curious about the workflow - `.github/workflows/docs-only-checks.yml`

1. I see the workflow runs the linter but the linter has commands which are related to python and some secret scan etc. How is this helpful for doc changes ?
2. The trigger condition has `.md` files, so this will be triggered for all the PRs to main which have a release note and we also run backend linters separately which do almost the same.
3. Can we run this for only 1 python version, in case this is needed for any reason ?

cc: @Huongg 

Thank you

SajidAlamQB (approved):
Looks good, thank you @ravi-kumar-pilla!

Issue #1899, opened 2024-05-13
Refactor Modular Pipelines in the Kedro-viz backend


rashidakanchwala:
## Description

The logic of how we calculate modular pipelines in Kedro-viz backend differs from how it is computed in Kedro. For instance, modular pipelines are not associated with namespaces, and neither are the datasets in a modular pipeline. Only nodes have namespaces, and based on the way the datasets are consumed by different nodes, these datasets can be regarded as input/output/children to each modular pipeline.

This PR is to revise the current logic and make it more like the way Kedro does it. Modular Pipelines have caused a lot of issues for the user based on the difference in approach. With this refactor, we hope to align the approaches.

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

## Checklist

- [ ] Include labels so that we can categorise your feature request

yury-fedotov:
Is this issue potentially resolving [that one](https://github.com/kedro-org/kedro-viz/issues/1814)? Just curious.

PR #1897, opened 2024-05-09
Refactor Namespace Pipelines 

rashidakanchwala:
## Description

Resolves #1899 , #1814 

## Development notes

To ease review process for - https://github.com/kedro-org/kedro-viz/pull/1897 , created the below PRs

- [ ] #1939  
- [ ] #1938 
- [ ] #1937  
- [ ] #1940 

## QA notes

**Example modular pipeline tree:**

```
"modular_pipelines": {
        "__root__": {
            "id": "__root__",
            "name": "__root__",
            "inputs": [],
            "outputs": [],
            "children": [
                {
                    "id": "feature_engineering",
                    "type": "modularPipeline"
                },
                {
                    "id": "b5609df0",
                    "type": "parameters"
                },
                {
                    "id": "f6d9538c",
                    "type": "data"
                },
            …
            ]
        },
        "feature_engineering": {
            "id": "feature_engineering",
            "name": "feature_engineering",
            "inputs": [
                "abed6a4d",
                "f063cc82",
            …
            ],
            "outputs": [
                "23c94afb",
                "1e3cc50a"
            …
            ],
            "children": [
                {
                    "id": "8e4f1015",
                    "type": "data"
                },
                {
                    "id": "04ba733a",
                    "type": "task"
                },
            …
            ]
        },
        …
    }
```

**Current issues in constructing the modular pipeline tree:**
1. How we determine internal_inputs/outputs, external_inputs/outputs based on namespace and not on what kedro returns. Since datasets do not have a namespace (i.e., only kedro node and pipeline have namespaces) this raised issues in determining the actual inputs/outputs of a nested modular pipeline.
2. Inheriting input/output datasets to parent modular pipeline when nested. This made few datasets to appear in the root modular pipeline even though they are not free output datasets.
3. Readability/Maintenance issues in case of nested modular pipelines, as we did not define rules in adding a modular pipeline child, inputs and outputs for a modular pipeline
4. On the UI, modular pipeline focus was missing associated inputs/outputs from getting highlighted in the node menu as dataset nodes do not have namespace, the associated modular_pipelines were always empty.

**Incorrect rendering of nodes :**

Issues raised by users -

- https://github.com/kedro-org/kedro-viz/pull/1651
- https://github.com/kedro-org/kedro-viz/issues/1814

**How does this PR resolve the issues:**

1. Determines inputs/outputs to a modular pipeline based on what kedro returns.
[...]

PR #1937, opened 2024-06-10
Refactor flowchart models for modular pipelines change

ravi-kumar-pilla:
## Description

Partially resolves - #1899 

## Development notes

* Removes `namespace` property from GraphNode and retain the property only for TaskNode following Kedro Framework
* The data type of `modular pipelines` is changed to `Optional[Set(str)]`
* Updated all the `create_node` class methods like (create_task_node, create_data_node, create_parameters_node etc) to accept modular_pipelines parameter as we are populating modular pipelines before creating the nodes and associating each node with the associated modular pipeline during node creation.
* Remove `internal_inputs/outputs` and `external_inputs/outputs` from ModularPipelineNode. We will only have inputs/outputs for a ModularPipelineNode (i.e., the datasets which are free inputs/outputs as per Kedro)
* Update tests

## QA notes

* This PR is part of a bigger refactor (https://github.com/kedro-org/kedro-viz/pull/1897) of modular pipelines and is created to ease review process.
* The CI build might fail as this PR is not self-sufficient

## Checklist

### DIFF
--- a/package/kedro_viz/models/flowchart.py
+++ b/package/kedro_viz/models/flowchart.py
@@ -459,8 +459,8 @@ class TaskNodeMetadata(GraphNodeMetadata):
         # if a node doesn't have a user-supplied `_name` attribute,
         # a human-readable run command `kedro run --to-nodes/nodes` is not available
         if cls.kedro_node._name is not None:
-            if cls.kedro_node.namespace is not None:
-                return f"kedro run --to-nodes={cls.kedro_node.namespace}.{cls.kedro_node._name}"
+            if cls.task_node.namespace is not None:
+                return f"kedro run --to-nodes={cls.task_node.namespace}.{cls.kedro_node._name}"
             return f"kedro run --to-nodes={cls.kedro_node._name}"
 
         return None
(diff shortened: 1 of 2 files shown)


merelcht (commented):
In the description you mention

> Removes namespace property from GraphNode and retain the property only for TaskNode following Kedro Framework

and I see a lot of methods related to `namespaces` have been removed. Can you explain  why that change was needed and how namespaces are now set and used in Viz?

ravi-kumar-pilla:
> In the description you mention
> 
> > Removes namespace property from GraphNode and retain the property only for TaskNode following Kedro Framework
> 
> and I see a lot of methods related to `namespaces` have been removed. Can you explain why that change was needed and how namespaces are now set and used in Viz?

So earlier we used to have `namespaces` at the GraphNode level (which includes TaskNode, DataNode and others). After having discussion with Ivan, we realized datasets were never a part of a namespace and to have the same schema as Kedro, we removed the namespace from GraphNode and retained it to TaskNode. 

For a TaskNode `namespace` is set as before via kedro object like -
```python
    @field_validator("namespace")
    @classmethod
    def set_namespace(cls, _, info: ValidationInfo):
        return info.data["kedro_obj"].namespace
```

For other nodes, we do not need namespace. Lot of methods related to namespace like `_expand_namespaces`, `_get_namespace` in GraphNode are not required as they were used to calculate the modular_pipelines each node belongs to . Since we shifted the logic ( as you mentioned [here](https://github.com/kedro-org/kedro-viz/pull/1939#issuecomment-2180248019) ) before creating the nodes, we do not need them now.

Thank you

idanov (approved):
Nice :)

merelcht (approved):

PR #1938, opened 2024-06-10
Refactor data access repositories for modular pipeline change

### LINKED ISSUE #1899: Refactor Modular Pipelines in the Kedro-viz backend

ravi-kumar-pilla:
## Description

Partially resolves #1899 

## Development notes

* Create a dictionary of node and modular pipeline mapping for faster search
* Add parameters set to keep track of parameters (As parameter nodes do not have modular pipelines)
* Add `populate_tree` method which takes care of inputs/outputs and children for all the modular pipelines within a registered pipeline
* Add `explode_namespace` method which takes care of expanding a nested namespace.
* Modify add_inputs and add_outputs methods to take a `Set[str]` of inputs and outputs that are calculated in `populate_tree` method
* Add `add_children` method which resolves the modular pipeline child nodes
* Add `_add_children_to_parent_pipeline` method which resolves adding children to parent modular pipeline in-case of nesting
* Add `_add_datasets_as_children` method which adds datasets as children to the modular pipeline
* Update tests

## QA notes

- This PR is part of a bigger refactor (https://github.com/kedro-org/kedro-viz/pull/1897) of modular pipelines and is created to ease review process.
- The CI build might fail as this PR is not self-sufficient

## Checklist

### DIFF
--- a/package/kedro_viz/data_access/repositories/modular_pipelines.py
+++ b/package/kedro_viz/data_access/repositories/modular_pipelines.py
@@ -1,11 +1,12 @@
 """`kedro_viz.data_access.repositories.modular_pipelines` defines
 repository to centralise access for modular pipelines data."""
 
-import hashlib
+
+from collections import defaultdict
 from typing import Dict, List, Set, Tuple, Union
 
 from kedro.pipeline import Pipeline as KedroPipeline
-from kedro.pipeline.node import Node
+from kedro.pipeline.node import Node as KedroNode
 
 from kedro_viz.constants import ROOT_MODULAR_PIPELINE_ID
 from kedro_viz.models.flowchart import (
@@ -14,20 +15,7 @@ from kedro_viz.models.flowchart import (
     ModularPipelineChild,
     ModularPipelineNode,
 )
-from kedro_viz.utils import TRANSCODING_SEPARATOR, _strip_transcoding, is_dataset_param
-
-
-def _hash(value: str):
-    return hashlib.sha1(value.encode("UTF-8")).hexdigest()[:8]
-
-
-def _hash_input_output(item: str) -> str:
-    """Hash the input/output dataset."""
-    return (
-        _hash(_strip_transcoding(item))
-        if TRANSCODING_SEPARATOR in item
-        else _hash(item)
-    )
+from kedro_viz.utils import _hash, _hash_input_output, is_dataset_param
 
 
 class ModularPipelinesRepository:
@@ -45,10 +33,7 @@ class ModularPipelinesRepository:
                 ROOT_MODULAR_PIPELINE_ID
             )
         }
-        self.node_mod_pipeline_map: Dict[
-            str, Set[str]
-        ] = {}  # Updated to map node_id to a set of modular_pipeline_ids
-
+        self.node_mod_pipeline_map = defaultdict(set)
         self.parameters = set()
 
     def populate_tree(self, pipeline: KedroPipeline) -> None:
@@ -60,29 +45,41 @@ class ModularPipelinesRepository:
         Args:
             pipeline (KedroPipeline): The Kedro pipeline to populate the tree from.
         """
+
+        # Extract namespaces from the nodes in the pipeline
         namespaces = sorted(
             {node.namespace for node in pipeline.nodes if node.namespace}
         )
 
+        # Get all modular pipeline IDs by exploding namespaces
         modular_pipeline_ids = {
             modular_pipeline_id
             for ns in namespaces
             for modular_pipeline_id in self._explode_namespace(ns)
         }
 
+        # Create and populate each modular pipeline
         for modular_pipeline_id in sorted(modular_pipeline_ids):
             self.get_or_create_modular_pipeline(modular_pipeline_id)
 
+            # Extract nodes within the current namespace to form a sub_pipeline
             sub_pipeline = pipeline.only_nodes_with_namespace(modular_pipeline_id)
             rest_of_the_pipeline = pipeline - sub_pipeline
 
+            # Identify free inputs of the sub_pipeline
             free_inputs = sub_pipeline.inputs()
+
+            # Identify free outputs of the sub_pipeline, including intermediary outputs
             free_outputs = sub_pipeline.outputs() | (
                 rest_of_the_pipeline.inputs() & sub_pipeline.all_outputs()
             )
 
+            # Add identified inputs and outputs to the modular pipeline
             self._add_inputs(modular_pipeline_id, free_inputs)
             self._add_outputs(modular_pipeline_id, free_outputs)
+
+            # Add nodes(and datasets of that node) of the sub_pipeline as
+            # children to the modular pipeline
             self._add_children(modular_pipeline_id, sub_pipeline.nodes)
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
+            List[str]: A list of expanded namespaces.
+            Example:
+                >>> nested_namespace = "uk.data_processing.internal"
+                >>> modular_pipeline_repo_obj = ModularPipelinesRepository()
+                >>> expanded_namespace =
+                        modular_pipeline_repo_obj._explode_namespace(nested_namespace)
+                >>> expanded_namespace
+                ["uk", "uk.data_processing", "uk.data_processing.internal"]
+        """
+        if not nested_namespace or "." not in nested_namespace:
```
merelcht, line 98: Why is it necessary to check for `if not nested_namespace`? And then again below: `[nested_namespace] if nested_namespace else []`?

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
+        ns_parts = nested_namespace.split(".")
+        return [".".join(ns_parts[: i + 1]) for i in range(len(ns_parts))]
 
     def get_or_create_modular_pipeline(
         self, modular_pipeline_id: str
     ) -> ModularPipelineNode:
-        """Get the modular pipeline node with the given ID from the repository.
-        If it doesn't exist, create the node, add to the repository and return the instance.
+        """
+        Retrieve or create a modular pipeline node.
```
merelcht, line 107: Why change the doc string? The behaviour is still the same right? The previous one also seems to provide more explanation.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
-        input_node: Union[DataNode, TranscodedDataNode, ParametersNode],
-    ) -> None:
-        """Add an input to a modular pipeline based on whether it's an internal or external input.
-        The input to a modular pipeline can only be a data node or parameter node.
-        The input also has knowledge of which modular pipelines it belongs to
-        based on its namespace. This information can be accessed with through
-        `input_node.modular_pipelines`.
+    def add_inputs(self, modular_pipeline_id: str, inputs: Set[str]) -> None:
+        """
+        Add input datasets to the modular pipeline.
```
merelcht, line 124: In the code there's also a special case for `parameters`. I think it makes sense to mention that in the docstring.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
+        """
+        hashed_outputs = {self._hash_input_output(output) for output in outputs}
+        self.tree[modular_pipeline_id].outputs = hashed_outputs
 
-        is_internal_output = modular_pipeline_id in output_node.modular_pipelines
-        if is_internal_output:
-            self.tree[modular_pipeline_id].internal_outputs.add(output_node.id)
-        else:
-            self.tree[modular_pipeline_id].external_outputs.add(output_node.id)
+    def _hash_input_output(self, item: str) -> str:
```
merelcht, line 151: Why does this need to take `self`? As far as I understand this method just hashes input/output so it doesn't need an instance of `ModularPipelinesRepository`

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
+            free_inputs = sub_pipeline.inputs()
+            free_outputs = sub_pipeline.outputs() | (
+                rest_of_the_pipeline.inputs() & sub_pipeline.all_outputs()
+            )
+
+            self.add_inputs(modular_pipeline_id, free_inputs)
+            self.add_outputs(modular_pipeline_id, free_outputs)
+            self.add_children(modular_pipeline_id, sub_pipeline.nodes)
+
+    def _explode_namespace(self, nested_namespace: str) -> List[str]:
```
merelcht, line 81: Why does this need to take `self`? This doesn't need an instance of `ModularPipelinesRepository` so it can just be a static method or it can be defined outside the `ModularPipelinesRepository` class altogether.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
-            self.tree[modular_pipeline_id].external_outputs.add(output_node.id)
+    def _hash_input_output(self, item: str) -> str:
+        """Hash the input/output dataset."""
+        return (
+            _hash(_strip_transcoding(item))
+            if TRANSCODING_SEPARATOR in item
+            else _hash(item)
+        )
+
+    def add_children(self, modular_pipeline_id: str, kedro_nodes: List[Node]):
```
merelcht, line 159: Why are this method and `add_outputs` and `add_inputs` public?

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
+                and io_id not in self.parameters
+            ):
+                modular_pipeline.children.add(
+                    ModularPipelineChild(id=io_id, type=GraphNodeType.DATA)
+                )
+                if io_id not in self.node_mod_pipeline_map:
+                    self.node_mod_pipeline_map[io_id] = set()
+                self.node_mod_pipeline_map[io_id].add(modular_pipeline.id)
+
+    def has_modular_pipeline(self, modular_pipeline_id: str) -> bool:
```
merelcht, line 309: I'm not sure this method is necessary, it's only called in one place as far as I can tell and IMO actually makes it more effort to figure out what's going on.

merelcht (commented):
I've done an initial review of this part, but want to have another look after reviewing the other parts. I've left comments mostly related to readability of the code.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,246 +1,324 @@
...
-    TranscodedDataNode,
 )
+from kedro_viz.utils import TRANSCODING_SEPARATOR, _strip_transcoding
+
+
+def _hash(value: str):
+    return hashlib.sha1(value.encode("UTF-8")).hexdigest()[:8]
+
+
+def is_dataset_param(dataset_name: str) -> bool:
```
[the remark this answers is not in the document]
ravi-kumar-pilla: Not really. I observed this method is used via Catalog Repository as well. I am shifting this to utils.py which will help in removing code duplication. Thank you

package/kedro_viz/data_access/repositories/modular_pipelines.py
ravi-kumar-pilla: Yes, converted to static method. Thank you

package/kedro_viz/data_access/repositories/modular_pipelines.py
ravi-kumar-pilla: I think we should only check for the below condition -

```python
if not nested_namespace:
    return []
```
Thanks

package/kedro_viz/data_access/repositories/modular_pipelines.py
ravi-kumar-pilla: Yes, it is the same. Reverting back this change.

package/kedro_viz/data_access/repositories/modular_pipelines.py
ravi-kumar-pilla: Yes there is a special case for parameters and that is while adding the datasets as children. That is in the docstring for `add_children`, `_add_children_to_parent_pipeline` and `_add_datasets_as_children`. The method here is to add_inputs to the modular pipeline (like the free_inputs hanging outside the modular pipeline)

package/kedro_viz/data_access/repositories/modular_pipelines.py
ravi-kumar-pilla: Yes, I would like to know where should the helper functions like this be placed in ? I used to place all the helper functions in utils as public. But there should be some rules which I am not aware of. Could you please let me know ? Thanks

Placed in -

1. A class as static method (public/private)
2. Outside of a class but in same file (public/private)
3. Inside utils (public/private)

package/kedro_viz/data_access/repositories/modular_pipelines.py
ravi-kumar-pilla: I do not see them being used outside this class. They were public before the code change as they were used in managers.py. But now it is not needed. Thank you for pointing it out

package/kedro_viz/data_access/repositories/modular_pipelines.py
merelcht: This stackoverflow thread, specifically the first two answers, describe quite well what the general rule of thumb is when it comes to static methods, class methods and module functions. https://stackoverflow.com/questions/11788195/module-function-vs-staticmethod-vs-classmethod-vs-no-decorators-which-idiom-is

In the Kedro codebase we generally apply the following:
1. Inside a class as static method: any method that is only used by that class, but the difference with a module function is that these types of methods are specific to this class and don't have any meaning outside of it. (e.g. https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L336-L338)
2. Outside of a class but in same file: any helper method that is only used in that specific file should be a module function at the top of the file. Note that these methods are only used in that file/class, but don't require any "knowledge" about the class: no need to access a class instance and the method has meaning outside of the context of the class (e.g. https://github.com/kedro-org/kedro/blob/main/kedro/framework/session/session.py#L33-L56) 
3. Inside utils: when it's a helper method used in multiple files and not specific to a certain class. 

When it comes to making things public/private it's a bit different for Viz vs. Kedro but in the framework anything that's _public_ is part of the public API. So it can (and will) be used by outside users directly accessing Kedro as a library. So that means, unless a user should have access to the method, we make it _private_. My personal opinion is that for Viz we should make things _public_ when it's accessed in multiple classes/files.

ravi-kumar-pilla: Thank you for the explanation Merel 💯

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
-        """
+                modular_pipeline.tags.update(kedro_node.tags)
 
-        # There is no need to extract modular pipeline from parameters
-        # because all valid modular pipelines are encoded in either a TaskNode or DataNode.
-        if isinstance(node, ParametersNode):
-            return None
+                # add input/output datasets of a kedro node as a child
+                # to the modular pipeline
+                self._add_datasets_as_children(
```
rashidakanchwala, line 204: I was thinking we could add datasets that have no modular pipeline ID associated with them (and thus belong to the root modular pipeline) in this function instead of in managers.py.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
+                self.node_mod_pipeline_map[dataset] = set()
+            self.node_mod_pipeline_map[dataset].add(modular_pipeline_id)
+
+    def _add_datasets_as_children(
+        self,
+        modular_pipeline: ModularPipelineNode,
+        kedro_node: Node,
+        modular_pipeline_inputs_outputs: Set[str],
+    ):
+        """Helper to add datasets (not parameters) related to task nodes as children.
```
rashidakanchwala, line 295: I was thinking we could add datasets that have no modular pipeline ID associated with them (and thus belong to the root modular pipeline) in this function instead of in managers.py.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,79 +1,124 @@
...
+    return hashlib.sha1(value.encode("UTF-8")).hexdigest()[:8]
+
+
+def _hash_input_output(item: str) -> str:
+    """Hash the input/output dataset."""
+    return (
+        _hash(_strip_transcoding(item))
+        if TRANSCODING_SEPARATOR in item
+        else _hash(item)
+    )
```
idanov, line 30: I would put these in `utils` and import them here, rather than the transcoding logic. Hide as much as possible behind a smaller API footprint.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
-            )
+        hashed_outputs = {_hash_input_output(output) for output in outputs}
+        self.tree[modular_pipeline_id].outputs = hashed_outputs
 
-        is_internal_output = modular_pipeline_id in output_node.modular_pipelines
-        if is_internal_output:
-            self.tree[modular_pipeline_id].internal_outputs.add(output_node.id)
-        else:
-            self.tree[modular_pipeline_id].external_outputs.add(output_node.id)
+    def _add_children(self, modular_pipeline_id: str, kedro_nodes: List[Node]):
```
idanov, line 168: This whole function is overly complicated, very wide open if condition and too much nestedness. Could you try to refactor it into a more concise and easier to follow one?

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
-    def extract_from_node(self, node: GraphNode) -> Optional[str]:
-        """Extract the namespace from a graph node and add it as a modular pipeline node
-        to the modular pipeline repository.
+        for kedro_node in kedro_nodes:
+            # add kedro node as a child to the modular pipeline
+            if kedro_node.namespace == modular_pipeline_id:
+                kedro_node_id = _hash(str(kedro_node))
+                modular_pipeline.children.add(
+                    ModularPipelineChild(id=kedro_node_id, type=GraphNodeType.TASK)
+                )
```
idanov, line 198: Instead of nesting this with 2 extra levels, you could do something like:

```python
children = [n for n in kedro_nodes if n.namespace == modular_pipeline_id]
modular_pipeline.children = [ModularPipelineChild(id=_hash(str(n)), type=GraphNodeType.TASK) for n in children]
```

Much more readable and explainable:
1. We get all the nodes matching exactly the pipeline namespace
2. Add them as children to the pipeline (this one could be `.extend` instead of `=` if there's other places that add to children called before this one)

Two actions, two lines of code...

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
+                    modular_pipeline, kedro_node, modular_pipeline_inputs_outputs
+                )
 
-        modular_pipeline_id = node.namespace
-        if not modular_pipeline_id:
-            return None
+                # update node_modular_pipeline lookup
+                if kedro_node_id not in self.node_mod_pipeline_map:
+                    self.node_mod_pipeline_map[kedro_node_id] = set()
+                self.node_mod_pipeline_map[kedro_node_id].add(modular_pipeline_id)
```
idanov, line 211: If `self.node_mod_pipeline_map` was a `defaultdict(set)`, these 3 lines can be simplified to:

```
self.node_mod_pipeline_map[kedro_node_id].add(modular_pipeline_id)
```

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
+            parent_modular_pipeline.tags.update(modular_pipeline.tags)
 
-        modular_pipeline.tags.update(node.tags)
+            # add current modular pipeline and input/output datasets
+            # of a modular pipeline as a child to the parent modular
+            # pipeline based on the rules
+            self._add_children_to_parent_pipeline(
+                parent_modular_pipeline,
+                modular_pipeline_id,
+                modular_pipeline_inputs_outputs,
```
idanov, line 235: This is very unintuitive - the method is called `_add_children`, but in it we have the totally unrelated code for adding the modular pipeline itself to it's parent's children. Could this be factored out as a separate method which is called right after `_add_children` instead?

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
+                dataset not in parent_node.inputs
+                and dataset not in parent_node.outputs
+                and dataset not in self.parameters
+            ):
+                parent_node.children.add(
+                    ModularPipelineChild(id=dataset, type=GraphNodeType.DATA)
+                )
+            if dataset not in self.node_mod_pipeline_map:
+                self.node_mod_pipeline_map[dataset] = set()
+            self.node_mod_pipeline_map[dataset].add(modular_pipeline_id)
```
idanov, line 287: `defaultdict` to the rescue here...

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
+            if (
+                io_id not in modular_pipeline_inputs_outputs
+                and io_id not in self.parameters
+            ):
+                modular_pipeline.children.add(
+                    ModularPipelineChild(id=io_id, type=GraphNodeType.DATA)
+                )
+                if io_id not in self.node_mod_pipeline_map:
+                    self.node_mod_pipeline_map[io_id] = set()
+                self.node_mod_pipeline_map[io_id].add(modular_pipeline.id)
```
idanov, line 316: `defaultdict`...

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
+        """
+        hashed_io_ids = {
+            _hash_input_output(io)
+            for io in set(kedro_node.inputs) | set(kedro_node.outputs)
+        }
+        for io_id in hashed_io_ids:
+            if (
+                io_id not in modular_pipeline_inputs_outputs
+                and io_id not in self.parameters
+            ):
```
idanov, line 310: This filter could've just been added to the previous `for`-comprehension, or as a separate one on its own and then we don't need the extra nestedness.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +128,201 @@ def get_or_create_modular_pipeline(
...
+                )
+                if io_id not in self.node_mod_pipeline_map:
+                    self.node_mod_pipeline_map[io_id] = set()
+                self.node_mod_pipeline_map[io_id].add(modular_pipeline.id)
+
+    def get_node_and_modular_pipeline_mapping(
+        self, node
+    ) -> Tuple[str, Union[Set[str], None]]:
+        """Get the modular pipeline(s) to which the given node belongs."""
+        node_id = (
```
idanov, line 322: To avoid things like this, I would've created a bit more clever `_hash` helper, which can correctly hash the dataset or the node.

idanov (commented):
From a correctness point of view, this looks good as it is as far as I can tell. There's a few things that can be done to make the style of the code a bit more readable. I've added some comments here and there with pointers. Happy to discuss on a call if you have questions.

package/kedro_viz/data_access/repositories/modular_pipelines.py
rashidakanchwala: dataset or task node*   -- i tried to make this function a bit better but in this function a  'node' can be both dataset or task node. But u r right, the use of 'node' is getting a bit confusing.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -1,79 +1,124 @@
...
+def _hash_input_output(item: str) -> str:
+    """Hash the input/output dataset."""
+    return (
+        _hash(_strip_transcoding(item))
+        if TRANSCODING_SEPARATOR in item
+        else _hash(item)
+    )
 
 
 class ModularPipelinesRepository:
```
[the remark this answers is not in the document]
rashidakanchwala: This is a problem throughout the codebase. I think we need to clarify definitions here and create another refactor PR that tackles this bit.

package/kedro_viz/data_access/repositories/modular_pipelines.py
idanov: I also like the reversal here, start with the shorter one and finish with the longer one.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -165,162 +162,146 @@ def _add_outputs(self, modular_pipeline_id: str, outputs: Set[str]) -> None:
...
-            parent_node (ModularPipelineNode): The parent modular pipeline node.
+            modular_pipeline (ModularPipelineNode): The modular pipeline node.
             modular_pipeline_id (str): The ID of the modular pipeline to add as a child.
-            modular_pipeline_inputs_outputs (Set[str]): A set of inputs/outputs
-                    to/from the modular pipeline
+            modular_pipeline_inputs_outputs (Set[str]): A set of inputs/outputs to/from
+            the modular pipeline.
         """
+        # Determine the parent modular pipeline ID
+        parent_modular_pipeline_id = (
```
idanov, line 265: I'd simply call this `parent_id`, variables have context and in this function the only thing that can be a parent is a modular pipeline, thus it's implied and doesn't need to be part of the variable name.

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -165,162 +162,146 @@ def _add_outputs(self, modular_pipeline_id: str, outputs: Set[str]) -> None:
...
-            # add current modular pipeline and input/output datasets
-            # of a top-level modular pipeline as a child to ROOT modular
-            # pipeline
-            self._add_children_to_parent_pipeline(
-                self.tree[ROOT_MODULAR_PIPELINE_ID],
-                modular_pipeline_id,
-                modular_pipeline_inputs_outputs,
+            modular_pipeline.tags.update(node.tags)
+
+            hashed_io_ids = {
```
idanov, line 222: I would name this just `io_ids`.

idanov (approved):

package/kedro_viz/data_access/repositories/modular_pipelines.py
merelcht: I still think it's a good idea to describe the following code in the docstring:

```
 if is_dataset_param(_input):
                self.parameters.add(hashed_input)
```

package/kedro_viz/data_access/repositories/modular_pipelines.py
```diff
@@ -83,163 +125,185 @@ def get_or_create_modular_pipeline(
...
-        )
-        return modular_pipeline_id
+    def _add_children_to_parent_pipeline(
+        self,
+        modular_pipeline: ModularPipelineNode,
+        modular_pipeline_id: str,
+        modular_pipeline_inputs_outputs: Set[str],
+    ):
+        """
+        Helper to add modular_pipeline children correctly to parent modular pipelines
```
merelcht, line 248:
```suggestion
        Helper to add modular pipeline children correctly to parent modular pipelines

```

merelcht (approved):
Left some final minor comments, but nothing blocking!

PR #1939, opened 2024-06-10
Refactor data access manager for modular pipelines change

### LINKED ISSUE #1899: Refactor Modular Pipelines in the Kedro-viz backend

ravi-kumar-pilla:
## Description

Partially resolves #1899 

## Development notes

* Modify `add_pipeline` method to use `populate_tree` in the modular_pipelines repository to create a modular pipeline tree using KedroPipeline
* Modify `add_node_input` and `add_node_output` to take the `modular_pipelines_repo_obj` which contains the pre-populated modular pipelines tree (earlier we used to construct the tree at a later stage using expand_tree method of - package/kedro_viz/services/modular_pipelines.py)
* Modify `add_node` to take `modular_pipelines_repo_obj` as input
* Modify `add_dataset` to take `modular_pipelines_repo_obj` as input
* Remove the `expand_tree` service call from `create_modular_pipelines_tree_for_registered_pipeline` method as the tree is already constructed
* Update tests to use the edge case scenarios mentioned in :
      - https://github.com/kedro-org/kedro-viz/pull/1651 
      - https://github.com/kedro-org/kedro-viz/issues/1814

## QA notes
- This PR is part of a bigger refactor (https://github.com/kedro-org/kedro-viz/pull/1897) of modular pipelines and is created to ease review process.
- The CI build might fail as this PR is not self-sufficient

## Checklist


merelcht:
Can you explain why it is now necessary to pass the `modular_pipelines_repo_obj` of type `ModularPipelinesRepository` in some of the methods here? With the refactor it seems like the `ModularPipelinesRepository` has a different function than the other `repository` classes/objects.

ravi-kumar-pilla:
> Can you explain why it is now necessary to pass the `modular_pipelines_repo_obj` of type `ModularPipelinesRepository` in some of the methods here? With the refactor it seems like the `ModularPipelinesRepository` has a different function than the other `repository` classes/objects.

Hi @merelcht , Every node in the flowchart model of Kedro-Viz has a 2 way mapping i.e., modular pipeline repository contains a map - `node_mod_pipeline_map` and the node contains a field `modular_pipelines`. Earlier we used to have the below logic to gather the modular_pipelines. With the refactor, we are constructing the modular pipelines ahead of creating the nodes and we build the dict - `node_mod_pipeline_map`. This dict already contains the details of the modular pipelines the node belongs to. So we are passing the `modular_pipelines_repo_obj` while creating the nodes.

```python
    @field_validator("modular_pipelines")
    @classmethod
    def set_modular_pipelines(cls, _, info: ValidationInfo):
        return cls._expand_namespaces(info.data["kedro_obj"].namespace)
```

merelcht:
[WIP] Thanks @ravi-kumar-pilla this is clearer now. My main concern was whether the flow of data still adheres to the architecture (https://miro.com/app/board/uXjVOktl9TY=/ and https://user-images.githubusercontent.com/2032984/118201082-98dcfb80-b44e-11eb-8e63-d7cc7254d1d1.png) and as far as I can tell it does. 

From what I understand: before the refactoring the modular pipelines where populated as part of a call [to the REST API](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/api/rest/responses.py#L301). This would in turn call [the `expand_tree()` method in the `services/modular_pipelines.py`](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/services/modular_pipelines.py#L16). 

Now instead the modular pipelines are populated in the same way as the "regular" pipelines, the catalog

ravi-kumar-pilla:
> Thanks @ravi-kumar-pilla this is clearer now. My main concern was whether the flow of data still adheres to the architecture (https://miro.com/app/board/uXjVOktl9TY=/ and https://user-images.githubusercontent.com/2032984/118201082-98dcfb80-b44e-11eb-8e63-d7cc7254d1d1.png) and as far as I can tell it does.

Yes @merelcht , the data flow is not changed.
 
> From what I understand: before the refactoring the modular pipelines where populated as part of a call [to the REST API](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/api/rest/responses.py#L301). This would in turn call [the `expand_tree()` method in the `services/modular_pipelines.py`](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/services/modular_pipelines.py#L16).

Yes

> Now instead the modular pipelines are populated in the same way as the "regular" pipelines, the catalog and dataset stats from inside `populate_data()` that happens on the server side when the server is started. In fact this happens at the same time as the regular pipelines are populated. The reason why the `ModularPipelinesRepository` object is then used as an argument to subsequent methods in the call stack, is because it's used to determine the nodes + inputs + outputs.

Yes, the `ModularPipelinesRepository` object is now passed to the create_node methods to assign the `modular_pipelines` property on the GraphNode.

package/kedro_viz/data_access/managers.py
```diff
@@ -149,15 +149,18 @@ def add_pipeline(self, registered_pipeline_id: str, pipeline: KedroPipeline):
...
         self.registered_pipelines.add_pipeline(registered_pipeline_id)
+        modular_pipelines_repo_obj = self.modular_pipelines[registered_pipeline_id]
+        modular_pipelines_repo_obj.populate_tree(pipeline)
+
         free_inputs = pipeline.inputs()
 
         for node in pipeline.nodes:
-            task_node = self.add_node(registered_pipeline_id, node)
-            self.registered_pipelines.add_node(registered_pipeline_id, task_node.id)
+            task_node = self.add_node(
```
rashidakanchwala, line 159: since there are two `add_node` functions; it does affect readability. Maybe we add comments before the add_node functions explaining what happens?

package/kedro_viz/data_access/managers.py
```diff
@@ -282,27 +307,60 @@ def add_dataset(
...
+                modular_pipelines_repo_obj.get_or_create_modular_pipeline(
+                    ROOT_MODULAR_PIPELINE_ID
+                )
+            )
+            root_modular_pipeline_node.children.add(
+                ModularPipelineChild(id=dataset_id, type=GraphNodeType.DATA)
+            )
+
+            # update the node_mod_pipeline_map
+            if dataset_id not in modular_pipelines_repo_obj.node_mod_pipeline_map:
```
rashidakanchwala, line 339:
```
if dataset_id not in modular_pipelines_repo_obj.node_mod_pipeline_map:
    modular_pipelines_repo_obj.node_mod_pipeline_map[dataset_id] = {ROOT_MODULAR_PIPELINE_ID}
```

package/tests/test_data_access/test_managers.py
```diff
@@ -375,6 +471,348 @@ def test_add_pipelines(
...
+            )
+            all_node_names |= {node.name for node in _pipeline.nodes}
+
+        assert sorted(
+            {
+                f"{n.namespace}.{n.name}" if n.namespace else n.name
+                for n in data_access_manager.nodes.as_list()
+                if n.type == GraphNodeType.TASK
+            }
+        ) == sorted(list(all_node_names))
```
merelcht, line 530: I find this test quite hard to follow because both sides of the assert are generated in the test. If you look at other tests in this file e.g. https://github.com/kedro-org/kedro-viz/blob/main/package/tests/test_data_access/test_managers.py#L346-L354 you see that one side of the assert is generated whereas the other is very explicit, which makes it easier to reason about IMO.

merelcht (commented):
I think this is nearly there. My main suggestion is to make the asserts in the tests more explicit so instead of something like:
```
assert list(pipelines) == create_list(pipelines)
```

you do:
```
assert ["car_pipeline", "plane_pipeline"] == create_list(pipelines)
```
This explicitness makes the tests easier to read.

ravi-kumar-pilla:
> I think this is nearly there. My main suggestion is to make the asserts in the tests more explicit so instead of something like:
> 
> ```
> assert list(pipelines) == create_list(pipelines)
> ```
> 
> you do:
> 
> ```
> assert ["car_pipeline", "plane_pipeline"] == create_list(pipelines)
> ```
> 
> This explicitness makes the tests easier to read.

Hi @merelcht , 

I understand the edge case test cases are not straightforward to read in the assert statements like you pointed here. I tried to add additional comments to improve readability in the [new commit](https://github.com/kedro-org/kedro-viz/pull/1939/commits/7eb99f370bc4c52c1acd2252814ffd1bc00b9bbc). 

Since these tests are drawing the expected json from the [expected_tree_for_edge_cases](https://github.com/kedro-org/kedro-viz/pull/1939/files#diff-b93828acd96512e087942abebb158d98ed23e4987280bd3c6010d20d675acf9f) , it will be lot of hardcoding of strings in the test file if I extract the details from the json. 

I find this easy to manage in-case there is a change in the modular pipeline tree. Please have a look at the changes in my new commit and let me know if you feel there could be a better way to handle this.

Thank you

merelcht:
> I understand the edge case test cases are not straightforward to read in the assert statements like you pointed here. I tried to add additional comments to improve readability in the [new commit](https://github.com/kedro-org/kedro-viz/pull/1939/commits/7eb99f370bc4c52c1acd2252814ffd1bc00b9bbc).
> 
> Since these tests are drawing the expected json from the [expected_tree_for_edge_cases](https://github.com/kedro-org/kedro-viz/pull/1939/files#diff-b93828acd96512e087942abebb158d98ed23e4987280bd3c6010d20d675acf9f) , it will be lot of hardcoding of strings in the test file if I extract the details from the json.
> 
> I find this easy to manage in-case there is a change in the modular pipeline tree. Please have a look at the changes in my new commit and let me know if you feel there could be a better way to handle this.

I think this is a step into the right direction. Personally, I think hardcoding to enhance test readability is better and forces you to check that the code is still correct when you make changes, because you'll have to manually update your tests. However, since I'm not working on this codebase full time I leave the final decision on how much further to improve the tests up to you and the rest of the Viz team.

package/tests/test_data_access/test_managers.py
```diff
@@ -509,25 +506,37 @@ def test_task_nodes_for_modular_pipeline_edge_cases(
...
+        # setup
         data_access_manager.add_catalog(example_catalog, edge_case_example_pipelines)
         data_access_manager.add_pipelines(edge_case_example_pipelines)
 
-        # testing task nodes in GraphNodesRepository
-        all_node_names = set()
+        # Testing for task node names in NodesRepository
+
+        # Extract the task node names that are present in edge_case_example_pipelines
+        all_task_node_names_from_edge_case_example_pipelines = set()
```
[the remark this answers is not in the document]
ravi-kumar-pilla: This is already coming from a fixture, that is reading the json file. As @merelcht pointed if you feel it is hard to follow, I can hardcode the required part of the json that is present [here](https://github.com/kedro-org/kedro-viz/pull/1939/files#diff-b93828acd96512e087942abebb158d98ed23e4987280bd3c6010d20d675acf9f) for every test.

ravi-kumar-pilla:
> I think this is a step into the right direction. Personally, I think hardcoding to enhance test readability is better and forces you to check that the code is still correct when you make changes, because you'll have to manually update your tests. However, since I'm not working on this codebase full time I leave the final decision on how much further to improve the tests up to you and the rest of the Viz team.

Thank you @merelcht for taking time and reviewing the PR. Rashida and I had a discussion and she suggested of having one integration test for the edge cases to check the entire api response instead of checking individual properties. I have made the changes in this [commit](https://github.com/kedro-org/kedro-viz/pull/1940/commits/e06be430cb67d4a2073ec712381b05275b9a4d2b) .

@rashidakanchwala , I removed the individual tests [here](https://github.com/kedro-org/kedro-viz/pull/1939/commits/0c9099361f40ef24163b98c80786b7f5ac58489b) as we already test for `example_pipelines` in `test_add_pipelines`. Please have a look. Thank you

merelcht (approved):

PR #1876, opened 2024-04-24
CircleCI to Github actions migration for Kedro-Viz (Release)

jitu5:
## Description

Resolves https://github.com/kedro-org/kedro-viz/issues/1785

### Workflows:

*  #### `deploy-demo.yml`:
    * Triggered by push on the demo branch.
    * Sets up the environment, installs AWS CLI, builds a Docker image of the demo project, and pushes it to AWS ECR.
    * The script for building the Docker image and pushing it to AWS ECR is embedded in the Build demo container image step.
    * Creates a new AWS Lightsail deployment using the lightsail.json config file.
    * It deploys the demo project to AWS Lightsail.

*  #### `check-release.yml`:
    * Triggered by push on main branch.
    * The first job - `check-version` checks if the version specified on the kedro_viz/\_\_init\_\_.py file is on PyPI already or not. If it isn't, that means a release should be triggered.
    * The script for checking of versions is tools/github_actions/check_version.py
    * The second job -` test-kedro-viz` runs the test suite with all-checks.yml
    * The third job - `prepare-release`, extracts release notes from RELEASE.md(using tools/github_actions/extract_release_notes.py) and creates a GitHub release.

*  #### `release.yml`:
    * Triggered by a GitHub release publish event.
    * The first job - `release_to_npm` checks out the code, sets up Node.js, installs Node dependencies, builds the React application, and then publishes the package to npm.
    * The second job - `release_to_pypi` checks out the code, sets up tests, makes a Python package, installs twine, and then publishes the package to PyPI.

Screenshot 2024-04-24 at 2 20 15 p m

[Miro board link](https://miro.com/app/board/uXjVKZUQz8Y=/)

#### Notes 
- For demo-project/Dockerfile file, python version updated from 3.8 to 3.9 
- CircleCI config files removed.
- Now publishing to PyPi & npm happening in `check-release -> tests (lint, unit, e2e, js-lint-test) -> tags the release -> releases`
- I successfully ran all the workflow in private repo. For reference:
  For PyPi
  https://github.com/jitu5/kedro-viz/actions/runs/8816265319/job/24200003147
  https://test.pypi.org/project/kedro-vizzz/9.0.0/

  For npm
  https://github.com/jitu5/kedro-viz/actions/runs/8816265319/job/24200002881
  https://www.npmjs.com/package/kedro-vizzz

  For deploy demo
  https://github.com/jitu5/kedro-viz/actions/runs/8805821140/job/24169247490  
[...]

PR #1949, opened 2024-06-18
Fix numpy 2.0 e2e test fail with lower requirements

ravi-kumar-pilla:
## Description

Resolves https://github.com/kedro-org/kedro-viz/actions/runs/9548598519/job/26316143362?pr=1935
```bash
 Full exception: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.                  
```

## Development notes

- Pin numpy in `package/features/steps/lower_requirements.txt` at `1.26.4`
- Pin numpy in `package/features/steps/cli_steps.py` at `1.26.4`
- Doc link fix at `docs/source/publish_and_share_kedro_viz_on_azure.md`

## QA notes

- All tests should pass

## Checklist


ravi-kumar-pilla:
Hi Team,

I am trying to fix e2e test failures on kedro-viz. For both the cases below, pinning `numpy < 2` resolved the issue. But I am not certain to go ahead with this fix as the actual fix might be in the starters. Could someone please confirm if the starters are working fine with latest release of numpy. 

https://github.com/kedro-org/kedro-viz/actions/runs/9548599222/job/26316144191?pr=1935
https://github.com/kedro-org/kedro-viz/actions/runs/9571094566/job/26387337026

**NOTE:** Kedro-Viz itself does not use numpy but seems like pandas uses numpy. Compatibility matrix mentioned [here](https://github.com/numpy/numpy/issues/26191) might help.

cc: @ankatiyar @SajidAlamQB 

Thank you

ankatiyar:
It makes sense to cap `numpy` in `lower_requirements.txt` for this failure - https://github.com/kedro-org/kedro-viz/actions/runs/9548599222/job/26316144191?pr=1935 
Since `pandas==1.15` does not have an upper bound on `numpy`

But why does it need to be pinned in `package/requirements.txt`?

ravi-kumar-pilla:
> But why does it need to be pinned in `package/requirements.txt`? What commit is this failure related to - https://github.com/kedro-org/kedro-viz/actions/runs/9571094566/job/26387337026? It seems like a missing dependency from `kedro-datasets` 👀 @ravi-kumar-pilla

Hi @ankatiyar , Thank you for looking over the PR. This is not related to any new commit. Our builds were failing all of a sudden without any code change (even the old builds if re-run are failing). The e2e test that fails now is for the below scenario -

    Scenario: Execute viz with the earliest Kedro version that it supports 
        Given I have installed kedro version "0.18.3"
        And I have run a non-interactive kedro new with pandas-iris starter
        And I have installed the project's requirements
        When I execute the kedro viz run command
        Then kedro-viz should start successfully
        
It might be some issue with the `pandas-iris` starter requirements as Kedro-Viz works fine with the `demo-project` we have.  I will change the starter for the test and try again. But I wanted to bring this to your attention if we need to push changes to the starters due to the numpy release.

Thank you

ankatiyar:
Okay I created a project with 0.18.3 and tried it and I see what is happening -
For this particular test, when a Kedro project's requirements are installed, it does `pandas==1.15.x` which installs `numpy==2.0.0` because it's not capped. It's the same problem but I don't think we should pin the `numpy` version in `package/requirements.txt` for all of kedro-viz but only for _this_ specific e2e test. 

I don't think anything of Kedro side is breaking since we run our tests only with the latest versions of `pandas` which works fine with the latest `numpy`.

package/features/steps/cli_steps.py
```diff
@@ -80,6 +80,15 @@ def install_project_requirements(context):
     """Run ``pip install -r requirements.txt``."""
     if context.kedro_version != "latest":
         requirements_path = str(context.root_project_dir) + "/src/requirements.txt"
+
+        with open(requirements_path, "r") as req_file:
```
ankatiyar, line 84: I would also add a comment here that this is because `numpy 2.0` breaks with old versions of `pandas` and this could be removed when the lowest version supported is updated

package/features/steps/lower_requirements.txt
```diff
@@ -13,3 +13,4 @@ strawberry-graphql==0.192.0
 networkx==2.5
 orjson==3.9
 secure==0.3.0
+numpy==1.26.4
```
ankatiyar, line 16: Also add a comment here with a reason

RELEASE.md
```diff
@@ -8,16 +8,15 @@ Please follow the established format:
...
 - Enable/disable preview for all the datasets when publishing Kedro-Viz from UI. (#1895)
-
-## Bug fixes and other changes
-
-## Major features and improvements
 - Display published URLs. (#1907)
 
 ## Bug fixes and other changes
 
+- Fix numpy 2.0 e2e test fail with lower requirements. (#1949)
```
ankatiyar, line 18: Doesn't need to be in the release notes I think as it's not a user facing change!

ravi-kumar-pilla: I am not sure on this. We do mention changes in the - ## Bug fixes and other changes section which are not user facing (like - - Migrate from CircleCi to GitHub Actions. (#1876)). But will remove it for this PR. Need to know what to mention in this section of the release note.

ankatiyar (approved):
💯

Huongg (approved):
thanks @ravi-kumar-pilla, looking good

PR #1959, opened 2024-06-27
Add favicon to kedro-viz documentation

### LINKED ISSUE #1724: Documentation: Catch up documentation with Framework
## Description
I've recently made some improvements to the framework documentation to add some sphinx extensions with corresponding features. We should have the same on the Viz (and datasets) docs. Please can you factor in some time in the team to do the following:

* Add the video extension for youtube videos (e.g. if you want to include any of Juan Luis' videos). See [the Framework PR](https://github.com/kedro-org/kedro/pull/3542) for details.
* Add a favicon to the Viz docs + add "last updated" to the Viz docs: [one PR here](https://github.com/kedro-org/kedro/pull/3547)
* Update `robots.txt` to [remove older versions](https://github.com/kedro-org/kedro/pull/3556) from google indexing and also [remove `latest`](https://github.com/kedro-org/kedro/pull/3564)

Huongg:
## Description
Fixes #1724 , as a part of the ticket is to add the favicon to our documentation. Ideally we should do it directly in kedro-sphinx-theme.

However @astrojuanlu pointed out that there's some issue with Sphinx theme don't normally have their own conf.py ([issue here](https://github.com/kedro-org/kedro-sphinx-theme/issues/8)), therefore the fix for now will be directly inside kedro-viz doc, and the same for kedro datasets

The favicon is now shown through the doc build https://kedro--1959.org.readthedocs.build/projects/kedro-viz/en/1959/
## Checklist

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -41,8 +41,12 @@ extensions = [
     "myst_parser",
     "notfound.extension",
     "sphinxcontrib.jquery",
+    'sphinx_favicon',
 ]
 
+favicons = [
+    "https://kedro.org/images/favicon.ico",
+]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,6 +15,7 @@ Please follow the established format:
 - Conditionally move session store and stats file to .viz directory. (#1915)
 
 ## Bug fixes and other changes
+- Add favicon to kedro-viz documentation. (#1959)
 
 # Release 9.1.0
 
--- a/package/setup.py
+++ b/package/setup.py
@@ -52,7 +52,7 @@ setup(
     },
     extras_require={
         "docs": [
-            "kedro-sphinx-theme==2024.4.0",
+            "kedro-sphinx-theme @ git+https://github.com/kedro-org/kedro-sphinx-theme.git@fix-favicon",
         ],
         "aws": ["s3fs>=2021.4"],
         "azure": ["adlfs>=2021.4"],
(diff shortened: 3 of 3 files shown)


astrojuanlu (approved):
LGTM!

docs/source/conf.py
```diff
@@ -41,8 +41,12 @@
     "myst_parser",
     "notfound.extension",
     "sphinxcontrib.jquery",
+    'sphinx_favicon',
 ]
 
+favicons = [
+    "https://kedro.org/images/favicon.ico",
```
jitu5, line 48: @Huongg Is it possible to add favicon locally from repo instead of downloading from https://kedro.org/images/favicon.ico

Huongg: hey @jitu5  it's possible but it means we have to do the same download for framework and kedro-datasets, at the moment 3 repos are using the same image from the link

SajidAlamQB (approved):
Thank you @Huongg!

docs/source/conf.py
ravi-kumar-pilla: I feel having it in one place for all the repos is easy for maintenance. Also, where are we actually downloading this favicon, here we are just defining the source. Do you know where is the download code ? or does sphinx do it internally

jitu5: If we are already following this then its fine. I just asked because if the link somehow gets broken, it will break everywhere, and we won't know about it.

ravi-kumar-pilla (approved):
LGTM 👍 ... thank you @Huongg

docs/source/conf.py
Huongg: @ravi-kumar-pilla i dont think it's just accessing the link, so the proper fix will be to get all of this handled inside kedro-sphinx-theme but it requires a lot of investigation i think. Juanlu has already opened the ticket for it so hopefully we get it fixed soon

PR #1920, opened 2024-05-22
Refactor CLI with lazy subcommands and deferring imports

ankatiyar:
## Description

Related to https://github.com/kedro-org/kedro/pull/3883 and resolves a follow up action mentioned [here](https://github.com/kedro-org/kedro/issues/1476#issuecomment-2227978718)

Thanks to @ankatiyar for suggesting the changes and making the POC

## Development notes

- Removed single `package/kedro_viz/launchers/cli.py` file which handles all Kedro-Viz CLI commands 
- Created multiple files to handle each Kedro-Viz CLI command (like package/kedro_viz/launchers/cli/build.py, run.py etc)
- Deferred dependencies until they are required
- Created LazyDefaultGroup class as mentioned [here](https://click.palletsprojects.com/en/8.1.x/complex/#lazily-loading-subcommands)
- Updated tests

> [!Tip]
> Read on the docs for [lazy-loading-subcommands](https://click.palletsprojects.com/en/8.1.x/complex/#lazily-loading-subcommands) and learn more about the initial proposal [here](https://github.com/kedro-org/kedro/issues/1476#issuecomment-2139515747)

## QA notes

- All tests should pass
- Test Kedro-Viz CLI commands (`kedro viz -h`, `kedro viz`, `kedro viz run`, `kedro viz build`, `kedro viz deploy`) with and without options
- Make observations similar to the ones made below -

The lazy subcommand loading and refactoring of viz CLI helped in `kedro` init time as viz cli had some expensive imports all in one file `cli.py`. 

Few observations on demo-project with my system running on `macOS Sonoma (2.4 GHz 8-Core Intel i9, 64GB)` -

**( Comparing after system warmup (initial run) as I am not sure how to clear cache or make this a cold start in the CLI. Please let me know if there is a way to make a cold start again )**

**Before -**

Time taken for `kedro -h` - **4sec**

image

Time taken for `kedro registry list` - **8sec**

image

Time taken for `kedro viz run` to reach the actual `run` method - **6sec**

To run this, you need [pyinstrument](https://github.com/joerick/pyinstrument) and [speedscope](https://github.com/jlfwong/speedscope)

```bash
# Commands used for profiling

pyinstrument -r speedscope -o profile.speedscope.json -m kedro viz run
speedscope profile.speedscope.json
```

image

---------------------

**After -**

Time taken for `kedro -h` - **2sec**

image

Time taken for `kedro registry list` - **6sec**

image

[...]

Issue #1742, opened 2024-02-08
Kedro-viz --lite  : Build DAG without importing the code


astrojuanlu:
## Description
kedro-viz has lots of heavy dependencies. At the same time, it needs to `import` the pipeline code to be able to function, even when doing an initial export with `--save-file`. This means that sometimes using Kedro Viz is difficult or impossible if Viz dependencies clash with the project dependencies, which can happen often.

One outstanding example of that has been the push for Pydantic v2 support #1603.

Another example, @inigohidalgo says "due to the heavy deps from viz i usually have my dev venv but I create another one just for viz where i just install viz over whatever project I have installed, overriding the project's dependencies with viz's" and asks "do you know if anybody has tested using kedro viz as an "app", so installing it through pipx or smth similar? is that even possible with how viz works?". https://linen-slack.kedro.org/t/16380121/question-regarding-kedro-viz-why-is-there-a-restriction-on-p#38213e99-ba9d-4b60-9001-c0add0e2555b

## Possible Implementation
One way to do it is to tell Kedro users to write their pipelines in YAML https://github.com/kedro-org/kedro/issues/650, https://github.com/kedro-org/kedro/issues/1963

## Possible Alternatives
Another way would be to do some sort of AST scanning of the Python code, assuming that in some cases this would fail or not be accurate.

Yet another way would be to extract the minimal amount of code that does the `--save-file` and decouple it from the web application that serves it with `--load-file`.

There are possibly other alternatives.

## Checklist

- [x] Include labels so that we can categorise your feature request

astrojuanlu:
Tangentially related: https://openlineage.io/ (as a means to export Kedro pipelines)

noklam:
I've seen Openlineage in a few issues, but is it related to this? From what I understand it's more about understanding the lineage between systems, how data flows from different databases/table to downstream application etc.

datajoely:
The acceptance criteria for this is simple - As a user I shouldn't need a full Spark installation to view Kedro-Viz for a project which uses Spark to process data.

ravi-kumar-pilla:
> The acceptance criteria for this is simple - As a user I shouldn't need a full Spark installation to view Kedro-Viz for a project which uses Spark to process data.

Hi @datajoely , 

[...]

Issue #1603, opened 2023-10-24
Support for Pydantic V2 [Planned for Q1 2024]


ravi-kumar-pilla:
## Description

* This will bring pydantic v2 features to Kedro Viz

## Context
* Pydantic v2 has changed its implementation to use Rust under the hood which makes it faster

## Possible Implementation
https://docs.pydantic.dev/2.0/migration/

Timelines - We will work on the migration to pydantic v2 next year (Q1, 2024)

## Possible Alternatives
NA

## Checklist

- [x] Include labels so that we can categorise your feature request

antonymilne:
I think this is not the right way to do things (or at least not for a long time). While the addition of `model_post_init` in pydantic v2 is very nice and would make it very easy to convert from dataclasses `__post_init__`, doing this would mean dropping support for pydantic v1, which I don't think we want to do for e.g. 3-6 months at least. The performance benefit of pydantic v2 will really be insignificant for kedro-viz and we don't especially need any of the features that pydantic v2 brings.

Besides, many of the things that are doing in `__post_init__` in dataclasses are actually better suited to [validators](https://docs.pydantic.dev/1.10/usage/validators/) (note v1 documentation). Let me leave a few comments on https://github.com/kedro-org/kedro-viz/pull/1565/ to illustrate.

So what I would recommend is to support both pydantic v1 and v2 but use only the v1 API, as in https://github.com/kedro-org/kedro-viz/pull/1529. Then in due course convert from pydantic v1 to v2 API, which means you can then use `model_post_init` where useful, and drop support for pydantic v1 entirely.

astrojuanlu:
> This is because apache-airflow has moved to pydantic v2 in 2.8.0

https://github.com/kedro-org/kedro-plugins/pull/511#issue-2071944428

There is more pressure from upstream dependencies to support Pydantic v2.

astrojuanlu:
Also to make things clear, supporting Pydantic v2 is not about performance (which for Kedro-Viz will have a negligible impact), but about allowing Kedro-Viz to coexist with Pydantic v2 in the same environment.

butterlyn:
@astrojuanlu Hi, big fan of Kedro (and your latest YouTube video series), trying to convince my organisation to adopt the platform in day-to-day operations.

[...]

PR #1966, opened 2024-07-03
Lightweight Kedro Viz Experimentation using AST

ravi-kumar-pilla:
## Description

Related to #1742 

Kedro-Viz has lots of heavy dependencies. At the same time, it needs to import the pipeline code to be able to function, even when doing an initial export with --save-file. This means that sometimes using Kedro-Viz is difficult or impossible if Viz dependencies clash with the project dependencies, which can happen often.

One example of that has been the push for Pydantic v2 support https://github.com/kedro-org/kedro-viz/issues/1603.

Another example, @inigohidalgo says "due to the heavy deps from viz i usually have my dev venv but I create another one just for viz where i just install viz over whatever project I have installed, overriding the project's dependencies with viz's" and asks "do you know if anybody has tested using kedro viz as an "app", so installing it through pipx or smth similar? is that even possible with how viz works?". https://linen-slack.kedro.org/t/16380121/question-regarding-kedro-viz-why-is-there-a-restriction-on-p#38213e99-ba9d-4b60-9001-c0add0e2555b

The acceptance criteria for this is simple - As a user I shouldn't need a full Spark installation to view Kedro-Viz for a project which uses Spark to process data.

## Development notes

Added an option `--lite` in Kedro-Viz CLI. When users execute the command `kedro viz --lite` , it takes the approach mentioned below -

**Using AST + Mock Imports :**

 **Steps:**
    1. Parse the Kedro project using AST
    2. Locate all the import statements
    3. Try importing the located statements
    4. Mock the dependencies in-case of an import error
    5. Patch sys modules with the mocked modules before retrieving the pipelines information from Kedro.

**Testing:**

I have tested basic Kedro projects with -

- [x] Dataset factory patterns
- [x] Starter projects
- [x] Demo-Project on Kedro-Viz
- [x] [MLOps project](https://github.com/astrojuanlu/workshop-from-zero-to-mlops/)
- [x] Spark pet project with `spark.driver.host` configured to localhost. The idea was to test spark initialization both via hooks and outside of hooks
- [x] [Dynamic Pipelines project](https://getindata.com/blog/kedro-dynamic-pipelines/) getindata tutorial

**Observations:**

On macOS Sonoma (2.4 GHz 8-Core Intel i9, 64GB) - These observations might differ as my system was a bit slow while doing tests. But this should give a basic idea of performance. All the tests are run using `time <command>`. To summarize, `kedro viz --lite` was faster (~10-15sec) though `get_mocked_modules()` took (~1-2sec), initializing `DataCatalogLite` instead of `DataCatalog` saved time.

```bash
**Project: demo_project**

Run 1 - kedro viz  2.69s user 0.71s system 7% cpu 47.519 total
Run 2 - kedro viz  2.78s user 0.75s system 7% cpu 44.181 total

Run 1 - kedro viz --lite  2.79s user 0.74s system 7% cpu 46.126 total
Run 2 - kedro viz --lite  2.76s user 0.74s system 7% cpu 44.393 total

In env with no dependencies installed -

Run 1 - kedro viz --lite  2.43s user 0.93s system 9% cpu 37.082 total
Run 2 - kedro viz --lite  2.32s user 0.67s system 10% cpu 28.553 total

Lite_Parser.get_mocked_modules() - python3 -m lite_parser  1.76s user 0.35s system 86% cpu 2.433 total

**Project: MLOps**

Run 1 - kedro viz  4.48s user 1.86s system 11% cpu 54.817 total
Run 2 - kedro viz  4.52s user 1.59s system 13% cpu 46.505 total

Run 1 - kedro viz --lite  3.98s user 1.42s system 24% cpu 22.159 total
Run 2 - kedro viz --lite  3.99s user 1.42s system 19% cpu 28.372 total

In env with no dependencies installed -

Run 1 - kedro viz --lite  2.35s user 0.71s system 11% cpu 27.291 total
Run 2 - kedro viz --lite  2.34s user 0.69s system 13% cpu 23.242 total

Lite_Parser.get_mocked_modules() - python3 -m lite_parser  0.14s user 0.12s system 15% cpu 1.658 total

**Project: Pyspark dummy project**

Run 1 - kedro viz  3.66s user 1.32s system 6% cpu 1:14.60 total
Run 2 - kedro viz  3.22s user 1.05s system 32% cpu 13.233 total

Run 1 - kedro viz --lite  3.70s user 1.18s system 16% cpu 29.910 total
Run 2 - kedro viz --lite  3.18s user 1.07s system 32% cpu 13.083 total

In env with no dependencies installed -

Run 1 - kedro viz --lite  2.38s user 0.75s system 10% cpu 28.534 total
Run 2 - kedro viz --lite  2.03s user 0.60s system 23% cpu 11.011 total

Lite_Parser.get_mocked_modules() - python3 -m lite_parser  0.51s user 0.18s system 54% cpu 1.269 total

**Project: spaceflights with dynamic pipelines**

Run 1 - kedro viz  2.73s user 0.89s system 7% cpu 46.366 total
Run 2 - kedro viz  2.41s user 0.60s system 20% cpu 14.372 total

Run 1 - kedro viz --lite  2.73s user 0.73s system 9% cpu 35.341 total
Run 2 - kedro viz --lite  2.42s user 0.62s system 11% cpu 27.509 total

In env with no dependencies installed -

Run 1 - kedro viz --lite  2.00s user 0.57s system 16% cpu 15.596 total
Run 2 - kedro viz --lite  1.98s user 0.57s system 21% cpu 11.801 total

Lite_Parser.get_mocked_modules() - python3 -m lite_parser  0.52s user 0.21s system 53% cpu 1.369 total
```

>[!NOTE] 
> I have also performed monitoring using `line_profiler` which gave similar results. This ticket may not improve `kedro viz` performance but makes it run with missing external dependencies. However, we can improve the overall performance once https://github.com/kedro-org/kedro-viz/pull/1920 and https://github.com/kedro-org/kedro-viz/pull/1920#issuecomment-2292672816 are implemented

**Limitations:**
    1. If the datasets are not resolved in the catalog, they will be defaulted to MemoryDataset
    2. Since MemoryDatasets do not have layer information, the layers will not be shown in the flowchart if the datasets are 
        not resolved
    3. Experiment Tracking will not work if the datasets are not resolved and the pre-requisite of having kedro-datasets 
        version 2.1.0 and above is not met.
    4. The metadata panel for a data node shows the data node type as MemoryDataset if the dataset is not resolved

 **Next Steps:**
- [x] Once this PR is reviewed and good to go, I will add documentation as needed in a separate PR.
- [x] MemoryDataset to custom ImportErrorDataset in-case of missing dataset dependencies
- [x] CLI showing a warning message with missing imports (if there are any missing imports) - This is already done, but 
need to confirm with @stephkaiser 
- [x] FE notice with a warning of missing imports. This needs a separate ticket and discussion with @stephkaiser
- [ ] Word about the `--lite` flag. Once this PR is merged and we have the above tasks complete, I will demo this feature 
        in the Coffee chat (Sep 1st or 2nd week). 

## QA notes

**Steps to test -**
1. Create a new conda env - `conda create -n viz-parser-test python=3.11`
5. Activate the created env - `conda activate viz-parser-test`
6. Install Kedro - `pip install kedro`
7. Install Kedro-Viz (current parser branch) -

```bash
git clone https://github.com/kedro-org/kedro-viz.git
cd kedro-viz
git checkout feature/kedro-viz-lite

pip install -e package
```
8. Create a spaceflights starter project - `kedro new --starter=spaceflights-pandas`
9. Navigate to your Kedro project - `cd spaceflights-pandas`
10. Run `kedro viz`
11. It throws an error -

```bash
raise DatasetError(
kedro.io.core.DatasetError: An exception occurred when parsing config for dataset 'companies':
Class 'pandas.CSVDataset' not found, is this a typo?
```
12. Run `kedro viz --lite`
13. Kedro Viz should start successfully 

image

## Credits:
1. Thank you Kedro TSC for suggesting the approach taken in this PR
2. Thank you @noklam for having an initial look at the PR and suggesting the sys modules patch and custom DataCatalog implementation
4. Thank you Kedro Team for writing an awesome TestSuite for DataCatalog which I reused in DataCatalogLite for coverage. (Let me know if there is a better way to do instead of duplicating the test code)

## Checklist


package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -91,30 +92,54 @@ def load_data(
...
+        context = KedroContext(
+            package_name="{{ cookiecutter.python_package }}",
+            project_path=project_path,
+            config_loader=OmegaConfigLoader(conf_source=str(project_path)),
+            hook_manager=_VizNullPluginManager(),
+            env=env,
+        )
+
+        # [TODO: Confirm on the session store creation]
+        session_store = None
```
noklam, line 106: couple of things to consider:
- `settings.py` are not read because we are skipping `bootstrap_project` or `configure_project`. See https://docs.kedro.org/en/stable/kedro_project_setup/session.html#bootstrap-project-and-configure-project . We need to read `settings.py` to ensure `CONFIG_LOADER_ARGS` are read.
- OmegaConfigLoader need to takes argument. In framework settings it's usually `OmegaConfigLoader(conf_source="conf", base_env="base", default_run_env="local")`, but as mentioned above we need to take the settings from `settings.py` if provided. Otherwise use the default as described.

package/kedro_viz/integrations/kedro/lite_parser.py
```diff
@@ -0,0 +1,242 @@
...
+            # dump the visiting node for debugging]
+            logger.error(exc)
+            logger.info(ast.dump(node, indent=2))
+
+
+class KedroPipelineExplorer(ast.NodeVisitor):
+    # [TODO: Current explorer only serves for 1 pipeline() function within a create_pipeline def]
+    def __init__(self):
+        # keeping these here for future use-case
+        # when dealing with multiple `pipeline()` functions
```
noklam, line 68: As you cannot use `find_pipeline`, you will likely need something like this to discover the pipeline in a similar way. 

https://github.com/kedro-org/vscode-kedro/blob/91e14d320e0d94b5bc3dc05148955f0f16552308/bundled/tool/lsp_server.py#L482-L499

noklam (commented):
I only review the high level approach without diving into details yet. Can you tell me how should I test this PR or testing the parser separately?

astrojuanlu:
Thanks a lot @ravi-kumar-pilla for the super detailed writeups!! 👏🏼 

One question on the approach, just by quickly reading your first comment:

>  4. Mock the dependencies in-case of an import error

Have you considered _unconditionally_ mock the dependencies? So that `--lite` is _always_ faster than the regular version, whether the user has the dependencies or not.

ravi-kumar-pilla:
> One question on the approach, just by quickly reading your first comment:
> 
> > 4. Mock the dependencies in-case of an import error
> 
> Have you considered _unconditionally_ mock the dependencies? So that `--lite` is _always_ faster than the regular version, whether the user has the dependencies or not.

Good question @astrojuanlu, Yes I did consider that, but I was thinking of partial environments where users have some dependencies resolved. But we can mock everything if we as a team, decide that the `--lite` version should do so. I personally feel we can mock only the missing imports. Thank you

astrojuanlu:
> One challenge would be to not mock pipelines that are imported from relative or other projects.

Yeah, good point...

My concern here is that, if I understand correctly, users with fully working environments won't see any differences between `--lite` and regular, am I right?

astrojuanlu:
Maybe we should implement both modes then:

- (Current)
- "Lite": mock imports if they fail
- "Litest": mock all imports except pipelines

and try to get users to test both.

In any case, I understand the difficulty of the "litest" mode, so it can be left for a future PR. On the other hand, we shouldn't change the default mode just yet, and make `--lite` opt-in for now until we're confident that it helps.

noklam:
Please re-request me when this is ready to be reviewed.

noklam:
I managed to fix the build issues, quickly test on it:
image

- I don't see the MagicMock function anymore, and seems like all the mocking is done correctly.
- "Show Code" feature works properly.

noklam:
Moving some conversation here for records:
> since demo_project (any kedro project on which we run kedro viz --lite) is not available in the python environment, all absolute paths of the package like demo_project.pipelines.reporting.nodes will say No module named 'demo_project' when using importlib.util.find_spec(module_name)
> To resolve this, I tried doing bootstrap_project(project_path). This helped in getting demo_project recognizable as a module

The important thing here is making the package is in `sys.path`, which is what `bootstrap_project` do. bootstrap_project will however load the project via `configure_project`. I suspect this is the reason why `sklearn` is imported.

image

Is this intended? Are you only mocking in case there are missing dependencies now? (there were some discussion in TD before so I am not sure what's the latest design). I think this is fine but just want to confirm if it's expected behavior.

noklam:
Overall I think this is very close already! I find a way to break it.

in `nodes.py` I add:
```
from ...pkl import abc # doesn't exist
```

Then I get an error:
> ModuleNotFoundError: No module named 'demo_project.pkl'

This is not a common case so maybe it's not necessary to fix it within the PR.  I think the strategy here is:
1. `first check if `demo_project.pkl` exists
2. If it exists, mock `demo_project.pkl.abc`, if not, mock `demo_project.pkl` and `demo_project.pkl.abc`

ravi-kumar-pilla:
> Is this intended? Are you only mocking in case there are missing dependencies now? (there were some discussion in TD before so I am not sure what's the latest design). I think this is fine but just want to confirm if it's expected behavior.

Yes, we are mocking only missing external dependencies

package/kedro_viz/integrations/kedro/lite_parser.py
```diff
@@ -0,0 +1,175 @@
...
+                if (self._package_name and self._package_name in module_name) or (
+                    # dev or standalone mode
+                    not self._package_name
+                    and self._is_relative_import(module_name)
+                ):
+                    continue
+
+                # absolute modules in the env
+                if level == 0:
+                    self._create_mock_imports(module_name, mocked_modules)
```
noklam, line 148: Can you add some comments for these cases?

noklam (commented):
Overall I can approve this as is now, and leave some refactoring/minor improvement for follow up

Minor comments: I found it slightly hard to follow the code when I read them, for example:

`_mock_missing_dependencies` and `_create_mock_imports` sounds like exactly the same thing. It may be easier to understand to separate it into two steps, one to create a collection of module to be mocked. Then simply look through that collections and create mocks. It probably makes this a bit easier to test as well.

ravi-kumar-pilla:
> Overall I think this is very close already! I find a way to break it.
> 
> in `nodes.py` I add:
> 
> ```
> from ...pkl import abc # doesn't exist
> ```
> 
> Then I get an error:
> 
> > ModuleNotFoundError: No module named 'demo_project.pkl'
> 
> This is not a common case so maybe it's not necessary to fix it within the PR. I think the strategy here is:
> 
> 1. `first check if `demo_project.pkl` exists
> 2. If it exists, mock `demo_project.pkl.abc`, if not, mock `demo_project.pkl` and `demo_project.pkl.abc`

Yes, I tried mocking relative imports by converting them to absolute and did try different variations. You can find some [here](https://github.com/kedro-org/kedro-viz/pull/1966/commits/81621479992bc0c04eeeb7331c0d827e3cc67fe1#diff-08bcfa0f66f00b2b636e997393c802ca88f6f6ff0637b178d18838e542f5c12dR180) . So there are few ways we can handle relative imports -

1. Only check if the path exists
2. Actually try importing the module (similar to your suggestion above)

The issue with the approach of importing the module (if you remember from our conversation) where I mentioned -
> Now if  demo_project.pipelines.reporting.nodes  file has an import statement like import seaborn as sn and if seaborn is not available, the module demo_project.pipelines.reporting.nodes  also gets mocked as it is not importable (which was happening this morning).

So if we try to check for relative imports, there is a chance we mock the relative import just because there is an un-resolved external dependency in the relative module. Now there are work-arounds for this, (like keep track of the files which have mocked absolute imports etc)

**What can we do -**

1. If we decide on handling all the import types external, internal (relative imports that start with dot (eg., ..nodes, .nodes etc) and relative imports from package root dir (eg., demo_project.pipelines.reporting.nodes etc), this may also increase the time it takes for mocking
2. Try to mock external missing dependencies and leave the responsibility of having relative imports resolvable to the user (like if the user has an import statement like ..nodes from a file, that file should be present at the least). The current status of the PR does this.

Thank you

ravi-kumar-pilla:
> `_mock_missing_dependencies` and `_create_mock_imports` sounds like exactly the same thing. It may be easier to understand to separate it into two steps, one to create a collection of module to be mocked. Then simply look through that collections and create mocks. It probably makes this a bit easier to test as well.

Yes, I will rename and add comments for better understanding. The collection approach is what I initially did. But, I felt the storing of missing dependencies and looping over them again is two-fold and not necessary. In the new approach, I mock the modules on the initial AST parsing.

Also, I will be adding performance metrics as I tested with different use cases -

1. MLOps - This was done on Juan Lu repo - https://github.com/astrojuanlu/workshop-from-zero-to-mlops/
2. Spark - I created a dummy project with spark.yml configured to localhost. The idea was to test spark initialization both via hooks and outside of hooks. Since we have --ignore-hooks by default.
3. Dynamic pipelines - Using getInData tutorial project https://getindata.com/blog/kedro-dynamic-pipelines/

Thank you

package/kedro_viz/integrations/kedro/lite_parser.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def get_mocked_modules(self) -> Dict[str, MagicMock]:
+        """Returns mocked modules for all the dependency errors
+        as a dictionary for each file in your Kedro project
+        """
+        mocked_modules: Dict[str, MagicMock] = {}
+
+        for file_path in self._project_file_paths:
+            with open(file_path, "r", encoding="utf-8") as file:
```
noklam, line 181: Nit: Can it be less couple with viz/project structure. Let say I want to mock an arbitary file, I should be able to reuse the same class.

i.e maybe it could be `def get_mocked_modules(self, paths)` instead? This should make testing much easier as you can test on a single file.

noklam (approved):
Great work! Really happy to see this is getting to the finish line. This would be useful in many ways and speed up kedro viz a lot!

I approved with a minor comment as I don't want to block the PR.

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -142,10 +157,19 @@ def load_data(
         bootstrap_project(project_path)
 
     if is_lite:
-        lite_parser = LiteParser(project_path, package_name)
-        mocked_modules = lite_parser.get_mocked_modules()
+        lite_parser = LiteParser(package_name)
+        unresolved_imports = lite_parser.parse(project_path)
```
noklam, line 161: ⭐️

noklam (approved):

package/kedro_viz/integrations/kedro/lite_parser.py
```diff
@@ -0,0 +1,265 @@
...
+                unresolved_imports[str(target_path)] = missing_dependencies
+            return unresolved_imports
+
+        # handling directories
+        _project_file_paths = set(target_path.rglob("*.py"))
+
+        for file_path in _project_file_paths:
+            missing_dependencies = self._get_unresolved_imports(
+                file_path, _project_file_paths
+            )
```
astrojuanlu, line 261: This is traversing everything, including virtual environments

ravi-kumar-pilla: The check is inside _get_unresolved_imports, will move it out. But I wonder why it is still parsing venv

ravi-kumar-pilla: @astrojuanlu fixed this. Please review it when you have some time. Thank you

astrojuanlu (approved):
I gave this a quick test and it works! 💯 Thanks @ravi-kumar-pilla!

package/kedro_viz/data_access/managers.py
```diff
@@ -316,7 +318,17 @@ def add_dataset(
...
+        try:
+            obj = self.catalog.get_dataset(dataset_name)
+        except DatasetError:
+            # This is to handle dataset factory patterns when running
+            # Kedro Viz in lite mode. The `get_dataset` function
+            # of DataCatalog calls AbstractDataset.from_config
+            # which tries to create a Dataset instance from the pattern
+
+            # pylint: disable=abstract-class-instantiated
+            obj = MemoryDataset()  # type: ignore[abstract]
```
rashidakanchwala, line 330: For now, this approach works, but in the future, we may need to evaluate whether to represent this as a `MemoryDataset` or as a new term like  `UnavailableDataset` (or similar). The latter would specifically denote a dataset that cannot be loaded due to missing dependencies

package/tests/test_data_access/test_managers.py
```diff
@@ -27,25 +28,6 @@ def identity(x):
     return x
 
 
-def assert_expected_modular_pipeline_values_for_edge_cases(
```
rashidakanchwala, line 30: why do we remove this?

package/tests/test_integrations/test_data_catalog_lite.py
```diff
@@ -0,0 +1,323 @@
...
+            # DataCatalogLite should not raise DatasetError
+            DataCatalogLite.from_config(**sane_config)
+        except DatasetError:
+            pytest.fail("DataCatalogLite.from_config raised DatasetError unexpectedly")
+
+    @pytest.mark.skipif(
+        sys.version_info < (3, 9),
+        reason="for python 3.8 kedro-datasets version 1.8 is used which has the old spelling",
+    )
+    def test_config_incorrect_spelling(self, sane_config):
```
rashidakanchwala, line 147: nice!

package/kedro_viz/data_access/managers.py
rashidakanchwala: Do we need to do this here, as `get_dataset` handles `DatasetNotFoundError` with `MemoryDataset` anyway?

rashidakanchwala: also would it make sense to define catalog as a Union[DatasetCatalog, DatasetCatalogLite] -- i am not completely sure how much it would make the code better .

package/tests/test_integrations/test_data_catalog_lite.py
ravi-kumar-pilla: thanks to Kedro team :D

package/kedro_viz/data_access/managers.py
ravi-kumar-pilla: > we may need to evaluate whether to represent this as a MemoryDataset or as a new term like UnavailableDataset (or similar). This would specifically denote a dataset that cannot be loaded due to missing dependencies

Yes I wanted to have a discussion and we can handle this here - https://github.com/kedro-org/kedro-viz/issues/2058

> Do we need to do this here, as get_dataset handles DatasetNotFoundError with MemoryDataset anyway?
Also do we need to add the MemoryDataset logic to _get_dataset in DataCatalogLite?

So the catalog here is DataCatalog instance and not DataCatalogLite. We can either convert the CatalogRepository to hold DataCatalogLite instance or handle it here. I felt since we have 2 paths - normal and lite, for now we can handle this here which is much cleaner way. In future if we decide to go with DataCatalogLite for Kedro-Viz as a normal path, we can make the change. Wdyt ?

package/kedro_viz/integrations/kedro/data_catalog_lite.py
```diff
@@ -0,0 +1,76 @@
...
+        catalog = copy.deepcopy(catalog) or {}
+        credentials = copy.deepcopy(credentials) or {}
+        save_version = save_version or generate_timestamp()
+        load_versions = copy.deepcopy(load_versions) or {}
+        user_default = {}
+
+        for ds_name, ds_config in catalog.items():
+            if not isinstance(ds_config, dict):
+                raise DatasetError(
+                    f"Catalog entry '{ds_name}' is not a valid dataset configuration. "
```
rashidakanchwala, line 38: can you pls explain this part -- a bit confused when this error happens ?

package/kedro_viz/integrations/kedro/data_catalog_lite.py
```diff
@@ -0,0 +1,76 @@
...
+        dataset_patterns = {}
+        catalog = copy.deepcopy(catalog) or {}
+        credentials = copy.deepcopy(credentials) or {}
+        save_version = save_version or generate_timestamp()
+        load_versions = copy.deepcopy(load_versions) or {}
+        user_default = {}
+
+        for ds_name, ds_config in catalog.items():
+            if not isinstance(ds_config, dict):
+                raise DatasetError(
```
rashidakanchwala, line 37: Could you pls explain this error handling more ? as i am a bit confused ... when it happens, how the user will understand?

ravi-kumar-pilla: This is from the original DataCatalog implementation. This might happen when the config provided is not a dictionary, lets say 

```yaml
companies: "type: pandas.csvdataset"
```

package/kedro_viz/integrations/kedro/data_loader.py
```diff
@@ -109,12 +109,81 @@ def load_data(
...
         # Pipelines is a lazy dict-like object, so we force it to populate here
         # in case user doesn't have an active session down the line when it's first accessed.
         # Useful for users who have `get_current_session` in their `register_pipelines()`.
         pipelines_dict = dict(pipelines)
         stats_dict = _get_dataset_stats(project_path)
-
     return catalog, pipelines_dict, session_store, stats_dict
+
+
+def load_data(
```
rashidakanchwala, line 129: does it make sense to break down the function like this 

public method `load_data()` which called private `_load_data_lite` if `is_lite` else `_load_data`

ravi-kumar-pilla: Yes I initially did that when I created a dummy session, had a different config loader etc. 

But now we have most of the parts like session, context etc same in both normal and lite version. Also since lite version supports partial rendering (i.e., if some dependencies are present we stick to that), the only change we did is to initialize DatCatalogLite and patch the mocked dependencies. 

I felt this is a good way to breakdown to reuse session creation and other necessary objects

rashidakanchwala (approved):
AMAZZING!!!

Issue #1921, opened 2024-05-22
Enable user to disable preview for all from setting panel(UI) when running `kedro viz run`


jitu5:
## Description
This feature will help users to control data preview for all the nodes from setting panel of Kedro-Viz when running `kedro viz run`.

## Context
After running Kedro-Viz locally by `kedro viz run` command, Dataset preview for all the nodes will be enabled by default but user will be about to enable/disable from Kedro-Viz setting panel.

## Possible Implementation
-Add toggle button option in setting panel
-Pass current toggle value to getNodeById API call

## Checklist

- [X] Include labels so that we can categorise your feature request

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro-viz/pull/1977

PR #1977, opened 2024-07-10
Add disable preview toggle in settings panel

SajidAlamQB:
## Description
Related to: #1921

This feature enables users to control data previews for all nodes from the settings panel of Kedro-Viz when running kedro viz locally.

After running Kedro-Viz locally, dataset previews for all nodes will be enabled by default. However, users will have the ability to enable or disable this feature from the Kedro-Viz settings panel.

## Development Notes

**Backend Changes:**

- Introduced a new API endpoint `/api/preferences` to handle user preferences updates.
- Created a `UserPreference` model to define the preference schema.
- Implemented a new function in `router.py` to update the `showDatasetPreviews` preference and set the class variable accordingly.
- Adjusted the logic in `responses.py` to use the new preference for determining dataset preview settings.

**Frontend Changes:**

- Added a new `showDatasetPreviews` toggle in the Redux state to manage the setting.
- Updated the `SettingsModal` component to include the `showDatasetPreviews` toggle.

**How to Test:**

- Run kedro viz locally via
- Open the Kedro-Viz settings panel.
- Toggle the Dataset Previews setting on and off.
- Verify that the dataset previews for all nodes are enabled/disabled based on the setting.
- Check the network tab to confirm that Preferences API is called and that the preferences are correctly updated on the backend. (Might need to toggle preserve logs to view it as saving in the settings menu refreshes the page and clears network logs)

New dataset preview toggle:

image

New API endpoint:

image

## Checklist


Huongg:
hey @SajidAlamQB what's the best way to test your branch locally? I run `kedro viz run` in demo-project but I guess it won't reflect any changes from the FE

SajidAlamQB:
> hey @SajidAlamQB what's the best way to test your branch locally? I run `kedro viz run` in demo-project but I guess it won't reflect any changes from the FE

Hi, I'm making some changes at the moment, but I've been using `make build` and `make run` to test the changes locally.

Huongg:
> > hey @SajidAlamQB what's the best way to test your branch locally? I run `kedro viz run` in demo-project but I guess it won't reflect any changes from the FE
> 
> Hi, I'm making some changes at the moment, but I've been using `make build` and `make run` to test the changes locally.

oh i see, maybe i got it wrong but from the description it says when you run "kedro viz run". So i expect if i run "kedro viz run" the behaviour will be different from if i run "kedro viz build" etc. Is it the case?

SajidAlamQB:
> oh i see, maybe i got it wrong but from the description it says when you run "kedro viz run". So i expect if i run "kedro viz run" the behaviour will be different from if i run "kedro viz build" etc. Is it the case?

It should be the same behaviour, I've made quite a few changes and updated the description might be worth testing again.

src/actions/index.js
```diff
@@ -141,6 +141,51 @@ export function toggleShowFeatureHints(showFeatureHints) {
   };
 }
 
+export const UPDATE_PREFERENCES = 'UPDATE_PREFERENCES';
```
rashidakanchwala, line 144: I wouldn't add this as a redux action. I would just have a useEffect in settings.modal that would track changes to `setShowDataSetPreview` and make the below call.

src/reducers/index.js
```diff
@@ -81,6 +82,11 @@ const combinedReducer = combineReducers({
     TOGGLE_SHOW_FEATURE_HINTS,
     'showFeatureHints'
   ),
+  showDatasetPreviews: createReducer(
```
Huongg, line 85: hey if we can have a test here to cover this change, would be great

src/components/settings-modal/settings-modal.js
```diff
@@ -80,19 +90,49 @@ const SettingsModal = ({
...
+        if (!response.ok) {
+          throw new Error('Failed to update preferences');
+        }
+
+        const result = await response.json();
+        console.log('Preferences updated successfully:', result);
+      } catch (error) {
+        console.error('Error updating preferences:', error);
+      }
+    };
```
jitu5, line 124: Please move this api call to separate file.

src/components/settings-modal/settings-modal.js
```diff
@@ -80,19 +90,49 @@ const SettingsModal = ({
...
+
+    updatePreferences();
+  }, [showDatasetPreviewsValue]);
```
jitu5, line 127: I see few problems related to api calls
1. On first load every time it make `api/preferences` api call
2. Whenever user clicks on toggle button it make `api/preferences `api call, ideally it should call only when clicked on save button.
3. When user clicks on toggle button and it saves the preferences even after user clicks on cancel button.

src/components/settings-modal/settings-modal.js
```diff
@@ -65,6 +74,7 @@ const SettingsModal = ({
 
         onToggleIsPrettyName(isPrettyNameValue);
         onToggleShowFeatureHints(showFeatureHintsValue);
+        onToggleShowDatasetPreviews(showDatasetPreviewsValue);
```
Huongg, line 77: do you think we can have the  update state `onToggleShowDatasetPreviews(showDatasetPreviewsValue);` and API call together? 

perhaps something here like 

`updateShowDatasetPreviews()`

```
const updateShowDatasetPreviews = () => {
    // to update redux state
    onToggleShowFeatureHints(showFeatureHintsValue);

   // to update BE
   const updatePreferences = async () => {
      try {
        const response = await fetch('/api/preferences', {
          method: 'POST',
          headers: {
            'Content-Type': 'application/json',
          },
          body: JSON.stringify({
            showDatasetPreviews: showDatasetPreviewsValue,
          }),
        });

        if (!response.ok) {
          throw new Error('Failed to update preferences');
        }

        const result = await response.json();
        console.log('Preferences updated successfully:', result);
      } catch (error) {
        console.error('Error updating preferences:', error);
      }
    };

    updatePreferences();
}
```

It's just easier to know that all the actions related to update dataset preview are in one place

package/kedro_viz/api/rest/router.py
```diff
@@ -49,6 +50,18 @@ async def get_single_node_metadata(node_id: str):
     return get_node_metadata_response(node_id)
 
 
+@router.post("/preferences")
```
jitu5, line 53: Any reason why we created a new api endpoint for setting preferences instead of using existing api/nodes/ by passing user preferences set in redux state to api/nodes/

Huongg (commented):
thank you @SajidAlamQB. I've left a few comments. Let me know if you have any questions about them

stephkaiser (approved):

src/components/settings-modal/settings-modal.js
SajidAlamQB: Moved it to `src/utils/preferences-api.js`.

src/components/settings-modal/settings-modal.js
SajidAlamQB: The API call moved to a separate file in `src/utils/preferences-api.js`.

src/components/settings-modal/settings-modal.js
SajidAlamQB: I moved the API call to update preferences inside the `useEffect` that is triggered when the "Save and apply" button is clicked.

package/kedro_viz/api/rest/router.py
```diff
@@ -49,6 +50,18 @@ async def get_single_node_metadata(node_id: str):
...
+async def update_preferences(preferences: UserPreference):
+    try:
+        DataNodeMetadata.set_is_all_previews_enabled(preferences.showDatasetPreviews)
+        return {"message": "Preferences updated successfully"}
+    except Exception as exception:
+        logger.error("Failed to update preferences: %s", str(exception))
+        raise HTTPException(
```
ravi-kumar-pilla, line 60: since we are handling the exception, it would be nice to return a JSONResponse as below -
return JSONResponse(status_code=500, content={"message": f"{exc}"}) if we want the users to see the err trace message on the console. If not we can have a generic message and just log the exception.

src/components/settings-modal/settings-modal.js
```diff
@@ -40,12 +44,31 @@ const SettingsModal = ({
...
     setShowFeatureHintsValue(showFeatureHints);
   }, [showFeatureHints]);
 
+  useEffect(() => {
+    setShowDatasetPreviewsValue(showDatasetPreviews);
+  }, [showDatasetPreviews]);
+
+  useEffect(() => {
+    setToggleFlags(flags);
+  }, [flags]);
```
ravi-kumar-pilla, line 61: I see this effect was missing earlier. Why is this change needed ?

package/kedro_viz/api/rest/router.py
```diff
@@ -49,6 +50,18 @@ async def get_single_node_metadata(node_id: str):
     return get_node_metadata_response(node_id)
 
 
+@router.post("/preferences")
+async def update_preferences(preferences: UserPreference):
+    try:
+        DataNodeMetadata.set_is_all_previews_enabled(preferences.showDatasetPreviews)
+        return {"message": "Preferences updated successfully"}
```
ravi-kumar-pilla, line 57: You need to send the success message as - `return JSONResponse(status_code=200, content={"message": "Preferences updated successfully"})`. If not this will be printed as a message in the browser console.

src/store/store.js
```diff
@@ -57,6 +57,7 @@ const saveStateToLocalStorage = (state) => {
     theme: state.theme,
     isPrettyName: state.isPrettyName,
     showFeatureHints: state.showFeatureHints,
+    showDatasetPreviews: state.showDatasetPreviews,
```
ravi-kumar-pilla, line 60: Hi @SajidAlamQB, 

There is an issue setting `showDatasetPreviews` this way. For other flags like `showFeatureHints` we are not sending a request to backend for updating the toggleStatus. So either having the default or reading from localStorage works. 

But for this specific flag, we need to get the initial status from the backend. This would require you to create another api `/preferences` which will be a GET request. This way the FE and backend will be in sync even when the browser is changed or the localStorage is deleted. Let me know if this is not clear. Thank you

SajidAlamQB: Could you elaborate on this a bit more, don't we assume that `showDatasetPreviews` is enabled by default?

ravi-kumar-pilla: Let's connect and discuss

src/config.js
```diff
@@ -77,8 +78,23 @@ export const settings = {
...
 
+// Fetch preferences from the backend and update settings
+(async () => {
+  try {
+    const preferences = await fetchPreferences();
+    settings.showDatasetPreviews.default = preferences.showDatasetPreviews;
+  } catch (error) {
+    console.error('Error fetching initial preferences:', error);
+  }
+})();
```
jitu5, line 96: @SajidAlamQB I think this might not be the right place to make call, as its async, it may or may not update Redux state, I think you should explore redux-thunk.

ravi-kumar-pilla: @SajidAlamQB, As I was thinking the api call can be delayed and it can be called when someone clicks on the settings icon (before we open the settings modal). That way you can dispatch an action when someone clicks on settings icon -> get the preferences json -> update the toggle accordingly -> open the modal popup which has the updated state. The other way is to call the api along with /main on first load, similar to /api/package-compatibilities. 

This file can be left with the default as before. We will only update redux store.

src/components/settings-modal/settings-modal.js
```diff
@@ -40,12 +46,33 @@ const SettingsModal = ({
...
+  useEffect(() => {
+    if (visible.settingsModal) {
+      getPreferences();
+    }
+  }, [visible.settingsModal, getPreferences]);
+
+  const handleSavePreferences = useCallback(async () => {
+    try {
+      const result = await updatePreferences(showDatasetPreviewsValue);
+      console.log('Preferences updated successfully:', result);
```
jitu5, line 70: Please remove success console.

src/actions/index.js
```diff
@@ -141,6 +141,19 @@ export function toggleShowFeatureHints(showFeatureHints) {
   };
 }
 
+export const TOGGLE_SHOW_DATASET_PREVIEWS = 'TOGGLE_SHOW_DATASET_PREVIEWS';
```
ravi-kumar-pilla, line 144: can we have all the actions related to preferences in 1 file like - `src/actions/preferences.js` as you added preferencesLoaded ?

src/components/settings-modal/settings-modal.js
```diff
@@ -40,12 +46,32 @@ const SettingsModal = ({
...
   useEffect(() => {
     setShowFeatureHintsValue(showFeatureHints);
   }, [showFeatureHints]);
 
+  useEffect(() => {
+    setShowDatasetPreviewsValue(showDatasetPreviews);
+  }, [showDatasetPreviews]);
+
+  useEffect(() => {
+    if (visible.settingsModal) {
```
ravi-kumar-pilla, line 62: is this check required ? Also, I wonder how the toggle changes if the api takes time to return. I felt we can get the state before settingsModal is open. Based on the code, it seems the state is fetched after the modal is visible

src/components/settings-modal/settings-modal.js
```diff
@@ -209,13 +251,17 @@ export const mapStateToProps = (state) => ({
   flags: state.flags,
   showFeatureHints: state.showFeatureHints,
   isPrettyName: state.isPrettyName,
+  showDatasetPreviews: state.preferences.showDatasetPreviews,
```
jitu5, line 254: It should be like this right?
`showDatasetPreviews: state.showDatasetPreviews,` 
or I am missing something ?

src/components/settings-modal/settings-modal.js
jitu5: @ravi-kumar-pilla Yes `if (visible.settingsModal)` this check will make sure that we get preferences when we open it and not when we close. Current setup is fine as `visible.settingsModal` get true it make api call, and therefore `showDatasetPreviews` changes and this triggers `setShowDatasetPreviewsValue` to update. 

```
  useEffect(() => {
    setShowDatasetPreviewsValue(showDatasetPreviews);
  }, [showDatasetPreviews]);
```

```
  useEffect(() => {
    if (visible.settingsModal) {
      getPreferences();
    }
  }, [visible.settingsModal, getPreferences]);
```

src/components/settings-modal/settings-modal.js
ravi-kumar-pilla: I tested locally and surprisingly it works 🗡️  .... not sure if the prop is used to set the local state of the toggle

SajidAlamQB: This just keeps `showDatasetPreviews` nested within the preferences state its just to keep it more distinct incase in the future we want to add more preferences. We can also go with the original global state version as well.

src/store/initial-state.js
```diff
@@ -24,6 +24,7 @@ export const createInitialState = () => ({
   expandAllPipelines: false,
   isPrettyName: settings.isPrettyName.default,
   showFeatureHints: settings.showFeatureHints.default,
+  showDatasetPreviews: settings.showDatasetPreviews.default,
```
ravi-kumar-pilla, line 27: It is good to have a nesting, but how are you able to nest if the initial state is at the same level as flags ?
https://github.com/kedro-org/kedro-viz/pull/1977/files#r1688315574

src/components/settings-modal/settings-modal.js
ravi-kumar-pilla: it is good to have a nesting, but I guess the question is how showDatasetPreviews is under preferences when the initial state does not have a nesting

src/store/initial-state.js
SajidAlamQB: Fair you are right, we can go back to original.

src/reducers/index.js
```diff
@@ -96,6 +100,16 @@ const combinedReducer = combineReducers({
     TOGGLE_SHOW_FEATURE_HINTS,
     'showFeatureHints'
   ),
+  showDatasetPreviews: (state = true, action) => {
```
ravi-kumar-pilla, line 103: can we have the block inside its own reducer as before ? This is more of a reducer block and best kept separate, instead of keeping it inside combineReducers to be consistent. Let me know if this was needed or if I am missing something here.

src/actions/preferences.js
```diff
@@ -0,0 +1,25 @@
...
+// Action Creators
+export const preferencesLoaded = (preferences) => ({
+  type: PREFERENCES_LOADED,
+  payload: preferences,
+});
+
+export const toggleShowDatasetPreviews = (showDatasetPreviews) => ({
+  type: TOGGLE_SHOW_DATASET_PREVIEWS,
+  showDatasetPreviews,
+});
```
jitu5, line 16: @SajidAlamQB 
Can we combine `preferencesLoaded` and `toggleShowDatasetPreviews` to have single action like `updateDatasetPreviews`/ `UPDATE_DATASETS_PREVIEWS` as both are updating `showDatasetPreviews` in Redux state.

jitu5 (approved):
LGTM! Thanks @SajidAlamQB

src/actions/preferences.js
```diff
@@ -0,0 +1,19 @@
...
+// Action Creators
+export const updateDatasetPreviews = (showDatasetPreviews) => ({
+  type: UPDATE_DATASET_PREVIEWS,
+  payload: { showDatasetPreviews },
+});
+
+export const getPreferences = () => async (dispatch) => {
+  try {
+    const preferences = await fetchPreferences();
+    dispatch(updateDatasetPreviews(preferences.showDatasetPreviews));
```
ravi-kumar-pilla, line 15:
```suggestion
    dispatch(updateUserPreferences(preferences));
```

src/reducers/index.js
```diff
@@ -81,6 +82,7 @@ const combinedReducer = combineReducers({
   modularPipeline,
   visible,
   runsMetadata,
+  showDatasetPreviews,
```
ravi-kumar-pilla, line 85:
```suggestion
  userPreferences
```

src/reducers/preferences.js
```diff
@@ -0,0 +1,14 @@
+import { UPDATE_DATASET_PREVIEWS } from '../actions/preferences';
+
+const initialState = true;
+
+const showDatasetPreviews = (state = initialState, action) => {
+  switch (action.type) {
+    case UPDATE_DATASET_PREVIEWS:
```
ravi-kumar-pilla, line 7:
```suggestion
    case UPDATE_USER_PREFERENCES:
```

src/actions/preferences.js
```diff
@@ -0,0 +1,19 @@
+import { fetchPreferences } from '../utils/preferences-api';
+
+// Action Types
+export const UPDATE_USER_PREFERENCES = 'UPDATE_USER_PREFERENCES';
+
+// Action Creators
+export const updateUserPreferences = (preferences) => ({
+  type: UPDATE_USER_PREFERENCES,
+  payload: { preferences },
```
ravi-kumar-pilla, line 9:
```suggestion
  payload: preferences,
```

ravi-kumar-pilla: preferences should already be an object. Do we need to wrap it around { } ?

src/components/settings-modal/settings-modal.test.js
```diff
@@ -73,5 +74,11 @@ describe('SettingsModal', () => {
       type: 'TOGGLE_IS_PRETTY_NAME',
       isPrettyName: false,
     });
+
+    mapDispatchToProps(dispatch).onToggleShowDatasetPreviews(false);
+    expect(dispatch.mock.calls[3][0]).toEqual({
+      type: 'UPDATE_USER_PREFERENCES',
+      payload: { preferences: { showDatasetPreviews: false } },
```
ravi-kumar-pilla, line 81: UPDATE_USER_PREFERENCES should take an object preferences. The key preferences is not needed. The object is a preferences object as a whole

```
{ 
  showDatasetPreviews: false 
}
```

src/reducers/preferences.js
```diff
@@ -0,0 +1,19 @@
...
+const initialState = {
+  showDatasetPreviews: true,
+};
+
+const userPreferences = (state = initialState, action) => {
+  switch (action.type) {
+    case UPDATE_USER_PREFERENCES:
+      return {
+        ...state,
+        showDatasetPreviews: action.payload.preferences.showDatasetPreviews,
```
ravi-kumar-pilla, line 12:
```suggestion
        showDatasetPreviews: action.payload.showDatasetPreviews,
```

ravi-kumar-pilla: The payload for userPreferences should be a preferences object. i.e., 
```
{
     showDatasetPreviews: false
}
```
It need not contain the preferences key inside like -
```
{
     preferences: {
           showDatasetPreviews: false
      }
}
```
I think there is some confusion around nesting. If we have payload as - `payload = showDatasetPreviews` (this means no nesting and cannot be extended in future)

`payload = {showDatasetPreviews: false}` , there is a nested prop and the payload becomes an object. Here, the object will have fields of User Preferences.

src/store/initial-state.js
```diff
@@ -24,6 +24,9 @@ export const createInitialState = () => ({
   expandAllPipelines: false,
   isPrettyName: settings.isPrettyName.default,
   showFeatureHints: settings.showFeatureHints.default,
+  userPreferences: {
```
ravi-kumar-pilla, line 27: This is perfect 💯

ravi-kumar-pilla (approved):
Thank you for addressing all the PR comments @SajidAlamQB . LGTM and works well 💯

PR #1766, opened 2024-02-26
Fix packaging

astrojuanlu:
## Description

Make sdists self-contained, hence fix #1267

## Development notes

Commits:

1. cf738296 Remove direct invokations of `setup.py`, which have been deprecated for about 2 years https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
2. 44538613 Migrate to `pyproject.toml` using @MehdiNV work in #1584 as baseline with some updates, hence fix #1527
3. 495a54b1 Migrate from `setuptools` to `hatchling` https://hatch.pypa.io/, same backend vizro uses (cc @antonymilne), as a preparation to eventually address #1611 and also fix #1267
4. e22a6ba0 Move `README.md` to `package/` and create root symlink to retain GitHub preview

## QA notes

1. Run `make package` in `main` and with this PR, save the resulting sdist and wheel in `dist.old` and `dist` respectively
2. Unzip the resulting `.tar.gz`, and compare the `PKG-INFO` files of both sdists - there should be only lines out of order and other non consequential changes directories with `main` and with this PR.
3. Compare the `tree` structure of both sdists - only metadata files should have changed.

```diff
--- old.tree	2024-02-27 00:41:59
+++ new.tree	2024-02-27 00:42:29
@@ -1,5 +1,6 @@
-dist.old/kedro-viz-7.1.0/
+dist/kedro_viz-7.1.0/
 ├── PKG-INFO
+├── README.md
 ├── kedro_viz
 │   ├── __init__.py
 │   ├── api
@@ -81,10 +82,7 @@
 │       ├── __init__.py
 │       ├── layers.py
 │       └── modular_pipelines.py
-├── setup.cfg
-├── setup.py
-└── tests
-    ├── test_import.py
-    └── test_server.py
+├── pyproject.toml
+└── requirements.txt
 
-19 directories, 69 files
+18 directories, 68 files
```

4. Unzip both wheel files and compare the `*.dist-info` directories - there should be only lines out of order but the number of files in `RECORD` should be basically the same.

```diff
...
diff --color=auto -u dist.old/kedro_viz-7.1.0.dist-info/RECORD dist/kedro_viz-7.1.0.dist-info/RECORD
--- dist.old/kedro_viz-7.1.0.dist-info/RECORD   2024-02-26 23:37:50
+++ dist/kedro_viz-7.1.0.dist-info/RECORD       2020-02-02 00:00:00
@@ -62,8 +62,7 @@
[...]

PR #1978, opened 2024-07-11
Adjust requirements file and dependabot versioning strategy

### LINKED ISSUE #1967: Relax dependabot strategy
## Description
In https://github.com/kedro-org/kedro-viz/issues/1582#issuecomment-1764451560 I wrote

> Comment from the peanut gallery: since Kedro-Viz dependabot is using the default configuration, `versioning-strategy` is set to `auto`, meaning
> 
> > Try to differentiate between apps and libraries. Use `increase` for apps and `widen` for libraries.
> 
> https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy
> 
> and since Kedro-Viz uses `requirements.txt` for its Python requirements, it's probably applying the `increase` strategy, which is arguably too aggressive for a component like Kedro-Viz. In fact, since Kedro switched to `pyproject.toml`, dependabot PRs have been using the `widen` strategy.
> 
> Viz can either
> 
>     * Do nothing and keep aggressively upgrading dependencies, but this can lead to problems with people that have environments with lots of dependencies.
> 
>     * Configure dependabot to use `widen`.
> 
>     * Migrate dependencies to `pyproject.toml` [Move to pyproject.toml #1527](https://github.com/kedro-org/kedro-viz/issues/1527) and keep the `auto` config.

Since the migration to `pyproject.toml` hasn't happened, Kedro-Viz continued to use the `increase` dependabot strategy, hence making too aggressive updates to `requirements.txt` files.

## Context
This has caused trouble to me. Today I was trying to install Kedro-Viz alongside Prefect 3.0.0rc10, and got this error:

```
$ uv pip compile --universal requirements.in -o requirements.txt
  × No solution found when resolving dependencies:
  ╰─▶ Because kedro-viz==9.1.0 depends on uvicorn[standard]>=0.29.0,<0.30.dev0 and only the following versions
      of uvicorn[standard] are available:
          uvicorn[standard]<=0.29.0
          uvicorn[standard]>=0.30.dev0
      we can conclude that kedro-viz==9.1.0 depends on uvicorn[standard]==0.29.0.
      And because prefect==3.0.0rc10 depends on one of:
          uvicorn>=0.14.0,<0.29.0
          uvicorn>0.29.0
      and only the following versions of prefect are available:
          prefect<=3.0.0rc10
          prefect>=3.1.dev0
      we can conclude that kedro-viz==9.1.0 and prefect>=3.0.0rc10,<3.1.dev0 are incompatible.
      And because you require kedro-viz==9.1.0 and prefect>=3.0.0rc10, we can conclude that the requirements
      are unsatisfiable.
```

Kedro-Viz has such strict dependency pinnings because of Dependabot https://github.com/kedro-org/kedro-viz/pull/1839

The dependency was bumped in the `main` branch already https://github.com/kedro-org/kedro-viz/pull/1935 but since development installs are tricky https://github.com/kedro-org/kedro-viz/issues/1611 I cannot just run `uv pip install "kedro-viz @ git+https://github.com/kedro-org/kedro-viz.git@main#subpackage=package"`, because that yields a broken installation:

```
$ uv pip install "kedro-viz @ git+https://github.com/kedro-org/kedro-viz.git@main#subdirectory=package"
 Updated https://github.com/kedro-org/kedro-viz.git (35d351f)
Resolved 103 packages in 4.48s
   Built kedro-viz @ git+https://github.com/kedro-org/kedro-viz.git@35d351f03172032876a9f6571f9b9c31145d7417#subdirec
Prepared 1 package in 1.49s
Uninstalled 2 packages in 29ms
Installed 2 packages in 16ms
 - kedro-viz==9.1.0
 + kedro-viz==9.1.0 (from git+https://github.com/kedro-org/kedro-viz.git@35d351f03172032876a9f6571f9b9c31145d7417#subdirectory=package)
 - uvicorn==0.29.0
 + uvicorn==0.30.1
$ kedro viz run
...
    FileNotFoundError: [Errno 2] No such file or directory:                             
                             '/Users/juan_cano/Projects/QuantumBlackLabs/workshop-from-zero                      
                             -to-mlops/.venv/lib/python3.11/site-packages/kedro_viz/html/in                      
                             dex.html' 
```

There are two solutions:

- Review & merge https://github.com/kedro-org/kedro-viz/pull/1766 (will require some updates because it's been a while) so that Dependabot adjusts automatically
- or, alternatively, configure Dependabot to change the strategy, while keeping `requirements.txt` for now

ravi-kumar-pilla:
## Description

Resolves #1967 

## Development notes

- Following the doc mentioned [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) , the dependabot versioning-strategy `widen` would include both the new and old versions when possible, allowing users the flexiblility to install Kedro-Viz. Unfortunately, the `pip` ecosystem does not support `widen`. We will use `auto` (which is widen for libraries)
- Relaxed the allowed versions in `requirements.txt` file to have a range with upper-bound corresponding to a major release of the package (which corresponds to a non-backward compatible change, hence needs more attention when upgrading).

## QA notes

- No change in the app behavior
- All tests should pass
- Try installing Kedro-Viz and prefect>=3.0.0rc10. This should not raise any issues as mentioned in #1967 . Note that there is an issue with watchgod not having support for anyio >= 4 and prefect >=3.0.0rc10 requires anyio >= 4. There is an open [discussion](https://github.com/samuelcolvin/watchfiles/issues/254) in watchgod.

## Checklist


astrojuanlu:
This might be bad docs though:

> The service currently does not accept that users configure the `widen` strategy in their configuration file.
> 
> However, we do have `widen` support in Python, since it's the default strategy used for libraries.

https://github.com/dependabot/dependabot-core/issues/6630

So `auto` + `pyproject.toml` = `widen`.

We might see progress on this soon though? https://github.com/dependabot/dependabot-core/issues/6630#event-13336645619

The alternative is to merge https://github.com/kedro-org/kedro-viz/pull/1766 and then migrate the dependencies to `pyproject.toml`. But probably that has implications for CI etc.

astrojuanlu:
Sounds good!

rashidakanchwala (approved):

astrojuanlu (approved):

Issue #1973, opened 2024-07-08
As a Kedro user, I want to be able to select multiple nodes within the flowchart so that I can slice the pipeline I am currently viewing.


Huongg:
![image](https://github.com/kedro-org/kedro-viz/assets/32060364/c904eb4b-db27-4537-ab57-17862579f251)

### Acceptance criteria:

- [ ] **Notification on First Node Click**: When I click on the first node, a notification should appear. This behavior should not occur when clicking on a pipeline (is it referred to as "pipeline" or "modular pipeline"?).

- [ ] **Second Node Click with Shift**: If I hold the Shift key and click on a second node, the metadata should close immediately, and the notification should disappear.

- [ ] when selecting a node from bottom to top, the order should be updated according to `--from-node=bottom-node-id --to-node=top-node-id`

Figma: https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=1469-56945&t=gAB4IcCLRWoEGtZQ-0

PR #1980, opened 2024-07-11
Feature/slicing pipeline

### LINKED ISSUE #1974: As a Kedro user, I want to preview which nodes are included in my selection for slicing so that I can understand what my slice will be before creating a new slice.
image

### Acceptance criteria:

- [ ] Chart Highlighting: The two nodes selected with Shift held should be highlighted with a darker blue background, and any nodes in between should be highlighted with a lighter blue background.
- [ ] Nodes List Highlighting: The nodes list on the left-hand side should also highlight all the nodes included in the newly selected pipeline.
- [ ] Call to Action: A button at the bottom should display the number of nodes selected and include a "Slice" button.
- [ ] Slice Button Functionality: Clicking the "Slice" button should navigate to a new view displaying only the selected nodes.
- [ ] Outside Click: 

Clicking outside the selection will un-select all the nodes, and go back to the default/previous page

Huongg:
## Description
Fixes #1974, 

**User Story**: As a Kedro user, I want to preview which nodes are included in my selection for slicing so that I can understand what my slice will be before creating a new slice

## Development notes

This is a feature branch, and won't merge to main until all the other functionalities are merged and ready. 

**Architecture diagram as below:**

Ideation session_ pipeline slicing and filtering - Data flow between Flowchart and draw and node-list

## QA notes

**When testing on your machine locally, or through gitpod, please note that this PR only covers the functionalities below:**

- [x] Chart Highlighting: The two nodes selected with Shift held should be highlighted with a darker blue background, and any nodes in between should be highlighted with a lighter blue background.
- [x]  Nodes List Highlighting: The nodes list on the left-hand side should also highlight all the nodes included in the newly selected pipeline.
- [x]  Call to Action: A button at the bottom should display the number of nodes selected and include a "Slice" button.
- [x]  Outside Click: Clicking outside the selection will un-select all the nodes, and go back to the default/previous page. This includes clicking on any other nodes outside the sliced pipeline

**Pending design tasks that are waiting for @stephkaiser . Please ignore them when you review the PR**

- [x] Node List Highlighting: the issue happens when the node list is currently highlighting, and you try to hover on other node. The hover style will cover the highlighting style
Screenshot 2024-07-16 at 10 38 25

- [ ] The slice button icon

## Checklist

### DIFF
--- a/src/components/node-list/styles/_row.scss
+++ b/src/components/node-list/styles/_row.scss
@@ -13,19 +13,6 @@
   transform: translate(0, 0);
   cursor: default;
 
-  // overwrite material-ui selected row background color settings
-  &--overwrite {
-    .Mui-selected & {
-      .kui-theme--dark & {
-        background-color: #{colors.$slate-200};
-      }
-
-      .kui-theme--light & {
-        background-color: #{colors.$white-0};
-      }
-    }
-  }
-
   // overwrite with highlighted background for row on hover
   &--overwrite:hover {
     .Mui-selected & {


src/components/flowchart/components/slice-pipeline-action/slice-pipeline-action.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Button from '../../../ui/button';
+
+import './slice-pipeline-action.scss';
+
+export const SlicePipelineAction = ({ chartSize, filteredPipeline }) => {
```
rashidakanchwala, line 6: The name ix a bit confusing because it made me think it was an action rather than a component. Perhaps we could rename it to something like PipelineSliceIndicator ?

Huongg: hey i agree the name is a bit confusing, especially later on it shows the run command too. How about `FilteredPipelineActionBar` ? It's following what Steps used in Figma, and it's also consistent with selectors `FilteredPipeline`

src/components/flowchart/draw.js
```diff
@@ -136,6 +150,16 @@ export const drawNodes = function (changed) {
     focusMode,
     hoveredFocusMode,
   } = this.props;
+  const { filteredPipelineState } = this.state;
+
+  const fromToFilteredPipeline = createNodeStateMap(nodes, [
+    filteredPipelineState.from,
+    filteredPipelineState.to,
+  ]);
+  const nodesFromFilteredPipeline = createNodeStateMap(
```
rashidakanchwala, line 159: nit 

```suggestion
  const rangeFilteredPipeline = createNodeStateMap(
```

src/components/flowchart/draw.js
```diff
@@ -136,6 +150,16 @@ export const drawNodes = function (changed) {
     focusMode,
     hoveredFocusMode,
   } = this.props;
+  const { filteredPipelineState } = this.state;
+
+  const fromToFilteredPipeline = createNodeStateMap(nodes, [
```
rashidakanchwala, line 155: these names are a bit confusing. I cannot come up with better names but below is the best. 
I am also ok if we stick to these names  but lets leave a comment on top to define what they do

```
// Map the state of nodes selected by the user (from and to nodes)
const selectedNodesState = createNodeStateMap(nodes, [
  filteredPipelineState.from,
  filteredPipelineState.to,
]);

// Map the state of nodes that fall within the range between the selected nodes
const nodesInRangeState = createNodeStateMap(
  nodes,
  filteredPipelineState.range
);
```

src/components/flowchart/draw.js
```diff
@@ -24,6 +24,20 @@ const toSinglePoint = (value) => parseFloat(value).toFixed(1);
  */
 const limitPrecision = (path) => path.replace(matchFloats, toSinglePoint);
 
+/**
+ * Creates a mapping of node IDs to a boolean indicating if the node ID is included in the given values.
+ * @param {Array} nodes - Array of nodes to process.
+ * @param {Array} values - Array of values to check against node IDs.
+ * @returns {Object} An object mapping node IDs to booleans.
+ */
+function createNodeStateMap(nodes, values) {
```
rashidakanchwala, line 33: nice!

src/components/flowchart/flowchart.js
```diff
@@ -90,6 +102,34 @@ export class FlowChart extends Component {
 
   componentDidUpdate(prevProps) {
     this.update(prevProps);
+
+    const isFilteredPipelineChanged =
```
rashidakanchwala, line 106: can we add this code logic to update(prevProps = {}) in line 138?

src/components/flowchart/flowchart.js
```diff
@@ -451,23 +491,66 @@ export class FlowChart extends Component {
...
+    toSelectedNode(node);
+
+    const { from, to } = this.state.filteredPipelineState;
+    // Reset or set the first node as the 'from' node based on current state
+    const newState =
+      from !== null && to !== null
+        ? { ...this.state.filteredPipelineState, from: null, to: null }
+        : { ...this.state.filteredPipelineState, from: node.id };
+
+    this.setState({ filteredPipelineState: newState });
```
rashidakanchwala, line 529: this code is being used multiple declare a newState, and then update the filteredState ... maybe we can have a function `updateFilteredPipelineState` that can reduce the redudancy

Huongg: yup good idea, I've created one now. Thank you Rashida 💯

src/components/flowchart/draw.js
Huongg: let me have a think, I'll come back to you on this

src/store/normalize-data.js
```diff
@@ -79,6 +79,11 @@ export const createInitialPipelineState = () => ({
     active: {},
     enabled: {},
   },
+  filters: {
```
ravi-kumar-pilla, line 82: we have a filters section in the UI,[Nit] When I saw this, it reminded me of those filters which we have above. If possible we can rename this to pipelineSlice or slice ?

src/selectors/filtered-pipeline.js
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Selector to filter nodes that are connected between two specified node IDs.
+ * @param {Object} edgesByNode - Edges organized by node IDs.
+ * @param {string} startID - Starting node ID.
+ * @param {string} endID - Ending node ID.
+ * @returns {Array} Array of node IDs that are connected from startID to endID.
+ */
+
+export const getFilteredPipeline = createSelector(
```
ravi-kumar-pilla, line 111:
```suggestion
export const getSlicedPipeline = createSelector(
```

ravi-kumar-pilla: I see we used filter terminology in all the places, if this is something the team feels fine to go ahead, then lets do that. For me this confuses with the nodeType, tags filter

src/store/normalize-data.js
Huongg: hey i think the idea is later on we might merge those filters in the UI together with slicing function. Hence I thought it makes sense to put it here.

src/selectors/filtered-pipeline.js
Huongg: yeah i dont mind using sliced Pipeline. Let see what the team thinks in standup

rashidakanchwala: I also prefer using "slice" because it matches Kedro terminology and we have filters that behave differently. My original PR used "filtered" everywhere, which is probably why we followed that.

Huongg: cool happy to change to sliced Pipeline 👍

Huongg: DONE 👍 @rashidakanchwala @ravi-kumar-pilla

src/components/flowchart/flowchart.js
```diff
@@ -451,32 +495,91 @@ export class FlowChart extends Component {
...
     if (this.props.clickedNode) {
       this.props.onLoadNodeData(null);
       // To reset URL to current active pipeline when click outside of a node on flowchart
       this.props.toSelectedPipeline();
     }
+
+    // Determine if the click event occurred on the slice button.
+    const isSliceButtonClicked =
+      this.slicedPipelineActionRef.current &&
+      this.slicedPipelineActionRef.current.contains(event.target);
```
jitu5, line 574: Any specific reason why we are using `ref` to determine click?  how about `onClick`  on `SlicedPipelineActionBar`?

src/components/flowchart/flowchart.js
```diff
@@ -451,32 +495,91 @@ export class FlowChart extends Component {
...
+
+    // Check if the pipeline is sliced, no slice button is clicked, and no filters are applied
+    if (this.props.slicedPipeline && !isSliceButtonClicked) {
+      this.props.onResetSlicePipeline();
+      this.updateSlicedPipelineState(null, null, []);
+      // To reset URL to current active pipeline when click outside of a node on flowchart
+      this.props.toSelectedPipeline();
```
jitu5, line 581: On click on Slice button nothing is happening. Is it expected just for this PR ?

jitu5:
@Huongg I observed few things when testing locally.
- I can selected to same level nodes, Is it expected in this PR and we will handle this in separate PR?
Screenshot 2024-07-26 at 4 56 51 p m

- Not sure about expected behaviour, If I try to hide the visibility of currently highlighted nodes from node list, it does work so we might need to disable visibility button of highlighted nodes?

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Button from '../../../ui/button';
+
+import './sliced-pipeline-action-bar.scss';
+
+export const SlicedPipelineActionBar = ({ chartSize, slicedPipeline }) => {
+  const { outerWidth, sidebarWidth } = chartSize;
+  const sidebarVisible = sidebarWidth > 140;
```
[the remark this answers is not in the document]
Huongg: ah yeah sure, I prob changed it by accident.

.eslintrc.json
```diff
@@ -4,8 +4,7 @@
     "curly": ["error"],
     "valid-typeof": ["error"],
     "camelcase": "error",
-    "id-length": ["error", { "min": 3, "exceptions": ["_","a","b","d","e","i","j","k","x","y","id","el","pi","PI","up"] }],
-    "no-var": ["error"],
+"id-length": ["error", { "min": 3, "exceptions": ["_","a","b","d","e","i","j","k","x","y","id","el","pi","PI","up","to"] }],    "no-var": ["error"],
```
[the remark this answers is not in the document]
Huongg: I included "to" in the list

src/actions/slice.js
```diff
@@ -0,0 +1,28 @@
...
+    });
+  };
+};
+
+export const RESET_SLICE_PIPELINE = 'RESET_SLICE_PIPELINE';
+
+export const resetSlicePipeline = () => ({
+  type: RESET_SLICE_PIPELINE,
+  slice: { from: null, to: null },
+});
```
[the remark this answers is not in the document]
Huongg: hey i think having a separate action to reset makes it quite clear, as I can imagine if we just use `slicePipeline(null, null)`, as one point would need a comment to explain what that does. But now it's clear by just reading `reset...`.What do you think? @jitu5

src/components/flowchart/flowchart.js
Huongg: yeah, no functionality for the slice button in this PR. it will be in a separate PR

Huongg:
> @Huongg I observed few things when testing locally.
> 
> * I can selected to same level nodes, Is it expected in this PR and we will handle this in separate PR?
> 
> Screenshot 2024-07-26 at 4 56 51 p m
> * Not sure about expected behaviour, If I try to hide the visibility of currently highlighted nodes from node list, it does work so we might need to disable visibility button of highlighted nodes?

Hi @jitu5 this is expected, it's the same behaviour from framework side. I've checked with @noklam and he confirmed it. So when user slices 2 nodes not linked to each other, this is what will be shown on the chart

src/components/flowchart/flowchart.js
Huongg: hey because the way the flowchart setup, any click outside the element will be considered as "click outside" then it will clear the flowchart. We don't want it to reset everything when clicking on the slicing action (eg: copy command line, click slice button". So this ref is for that reason, hope it makes sense

jitu5: @Huongg 

In `flowchart.js`

```
  onSlicePipelineClick = (event) => {
    console.log('Slice pipeline clicked', event);
  };
  
  ...
  
<div ref={this.slicedPipelineActionRef}>
  <SlicedPipelineActionBar
    chartSize={chartSize}
    onSlicePipelineClick={this.onSlicePipelineClick}
    slicedPipeline={slicedPipelineState.range}
  />
</div>
```

In `sliced-pipeline-action-bar.js`

```
export const SlicedPipelineActionBar = ({ chartSize, slicedPipeline, onSlicePipelineClick }) => {

...

<div className="sliced-pipeline-action-bar--cta">
  <Button onClick={e => onSlicePipelineClick(e)}>Slice</Button>
</div>
```

I did the above changes in you branch and I clicked in `Slice` button and this does not make "click outside" flowchart. Did you tried this? or I am missing something?

jitu5: @Huongg Yes I got it, after removing `ref` it clears the chart. but we can fix this by simply adding `event.stopPropagation();` in `onSlicePipelineClick` in `flowchart.js` to stop click event propagation to parent i our case its flowchart.

```
  onSlicePipelineClick = (event) => {
    event.stopPropagation();
    console.log('Slice pipeline clicked', event);
  }; 
```

jitu5: @Huongg 
In the `SlicedPipelineActionBar` component, if you click on Slice button its already handled and other area like anywhere or on '25 selected', nothing will happen as there is no click event attached, and if you add anything to `SlicedPipelineActionBar` like the run command component or copy button you can alway add `stopPropagation ` to click handler.

The only problem with `ref` approach is its not straight forward  and `onClick` is simple and easy to understand. let me know what do you think.

Screenshot 2024-07-29 at 5 13 41 p m

Huongg: hey @jitu5 lets have a chat as maybe it's easier. 
I've attached a screen recording of other area would also need to the `event.stopPropagation` if we don't use the ref approach, especially then `CommandCopier` as there's a lot of things happened there and it doesn't always have the event passing around. 

For me it's easier to handle this in flowchart as it's the parent component, rather than `event.stopPropagation` in other places like ActionBar, CommandCopier, and there might be other places too. 


https://github.com/user-attachments/assets/62a37bf3-8f0d-40c8-8a5e-753cd710fd36

Huongg: I'm not very familiar with the issue but what would be the downside if we use `ref`?

Huongg: @jitu5 and I agree that let's keep the ref for now, and we will revisit it again in the second PR with the run command and everything in the ActionBar

ravi-kumar-pilla:
Hi @Huongg , This looks great. I have few questions, which I am not sure if it is by design -

1. Should the `from` node be always on an upper level compared to `to` node ? i.e., in our demo_project can I select `from` as `split_data` and `to` as `companies` ? 
2. When I select a modular_pipeline using shift it gets expanded but none of the nodes are selected. Is it by design that shift only works with nodes ?

Thank you

src/actions/slice.js
```diff
@@ -0,0 +1,28 @@
...
+    dispatch({
+      type: SLICE_PIPELINE,
+      slice: { from, to },
+    });
+  };
+};
+
+export const RESET_SLICE_PIPELINE = 'RESET_SLICE_PIPELINE';
+
+export const resetSlicePipeline = () => ({
```
ravi-kumar-pilla, line 25: [Nit] - just to have an opposite for applySlice

export const resetSlice = () => ({
  type: RESET_SLICE,
  slice: { from: null, to: null },
});

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
```diff
@@ -0,0 +1,31 @@
...
+          {`${slicedPipeline.length} selected`}
+        </div>
+        <div className="sliced-pipeline-action-bar--cta">
+          <Button>Slice</Button>
+        </div>
+      </div>
+    );
+  } else {
+    return null;
+  }
```
ravi-kumar-pilla, line 30: [Nit suggestion] In general, it would be nice to handle the negative cases first and then handle the positive case. Example -

if (!slicedPipeline || slicedPipeline.length <= 0) {
return null
}
....then the positive case, no need of else

ravi-kumar-pilla: Also, if we are returning null here in-case slicedPipeline.length <= 0. Can we mount or call SlicedPipelineActionBar in the parent only when slicedPipeline.length > 0 ? Is there a reason to pass the prop and then check it in the child component ? Thank you

src/components/node-list/index.js
```diff
@@ -67,6 +68,7 @@ const NodeListProvider = ({
   focusMode,
   disabledModularPipeline,
   inputOutputDataNodes,
+  onResetslicePipeline,
```
ravi-kumar-pilla, line 71:
```suggestion
  onResetSlicePipeline,
```

src/components/node-list/index.js
```diff
@@ -109,6 +111,7 @@ const NodeListProvider = ({
       } else {
         onToggleNodeSelected(item.id);
         toSelectedNode(item);
+        onResetslicePipeline();
```
ravi-kumar-pilla, line 114:
```suggestion
        onResetSlicePipeline();
```

src/components/node-list/index.js
```diff
@@ -363,6 +366,9 @@ export const mapDispatchToProps = (dispatch) => ({
   onToggleFocusMode: (modularPipeline) => {
     dispatch(toggleFocusMode(modularPipeline));
   },
+  onResetslicePipeline: () => {
```
ravi-kumar-pilla, line 369:
```suggestion
  onResetSlicePipeline: () => {
```

src/store/normalize-data.js
```diff
@@ -79,6 +79,11 @@ export const createInitialPipelineState = () => ({
     active: {},
     enabled: {},
   },
+  slice: {
+    from: null,
+    to: null,
+    apply: false,
```
ravi-kumar-pilla, line 85: Does apply indicate whether from and to are not null or is there any other reason for storing this key ? Are we using apply for other reasons in the store ?

src/reducers/slice.js
```diff
@@ -0,0 +1,28 @@
...
+      return updateState({
+        apply: action.apply,
+      });
+    case SLICE_PIPELINE:
+      return updateState({
+        from: action.slice.from,
+        to: action.slice.to,
+      });
+    case RESET_SLICE_PIPELINE:
+      return {};
```
ravi-kumar-pilla, line 22: should this return  updateState({
    from: null,
    to: null,
    apply: false,
    }) ?

src/selectors/disabled.js
```diff
@@ -91,12 +95,19 @@ export const getNodeDisabled = createSelector(
...
+    disabledModularPipeline,
+    slicedPipeline,
+    isSliceApplied
   ) =>
     arrayToObject(nodeIDs, (id) => {
       let isDisabledViaModularPipeline =
         disabledModularPipeline[nodeModularPipelines[id]];
 
+      let isDisabledViaFilters = false;
+      if (slicedPipeline.length > 0 && isSliceApplied) {
```
ravi-kumar-pilla, line 107:
```suggestion
      if (isSliceApplied && slicedPipeline.length > 0) {
```

src/selectors/disabled.js
```diff
@@ -91,12 +95,19 @@ export const getNodeDisabled = createSelector(
...
-    disabledModularPipeline
+    disabledModularPipeline,
+    slicedPipeline,
+    isSliceApplied
   ) =>
     arrayToObject(nodeIDs, (id) => {
       let isDisabledViaModularPipeline =
         disabledModularPipeline[nodeModularPipelines[id]];
 
+      let isDisabledViaFilters = false;
```
ravi-kumar-pilla, line 106:
```suggestion
      let isDisabledViaSlicedPipeline = false;
```

src/selectors/disabled.js
ravi-kumar-pilla: could you please let me know what happens if getNodeDisabled returns True/False and how can a node be disabled via a slice ? Is this used for a sliced view ?

Huongg:
> Hi @Huongg , This looks great. I have few questions, which I am not sure if it is by design -
> 
> 1. Should the `from` node be always on an upper level compared to `to` node ? i.e., in our demo_project can I select `from` as `split_data` and `to` as `companies` ?
> 2. When I select a modular_pipeline using shift it gets expanded but none of the nodes are selected. Is it by design that shift only works with nodes ?
> 
> Thank you

hey @ravi-kumar-pilla,

first point is covered in this user story [1973](https://github.com/kedro-org/kedro-viz/issues/1973), we're going to explore changing the order in the command line and let the user selects the bottom node to top node.

for second point, currently it's only working with node and dataset in the MVP 1 i think, we can certainly explore the modular_pipeline but will need to check with the framework team to see if this is something would work in framework too

src/selectors/disabled.js
Huongg: good spot, thank you

src/selectors/disabled.js
Huongg: hey i dont think `getNodeDisabled` would return true or false, it will return an object with IDs and its value, eg:

```
{
  1e3cc50a :false
  3b199c6b: false
  3fb71518: true
}
```

so if its false, the node/element on the chart will be faded out. So any nodes in slicedPipeline will have the value of true, and the rest will be false. Does it make sense?

src/reducers/slice.js
Huongg: yup i think it should 👍

src/store/normalize-data.js
Huongg: `apply` will help the flowchart know when to actually render just the slicedPipeline or not, it does not affect whether from and to are null or valid

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
Huongg: yeah good point, originally i was thinking the flowchart is already too big so I'd like to keep the logic inside the child component to keep it tidy. But think we can move it to the parent so dont need to pass the props to the child if dont need to

src/components/flowchart/flowchart.js
```diff
@@ -84,12 +97,43 @@ export class FlowChart extends Component {
...
+  }
+
   componentWillUnmount() {
     this.removeGlobalEventListeners();
   }
 
   componentDidUpdate(prevProps) {
     this.update(prevProps);
+
+    const { from, to } = this.state.slicedPipelineState;
```
rashidakanchwala, line 121: could we move this code to update function?

src/components/flowchart/flowchart.js
```diff
@@ -84,12 +97,43 @@ export class FlowChart extends Component {
...
+
+    if (isSlicedPipelineChanged) {
+      // Reset local state to null if the redux state's SlicedPipeline is empty,
+      // but the local state still has 'from' and 'to' values defined.
+      if (isSlicedPipelineEmpty && isSlicedPipelineStateDefined) {
+        this.updateSlicedPipelineState(null, null, []);
+      } else {
+        this.updateSlicedPipelineState(from, to, this.props.slicedPipeline);
+      }
+    }
```
rashidakanchwala, line 136:
```suggestion
    const isSlicedPipelineChanged =
      this.props.slicedPipeline !== prevProps.slicedPipeline;

    if (isSlicedPipelineChanged) {
      if (this.props.slicedPipeline.length === 0) {
        // If the slicedPipeline is now empty, reset the local state
        if (from !== null || to !== null) {
          this.updateSlicedPipelineState(null, null, []);
        }
      } else {
        this.updateSlicedPipelineState(from, to, this.props.slicedPipeline);
      }
    }

    // Hide slicing notification if metadata panel is closed using button
    if (!this.props.clickedNode && prevProps.clickedNode && this.props.slicedPipeline.length === 0) {
      this.updateSlicedPipelineState(null, null, []);
      this.setState({ showSlicingNotification: false });
    }
```

rashidakanchwala (approved):
Look's great @Huongg . I have created #2037  and #2038 as I found a few design inconsistencies when reviweing the PR (not related to slicing but from before) and we can work on it eventually.

ravi-kumar-pilla:
Awesome feature for Viz and looking forward to the release. Thank you @Huongg .... looks great !! 💯

ravi-kumar-pilla (approved):

jitu5 (approved):
Some point we discussed during the review process like click on action bar event and slicing during embedded mode. May be we can improvise on it later but for this PR Great work @Huongg !!

PR #1996, opened 2024-07-23
Feature/slicing pipeline run command

### LINKED ISSUE #1975: As a Kedro user, I want to view my sliced pipeline and its corresponding run command so that I can run this slice in the CLI for my Kedro project. I also want to be able to reset my sliced pipeline view so that I can go back to viewing the full Kedro pipeline.
Screenshot 2024-07-31 at 14 16 11

Screenshot 2024-07-31 at 14 16 18

### Acceptance criteria:

- [ ] Filtered Nodes View: A new view should display all the filtered nodes.
- [ ] Updated Node List: The node list should reflect all the newly filtered nodes.
- [ ] CLI Command Display: A command line at the bottom should show how to run the filtered nodes in the CLI.
- [ ] Close the sliced view: Click the exit button should take the user back to viewing the full kedro pipeline

Figma: https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=1469-56945&t=gAB4IcCLRWoEGtZQ-0

Huongg:
## Description
Fixes #1975 
User story: As a Kedro user, I want to view my sliced pipeline and its corresponding run command so that I can run this slice in the CLI for my Kedro project. I also want to be able to reset my sliced pipeline view so that I can go back to viewing the full Kedro pipeline

## Development:

https://github.com/user-attachments/assets/96b0f6b1-0375-4fa9-985f-2c279550f6e5

## QA notes:
When testing on your machine locally, or through gitpod, please note that this PR only covers the functionalities below:

- [x]  Filtered Nodes View: A new view should display all the filtered nodes.
- [x]  Updated Node List: The node list should reflect all the newly filtered nodes.
- [x]  CLI Command Display: A command line at the bottom should show how to run the filtered nodes in the CLI.
- [x]  Close the sliced view: Click the exit button should take the user back to viewing the full kedro pipeline

Figma: https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=1469-56945&t=gAB4IcCLRWoEGtZQ-0

### DIFF
--- a/src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
+++ b/src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
@@ -40,15 +40,15 @@ export const SlicedPipelineActionBar = React.forwardRef((props, ref) => {
     visibleSidebar,
   } = props;
   const { outerWidth: screenWidth } = chartSize;
-  const buffer = 200;
+  const transitionMargin = 200;
   const slicePipelineActionBarWidth =
     ref.current && ref.current.firstChild.getBoundingClientRect().width;
   const metaDataPanelWidth = displayMetadataPanel
-    ? metaSidebarWidth.open + buffer
+    ? metaSidebarWidth.open + transitionMargin
     : metaSidebarWidth.open;
   const nodeListWidth = visibleSidebar
-    ? sidebarWidth.open - buffer
-    : sidebarWidth.open - buffer / 2;
+    ? sidebarWidth.open - transitionMargin
+    : sidebarWidth.open - transitionMargin / 2;
   const minimumTransformX = visibleSidebar
     ? sidebarWidth.open
     : sidebarWidth.closed;
--- a/src/components/flowchart/flowchart.js
+++ b/src/components/flowchart/flowchart.js
@@ -506,7 +506,7 @@ export class FlowChart extends Component {
       // the hold shift only happens on clicking a node first
       // but only if no filters are currently applied.
       if (event.shiftKey && !this.props.isSlicingPipelineApplied) {
-        this.handleShiftClick(node);
+        this.handleMultipleNodesClick(node);
       }
     }
 
@@ -546,7 +546,7 @@ export class FlowChart extends Component {
     }
   };
 
-  handleShiftClick = (node) => {
+  handleMultipleNodesClick = (node) => {
     // Close meta data panel
     this.props.onLoadNodeData(null);
 


src/components/flowchart/flowchart.js
```diff
@@ -520,23 +529,21 @@ export class FlowChart extends Component {
...
+    // Handle metadata panel display or node click toggle
     displayMetadataPanel ? onLoadNodeData(id) : onToggleNodeClicked(id);
     toSelectedNode(node);
 
     const { from, to, range } = this.state.slicedPipelineState;
     // if both "from" and "to" are defined
     // then on a single node click, it should reset the sliced pipeline state
-    if (from !== null && to !== null) {
-      this.updateSlicedPipelineState(null, null, []);
+    if (from !== null && to !== null && !this.props.isSlicingPipelineApplied) {
```
jitu5, line 539: This condition is looks confusing as negative condition is checked first, We can keep the positive check first.

src/components/node-list/index.js
```diff
@@ -111,7 +112,11 @@ const NodeListProvider = ({
       } else {
         onToggleNodeSelected(item.id);
         toSelectedNode(item);
-        onResetSlicePipeline();
+
+        // Reset node filters only if filters are currently applied.
```
jitu5, line 116: When you say filter its filter by tag/nodeType or node slicing ?

src/components/node-list/node-list-tree.js
```diff
@@ -122,9 +122,14 @@ const TreeListProvider = ({
   expanded,
   onToggleNodeSelected,
   slicedPipeline,
+  isSlicingPipelineApplied,
 }) => {
   // render a leaf node in the modular pipelines tree
   const renderLeafNode = (node) => {
+    if (Object.keys(node).length === 0) {
+      return null;
+    }
```
jitu5, line 131: Why this check added? how it will relevant to slicing ?

src/components/node-list/node-list-tree.js
```diff
@@ -162,9 +168,16 @@ const TreeListProvider = ({
...
+
     if (!node) {
       return;
     }
+    const allChildrenDataEmpty = node.children.every(
+      (child) => Object.keys(child.data).length === 0
+    );
+    if (allChildrenDataEmpty) {
+      return;
+    }
```
jitu5, line 180: Could you please explain why we need `allChildrenDataEmpty` check ?

src/selectors/nodes.js
```diff
@@ -42,6 +43,16 @@ export const getGraphNodes = createSelector(
     }, {})
 );
 
+/**
+ * Filters the `nodes` object to include only the nodes whose IDs are present in the `filter` array.
+ * @param {Object} nodes - An object where keys are node IDs and values are node data objects.
+ * @param {Array} slicedPipeline - An array of node IDs to include in the sliced pipeline .
+ */
+export const filterNodes = (nodes, slicedPipeline) =>
```
jitu5, line 51: [Nit] `filterNodes` looks generic name, can we make it more specific to slicing Pipeline.

src/components/flowchart/flowchart.js
```diff
@@ -783,11 +792,17 @@ export class FlowChart extends Component {
           })}
           ref={this.layerNamesRef}
         />
-        {slicedPipelineState.range.length > 0 && (
+        {slicedPipeline.length > 0 && (
           <div ref={this.slicedPipelineActionRef}>
```
jitu5, line 796: `slicedPipelineActionRef` => `slicedPipelineActionBarRef`

src/components/node-list/node-list-tree.js
Huongg: hey good question, i'll add a comment to explain. But in short, because we need to filter out all the nodes from the node-list, that are not in slicing pipeline, and the way I filter it out is to assign an empty data object to its child. ([this change specifically](https://github.com/kedro-org/kedro-viz/blob/9d7d9cefd60364889919e8ea7c6f2fa64ccdf3ba/src/selectors/nodes.js#L291))

So the check here to make sure if the all the child under a modular pipeline are empty, then don't render that modular pipeline at all. Hope that makes sense?

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
```diff
@@ -1,27 +1,69 @@
...
+  } = props;
+  const { outerWidth: screenWidth } = chartSize;
+  const actionBarWidth =
+    ref.current && ref.current.firstChild.getBoundingClientRect().width;
+
+  const metaDataPanelWidth = displayMetadataPanel ? 600 : 400;
+  const nodeListWidth = visibleSidebar ? 200 : 400;
+
+  const transformX =
+    screenWidth - nodeListWidth - metaDataPanelWidth - actionBarWidth / 2;
```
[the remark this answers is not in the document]
Huongg: good spot, I'm meant to add the minimum value for transformX so it never gets hidden behind the nodelist bar. Updated the code now

src/selectors/run-command.js
```diff
@@ -0,0 +1,20 @@
+import { createSelector } from 'reselect';
+
+const getSlicedPipeline = (state) => state.slice;
+const getNodesNames = (state) => state.node.fullName;
```
ravi-kumar-pilla, line 4:
```suggestion
const getNodeName = (state) => state.node.fullName;
```

src/selectors/run-command.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const getRunCommand = createSelector(
+  [getSlicedPipeline, getNodesNames],
```
ravi-kumar-pilla, line 7:
```suggestion
  [getSlicedPipeline, getNodeName],
```

src/selectors/run-command.js
ravi-kumar-pilla, line 4: The naming of our redux store needs to be revisited. `state.node.fullName` sounds like a string but once I see the structure, it is a dictionary of node ids and names. Not related to this PR, but I would like us to work on this in future @Huongg

src/selectors/run-command.js
```diff
@@ -0,0 +1,20 @@
...
+  (slicedPipeline, nodesNames) => {
+    const { from, to } = slicedPipeline;
+
+    if (!from && !to) {
```
ravi-kumar-pilla, line 11: shouldn't this be OR like below -
```
if (!from || !to) {
}
```

src/selectors/nodes.js
```diff
@@ -255,23 +266,32 @@ export const getNodeDataObject = createSelector(
...
 export const getModularPipelinesTree = createSelector(
-  [(state) => state.modularPipeline.tree, getNodeDataObject],
-  (modularPipelinesTree, nodes) => {
+  [
+    (state) => state.modularPipeline.tree,
+    (state) => state.slice.apply,
+    getNodeDataObject,
+    getSlicedPipeline,
+  ],
+  (modularPipelinesTree, isSlicingPipelineApplied, nodes, slicedPipeline) => {
```
ravi-kumar-pilla, line 276: [Nit] isPipelineSliced

src/selectors/nodes.js
```diff
@@ -255,23 +266,32 @@ export const getNodeDataObject = createSelector(
...
     }
+
+    // Determine the relevant nodes based on whether slicing is applied
+    const relevantNodes = isSlicingPipelineApplied
+      ? filterNodesFromSlicingPipeline(nodes, slicedPipeline)
+      : nodes;
+
     for (const modularPipelineID in modularPipelinesTree) {
       modularPipelinesTree[modularPipelineID].data = {
         ...nodes[modularPipelineID],
```
ravi-kumar-pilla, line 288: Shouldn't this have ...relevantNodes[modularPipelineID] ?

src/selectors/nodes.js
```diff
@@ -42,6 +43,16 @@ export const getGraphNodes = createSelector(
     }, {})
 );
 
+/**
+ * Filters the `nodes` object to include only the nodes whose IDs are present in the `slicedPipeline` array.
+ * @param {Object} nodes - An object where keys are node IDs and values are node data objects.
+ * @param {Array} slicedPipeline - An array of node IDs to include in the sliced pipeline .
+ */
+export const filterNodesFromSlicingPipeline = (nodes, slicedPipeline) =>
```
ravi-kumar-pilla, line 51:
```suggestion
export const filterNodesFromSlicingPipeline = (nodes, slicedNodeIds) =>
```

src/components/node-list/node-list-tree.js
```diff
@@ -162,10 +170,20 @@ const TreeListProvider = ({
...
+
     // render each child of the tree node first
     const children = sortBy(
       node.children,
```
ravi-kumar-pilla, line 189: So we are checking if allChildren of modular pipelines are empty. If so, we return. Will there be a case where some children are empty ? If so, the line at 173 has `child.data.name` which will error out.

src/components/flowchart/flowchart.js
```diff
@@ -503,14 +504,22 @@ export class FlowChart extends Component {
       this.handleSingleNodeClick(node);
 
       // the hold shift only happens on clicking a node first
-      if (event.shiftKey) {
+      // but only if no filters are currently applied.
+      if (event.shiftKey && !this.props.isSlicingPipelineApplied) {
```
ravi-kumar-pilla, line 508: Is this to prevent slice on slice ? Do we in future expect to have a slice-on-slice feature (like multi-level-slicing) ? If so, how easy is it to extend this feature to multi-slice ?

src/components/node-list/node-list-tree.js
Huongg: i think it will return undefined if theres no name under child.data. But good point, I'm surprised that it didn't show any error. Should we add extra check there?

src/components/flowchart/flowchart.js
Huongg: hey that's correct, that's to prevent to slice on a sliced pipeline. I think what you pointed out there is a very extreme edge case and don't think we would every want to slice on a slice? What we might do is after slicing a pipeline, user can register it as a new pipeline. Then we just reset the slicing state and start again

src/components/flowchart/flowchart.js
```diff
@@ -570,15 +571,16 @@ export class FlowChart extends Component {
...
     // Check if the pipeline is sliced, no slice button is clicked, and no filters are applied
-    if (this.props.slicedPipeline && !isSliceButtonClicked) {
-      this.props.onResetSlicePipeline();
-      this.updateSlicedPipelineState(null, null, []);
-      // To reset URL to current active pipeline when click outside of a node on flowchart
-      this.props.toSelectedPipeline();
+    if (
+      this.props.slicedPipeline &&
+      !isSliceButtonClicked &&
+      !this.props.isSlicingPipelineApplied
```
ravi-kumar-pilla, line 581: I dont completely understand this check. I guess we are trying to see if we need to reset sliced pipeline on chart click. (Is this for the sliced view or before the sliced view ? )

Could you please explain why we need to check for all the 3 conditions when a chart click is triggered ? Also `isSlicingPipelineApplied` is true when someone clicks on slice button ? or before clicking ?

src/selectors/nodes.js
Huongg: hmm yeah I don't think it makes a different as at this level it is about modular pipeline, so nodes or relevant nodes will have the same data about modular pipeline

src/components/flowchart/flowchart.js
```diff
@@ -783,11 +790,17 @@ export class FlowChart extends Component {
...
+        {slicedPipeline.length > 0 && (
+          <div ref={this.slicedPipelineActionBarRef}>
             <SlicedPipelineActionBar
               chartSize={chartSize}
-              slicedPipeline={slicedPipelineState.range}
+              displayMetadataPanel={Boolean(clickedNode)}
+              isSlicingPipelineApplied={isSlicingPipelineApplied}
+              onApplySlicingPipeline={() => onApplySlice(true)}
+              onResetSlicingPipeline={this.resetSlicedPipeline}
+              ref={this.slicedPipelineActionBarRef}
```
ravi-kumar-pilla, line 801: Why do we have duplicate ref like the ref seems to be the same for div and the component. Any reason for this ?

src/components/flowchart/flowchart.js
Huongg: this is specifically for handling clicking outside the node. So we every time you click outside an element on the chart, it will trigger this `handleChartClick`, it will clear everything

We want to have this functionality for slicing view, lets say when you select one element, or 2 element, in the preview mode, when you click outside the elements, it should reset the chart. However we don't want this to happen in the SLICED view, because the only way to exit the sliced view is to click on the RESET button. And we also don't want this to happen when clicking on the action bar. 

But now I look at this, we  don't need to check `this.props.slicedPipeline` anymore. 


`isSlicingPipelineApplied is true when someone clicks on slice button` => this is correct

src/components/flowchart/flowchart.js
Huongg: ref for div is to create the ref. This `ref={this.slicedPipelineActionBarRef}` is to pass that ref we created from the div to the `SlicedPipelineActionBar` as a props

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
```diff
@@ -1,27 +1,83 @@
...
+    displayMetadataPanel,
+    isSlicingPipelineApplied,
+    onApplySlicingPipeline,
+    onResetSlicingPipeline,
+    runCommand,
+    slicedPipeline,
+    visibleSidebar,
+  } = props;
+  const { outerWidth: screenWidth } = chartSize;
+  const buffer = 200;
```
ravi-kumar-pilla, line 23: may be we can have these values in constant or config files to manage like SLICED_PIPELINE_BUFFER_WIDTH ? (not sure of the naming)

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
```diff
@@ -1,27 +1,83 @@
...
+        {`${slicedPipeline.length} ${
+          isSlicingPipelineApplied ? 'sliced' : 'selected'
+        }`}
       </div>
-      <div className="sliced-pipeline-action-bar--cta">
-        <Button>Slice</Button>
+      <div
+        className={classnames('sliced-pipeline-action-bar--run-command', {
+          'sliced-pipeline-action-bar--run-command-long':
+            runCommand.length > 85,
```
ravi-kumar-pilla, line 58: can we move 85 to config or constants file and also, how did we come up to this number ?

src/selectors/nodes.js
ravi-kumar-pilla: but relavant nodes are subset of nodes right ? So the relevant nodes might not have the modularPipelineID ?

src/components/node-list/node-list-tree.js
ravi-kumar-pilla: I think so may be child?.data?.name

src/components/flowchart/flowchart.js
ravi-kumar-pilla: On Viz side atleast, if we have an initialPipelineState, we can iteratively update the initialPipelineState = slicedState for recursive slicing. For now, I wanted to know if there is any idea of supporting this case in future :)

ravi-kumar-pilla:
Hi @Huongg , Overall the PR looks great. However, I observed few issues -

1. The SliceActionBar animates from left to right is not smooth. (It might depend on the system as well)
2. On bigger screens, the SliceActionBar is not centered. Can we try to justify the action bar to center irrespective of screen size ?
3. After clicking of Reset slice, sometimes I observed the transition is not smooth.

The functionality looks great, but there are some bumpy transitions on my machine. Thank you

src/components/flowchart/flowchart.js
Huongg: good suggestion 👍 for now as this is meant to be MVP1 and we want to keep them simple, and we will need to look at user data to see whether we want to improve the feature or not. This can be something taking into account for future discussion

src/selectors/nodes.js
Huongg: hey relevant nodes will have modularPipelineID , as well as nodes. That's because when we do the slicing, we only filtering out nodes but not modular pipelines

Huongg:
> Hi @Huongg , Overall the PR looks great. However, I observed few issues -
> 
> 1. The SliceActionBar animates from left to right is not smooth. (It might depend on the system as well)
> 2. On bigger screens, the SliceActionBar is not centered. Can we try to justify the action bar to center irrespective of screen size ?
> 3. After clicking of Reset slice, sometimes I observed the transition is not smooth.
> 
> The functionality looks great, but there are some bumpy transitions on my machine. Thank you

hey can you attach the screen recording from your side please? Also at the moment, the position of the bar is calculated based on screen size. So would be great to see what you experienced from your side 👍

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
Huongg: it's just a magic number. from look and feel anything more than 85 seems like a long command, and no need to truncate. I'll give it a name so it makes more sense we read it

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
Huongg: I could give it a better name rather just buffer? Im not sure if we should move it to config as it won't be shared between any other components 🤔

stephkaiser:
Thanks Huong! so excited for this, great work so far!

some design QA comments from me:

1. the paddings for the floating slice component doesnt seem to match design, let me know if you need my help in documenting the different paddings its a bit complicated
2. the hover states are missing for the copy button and the slice button. The reset slice button on hover should have a slightly darker red colour and keeps the white text
3. I forgot to add the gradient for long run commands on the left side as well when user scrolls to the right, could we add this as well? right now when user scrolls to the end on the right, there is still a gradient so its not clear if user has reached the end of the text.
Screenshot 2024-08-07 at 19 41 57

4. the "Copied!" tooltip position needs to be a bit higher - should be 4px above the entire component itself
5. the slice button and reset slice button have different heights, both should be 40px

Thank you!!

src/selectors/nodes.js
Huongg: hmm i think that name is quite confusing, because with this one we try to highlight whether the slicing Pipeline is applied or not, it's not really about whether the pipeline is sliced or not

Huongg:
hey @ravi-kumar-pilla as discussed one friday, I've now improved the transition of the action bar, and also fix the bug that you found when the first selected node is not highlighted. Let me know if you can see these changes from your side?

Huongg:
> Thanks Huong! so excited for this, great work so far!
> 
> some design QA comments from me:
> 
> 1. the paddings for the floating slice component doesnt seem to match design, let me know if you need my help in documenting the different paddings its a bit complicated
> 2. the hover states are missing for the copy button and the slice button. The reset slice button on hover should have a slightly darker red colour and keeps the white text
> 3. I forgot to add the gradient for long run commands on the left side as well when user scrolls to the right, could we add this as well? right now when user scrolls to the end on the right, there is still a gradient so its not clear if user has reached the end of the text.
> 
> Screenshot 2024-08-07 at 19 41 57
> 4. the "Copied!" tooltip position needs to be a bit higher - should be 4px above the entire component itself
> 5. the slice button and reset slice button have different heights, both should be 40px
> 
> Thank you!!

hey @stephkaiser I think I've addressed all of your comments apart from point 3, let's discuss which approach we should go with for this one 😄

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
rashidakanchwala: yeah makes sense to name it better. `buffer` did feel generic.

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
```diff
@@ -1,27 +1,102 @@
...
+        </div>
+      ) : (
+        <div
+          className="sliced-pipeline-action-bar--cta sliced-pipeline-action-bar--slice"
+          onClick={onApplySlicingPipeline}
+        >
+          <IconButton
+            ariaLabel="Copy run command to clipboard."
+            className="copy-button"
+            dataTest={`clicked.run_command`}
```
rashidakanchwala, line 94: need to update in contributing docs but the norm is dataTest = < parent component >-< component >-<ui element (button, input, dropdown)>-< state : optional (true/false) >

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
```diff
@@ -1,27 +1,102 @@
...
+        className={classnames('sliced-pipeline-action-bar--run-command', {
+          'sliced-pipeline-action-bar--run-command-long':
+            runCommand.length > 85,
+        })}
+      >
+        <CommandCopier command={runCommand} isCommand={true} />
       </div>
+      {isSlicingPipelineApplied ? (
+        <div className="sliced-pipeline-action-bar--cta sliced-pipeline-action-bar--reset">
+          <Button onClick={onResetSlicingPipeline}>Reset slice</Button>
```
rashidakanchwala, line 84: can we add dataTest here, and the next cta too.

src/components/flowchart/flowchart.js
```diff
@@ -503,14 +504,22 @@ export class FlowChart extends Component {
       this.handleSingleNodeClick(node);
 
       // the hold shift only happens on clicking a node first
-      if (event.shiftKey) {
+      // but only if no filters are currently applied.
+      if (event.shiftKey && !this.props.isSlicingPipelineApplied) {
         this.handleShiftClick(node);
```
rashidakanchwala, line 509: nit - i would call this handleMultipleNodeClick(node)

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
Huongg: hey yup will do once I get to the working on heaps analytics dashboard

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.js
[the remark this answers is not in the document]
jitu5: @Huongg on `demo-slicing-pipeline-final` branch I tested locally I still see action bar overlap with metadata panel.

Screenshot 2024-08-13 at 2 31 55 p m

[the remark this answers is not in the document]
jitu5: @Huongg I am fine with it. It would have been nice to fit action bar within available space, keeping command box as variable width container, may be we can keep this as future work after MVP.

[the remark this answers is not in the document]
Huongg: yeah interesting idea, i think it require further design thinking as in order to fix the whole action bar in the available space, we will need to reduce the width, in that case what the transition/animation would look like, etc

Agree that can be something to be considered in the future MVP

ravi-kumar-pilla (approved):
Great work @Huongg !! 💯

PR #2003, opened 2024-07-26
Feature/slicing pipeline add notification

SajidAlamQB:
## Description

This feature will show a notification to kedro users to `Hold Shift + Click on another node to slice pipeline`. Moreover, the order in which nodes are selected to should not matter and be resolved Programatically.

## Development notes
When testing on your machine locally, or through gitpod, please note that this PR only covers the functionalities below:

- [x]  Notification on First Node Click: When I click on the first node, a notification should appear. This behavior should not occur when clicking on a pipeline (is it referred to as "pipeline" or "modular pipeline"?).
- [x] Second Node Click with Shift: If I hold the Shift key and click on a second node, the metadata should close immediately, and the notification should disappear.
- [x] when selecting a node from bottom to top, the order should be updated according to --from-node=bottom-node-id --to-node=top-node-id

**Notification bar:**

image

## Checklist


src/components/flowchart/flowchart.js
```diff
@@ -711,6 +717,17 @@ export class FlowChart extends Component {
   /**
    * Render React elements
    */
+  renderNotification() {
```
Huongg, line 720: hey @SajidAlamQB perhaps `SlicedPipelineActionBar` can be reused here, rather you have to make a new component. You might need to change a bit logic inside `SlicedPipelineActionBar` but it has the styling and theme already set up in there

Huongg (commented):
hey @SajidAlamQB, The code is looking pretty good! 🔥 

I did notice one thing: the notification transition is a bit glitchy. In the video, it looks like it gets stuck and doesn't move to the left smoothly in one transition, but it seems fine when you shift-hold and click. 🤔 I wonder if the translateX isn't being applied immediately or if the transformX calculated number is changing too frequently.


https://github.com/user-attachments/assets/a9749003-92d2-4a53-ae3f-de6a0bc929e2

src/components/flowchart/flowchart.js
```diff
@@ -542,20 +545,34 @@ export class FlowChart extends Component {
...
+    const fromNodeId = fromNodeIdState || node.id;
     const toNodeId = node.id;
 
-    const newState = {
-      ...this.state.slicedPipelineState,
-      from: fromNodeId,
-      to: toNodeId,
-    };
+    // Get bounding client rects of nodes
+    const fromNodeElement = document.querySelector(`[data-id="${fromNodeId}"]`);
```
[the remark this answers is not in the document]
SajidAlamQB: That makes sense it should make it a lot cleaner now.

src/components/flowchart/flowchart.js
```diff
@@ -528,34 +529,70 @@ export class FlowChart extends Component {
...
+    }
+
+    const fromNodeRect = fromNodeElement.getBoundingClientRect();
+    const toNodeRect = toNodeElement.getBoundingClientRect();
+
+    // Ensure that 'from' node is higher (smaller Y-coordinate) than 'to' node
+    return fromNodeRect.y < toNodeRect.y
+      ? [initialFromNodeId, initialToNodeId] // 'from' node is higher
+      : [initialToNodeId, initialFromNodeId]; // 'to' node is higher
+  };
```
jitu5, line 571: This is cool 👍🏼

SajidAlamQB:
> I did notice one thing: the notification transition is a bit glitchy. In the video, it looks like it gets stuck and doesn't move to the left smoothly in one transition, but it seems fine when you shift-hold and click. 

I am not able to see this on my end, the transition seems smooth when we click away and when we shift + click.

Huongg:
> > I did notice one thing: the notification transition is a bit glitchy. In the video, it looks like it gets stuck and doesn't move to the left smoothly in one transition, but it seems fine when you shift-hold and click.
> 
> I am not able to see this on my end, the transition seems smooth when we click away and when we shift + click.

hey I've updated the transition timing from 0.7s to 0.2s and it looks a lot smoother from my side now. I think 0.7s was to slow so it's showing all the different movements, that's why i can see it's a bit glitchy from myside

stephkaiser:
thanks Sajid! some comments from me:

1. the text colour in the tooltip notification component should be black-900 on dark UI, and white-0 on light UI
2. there seems to be an odd line at the end of the component, not sure if this is a bug or not 
Screenshot 2024-08-08 at 12 32 58
Screenshot 2024-08-08 at 12 43 49

3. the tooltip notification closes when I click outside in the flowchart background area however, it doesn't get dismissed if i click on the X close button in the metadata panel, or if i click on another node in the list panel, or on the expand/collapse button. The notification is still there and persists
Screenshot 2024-08-08 at 12 33 26

4. when selecting from bottom to top - after having selected a from and to node, when holding shift to select a different to node, it keeps the second selection (to node) but changes the from node. I expected the to node to change instead and keep the first from node selection, which is the same behaviour if selecting from top to bottom.

src/components/flowchart/components/sliced-pipeline-action-bar/sliced-pipeline-action-bar.test.js
```diff
@@ -0,0 +1,48 @@
...
+      <SlicedPipelineActionBar
+        chartSize={{ outerWidth: 800 }}
+        slicedPipeline={[1, 2, 3]}
+        visibleSidebar={true}
+        notification={false}
+        displayMetadataPanel={false}
+        ref={ref}
+      />
+    );
+    const sliceButton = screen.getByRole('button', { name: /slice/i });
```
Huongg, line 45: another way you can test this is to just mount the component and find if the classname existed. The reason why it's better because we can be more flexibility with the classname, as it might not always be button 

eg

```
const wrapper = mount(<Container />);
expect(wrapper.find('classname')).toHaveLength(1);
```

src/components/flowchart/flowchart.js
```diff
@@ -528,34 +538,80 @@ export class FlowChart extends Component {
...
+    const toNodeRect = toNodeElement.getBoundingClientRect();
+
+    // Reorder the nodes based on their Y-coordinate
+    const [adjustedFromNodeId, newToNodeId] =
+      fromNodeRect.y < toNodeRect.y
+        ? [userSelectedFromNodeId, toNodeId] // Keep order
+        : [toNodeId, userSelectedFromNodeId]; // Swap if needed
+
+    return {
+      userVisualFromNodeId: userSelectedFromNodeId, // Keep user's selection visually as 'from'
```
Huongg, line 586: i don't think you need to return this to keep the users selection. Different approach as the comment below

src/components/flowchart/flowchart.js
```diff
@@ -528,34 +538,80 @@ export class FlowChart extends Component {
...
+  };
+
   handleShiftClick = (node) => {
     // Close meta data panel
     this.props.onLoadNodeData(null);
+    const { from: userSelectedFromNodeId, range } =
+      this.state.slicedPipelineState;
 
-    const fromNodeId = this.state.slicedPipelineState.from || node.id;
+    const fromNodeId = userSelectedFromNodeId || node.id; // Keep existing 'from' node or set the current one if not set
```
Huongg, line 598: I think the simpler way will be:
- whatever user choses, we update the state, because the state is simply just for changing the background colour of the element so it doesn't really matter if the from node is from the bottom
- we only change the order when we update the redux store. 

So something like this will be better I think:

```
const fromNodeId = fromNodeIdState || node.id;
const toNodeId = node.id;

// update the state from what user has clicked
this.updateSlicedPipelineState(fromNodeId, toNodeId, range);

const { updatedFromNodeId, updatedToNodeId } = this.determineNodesOrder(
      fromNodeId,
      toNodeId
 );

 // Slice the pipeline based on the determined node order
 // If the order could not be determined, use the original selection
 if (updatedFromNodeId && updatedToNodeId) {
      this.props.onSlicePipeline(updatedFromNodeId, updatedToNodeId);
  } else {
      this.props.onSlicePipeline(fromNodeId, toNodeId);
  }

this.props.onApplySlice(false);
this.setState({ showSlicingNotification: false }); // Hide notification after selecting the second node

```

Huongg: this way, `determineNodesOrder` only returns 2 updated values, rather than 3

src/components/flowchart/flowchart.js
```diff
@@ -577,6 +633,7 @@ export class FlowChart extends Component {
     if (this.props.slicedPipeline && !isSliceButtonClicked) {
       this.props.onResetSlicePipeline();
       this.updateSlicedPipelineState(null, null, []);
+      this.setState({ showSlicingNotification: false }); // Hide notification when clicking away
```
Huongg, line 636: do you think the `this.setState({ showSlicingNotification: false })` can be a part of `this.props.onResetSlicePipeline();`? I can imagine every time you resetPipeline you will need to clear the notification 🤔

src/components/flowchart/flowchart.js
Huongg: also i took me awhile to fully understand what these `adjustedFromNodeId and newToNodeId` mean. Maybe something simpler might be better, something to indicated that this `fromNodeId` and `toNodeId` has been updated/re-ordered their position?

src/components/flowchart/flowchart.js
SajidAlamQB: Thats a good point, I think that should be fine to do!

src/components/flowchart/flowchart.js
SajidAlamQB: Done, this has simplified it!

jitu5:
@SajidAlamQB @Huongg 
Have you considered the case where Kedro Viz used as React component? If I am not wrong we need to disable this functionality right ?

Huongg:
> @SajidAlamQB @Huongg Have you considered the case where Kedro Viz used as React component? If I am not wrong we need to disable this functionality right ?

good question, can we test it together tomorrow or sometimes this week? If it's working fine and this functionality is a part of flowchart, I don't see why we can't have this? 

Do you see any risks we will take if we don't disable it @jitu5 ?

jitu5:
> > @SajidAlamQB @Huongg Have you considered the case where Kedro Viz used as React component? If I am not wrong we need to disable this functionality right ?
> 
> good question, can we test it together tomorrow or sometimes this week? If it's working fine and this functionality is a part of flowchart, I don't see why we can't have this?
> 
> Do you see any risks we will take if we don't disable it @jitu5 ?

@Huongg In most of the case When Kedro viz is used in embedded mode the main purpose is readonly or only flowchart visualisation, But I am not fully sure should we disabled it or not. @rashidakanchwala wdyt ?

jitu5 (approved):
LGTM! Thanks @SajidAlamQB @Huongg

Huongg:
> > > @SajidAlamQB @Huongg Have you considered the case where Kedro Viz used as React component? If I am not wrong we need to disable this functionality right ?
> > 
> > 
> > good question, can we test it together tomorrow or sometimes this week? If it's working fine and this functionality is a part of flowchart, I don't see why we can't have this?
> > Do you see any risks we will take if we don't disable it @jitu5 ?
> 
> @Huongg In most of the case When Kedro viz is used in embedded mode the main purpose is readonly or only flowchart visualisation, But I am not fully sure should we disabled it or not. @rashidakanchwala wdyt ?

It's a bit tricky to determine if it's suitable for embedded mode without actually seeing it in action. In my mind, it seems like there wouldn't be any harm in including it in the standalone flowchart, since that's where most of the interaction happens. However, I can also understand the reasons for disabling it, especially if we decide to turn off the metadata panel.

src/components/flowchart/flowchart.js
```diff
@@ -134,6 +135,15 @@ export class FlowChart extends Component {
         this.updateSlicedPipelineState(from, to, this.props.slicedPipeline);
       }
     }
+
+    // Hide slicing notification if metadata panel is closed using button
```
ravi-kumar-pilla, line 139: I did not fully understand the condition here. Can you please add comments or let me know what these conditions are for. Thank you

SajidAlamQB: The first condition is checking if a previous node was clicked, `prevProps.clickedNode`, so the metadata panel was opened and it checks if a current node has not been clicked, `!this.props.clickedNode`,  which implies we have closed the metadata panel. This is the scenario where we have shift+clicked a single node which results in metadata-panel opening and user closing the metadata panel using the X button.

The second condition, `!this.props.slicedPipeline.length`, just makes sure we run this during slicing.

This scenario is where we have shift+clicked a single node which results in metadata-panel opening and user closing the metadata panel using the X button. We want to reset the slicing because without this it was keeping the node highlighted.

SajidAlamQB: @Huongg the bug where the node remains highlighted until we hover over a different node after metadata panel is closed is still present.

Huongg: hey @SajidAlamQB shouldn't it just to set the notification to false, we don't need to set the slicingPipelineState as well?

```
// Hide slicing notification when metadata panel is closed
if (this.props.clickedNode !== prevProps.clickedNode && !this.props.clickedNode) {
  this.setState({ showSlicingNotification: false });
}
```

Huongg: and yes that bug is fixed in the other PR, you can check the demo branch to see if you can still reproduce the bug

SajidAlamQB: Ah yes it works on the demo branch 💯 and we don't need `slicingPipelineState` it was left over from when I was trying to fix the bug.

ravi-kumar-pilla: > This scenario is where we have shift+clicked a single node which results in metadata-panel opening

Is this by design ? or can we do something like to enter the slice mode, user need to click on nodes using shift + click. Why should we open the metadata-panel when someone is using shift+click on a node ? Possibly they might want to use the slice functionality.

I would suggest -

1. In regular mode - 
   a. If the user clicks on a node without using shift, we open the metadata panel. 
   b. If the user clicks on a node using shift, even if it is a one node click, we do not open the metadata panel and wait for a second node click using shift
2. In slice mode view -
   a. If the user clicks on a node with/without using shift, we open the metadata panel (as we do not allow multi-slicing at this moment)
   
   In summary, I feel we should not open a metadata panel when someone is shift+clicking on a node.
   
   Thank you

ravi-kumar-pilla: After looking at this PR - https://github.com/kedro-org/kedro-viz/pull/1980
The scenario I suggested above already exists (if I click on first node using shift, it does not open metadata panel). 

What do you mean by - `This scenario is where we have shift+clicked a single node which results in metadata-panel opening` @SajidAlamQB in your response above ? Thank you

ravi-kumar-pilla: okay, I understand the condition now. But how should the notification behave in the below use case (in gif) -

1. I click on a node -> the notification appears + metadata panel opens
2. I shift+click on another node -> the notification now says number of nodes selected + metadata panel closes
3. Now I click on a different node -> At this moment, this opens the metadata panel but does not show the notification as it shows in step 1 (which seems inconsistent). If the behavior is consistent, this is just re-iteration of step 1, so it should do -> the notification appears + metadata panel opens

**For design -**
In the gif, if you see the initial notification appears at a different position and then after selection of 2 nodes, the notification position changes. I feel it would be better if we just change the text content of the already present notification bar and do not re-run the animation. Wdyt @stephkaiser 

shift_click_scenario_notification

Happy to discuss on this further. Thank you

SajidAlamQB: > What do you mean by - `This scenario is where we have shift+clicked a single node which results in metadata-panel opening` @SajidAlamQB in your response above ? Thank you

Sorry I meant just clicked node, not shift+click. When you click a node the metadata panel opened and if user closed it with 'X' the notification stayed. That's why the condition is there.

SajidAlamQB: > 1. I click on a node -> the notification appears + metadata panel opens
> 2. I shift+click on another node -> the notification now says number of nodes selected + metadata panel closes
> 3. Now I click on a different node -> At this moment, this opens the metadata panel but does not show the notification as it shows in step 1 (which seems inconsistent). If the behavior is consistent, this is just re-iteration of step 1, so it should do -> the notification appears + metadata panel opens

This is fixed please test it and lmk your thoughts.

ravi-kumar-pilla:
Hi @SajidAlamQB ,

As I was testing the fix for - https://github.com/kedro-org/kedro-viz/pull/2003#discussion_r1717355543
There seems to be another issue for the use case below [in gif] -

1. I click on a node -> the notification appears + metadata panel opens
2. I shift+click on another node -> the notification now says number of nodes selected + metadata panel closes
3. Now I click on a different node -> the notification appears + metadata panel opens (This was not working earlier and its fixed now. Thank you)
4. I shift+click on another node -> here is the **issue** [the notification now says 1 node selected + metadata panel closes]

shift_click_issue

SajidAlamQB:
> 4. I shift+click on another node -> here is the **issue** [the notification now says 1 node selected + metadata panel closes]

Should be fixed now.

ravi-kumar-pilla (approved):
LGTM !! Thank you @SajidAlamQB

PR #2004, opened 2024-07-26
Fixing highlight width inconsistencies in list panel

### LINKED ISSUE #2002: Fixing highlight width inconsistencies in list panel
## Description
While working on the slicing MVP, we noticed several inconsistencies with the highlighting in the left panel:

1. **When hovering on the list and on a node in the flowchart** - highlighting in the list panel is not the same width (one is longer, other is shorter)
2. **When clicking on the list and on a node in the flowchart** - highlighting in the list panel stretches to full-width (both are full-width)
3. **When hovering and clicking on pipeline groups** - 
**On the list** - hover and click has different highlight widths (one is longer, other is shorter)
**In the flowchart** - hover and click on pipeline groups doesn't highlight in the list panel **(this is a BUG)**

**Summary** - highlight widths are inconsistent across different interactions in the list panel. There are currently 3+ different widths in the list panel, and hover and selection states also have inconsistent highlight widths.

## Context
These issues should ideally be fixed before we can release the slicing MVP feature. 

## Suggested Implementation - New Design Updates

New design pattern for highlights in the left panel:

1. Now we only have 1 width (always full-width) for both hover and selection states.
2. Hovering on an expanded pipeline group will also highlight all it's children in the same highlight width.
3. Selected items will have a blue indication line on the right side, and a full-width highlight as well. Pipeline groups cannot be selected, only nodes can be selected. Clicking on a pipeline group will expand/collapse it (this is the same pattern currently and already exists). 
4. When a child node inside a pipeline is selected - if the pipeline group is expanded, the blue line will be shown at the child level. If the pipeline group is collapsed, the blue line will be shown at the parent pipeline level to indicate something is selected within that pipeline group. 

More detailed examples of the different states and various options can be found here - https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=1763-75495&t=yYd5EKuBdBCdDf3X-4

Any questions let me know!

rashidakanchwala:
## Description

Resolves #2002

## Development notes

**Known Issue**: When hovering over the 'Expand Pipeline' icon, the entire width is not highlighted. This issue arises due to the complexity of the MUI Tree structure. Attempting to fix this causes other elements to break. A follow-up ticket will be created to address this bug.

Steph: Given its complexity, please let me know if this issue should be considered a low priority, as it does not affect the slicing issue.
## Checklist


Huongg:
> Thanks Rashida! some comments from me:
> 
> 1. the full-width highlights looks great! it seems like parameters still have the shorter width, could we include parameters as well in this update?
> 2. selecting a child node and then closing the pipeline group doesnt bring the blue selection line to the parent level, can we also include this change? This would be useful in slicing when the contents of a pipeline are selected but not expanded
> 
> let me know if you want to discuss!

hey @stephkaiser, please see my reply as below:
1. what you found there is actually a bug, when the parameter tag is off, user shouldn't be able to hover over it. I've fixed it now. So now if you turn the parameter tag on, hovering over the parameter should have the full width
2. For this one, let me try to adjust this in the slicing PR rather than here. Because if we do it here it will introduce a new behaviour, as in now if you select the node, close the pipeline, the node will be deselected anyway, so it will be hard to highlight the pipeline. However in the slicing we keep the list of the node so should be able to do this one in that PR

stephkaiser:
> hey @stephkaiser, please see my reply as below:
> 
> 1. what you found there is actually a bug, when the parameter tag is off, user shouldn't be able to hover over it. I've fixed it now. So now if you turn the parameter tag on, hovering over the parameter should have the full width
> 2. For this one, let me try to adjust this in the slicing PR rather than here. Because if we do it here it will introduce a new behaviour, as in now if you select the node, close the pipeline, the node will be deselected anyway, so it will be hard to highlight the pipeline. However in the slicing we keep the list of the node so should be able to do this one in that PR

Thanks Huong!
1. I just had a look and this looks good to me! 
2. I'll review this part in the slicing PR, thanks

Huongg:
if so you're happy to approve the PR @stephkaiser so we can merge it?

stephkaiser (approved):

stephkaiser:
> if so you're happy to approve the PR @stephkaiser so we can merge it?

done! thank you :)

src/components/flowchart/flowchart.js
```diff
@@ -485,7 +488,11 @@ export class FlowChart extends Component {
    * @param {Object} node Datum for a single node
    */
   handleNodeMouseOver = (event, node) => {
-    this.props.onToggleNodeHovered(node.id);
+    if (node.type === 'modularPipeline') {
```
Huongg, line 491: hey @rashidakanchwala why do we need this? I thought this can be done via css?

rashidakanchwala: We need to add the --active attribute to modular pipelines and that only happens if you update the redux state.

rashidakanchwala: u can try it if possible via css

Huongg: https://github.com/user-attachments/assets/19177b08-c37e-4199-9773-27fe5585fb92


Am I missing something??  it's working fine without adding the active attribute here. If the video looks fine to you I'll remove this extra changes

rashidakanchwala: it's the other way round. hovering over the flowchart and see the effects on flowchart/nodelist. if you hover over the Modular Pipelines.

Huongg: ah I see, got it now. Let me test again

Huongg (approved):

Issue #2022, opened 2024-08-02
Introduce a callback as prop to React component to track node click in flowchart


jitu5:
## Description

Add a callback function as a prop to the Kedro-Viz React component. This callback will track node clicks, enabling developers to handle node click events outside of the component. This enhancement aims to provide more flexibility and control over interactions within the flowchart and facilitates communication with the Kedro extension to understand which node the user clicked and pass on the node data to the Kedro extension.

## Context
This feature will allow users of the Kedro-Viz component to define custom behavior when nodes are clicked. It will be particularly useful for applications that require interactions such as displaying detailed information about a node. Additionally, it enhances communication with the Kedro extension by providing node click data.

## Possible Implementation
- Add the `onNodeClick` prop to the Kedro-Viz component's prop types.
- Update the node click handler within the component to call `onNodeClick` with appropriate node data.
- Implement the logic to pass node data to the Kedro extension.

This issue is linked to parent ticket https://github.com/kedro-org/kedro-viz/issues/1968

## Checklist

- [ ] Include labels so that we can categorise your feature request

PR #2027, opened 2024-08-06
Introduce `onActionCallback` prop in Kedro-Viz react component

jitu5:
## Description

Resolves https://github.com/kedro-org/kedro-viz/issues/2022

This PR addresses a requirement for users of Kedro-Viz as a React component in embedded mode. It allows displaying node metadata in a custom UI. Additionally, it supports the VSCode extension use case where clicking on a node passes the node data to the extension to open the relevant file.

```
<KedroViz 
  data={data} 
  options={...}
  onActionCallback={(action) => {
    console.log('Action: ', action);
  }}
/>
```

Below is the action format, which will be pass as argument to call back function  
```
action = {
  type: <ACTION_NAME>,
  payload: <PAYLOAD_OF_ANY_TYPE>
}
```

Note: We can not put `onNodeClickCallback` under a `options` prop as `onActionCallback` is a callback function and options are Redux state, it should only contain plain JavaScript objects, arrays, and primitives.

## Development notes

- Callback is pass to store via `configureStore` in App.js
- Callback call happens in `createCallbackMiddleware`
- On `TOGGLE_NODE_CLICKED` action intercept in middleware, using clicked node id and `state.graph.nodes` entire node object is pass as an argument to callback.  

## QA notes

You can test new prop in src/components/container.js as <App /> is the entry point or top level component for a standalone use case.

```
import React from 'react';
import App from '../app';
import './container.scss';

/**
 * Top-level component for the use-case where Kedro-Viz is run as a standalone
 * app rather than imported as a library/package into a larger application.
 */
const Container = () => (
  <>
    <App
      options={{
        display: {
          globalNavigation: false,
          metadataPanel: false,
          sidebar: false,
        },
        theme: 'dark',
      }}
      onActionCallback ={(action) => {
        console.log('Action: ', action);
      }}
      data={getPipelineData()}
    />
  </>
);

export default Container;

```

## Checklist


src/store/middleware.js
```diff
@@ -0,0 +1,19 @@
+const createCallbackMiddleware =
```
Huongg, line 1: hey @jitu5 the code looks great, if you can add some context as a comment here would be great.

Huongg (commented):
hey generally the PR is looking pretty good and simple, thank you @jitu5 . My first thought would be what happens if we introduce another callback? eg: for slicing, run command etc. Would it be better to have a group for all the callback? Let me know your thought 

```
<KedroViz 
  data={data} 
  options={...}
  callbacks={
     onNodeClick={(node) => {
         console.log('Node clicked:', node);
      }} 
     onRunCommandClick={(runCommand) => ...}
  }
/>
```

jitu5:
> hey generally the PR is looking pretty good and simple, thank you @jitu5 . My first thought would be what happens if we introduce another callback? eg: for slicing, run command etc. Would it be better to have a group for all the callback? Let me know your thought
> 
> ```
> <KedroViz 
>   data={data} 
>   options={...}
>   callbacks={
>      onNodeClick={(node) => {
>          console.log('Node clicked:', node);
>       }} 
>      onRunCommandClick={(runCommand) => ...}
>   }
> />
> ```


@Huongg I made changes to accept any event or action if we want to use a callback, as shown below.

```
<KedroViz 
  data={data} 
  options={...}
  onActionCallback={(action) => {
    console.log('Action: ', action);
  }}
/>
``` 


Below is the action format, which will be pass as argument to call back function
```
action = {
  type: <ACTION_NAME>,
  payload: <PAYLOAD_OF_ANY_TYPE>
}
```

src/store/middleware.js
jitu5: Added.

rashidakanchwala (approved):
Look's great!. thanks @jitu5

Huongg (approved):
thanks for the update @jitu5 . looks good to me too

Issue #1997, opened 2024-07-24
Reload the page resets back the default pipeline


Huongg:
## Description
When you're in any other pipeline view rather the default pipeline, if you load the URL again, it will reset to the default page

https://github.com/user-attachments/assets/18414bc9-3468-49b2-9a76-3da826943f04

## Context
I'm unable to share a specific pipeline view with anyone through the URL as it always takes me back to the default pipeline

## Expected Result
If the pipeline view should be the same as `pid` when reload the page and `pid` should not always set back to be the default one

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version:
* Operating system and version:
* NodeJS version used (if relevant):
* Kedro version used (if relevant):
* Python version used (if relevant):

## Checklist
- [ ] Include labels so that we can categorise your issue

rashidakanchwala:
Look at https://github.com/kedro-org/kedro-viz/issues/1879

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro-viz/pull/2041

Issue #1879, opened 2024-04-26
Fix pipeline id which defaults to __default__ pipeline


ravi-kumar-pilla:
## Description

Kedro-Viz URL points to `__default__` pid even if there is no pipeline with the id `__default__`. This breaks the UI and shows as below -

![Image](https://github.com/kedro-org/kedro-viz/assets/87735534/890353ef-a4d8-4231-9478-5779b770b9a5)

## Context

I have a kedro project which returns 1 pipeline as below -

```python
generic_pipeline = gn.create_pipeline()

    return {
        "Generic": generic_pipeline
    }
```

The dropdown in the pipeline selector shows Generic but the URL pid points to `__default__`. This partially breaks Kedro-Viz

**NOTE:** This is working on Kedro-Viz < 9.0.0. This might be due to the stateful URL. However when I tried to re-produce this on latest Kedro-Viz, it started working. I would request anyone else (cc: @jitu5 ) to test this as I was unable to re-produce and I do not see any default pid on Frontend. Please feel free to close the ticket if everything looks good.

## Steps to Reproduce

1. Remove `__default__` pipeline from demo_project
2. Run `kedro viz run`
3. You will see the message `Oops, this URL isn't valid`

## Expected Result

- Kedro-Viz should show the pipeline which is selected in the dropdown
- The URL should point to the selected pipeline id in case there is no `__default__` pipeline 

## Actual Result

You will see the message `Oops, this URL isn't valid`

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version: Chrome
* Operating system and version: mac
* NodeJS version used (if relevant):
* Kedro version used (if relevant):
* Python version used (if relevant):

## Checklist
- [x] Include labels so that we can categorise your issue

PR #2041, opened 2024-08-16
Reload the page resets back the default pipeline

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/1997

When you're in any other pipeline view rather the default pipeline, if you load the URL again, it will reset to the default page.

## Development notes

- Introduced the `modularPipelineTree` logic to ensure that only the children of a modular pipeline are expanded.

- Added a check to avoid overwriting the active pipeline if one is already set.

- Removed the `urlParams` from `preparePipelineState` in the `getInitialState` method, as they were causing unexpected overwrites.

## Checklist


SajidAlamQB:
One issue I noticed was when you're in any other pipeline view except the default and you click a node when the page refreshes the node is not highlighted anymore and the metadata panel doesn't open. In the default view reloading with a node selected keeps it highlighted and opens metadata panel.

rashidakanchwala (approved):
works for me. thanks

Huongg:
hey @SajidAlamQB I think the pipeline bug is fixed now but as you said there's other bug about the selected node is not highlighted and metadata panel is not opened. And even in the default pipeline, the node is not highlighted even the meta data panel opens 🤔 

Screenshot 2024-08-19 at 17 00 18

SajidAlamQB:
> hey @SajidAlamQB I think the pipeline bug is fixed now but as you said there's other bug about the selected node is not highlighted and metadata panel is not opened. And even in the default pipeline, the node is not highlighted even the meta data panel opens 🤔

I'm not able to see the bug with the default pipeline, but it does exist with the other pipeline views. I'm not to sure the best fix for this do you have any ideas?

Huongg:
> > hey @SajidAlamQB I think the pipeline bug is fixed now but as you said there's other bug about the selected node is not highlighted and metadata panel is not opened. And even in the default pipeline, the node is not highlighted even the meta data panel opens 🤔
> 
> I'm not able to see the bug with the default pipeline, but it does exist with the other pipeline views. I'm not too sure what the best fix for this is, do you have any ideas?

hey I can still see the issue from my side. I think it might be related to your previous issue. Have you looked at `redirectToSelectedNode` in `FlowchartWrapper`? Perhaps the issue started there. This is where we get the params from the URL and load the node if nodeID existed etc. Perhaps something went missing there? 🤔 happy to jump on the call to do it together if it helps

SajidAlamQB:
> hey @SajidAlamQB I can still see the issue from my side. I think it might be related to your previous issue. Have you looked at `redirectToSelectedNode` in `FlowchartWrapper`? Perhaps the issue started there. 

@Huongg I will make this into a separate issue, its not related to the changes for this PR.

SajidAlamQB:
This PR also fixes two unrelated bugs:

1. **Flowchart View Splitting into Two Trees When `Expand Modular Pipelines` is Enabled**:
When the `Expand Modular Pipelines` button was enabled with `expandAllPipelines=true` in the query string, selecting a node and refreshing the page caused the flowchart view to break, splitting it into two separate trees (see the screenshot below):

image

I believe the cause was that modular pipelines was being expanded unnecessarily, regardless of the `expandAllPipelines` flag. I added an additional condition `!expandAllPipelines` to make sure that modular pipelines only expand when the "Expand All Pipelines" option is disabled:
```
if (hasModularPipeline && !isParameterType && !expandAllPipelines) {
     onToggleModularPipelineExpanded(modularPipeline);
}
```

2. **Node Not Highlighted After Page Refresh in Non-Default Pipeline views**:
The second bug happened when a node was not highlighted after refreshing the page in non-default pipeline views. This happened because the condition checking graph initialisation `graphRef.current === null` didn’t account for page refreshes. To fix this, I used the `PerformanceNavigationTiming` API to detect page reloads and added it as a check in the condition:
```
const navigationEntries = performance.getEntriesByType('navigation');
const isPageRefreshed = navigationEntries.length && navigationEntries[0].type === 'reload';

if (
   (graphRef.current === null || usedNavigationBtn || isInvalidUrl || isPageRefreshed) &&
   !isGraphEmpty
) {
```

Huongg (approved):
it's working from my side, thanks @SajidAlamQB for the changes 👍

SajidAlamQB:
> @SajidAlamQB have you tested with this scenario #1879 ?

I've tested it and it now opens on this page:

image

jitu5 (approved):
Thanks @SajidAlamQB and @rashidakanchwala.

Issue #2042, opened 2024-08-19
Refactor `preview toggle ` feature 


rashidakanchwala:
## Description

This ticket is to refactor the `preview datasets toggle` setting feature to handle logic for previews from FE instead of making BE calls. 

## Context
How has this bug affected you? What were you trying to accomplish?

## Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

## Expected Result
Tell us what should happen.

## Actual Result
Tell us what happens instead.

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

## Your Environment
Include as many relevant details as possible about the environment you experienced the bug in:

* Web browser system and version:
* Operating system and version:
* NodeJS version used (if relevant):
* Kedro version used (if relevant):
* Python version used (if relevant):

## Checklist
- [ ] Include labels so that we can categorise your issue

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro-viz/pull/2067

PR #2067, opened 2024-09-02
Refactor disable preview toggle to be only front end

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/2042

This ticket is to refactor the preview datasets toggle setting feature to handle logic for previews from FE instead of making BE calls.

This feature enables users to control data previews for all nodes from the settings panel of Kedro-Viz when running kedro viz locally.

After running Kedro-Viz locally, dataset previews for all nodes will be enabled by default. However, users will have the ability to enable or disable this feature from the Kedro-Viz settings panel.

## Development notes

We now have a check for the dataset preview toggle in `metadata.js` before rendering the previews.

## Checklist


src/components/metadata/metadata.js
```diff
@@ -266,7 +267,7 @@ const MetaData = ({
                   </>
                 )}
               </dl>
-              {hasPlot && (
+              {hasPlot && showDatasetPreviews && (
```
rashidakanchwala, line 270: maybe you add it just here 

const hasPreview = showDatasetPreviews && metadata?.preview;

rashidakanchwala (approved):
Just one comment otherwise looks good :)

ravi-kumar-pilla:
Hi @SajidAlamQB , Thank you for the refactor. Tested this on Gitpod and it works well. 

I just have 1 question for the flag in the UI. Do we show this flag even on the hosted Kedro-Viz instance ? Like if someone does a `kedro viz deploy` or `kedro viz build` without including preview, will Kedro-Viz has the flag in settings panel ? 

For me the flag does make sense in local as the user cannot disable preview on `kedro viz run`. So the flag is always `True` by default. But for the deployed version,  user can control this behavior using `--include-previews` . The flag will not have any effect if someone did not include preview when building the viz instance. cc: @rashidakanchwala @stephkaiser 

Thank you

SajidAlamQB:
> For me the flag does make sense in local as the user cannot disable preview on `kedro viz run`. So the flag is always `True` by default. But for the deployed version, user can control this behavior using `--include-previews` . The flag will not have any effect if someone did not include preview when building the viz instance. cc: @rashidakanchwala @stephkaiser

Good spot! As it is, the flag will be shown even if the `--include-previews` argument isn't used, meaning the flag might not function as expected without it. I think we can fix this by conditionally displaying the flag in the settings panel only if the `--include-previews` argument was used during deployment. What do you think? Is there any existing flags we do this for that we can take inspiration from?

jitu5 (approved):
Thanks @SajidAlamQB

ravi-kumar-pilla:
> Good spot! As it is, the flag will be shown even if the `--include-previews` argument isn't used, meaning the flag might not function as expected without it. I think we can fix this by conditionally displaying the flag in the settings panel only if the `--include-previews` argument was used during deployment. What do you think? Is there any existing flags we do this for that we can take inspiration from?

One way we can do this is using the viz_metadata_file that is available in the build/deploy artifacts. 

At this moment, it has timestamp and version keys. May be adding a json key like `is_all_previews_enabled` in this file and in the FE reading the json in deployment mode (i.e., not localhost) could be an option. But I am happy to hear from the team. Thank you

rashidakanchwala:
fair points, I would keep this simple for now the way it is. and wait for user to raise this if it's an issue and then try to fix it.

Issue #1712, opened 2024-01-18
Defining and using a custom resolver breaks the "kedro viz run" command


francisduval:
## Description
When using a custom resolver, the pipeline runs normally, but it becomes impossible to use the `kedro viz run` command.

## Context
I want to be able to use a custom resolver while still be able to visualize my pipelines.

## Steps to Reproduce

Define a custom resolver, i.e. put those lines in settings.py (this resolver transforms an activation function given as a string like 'ReLU' into a class object):

```
import torch
custom_resolvers = {'act_resolver': lambda liste: [getattr(torch.nn, element) for element in liste]}
config_loader = OmegaConfigLoader(conf_source='.', custom_resolvers=custom_resolvers)
```

In parameters.yml, use your resolver:

```
grid:
  batch_size: [64]
  optimizer__lr: [0.0001, 0.001]
  module__act: '${act_resolver: [ReLU, Tanh]}'
```

Then, run `kedro viz run`

## Expected Result

Should render the kedro visualization.

## Actual Result

The Kedro visualization does not render, and this error is thrown:

```
ValueError: [TypeError("'builtin_function_or_method' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')].
```

```
ERROR    Exception in ASGI application
Traceback:
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\uvicorn\protocols\http\httptools_impl.py:435 in run_asgi
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\uvicorn\middleware\proxy_headers.py:78 in __call__
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\uvicorn\middleware\message_logger.py:86 in __call__
 C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\uvicorn\middleware\message_logger.py:82 in __call__
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\fastapi\applications.py:1054 in __call__
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\starlette\applications.py:116 in __call__
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\starlette\middleware\errors.py:186 in __call__
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\starlette\middleware\errors.py:164 in __call__
C:\Users\XG04377\.conda\envs\monenv\lib\site-packages\starlette\middleware\exceptions.py:62 in __call__
[...]

PR #2122, opened 2024-10-02
Fix unserializable parameters value

### LINKED ISSUE #2010: Unable to serialize objects returned by custom resolvers in parameters
## Description

Originally raised by Erwin - https://kedro-org.slack.com/archives/C03RKP2LW64/p1722343503669499

## Context

From the user -

I just discovered that if I use a custom resolver in parameters, (using it in catalog have no problem) kedro-viz die. The object returned is not friendly for kedro-viz

"polars": lambda x: getattr(pl, x)

pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'polars.datatypes.classes.DataTypeClass'>
2024-07-30 09:15:55,489 - uvicorn.error - ERROR - Exception in ASGI application

Any workaround? Could be a bad practice to have parameters of type polars.datatypes.classes.DataTypeClass ? 

## Steps to Reproduce

Let's say we have the below resolver -

```
"custom_resolvers": {
        "add": lambda *my_list: sum(my_list),
        "polars": lambda x: getattr(pl, x)
    }
```

If the parameters have a resolver and used as below -

```
model_options:
  test_size: "${add:1,2,3}"
  random_state: 3
  data_info: "${polars:Float64}"
```

Kedro Viz constructs an object like -

```
parameters={
    'model_options': {
      'test_size': 6,
      'random_state': 3,
      'data_info': Float64,
}
```

Now `Float64` is not serializable by pydantic when it is sent as json to frontend. We either need to make str(Float64) or see what to do in case the parameter value is an object.

## Expected Result

Kedro-Viz should handle parameter values of object types and do not crash

## Actual Result

Kedro-Viz throws and error and does not start.

```
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'polars.datatypes.classes.DataTypeClass'>
2024-07-30 09:15:55,489 - uvicorn.error - ERROR - Exception in ASGI application
```

## Your Environment

* Web browser system and version: Chrome
* Operating system and version: macOS
* Kedro version used (if relevant): latest

## Checklist
- [] Include labels so that we can categorise your issue

ravi-kumar-pilla:
## Description

Resolves #2010 

## Development notes

* Added logic to serialize parameter_value in case of complex types
* Updated tests

## QA notes

* All tests should pass
* Consider testing parameters using custom resolver as mentioned in #2010 

## Checklist


noklam:
Add a note here: We discussed in private that this is not a new bug, but never the bug never surface before. We fixed an issue previously for non-primitive type, but this one is related to non-serialisation due to customer resolver, lambda etc

noklam (commented):
The fix looks good to me. Can you post a screenshot how does it looks like for the final result with this change? We should include this in the docs to inform user in case this happens.

rashidakanchwala:
Awesome, i also tried this with #1712 and it works!!

package/kedro_viz/models/flowchart.py
```diff
@@ -9,6 +9,7 @@
 from types import FunctionType
 from typing import Any, ClassVar, Dict, List, Optional, Set, Union, cast
 
+from fastapi.encoders import jsonable_encoder
```
rashidakanchwala, line 12: I wonder in future, if we can replace this with basic encoders. This is for removing reliance on FastAPI dependency when embedded Kedro-viz.

rashidakanchwala (approved):
The FE is a bit odd since we return a string but at least it doesn't break anymore. Maybe we can figure a better way in the future (Juanlu mentioned using getattr) 

Screenshot 2024-10-02 at 12 22 24

SajidAlamQB (approved):
Awesome, thank you @ravi-kumar-pilla!

ravi-kumar-pilla:
> The FE is a bit odd since we return a string but at least it doesn't break anymore. Maybe we can figure a better way in the future (Juanlu mentioned using getattr)

Yes, I was trying to avoid this and experimented different ways (like trying to iterate through the dict and only converting the dict value which is an object). So what I realized is, we can -
1. Try to come up with all different cases (like dict, set, list, custom object etc) and have helper functions which try to serialize them before returning to FE, which might get tricky when the values are combinations of complex types. We might also ask the users to have a __repr__ or __str__ implemented in-case of custom objects.
2. Try to use the jsonable_encoder provided by FASTAPI which tries its best to serialize the value and if not throws an exception.

Though this is not ideal, I could not think of any better way to handle all possible cases. 

@astrojuanlu could you please share any info on how to use `getattr` in resolving this issue (since the name of the attribute can be anything, I am not sure how to use getattr for this scenario)

Thank you

ravi-kumar-pilla:
> The fix looks good to me. Can you post a screenshot how does it looks like for the final result with this change? We should include this in the docs to inform user in case this happens.

For the example where parameters file is -

```yaml
split_options:
  test_size: 0.2
  random_state: 3
  target: 'price'
  data_info: "${polars:Float64}"
```

Custom Resolver as below -

```
CONFIG_LOADER_ARGS = {
      "base_env": "base",
      "default_run_env": "base",
      "custom_resolvers": {
        "add": lambda *my_list: sum(my_list),
        "polars": lambda x: getattr(pl, x)
    }
}
```

UI looks like -

image

Regarding the docs, we don't have much info about the metadata panel apart from stats. I can add a section about the parameter value here - https://docs.kedro.org/projects/kedro-viz/en/stable/kedro-viz_visualisation.html . What do you think ? @noklam 

Thank you

noklam:
image

I get this instead - where are those `__module__` coming from?

ravi-kumar-pilla:
> I get this instead - where are those `__module__` coming from?

image

This is how `jsonable_encoder` works when it encounters custom objects (like pl.Float64), it falls back on encoding the object's attributes (such as its class and module metadata), which is what you're seeing.

noklam:
In this case I think fastapi make this harder to understand as it creates additional key that doesn't exist. What if we don't use fastapi at all? Can you also give an example that using fastapi actually make this better?

ravi-kumar-pilla:
> In this case I think fastapi make this harder to understand as it creates additional key that doesn't exist. What if we don't use fastapi at all? Can you also give an example that using fastapi actually make this better?

I would not say it makes better but it will handle all edge cases that we might miss. But we can also have a custom serializer function something like below which will not return any unexpected value. I was not sure if this handles all the cases.

```python
def serialize_value(value: Any) -> Any: 
    """ Custom serialization function to handle different types of values, 
    including primitives, lists, dictionaries, and objects.""" 
    if isinstance(value, dict): 
        return {k: serialize_value(v) for k, v in value.items()} # Recursively serialize dictionaries 
    elif isinstance(value, list): 
        return [serialize_value(v) for v in value] # Recursively serialize lists 
    elif isinstance(value, (int, float, str, bool, type(None))): # Handle primitives 
        return value 
    # Fallback: convert unknown types to a string 
    return str(value)
```

noklam:
I see, what if we just naively use `str` instead when it's non-serialisable, otherwise just return it as is?

```
self.kedro_obj.load()
```
Do I understand this correctly, originally we only have this part and this will pass to a pydantic model and eventually go through the fastapi json encoder?

ravi-kumar-pilla:
> I see, what if we just naively use `str` instead when it's non-serialisable, otherwise just return it as is?

We can do that as well, but I thought since we have jsonable_encoder doing a better job than naively returning  a string, opted for it. But may be this is creating a bit of confusion. The basic serialization function also seems to handle most of the cases. Do you think we can go with the function I commented ?
 
> Do I understand this correctly, originally we only have this part and this will pass to a pydantic model and eventually go through the fastapi json encoder?

yes that is correct. So for example we have the below Response model and we allow parameters value to be Any. In case of objects, this throws error.

```python
class TaskNodeAPIResponse(BaseGraphNodeAPIResponse):
    parameters: Dict
    ...
 ```

noklam (approved):

Issue #1991, opened 2024-01-23
Unable to run Kedro in Azure with SQLiteSessionStore


Dekermanjian:
## Description
Whenever I try to load the catalog either using a jupyter lab or IPython interface I get the following error:
Failed to instantiate session store of type 'kedro_viz.integrations.kedro.sqlite_store.SQLiteStore'

## Context
I am not able to set up a new project on the cloud (Azure compute instance)

## Steps to Reproduce
1). Create a new kedro project on the cloud
2). Run kedro ipython

## Expected Result
The catalog/context should load like it normally does in local development

## Actual Result
An error message is thrown about a locked SQLAlchemy database with the following error:
Failed to instantiate session store of type 'kedro_viz.integrations.kedro.sqlite_store.SQLiteStore'

```
Failed to instantiate session store of type 'kedro_viz.integrations.kedro.sqlite_store.SQLiteStore'
```

## Your Environment
Brand new conda env with fresh installation of kedro 

* Kedro version used (`pip show kedro` or `kedro -V`): 0.19.2
* Python version used (`python -V`): 3.10.13
* Operating system and version: Cloud compute running linux

gyasibaaye:
Any updates on this issue? Commenting out the lines mentioned below does work, but that shouldn't be necessary to begin

> same here with version 0.19-3 (no pb with 0.18.14) on some cifs drives (linux), solved the problem by commenting in settings.py : from kedro_viz.integrations.kedro.sqlite_store import SQLiteStore # noqa: E402 SESSION_STORE_CLASS = SQLiteStore I guess that's not the right solution but so far....

ravi-kumar-pilla:
Hi , 

Sorry for the delayed response. Thanks for raising the issue. There seems to be multiple processes trying to access SQLite database that is initialized in `settings.py` for Kedro-Viz Experiment Tracking and SQLite allows one writer at a time.

At this moment, I do not have a permanent fix but we will take this as priority in the next sprint. For now, the work-around would be to comment out `SESSION_STORE_CLASS = SQLiteStore` instantiation in `settings.py`, if you are not using Experiment Tracking on Kedro-Viz.

Thank you

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro-viz/pull/2131

PR #2131, opened 2024-10-09
Fix Kedro in Azure ML with SQLiteSessionStore 

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/1991

This PR addresses issues with SQLite database locking in Azure Machine Learning (AML).

Reading online other users have also had this problem. It stems from the fact that in Azure ML the `/home` directory is mounted as CIFS filesystem which can't deal with SQLite3 lock. Read more here: https://stackoverflow.com/a/54051016

To resolve this, we can enable [Write-Ahead Logging (WAL)](https://sqlite.org/wal.html) mode to fix these locking issues. 

We can either:

1. **Adopt WAL for all environments**: If we go with this option the `session_store.db` file will have a small change in its internal structure to indicate that it is now using WAL mode. This change is stored in the database's header.

2. **Enable WAL mode conditionally for Azure ML:** So we activate WAL mode only when the code detects that it is running in an AML environment. (Currently implemented in this PR).

## Development notes

- Included `PRAGMA journal_mode=WAL;` to enable `WAL` mode.

- Implemented a condition to detect when it is running in an Azure Machine Learning environment using environment variables specific to Azure:`AZUREML_ARM_SUBSCRIPTION`, `AZUREML_ARM_RESOURCEGROUP`, or `AZUREML_RUN_ID`.

Provided two options:

1. Apply WAL mode universally across all environments.
2. Apply WAL mode only when the AML environment is detected, using the `is_azure_ml` flag.

## QA notes

I tested this in Azure Machine Learning studio and confirmed that the database operates in WAL mode without the locking issues.

Tested locally as well it works as expected with and without enabling WAL mode.

image

## Checklist


package/kedro_viz/database.py
```diff
@@ -10,9 +12,27 @@ def make_db_session_factory(session_store_location: str) -> sessionmaker:
     """SQLAlchemy connection to a SQLite DB"""
     database_url = f"sqlite:///{session_store_location}"
     engine = create_engine(database_url, connect_args={"check_same_thread": False})
-    session_class = sessionmaker(engine)
     # TODO: making db session factory shouldn't depend on models.
     # So want to move the table creation elsewhere ideally.
     # But this means returning engine as well as session class.
+
+    # Check if we are running in an Azure ML environment.
+    is_azure_ml = any(
```
rashidakanchwala, line 20: Looks good :) can we move the azure logic in another function apply_azure_ml_wal_mode ?

SajidAlamQB: That makes sense!

rashidakanchwala (approved):
LGTM. thanks

jitu5 (approved):

Huongg (approved):
thanks @SajidAlamQB , LGTM too 👍

PR #2146, opened 2024-10-22
Fix FE when `tag` name is `undefined`

### LINKED ISSUE #2106: Pipelines rendering breaks with exception, ver. 9.2.0, 10.0, yet works on 9.1.0
## Description
Slack discussion: https://kedro.hall.community/support-lY6wDVhxGXNY/kedroviz-crashes-randomly-with-javascript-type-error-after-adding-changes-to-pipeline-registry-yzcmqg0kbQqS

What first triggered this issue was a beningn attempt to reorder the pipelines (whole diff between commits that broke the kedro-viz in our project):

```diff
@@ -79,7 +79,16 @@ def register_pipelines() -> dict[str, Pipeline]:
             group_class, input_schema, pipelines[f"training_{group_class}"], pipelines[f"inference_{group_class}"]
         )
 
-    pipelines = {k: pipelines[k] for k in sorted(pipelines.keys())}  # sorting for kedro-viz selection box
-    pipelines["__default__"] = sum(pipelines[f"training_and_inference_{gc}"] for gc in GROUP_CLASSES)  # type: ignore

+    # pipelines = {k: pipelines[k] for k in sorted(pipelines.keys())}  # sorting for kedro-viz selection box
+    order = ["input", "data", "feature", "fgu", "layer", "pricing", "inference", "training"]
+    pipelines = {
+        k: pipelines[k]
+        for k in sorted(
+            pipelines.keys(),
+            key=lambda x: str(order.index(x.split("_")[0]) if x.split("_")[0] in order else len(order)) + x,
+        )
+    } # reordering pipelines in kedro-viz based on order list
+    pipelines["__default__"] = pipelines["input_na"]
```
^ this was in version 9.2.0

Currently we reversed to version 9.1.0 to keep working and then tried to updating to 10.0. It was working for a few minutes and without any further changes it started consistently crashing on start with the error posted at the very end of this report.

@ edit: after further investigation with dev kedro-viz, we found out that it's a separate issue at 10.0 and it's related to one tag not getting registered/having undefined "name" in js code. The kedro-viz breaks at the pipeline tagged with it and after refreshes/restarts automatically remembers the path to broken page and reloads that one, that's why it seemed to be okay and broke at some point. When manually editing the address bar we can make restore it from crash loop or when we comment out the offending tag from pipeline. We're not sure what separates this tag from others that it behaves differently - it's notable that it does not list in main.tags list of registered tags in generated page's JS.

## Context
Makes it unable to work with kedro-viz in newer versions.

## Steps to Reproduce

Not well known. In bigger project the kedro-viz just breaks at some random pipelines or at start with exceptions - difficult to track what was the origin/tipping point for the error to occur.

## Expected Result
Kedro viz working, stable as in 9.1.0.

## Actual Result
Kedro viz unusable.

In 9.2.0 the error of javascript when browsing was typerror exception
```
Uncaught TypeError: e is undefined
    Ri http://127.0.0.1:4141/static/js/main.f9e37320.js:2
    aA/r[n]< http://127.0.0.1:4141/static/js/main.f9e37320.js:2
    aA http://127.0.0.1:4141/static/js/main.f9e37320.js:2
    uA http://127.0.0.1:4141/static/js/main.f9e37320.js:2
    p http://127.0.0.1:4141/static/js/main.f9e37320.js:2
```

in 10.0, the error seems different, may be a different cause
```
main.a2a59735.js:2 
 TypeError: Cannot read properties of undefined (reading 'match')
    at Hi (main.a2a59735.js:2:4364976)
    at main.a2a59735.js:2:4848942
    at Array.map (<anonymous>)
    at LI (main.a2a59735.js:2:4848910)
    at main.a2a59735.js:2:4849263
    at main.a2a59735.js:2:4351469
    at c (main.a2a59735.js:2:4350427)
    at main.a2a59735.js:2:4351602
    at c (main.a2a59735.js:2:4350427)
    at main.a2a59735.js:2:4351568
ds	@	main.a2a59735.js:2
r.callback	@	main.a2a59735.js:2
```

## Your Environment
python 3.10, kedro 0.19.8, kedro viz tested 9.1-10.0, hooks enabled on kedro-viz loading (tried with and without),
OS: macos / should not be relevant
browser: firefox 130.0, edge 1.2.9

## Checklist

rashidakanchwala:
## Description

Partly resolves #2106 

## Development notes

In the user's pipeline, one of the nodes was tagged as `validation`. While this tag was mentioned in the `{ node: { tags: [] } }` response, it wasn't somehow mentioned in the `{ tags: {} }` response which caused the flowchart to break. We were unable to reproduce this issue on the backend, which would have been the optimal place to address it.

As a workaround, we fixed this on the frontend. When a tag is not defined  correctly in the backend, the frontend cannot link the tag to the corresponding tag name. Since the tag name and ID are the same, we added a fallback. If the tag name is undefined (due to an error we still need to investigate), we use the tag ID itself as a fallback. This prevents Kedro-Viz from breaking and ensures the flowchart runs correctly.

## QA notes

To test this fix using the demo project:

Run `kedro viz --save-file=test`. This will create a test folder with an api folder inside.
Move the `api` folder into the `kedro-viz/public` directory.
Remove 'company' dict from `tags:` in `api/main` folder
Run `npm run start:dev`.
Before this PR, you will see the app breaking with the error message the user encountered. After applying the PR, the issue will be fixed.

## Checklist

### DIFF
--- a/src/selectors/tags.js
+++ b/src/selectors/tags.js
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect';
 import { getPipelineTagIDs } from './pipeline';
+import { prettifyName } from '../utils';
 
 const getNodeTags = (state) => state.node.tags;
 const getTagName = (state) => state.tag.name;
@@ -14,7 +15,7 @@ export const getTagData = createSelector(
   (tagIDs, tagName, tagActive, tagEnabled) =>
     tagIDs.sort().map((id) => ({
       id,
-      name: tagName[id] || id,
+      name: tagName[id] || prettifyName(id),
       active: Boolean(tagActive[id]),
       enabled: Boolean(tagEnabled[id]),
     }))
--- a/src/selectors/tags.js
+++ b/src/selectors/tags.js
@@ -14,7 +14,7 @@ export const getTagData = createSelector(
   (tagIDs, tagName, tagActive, tagEnabled) =>
     tagIDs.sort().map((id) => ({
       id,
-      name: tagName[id],
+      name: tagName[id] || id,
       active: Boolean(tagActive[id]),
       enabled: Boolean(tagEnabled[id]),
     }))
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -120,7 +120,7 @@ export const stripNamespace = (str) => {
  * @param {String} str The string to check
  * @returns {String} The string with or without replaced values
  */
-export const prettifyName = (str) => {
+export const prettifyName = (str = '') => {
   const replacedString = str
     .replace(/-/g, ' ')
     .replace(/_/g, ' ')
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -120,7 +120,10 @@ export const stripNamespace = (str) => {
  * @param {String} str The string to check
  * @returns {String} The string with or without replaced values
  */
-export const prettifyName = (str = '') => {
+export const prettifyName = (str) => {
+  if (!str) {
+    return '';
+  }
   const replacedString = str
     .replace(/-/g, ' ')
     .replace(/_/g, ' ')


jitu5 (approved):

rashidakanchwala:
Kacper confirmed that this PR resolves the main issue i.e it doesn't break Kedro-viz; the outputs of pipelines are not entirely correct. and when he gets time he will create some example pipelines that we can use to reproduce and fix this bug. 

FYI @astrojuanlu

src/selectors/tags.js
```diff
@@ -14,7 +15,7 @@ export const getTagData = createSelector(
   (tagIDs, tagName, tagActive, tagEnabled) =>
     tagIDs.sort().map((id) => ({
       id,
-      name: tagName[id],
+      name: tagName[id] || prettifyName(id),
```
ravi-kumar-pilla, line 18: This looks good. I am not sure if there would be a case where the id is undefined ? In that case prettifyName(id) would error out.

rashidakanchwala: it's a bit weird. but id is read from node tags. but the user said the filtering wasn't correct at their end. so this works is a good fallback, but not entirely correct.

ravi-kumar-pilla: just to be sure can we add below check in prettifyName ? -
 
```python
export const prettifyName = (str) => {
  if (!str) {
    return ''; 
  }
  ...
```

rashidakanchwala: I'd like us to address this issue in the backend since you were able to reproduce it rather than now implementing this frontend fix, as it may hide the underlying backend error. Although I understand that breaking Kedro-Viz isn't ideal :)

jitu5: If you see in the screenshot in the `tag` array in validation object `id` is defined but good point we can add check in  `prettifyName `.

image (10)

ravi-kumar-pilla: > I'd like us to address this issue in the backend since you were able to reproduce it rather than now implementing this frontend fix, as it may hide the underlying backend error. Although I understand that breaking Kedro-Viz isn't ideal :)

Yes, I agree. We should do both as we dont want to break Viz even if the backend errors. We can go ahead with this fix or introduce ErrorBoundary in frontend. For now I feel FE fix should be merged, but resolving the root cause and returning the tags would be handled in the backend fix

ravi-kumar-pilla: > If you see in the screenshot in the `tag` array in validation object `id` is defined but good point we can add check in `prettifyName `.

Yup for this case it works. In general I think prettifyName should handle the edge case. Not related to this PR completely but since we are using the helper, it would be nice to add the check just to be sure.

src/utils/index.js
```diff
@@ -120,7 +120,7 @@ export const stripNamespace = (str) => {
  * @param {String} str The string to check
  * @returns {String} The string with or without replaced values
  */
-export const prettifyName = (str) => {
+export const prettifyName = (str = '') => {
```
ravi-kumar-pilla, line 123: The default value will only handle undefined, it will not handle cases of `null`. Though prettifyName will never be called with null, I would still suggest using an explicit check like below to handle all weird cases -
```python
export const prettifyName = (str) => {
  if (!str) {
    return ''; 
  }
  ...
```

SajidAlamQB (approved):
Thanks @rashidakanchwala!

PR #2161, opened 2024-10-29
Introduce `behaviour` prop object with `reFocus` prop

jitu5:
## Description

Resolves https://github.com/kedro-org/vscode-kedro/issues/126

This PR Introduce `behaviour` prop object with `reFocus` prop. When `behaviour.reFocus` is set to `false` in embedded mode and when user clicks on node it will not re-focus.

```
behaviour: { 
     reFocus: false, // default is true
}
```

### `FlowChart` component:

* [`src/components/flowchart/flowchart.js`](diffhunk://#diff-6a862e303b98da65fd66635d0e9bef64b7498a4f0d9fdb1b59a1620d2ee20826L223-L237): Refactored the conditions for redrawing nodes and edges to prevent unnecessary zooming when the metadata sidebar or code panel visibility changes, or when the clicked node changes without node refocusing.

### State management updates:

* [`src/reducers/index.js`](diffhunk://#diff-3527887dbf404f0c28be291b83274e5f361cab6b0e6e0bd5ed06f66dbd11c5b0R92): Added a new reducer for `behaviour` to manage the `reFocus` state.

* [`src/store/initial-state.js`](diffhunk://#diff-8aec283ba10f2d8f1aa71cf147f0b016e3491e28087e437043e921d4956b1da9R61-R63): Initialized the `behaviour` state with a default `reFocus` value of `true`.

### QA Notes:
You can test new options prop in `src/components/container.js` as `<App />` is the entry point or top level component for a standalone use case.

```
const Container = () => (
  <>
    <App
      options={{
        display: {
          globalNavigation: false,
          sidebar: false,
          metadataPanel: false,
        },
        behaviour: {
          reFocus: false,
        },
      }}
      data={getPipelineData()}
    />
  </>
);
```

## Checklist


rashidakanchwala (approved):
lgtm! thanks

src/reducers/index.js
```diff
@@ -89,6 +89,7 @@ const combinedReducer = combineReducers({
   // These props don't have any actions associated with them
   display: createReducer(null),
   dataSource: createReducer(null),
+  modeOptions: createReducer({}),
```
ravi-kumar-pilla, line 92: [Nit] modeOptions inside options seems a bit confusing to me. Was there any consideration of having reFocus inside display ? If not, was there any reason to separate this ? 
What does mode mean here ?

ravi-kumar-pilla: if we expand on how flowchart behaves, something like `behavior` or something along those lines might give a better idea of what the key value is. For me `modeOptions` does not seem fitting

src/components/flowchart/flowchart.js
```diff
@@ -220,22 +220,28 @@ export class FlowChart extends Component {
 
     if (changed('edges', 'nodes', 'layers', 'chartSize', 'clickedNode')) {
       // Don't zoom out when the metadata or code panels are opened or closed
-      if (prevProps.visibleMetaSidebar !== this.props.visibleMetaSidebar) {
+      const metaSidebarChanged =
```
ravi-kumar-pilla, line 223: [Nit] if this is to track visibility of sidebar then `const metaSidebarViewChanged` or `const metaSidebarVisibilityChanged` might fit

src/components/flowchart/flowchart.js
```diff
@@ -220,22 +220,28 @@ export class FlowChart extends Component {
...
+        prevProps.visibleMetaSidebar !== this.props.visibleMetaSidebar;
+
+      const codeChangedWithoutMetaSidebar =
```
ravi-kumar-pilla, line 226: [Nit] `const codeViewChangedWithoutMetaSidebar`

src/components/flowchart/flowchart.js
```diff
@@ -220,22 +220,28 @@ export class FlowChart extends Component {
 
     if (changed('edges', 'nodes', 'layers', 'chartSize', 'clickedNode')) {
       // Don't zoom out when the metadata or code panels are opened or closed
```
ravi-kumar-pilla, line 222: is this PR to not refocus flowchart due to any change or just the node click ? 

if it is not to refocus in anyway may be a check here like below should work without any other code change I guess ?

```javascript
if (!this.props.nodeRefocus) {
return;
}
```

ravi-kumar-pilla:
Hi @jitu5 , Thanks for the PR. I left some comments around naming but the functionality looks good. Thank you

src/reducers/index.js
jitu5: @ravi-kumar-pilla 
Please refer https://github.com/kedro-org/vscode-kedro/issues/126#issuecomment-2435157272

Huongg: hey since the stand-alone version primarily covers the flowchart, we might not need to specify 'flowchart.' Maybe something more general like `behaviour`, `interaction`, or `interactionMode` would work?

src/components/flowchart/flowchart.js
jitu5: @ravi-kumar-pilla I started with this approach, but it breaks the flowchart on the first load because `this.resetView` does not execute, as we are returning early. Here, `this.resetView` is responsible for centering and refocusing the flowchart.

Huongg (commented):
I tested it and it's all working fine for me. I agree with @ravi-kumar-pilla on the `modeFocus` name as it can be a bit confusing when you read it. Perhaps something about `behaviour` or `interaction` would work well in this context @jitu5 ?

src/components/flowchart/flowchart.js
```diff
@@ -220,22 +220,28 @@ export class FlowChart extends Component {
 
     if (changed('edges', 'nodes', 'layers', 'chartSize', 'clickedNode')) {
```
Huongg, line 221: Hey, it looks like this section of code is for different beahviour of the stand-alone app. Perhaps having a comment here could help clarify the purpose, so we'll have better context if we refactor this component in the future. What do you think?

jitu5:
> I tested it and it's all working fine for me. I agree with @ravi-kumar-pilla on the `modeFocus` name as it can be a bit confusing when you read it. Perhaps something about `behaviour` or `interaction` would work well in this context @jitu5 ?

@Huongg 

> `modeFocus`

 You mean `modeOptions` ? To avoid confusion we can think `reFocus` is one of the behaviours.
 
 ```
 behaviour: { 
     reFocus: false, // default is true
}
 ``` 
 
 @ravi-kumar-pilla @rashidakanchwala wdyt ?

src/components/flowchart/flowchart.js
jitu5: @Huongg Not exactly, those condition is also being used in normal mode, but yes refactoring `flowchart.js` component will really help overall.

ravi-kumar-pilla:
> ```
> behaviour: { 
>     reFocus: false, // default is true
> }
> ```

Since this is a user facing key (though we document it) this should be very clear on what it does. For me behavior seems very generic. `flowchartBehavior` seems more appropriate. Considering if we expand on other behaviors, having a behavior key might introduce additional nesting. But I am fine with either. Thank you

rashidakanchwala:
> > ```
> > behaviour: { 
> >     reFocus: false, // default is true
> > }
> > ```
> 
> Since this is a user facing key (though we document it) this should be very clear on what it does. For me behavior seems very generic. `flowchartBehavior` seems more appropriate. Considering if we expand on other behaviors, having a behavior key might introduce additional nesting. But I am fine with either. Thank you

nesting won't be a problem as we do deepmerge anyway right, @jitu5 ?

jitu5:
> > > ```
> > > behaviour: { 
> > >     reFocus: false, // default is true
> > > }
> > > ```
> > 
> > 
> > Since this is a user facing key (though we document it) this should be very clear on what it does. For me behavior seems very generic. `flowchartBehavior` seems more appropriate. Considering if we expand on other behaviors, having a behavior key might introduce additional nesting. But I am fine with either. Thank you
> 
> nesting won't be a problem as we do deepmerge anyway right, @jitu5 ?

@rashidakanchwala Yes, we do deepmerge.

ravi-kumar-pilla (approved):
Works well !! Thank you @jitu5

Issue #2052, opened 2024-08-22
Kedro-Viz mistakenly uses existing service


astrojuanlu:
## Description
While reviewing https://github.com/kedro-org/kedro-viz/pull/1966, I found something funny: if there is an existing Kedro Viz server running, `kedro viz run` will open that one.

## Context
I was trying to evaluate the behavior of `kedro viz run --lite`, and the result was that I was getting redirected to the non-lite version.

## Steps to Reproduce

1. Launch `kedro viz run` in one terminal
2. Launch `kedro viz run` in another terminal

## Expected Result
Either
- the second `kedro viz run` complains that the port is in use, or
- the second `kedro viz run` gets launched in another port

## Actual Result
`kedro viz run` redirects me to the existing server.

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro-viz/pull/2176

PR #2176, opened 2024-11-06
Add port in use check to prevent browser redirecting incorrectly for kedro viz

SajidAlamQB:
## Description
Related to: https://github.com/kedro-org/kedro-viz/issues/2052

If there is an existing Kedro Viz server running, kedro viz run will redirect the browser to that one when a second instance of kedro viz is launched using the same port and the error message from uvicorn is missed.

## Development notes

Check that the port is in use before running the server and increment it until a max 5 times to find an alternative free port, if none are found send error message to users no free ports available and suggest them to use `--port`.

## Checklist

### DIFF
--- a/package/kedro_viz/launchers/cli/run.py
+++ b/package/kedro_viz/launchers/cli/run.py
@@ -189,7 +189,7 @@ def run(
                 target=run_server, daemon=False, kwargs={**run_server_kwargs}
             )
 
-        display_cli_message(f"Starting Kedro Viz on port {port}...", "green")
+        display_cli_message("Starting Kedro Viz ...", "green")
         viz_process.start()
 
         _VIZ_PROCESSES[port] = viz_process
--- a/package/kedro_viz/launchers/cli/run.py
+++ b/package/kedro_viz/launchers/cli/run.py
@@ -116,6 +116,7 @@ def run(
         _PYPROJECT,
         _check_viz_up,
         _find_kedro_project,
+        _is_port_in_use,
         _start_browser,
         _wait_for,
         display_cli_message,
@@ -145,6 +146,18 @@ def run(
             "https://github.com/kedro-org/kedro-viz/releases.",
             "yellow",
         )
+
+    # Check if the port is already in use
+    if _is_port_in_use(host, port):
+        display_cli_message(
+            f"Error: Port {port} is already in use. Kedro Viz could not start.",
+            "red",
+        )
+        display_cli_message(
+            "Please specify a different port using the '--port' option.",
+            "red",
+        )
+        return
     try:
         if port in _VIZ_PROCESSES and _VIZ_PROCESSES[port].is_alive():
             _VIZ_PROCESSES[port].terminate()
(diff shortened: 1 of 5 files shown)


rashidakanchwala:
maybe we put a specific range from 4141, to 4145, and if all the ports in this range are busy, then they tell us which one through --port.

SajidAlamQB:
> maybe we put a specific range from 4141, to 4145, and if all the ports in this range are busy, then they tell us which one through --port.

This is a good middle approach we definitely can do.

rashidakanchwala:
@SajidAlamQB , thanks for this , will review this once tests are passing, can we also add unit-testing to this, if we can mock ports being occupied ?

package/kedro_viz/launchers/cli/run.py
```diff
@@ -145,6 +146,9 @@ def run(
             "https://github.com/kedro-org/kedro-viz/releases.",
             "yellow",
         )
+
+    port = _find_available_port(host, port)
```
ravi-kumar-pilla, line 150: This looks good but should we also investigate and fix shutting down of previous viz process efficiently once the user exits ?

rashidakanchwala: Good one, but that maybe out of scope for this particular ticket. and we will also need to think what would be the ideal user journey. Should we prompt them to kill the process, should we ask them for consent to kill the process etc. 

@ravi-kumar-pilla , could u create another issue
 for this?

ravi-kumar-pilla: > Good one, but that maybe out of scope for this particular ticket. and we will also need to think what would be the ideal user journey. Should we prompt them to kill the process, should we ask them for consent to kill the process etc.

Not about the user flow, I was saying may be we need to know the root cause of why a kedro viz process was hanging and we need to shutdown kedro viz gracefully (i.e., when someone did ctrl+c, make sure the process is shut down, this also includes kedro viz process started via jupyter notebook). Yes I can create a separate ticket but I thought it is related to #2052

rashidakanchwala: n the #2052  case described by Juanlu, he was probably trying to compare kedro-viz and kedro-viz --lite , so he wasn't terminating the previous process.

SajidAlamQB: In Juanlu's case he tried to start two viz processes and didn't exist the first instance, the second instance would fail to start on same port but still opened the browser and redirected to the first instance.

ravi-kumar-pilla: got it, thank you @rashidakanchwala @SajidAlamQB

astrojuanlu: > Should we prompt them to kill the process, should we ask them for consent to kill the process etc.

Notice that the process keeping the port busy can be anything. We shouldn't meddle with that, either say "sorry, that port isn't available, use another one" or choose another one ourselves (either at random or incrementing by +1 as I think this PR does)

rashidakanchwala: I completely agree that we must not interfere with the user's machine. So, ideally it would be informing them, and letting them close it themselves which this PR does.

ravi-kumar-pilla (approved):
Works well and LGTM. Thank you @SajidAlamQB 💯

rashidakanchwala (approved):
Awesome, thanks @SajidAlamQB

PR #2206, opened 2024-11-26
Fix `kedro viz --load-file` to run from any directory without requiring a Kedro project

### LINKED ISSUE #2045: `kedro viz --load-file` looks for project configurations.
## Description

`kedro-viz --load-file` looks for project configurations. Ideally it should be able to run directly using the json regardless if the Kedro project exists or not. 

Raised by @noklam 

## Steps to Reproduce

1. run `kedro viz --save-file=file` from demo-project directory
2. run `kedro viz --load-file=file` from demo-project directory 
3. now go to parent directory and run `kedro viz --load-file = demo-project/file' and it will give you below error 

Screenshot 2024-08-19 at 13 29 54

## Expected Result
It shouldn't throw that bug. 

## Actual Result
Tell us what happens instead.

```
```

```
```

## Your Environment

## Checklist

rashidakanchwala:
## Description

This PR fixes #2045.  

Previously, `kedro viz --load-file` checked if the command was run from within a Kedro project directory, which is not required. The fix ensures that `kedro viz --load-file` can run from anywhere as long as it has the correct path the API data directory . 

The key behavior is now as follows:

1. If the specified filepath does not exist, a ValueError is raised:
```
ValueError: The provided filepath 'hellod' does not exist.
```

2. If the filepath exists but does not contain the expected directory structure (api/main, api/pipelines/, and api/nodes/), a FileNotFoundError is raised:

`FileNotFoundError: [Errno 2] No such file or directory: '/path/to/xyz/api/main'`

3. If the filepath exists and has the required api directory structure and conent, `kedro viz --load-file` runs successfully using the API JSON data.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,15 @@ Please follow the established format:
 - Include the ID number for the related PR (or PRs) in parentheses
 -->
 
+# Release 10.2.0 
+
+
+## Major features and improvements
+
+## Bug fixes and other changes
+
+- Fix kedro viz `--load-file` to run from any directory without requiring a Kedro project. (#2206)
+
 # Release 10.1.0
 
 ## Major features and improvements
@@ -17,6 +26,7 @@ Please follow the established format:
 
 
 ## Bug fixes and other changes
+
 - Fix tag being undefined when pipeline are ordered differently (#2162, #2146)
 - Fix unserializable parameters value. (#2122)
 - Update kedro-viz lite banner icon and message. (#2196)


SajidAlamQB (approved):
LGTM, thank you @rashidakanchwala!

package/kedro_viz/launchers/cli/run.py
```diff
@@ -123,16 +123,21 @@ def run(
...
-
-    if kedro_project_path is None:
-        display_cli_message(
-            "ERROR: Failed to start Kedro-Viz : "
-            "Could not find the project configuration "
-            f"file '{_PYPROJECT}' at '{Path.cwd()}'. ",
-            "red",
-        )
-        return
+    kedro_project_path = None
```
Huongg, line 126: maybe this `kedro_project_path` does not need to be defined outside the if-else statement if it is only used within the else block?

rashidakanchwala: it is needed later, it is passed to run_server

Huongg (approved):
I tested it locally and its working as described in the ticket, thanks @rashidakanchwala

ravi-kumar-pilla:
Hi @rashidakanchwala , The PR looks good. Something not related to the ticket but if possible, can we update the docs [here](https://docs.kedro.org/projects/kedro-viz/en/stable/cli-docs.html#kedro-viz)

image

and [here](https://docs.kedro.org/projects/kedro-viz/en/stable/kedro-viz_visualisation.html#optional-arguments) with the directory structure that is needed to run viz when the user does `kedro viz --load-file` ?

Thank you

ravi-kumar-pilla (approved):
Left a doc comment but the PR works well... Thank you @rashidakanchwala

noklam (approved):

docs/source/cli-docs.md
```diff
@@ -44,10 +44,10 @@ kedro viz run [OPTIONS]
   - Whether to open the Kedro Viz interface in the default browser. The browser will open if the host is `localhost`. Defaults to `True`.
 
 - `--load-file <path>`
-  - Path to load Kedro Viz data from a directory. If provided, Kedro Viz will load the visualisation data from this path instead of generating it from the pipeline.
+  - Path to load Kedro Viz data from a [directory](https://docs.kedro.org/projects/kedro-viz/en/latest/cli-docs.html#kedro-viz-directory-structure-when-you-save-it-as-a-file). If provided, Kedro Viz will load the visualisation data from this path instead of generating it from the pipeline
```
astrojuanlu, line 47: These absolute links won't work, and also they introduce an inconsistency (they use `/en/latest` but maybe the user was browsing `/en/stable` or `/en/10.1.0`), we should use relative links

PR #2179, opened 2024-11-07
Handle Github pages 404 issue

jitu5:
## Description

Resolves https://github.com/kedro-org/kedro-viz/issues/2048

This pull request includes changes to fix a 404 error on the experiment tracking page, and to ensure the 404 page is correctly redirect to root and manage by React router internally by keeping the content of the index.html and 404.html same.

## Development notes

Redirect handling:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L16-R16): Modified the build script to copy `index.html` to `404.html` to ensure the 404 page is correctly served.

## QA notes

I have deployed the changes to Github pages https://jitu5.github.io to test.

## Checklist

- [X] Read the [contributing](/CONTRIBUTING.md) guidelines
- [X] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #2212, opened 2024-11-29
[Frontend] Remove experiment tracking

### LINKED ISSUE #2207: Remove ET front-end components
As a part of #2202, this ticket is to remove the FE components. This includes the UI, e2e tests, and Apollo queries.

tynandebold:
## Description

Resolves #2207

## Development notes

Removed all frontend references to experiment tracking.

Still to discuss if we want to:

- [ ] Remove all routing and `react-router-dom`
- [x] Remove the `sanitizedPathname()` function

## QA notes

Test the app locally. Everything should be the same, just without experiment tracking.

## Checklist


Huongg:
Hey, thanks for putting together this PR @tynandebold ! I'll take some time to review it this morning since it's still a draft. Here are my thoughts on the two points you mentioned:

- sanitizedPathname: I think we can remove this since we no longer need to switch between two routes, and it doesn’t seem like it would cause any breaking changes.
- react-router-dom: My initial thought is yes, we could remove it since we don't have multiple routes anymore. However, we still use some features from it, like `matchPath` and `useHistory`for dynamic URLs. Removing it might require refactoring for the URL handling. My gut feeling is that we should treat this separately and not include it in the current scope.

Let me know you thoughts 😄

Huongg:
Screenshot 2024-12-02 at 10 18 07

Also a question for @stephkaiser, since we're removing experiment tracking from the navigation bar, does it still make sense to highlight the flowchart?

Huongg:
hey overall the changes look good to me, i tested it locally and everything seems okay, apart from `/experiment-tracking` is still loading and not throwing error. I'll do another scan again after I've done with my current PR. Thank you very much @tynandebold

jitu5:
@tynandebold I have reviewed it and tested locally as well and all looks good. Just few note, 
1. If we hit `/experiment-tracking` url directly as Huong mentioned its not throwing error. 
2. `sanitizedPathname` is still there in `utils`, if we are planning to remove it.
3. Recent [fix](https://github.com/kedro-org/kedro-viz/pull/2179) related to 404 Github pages issue, we need to revert that as well.

.gitignore
```diff
@@ -65,3 +65,7 @@ coverage.xml
 
 # Kedro
 *.log
+
+# Python
+.venv/
+mise.toml
```
tynandebold, line 71: I've added this because I'm using [mise](https://mise.jdx.dev/lang/python.html) to manage my Python environment, not conda.

src/store/initial-state.js
```diff
@@ -167,18 +165,17 @@ const applyUrlParametersToNonPipelineState = (state, urlParams) => {
...
       delete localStorageState[key];
     }
   });
+
   const allLocalStorageState = {
     ...localStorageState,
-    ...{ runsMetadata: localStorageRunsMetadataState },
   };
+
   return deepmerge(state, allLocalStorageState);
```
Huongg, line 179: NIT: maybe since there is only one localStorage now, you can remove `allLocalStorageState` and pass `localStorageState` directly in the deepmerge?

Huongg (approved):
Hey thank you @tynandebold, amazing work and super quick turnaround! Everything looks great to me except for one small NIT point. It's all running okay on my local setup as well. As we discussed, `/experiment-tracking` doesn’t throw any errors on localhost. We’ll revisit this once I complete the backend removal, and we can re-test it on the feature_branch. I'll leave the note to remind us to test it in the feature_branch

jitu5 (approved):
Thanks @tynandebold Amazing...!

.gitignore
astrojuanlu: I was a mise user as well for a while! Until I discovered https://docs.astral.sh/uv/ 😄

astrojuanlu:
Thank you @tynandebold !

We received feedback from some team members that the decision to proceed with the removal was a bit opaque. So in light of the importance of this step, I think the right call is to request approval from the TSC. It will delay the PR a bit but does everyone agree?

jitu5:
> Thank you @tynandebold !
> 
> We received feedback from some team members that the decision to proceed with the removal was a bit opaque. So in light of the importance of this step, I think the right call is to request approval from the TSC. It will delay the PR a bit but does everyone agree?

@astrojuanlu Make sense to include TSC.

Huongg:
>  I think the right call is to request approval from the TSC. It will delay the PR a bit but does everyone agree?
hey @astrojuanlu That's a good idea to get approval from the TSC, the question is there will be at least 3-5 more PRs as a part of this work. Do we add them in every PR? Or should we add them to the feature_branch which will have the final changes both back-end, front-end, docs etc.

tynandebold:
> > I think the right call is to request approval from the TSC. It will delay the PR a bit but does everyone agree?
> 
> hey @astrojuanlu That's a good idea to get approval from the TSC, the question is there will be at least 3-5 more PRs as a part of this work. Do we add them in every PR? Or should we add them to the feature_branch which will have the final changes both back-end, front-end, docs etc.

A TSC vote makes sense and so does your point, Huong. 

I think the TSC approval should be on the [feature_branch/remove-experiment-tracking](https://github.com/kedro-org/kedro-viz/tree/feature_branch/remove-experiment-tracking) branch, which will be the final, all-encompassing PR.

src/store/initial-state.js
tynandebold: Done!

SajidAlamQB (approved):
Thanks for this, I've tested it manually and all seems good!

astrojuanlu:
> I think the TSC approval should be on the [feature_branch/remove-experiment-tracking](https://github.com/kedro-org/kedro-viz/tree/feature_branch/remove-experiment-tracking) branch, which will be the final, all-encompassing PR.

Sounds good, thanks 👍🏼 please keep it in mind before submitting the final PR to `main`

PR #2267, opened 2025-02-03
Fix %run_viz using old process in jupyter notebook

### LINKED ISSUE #1823: %run_viz Jupyter line magic inconsistency
## Description

Tried executing %run_viz line magic on Jupyter notebook to test `--inlcude-hooks` option, while the command gets passed appropriately, Open Viz link appears even before considering the flag. This shows, we are not terminating the previous session.

This inconsistency was also detected during https://github.com/kedro-org/kedro-viz/pull/1733

Image

The image above is from a debug session using print statements placed at -
1. run_viz function at `kedro_viz/launchers/jupyter.py` line 120
2. checked if viz is up for the second run by placing a print statement at `kedro_viz/launchers/jupyter.py` line 133
3. placed a print statement in load_data function at `kedro_viz/integrations/kedro/data_loader.py` line 99

## Context

I am not an expert in using Jupyter notebook but when I execute `%run_viz` line magic with differing options, each execution box should take in the options before restarting the viz server. Currently, we check if the viz is up and we show the Open Kedro Viz link which has a stale session. 

## Steps to Reproduce

1. Setup Kedro Viz and have a Kedro Project
2. Execute `kedro jupyter notebook`
3. Create a notebook and execute `%run_viz`
4. It takes some time and gives you a link to open kedro viz
5. Try executing `%run_viz --ignore-hooks` or any other option and debug it using print statements
6. The link to open kedro viz appears instantly, without restarting the server with new arguments

## Actual Result

Image

## Your Environment

* Web browser system and version: Chrome/Safari
* Operating system and version: macOS

## Checklist

ravi-kumar-pilla:
## Description

Resolves #1823 

**NOTE:** I increased the max attempts of ports to 10 but something seems off with the entire process. So the question is should we treat each jupyter cell a separate viz instance when someone does `%run_viz` ? With the current approach, the issue of using old running process is fixed but the users can run `%run_viz` only 10 times in a notebook session

image

## Development notes

* Add `_find_available_port` logic placed in `run` command to `%run_viz`
* Remove `allocate_port` method

## QA notes

* All tests should pass
* You can test %run_viz on notebook and it should use different ports for different cell runs. 

## Checklist


rashidakanchwala:
hi Ravi, i am debating whether we should spend more time on this. `%run_viz` is not used much. It was broken for 6-7 months and nobody complained. Also, given with the new way to 'Visualise pipelines in notebooks, this might become old' Let's increase it to max 10 and see if we get users complaining -- then we can decide what to do. wdyt @astrojuanlu ?

astrojuanlu (approved):
Agreed with @rashidakanchwala , in line with what I said in https://github.com/kedro-org/kedro-viz/issues/1823#issuecomment-2612566430, probably we will be de-emphasizing this over time. Still, if there's no more work needed here and given that this actually _removed_ code 😄 let's merge!

PR #2268, opened 2025-02-06
Add ESM bundle for Kedro-Viz

ravi-kumar-pilla:
## Description

Add an ESM Kedro-Viz bundle to be used directly in a html document.

## Development notes

- Created a folder esm in the current kedro-viz GH repository
- Upload the production bundles kedro-viz.production.mjs to the esm folder
- Automate the process of bundling and updating the bundles via `make version`, i.e., add a step to update this folder when we do a new release

## QA notes

* All tests should pass
* Test locally to check if the esm bundle is produced by running either `make version VERSION=10.3.0` or `npm run build:esm`. Discard all the changes after testing

## Checklist


rashidakanchwala:
this is awesome, few questions around 
- what are the bundle sizes now ? 
- u mentioned ESM wasn't working before, what change made it work ? 
- u mentioned lodash, and react-dom as heavy dependencies as well, does it make sense to put them in external too? 
- u mentioned adding stack traces and we decided not to, as the bundle size for UMD grew very large. is this the case with ESM too?

webpack.config.js
```diff
@@ -0,0 +1,58 @@
...
+        library: {
+            type: 'module',
+        },
+    },
+    externalsType: 'module',
+    externals: {
+        'plotly.js-dist-min': 'https://cdn.plot.ly/plotly-2.26.0.min.js',
+    },
+    experiments: {
+        outputModule: true,
```
rashidakanchwala, line 21: what does this do ?

astrojuanlu:
Thanks @ravi-kumar-pilla ! I tested this HTML with the bundle and it worked 👏🏼 

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Kedro-Viz</title>
  </head>
  <body>
    <div id="kedro-viz-1234" style="height: 600px"></div>
    <script src="https://cdn.jsdelivr.net/npm/react@18.2/umd/react.production.min.js"></script>
    <!-- script src="https://cdn.jsdelivr.net/npm/react-dom@18.2/umd/react-dom.production.min.js"></script -->
    <script type="module">
      import { KedroViz, createRoot } from './kedro-viz.production.mjs';

      const viz_container = document.getElementById('kedro-viz-1234');
      const options = {
        display: {
          expandPipelinesBtn: false,
          exportBtn: false,
          globalNavigation: false,
          labelBtn: false,
          layerBtn: false,
          metadataPanel: true,
          miniMap: false,
          sidebar: false,
          zoomToolbar: false,
        },
        expandAllPipelines: false,
        behaviour: {
          reFocus: false,
        },
        theme: 'dark',
      };

      fetch('spaceflights.mock.json')
        .then((response) => response.json())
        .then((json) => {
          if (createRoot && viz_container) {
            const viz_root = createRoot(viz_container);
            viz_root.render(
              React.createElement(KedroViz, {
                data: json,
                options: options,
              })
            );
          }
        })
        .catch((error) => console.error('Error loading data:', error));
    </script>
  </body>
</html>
```

- I did `npm run build:esm` and I see a diff in the `esm` folder. Is the expectation that the result should be identical to what's committed?
  - Related, should it be committed at all?
- I tried to not use `createRoot` from the bundle but rather from `react-dom`, to no avail. In line with what @rashidakanchwala says, is there a chance we can make `react-dom` external too?

webpack.config.js
ravi-kumar-pilla: Hi @rashidakanchwala , This instructs webpack to generate esm syntax in the output bundle and also tries to make smaller and cleaner bundles.

ravi-kumar-pilla:
Hi @astrojuanlu , Glad it worked :D

> * I did `npm run build:esm` and I see a diff in the `esm` folder. Is the expectation that the result should be identical to what's committed?

Yes, it is expected. `npm run build:esm` should be run only with a new release which will update our bundle. So I added it to `make version` command.
   
>   * Related, should it be committed at all?

We need our bundle to be hosted somewhere which can be served via CDN. I could find 3 places where we can have this -
1. npm package - Having a bundle here would increase our npm package size which is already quite big
2. Our GitHub repo - This is a low effort task and we can easily automate with the release workflow. Having it in repo does not effect our users (npm install or pip install wont have this bundle)
3. S3 bucket/Any cloud provider - A bit of effort and need to make changes to our CI workflow + maintenance of AWS/any cloud account.

> * I tried to not use `createRoot` from the bundle but rather from `react-dom`, to no avail. In line with what @rashidakanchwala says, is there a chance we can make `react-dom` external too?

As mentioned in the reply above - https://github.com/kedro-org/kedro-viz/pull/2268#issuecomment-2648206240

Thank you

rashidakanchwala (commented):
thanks Ravi!! :) let's merge! ⭐

astrojuanlu:
Thanks for the clarifications!

My only last comment is that maybe we can indeed publish the bundle to GitHub Releases _without_ having to commit the bundle to the `main` branch, right?

Other than that, I'll let one more engineer review this 👍🏼

ravi-kumar-pilla:
> My only last comment is that maybe we can indeed publish the bundle to GitHub Releases _without_ having to commit the bundle to the `main` branch, right?

I tried that @astrojuanlu uploading the bundle as an artifact in the github release. But I was unable to refer it via CDN like - https://cdn.jsdelivr.net/gh/kedro-org/kedro-viz@10.3.0/esm/kedro-viz.production.mjs. I do see some articles suggesting that. Let me try one more time. If anyone did this before, please let me know @rashidakanchwala @ankatiyar @SajidAlamQB  . Thanks

rashidakanchwala (approved):

jitu5 (approved):
Thanks @ravi-kumar-pilla!

Issue #2213, opened 2024-11-29
Enable Lazy loading of datasets feature from latest Kedro


rashidakanchwala:
## Description

Just a placeholder ticket. I shall fill it once I am back from PTO after a quick chat with @ElenaKhaustova . 

## Context
Changes made on catalog side - https://github.com/kedro-org/kedro/issues/3935

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

## Checklist

- [ ] Include labels so that we can categorise your feature request

astrojuanlu:
Is this connected with #1806, https://github.com/kedro-org/kedro/issues/2829, https://github.com/kedro-org/kedro/issues/3935 @rashidakanchwala ?

PR #2272, opened 2025-02-12
Enable lazy loading support for DataCatalog 2.0 on Kedro-Viz

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/2213

This PR integrates the new lazy-loading feature from Kedro’s updated DataCatalog API into Kedro-Viz. 

Specifically, it checks if the project’s catalog is an instance of `KedroDataCatalog`. This allows Kedro-Viz to show dataset details without requiring full dataset installation. It also prevents unintended data loads if a dataset is configured to be lazy.

## Development notes

- Added conditional imports for `KedroDataCatalog` with fallbacks
- Updated repository access methods to handle lazy datasets
- Made dataset parameters optional (`Optional[AbstractDataset]`)
- Removed assertions that would fail with lazy datasets
- Added null-safety checks in all validators and properties
- Modified to accept both catalog implementations
- Ensured dataset fetching works consistently for both types
- Updated type annotations for better API clarity

## QA notes

- Added tests that verify behavior with `KedroDataCatalog`
- Added tests for handling null cases in node models

## Checklist


package/kedro_viz/models/flowchart/nodes.py
```diff
@@ -246,16 +246,20 @@ class TaskNode(GraphNode):
         description="The original namespace on this node",
     )
 
+    # Remove forced assertion that kedro_obj is not None
     @model_validator(mode="before")
     @classmethod
     def check_kedro_obj_exists(cls, values):
-        assert "kedro_obj" in values
+        # assert "kedro_obj" in values
```
ravi-kumar-pilla, line 253: is kedro_obj not needed for TaskNode ?

package/kedro_viz/data_access/managers.py
```diff
@@ -82,6 +91,12 @@ def resolve_dataset_factory_patterns(
         """Resolve dataset factory patterns in data catalog by matching
         them against the datasets in the pipelines.
         """
+        if self._is_new_catalog:
+            logger.debug(
+                "Skipping dataset factory pattern resolution for KedroDataCatalog (lazy loading)."
+            )
+            return
```
ravi-kumar-pilla, line 98: If the log is not required, can we make this check before calling the method ?

SajidAlamQB: Yes, we can do this also.

package/kedro_viz/models/flowchart/nodes.py
SajidAlamQB: For the new catalogs, `kedro_obj` might be None for some nodes so we skip this. We store `None` for dataset_obj so we don’t cause an import or library load for the lazy loading approach.

package/kedro_viz/data_access/managers.py
```diff
@@ -6,6 +6,13 @@
 
 from kedro.io import DataCatalog
 
+try:
+    from kedro.io.kedro_data_catalog import KedroDataCatalog
+
+    IS_DATACATALOG_2 = True
```
merelcht, line 12: Why not just name this `IS_KEDRODATACATALOG` ?

package/kedro_viz/data_access/managers.py
```diff
@@ -104,7 +113,12 @@ def add_catalog(self, catalog: DataCatalog, pipelines: Dict[str, KedroPipeline])
             pipelines: A dictionary which holds project pipelines
         """
 
-        self.resolve_dataset_factory_patterns(catalog, pipelines)
+        self._is_new_catalog = IS_DATACATALOG_2 and isinstance(
+            catalog, KedroDataCatalog
+        )
```
merelcht, line 118: Why use so many different ways to refer to the `KedroDataCatalog`? New catalog, catalog 2... seems to overcomplicate things unnecessarily to me. I'd just use the actual name `KedroDataCatalog` and variations of that e.g. `kedro_datacatalog` eveywhere.

SajidAlamQB: Agreed, I will simplify it.

package/kedro_viz/data_access/managers.py
```diff
@@ -322,17 +347,15 @@ def add_dataset(
...
-            # of DataCatalog calls AbstractDataset.from_config
-            # which tries to create a Dataset instance from the pattern
-            obj = UnavailableDataset()
+        if self._is_new_catalog:
+            dataset_obj = None
+        else:
+            try:
+                dataset_obj = self.catalog.get_dataset(dataset_name)
+            except DatasetError:
+                dataset_obj = UnavailableDataset()
```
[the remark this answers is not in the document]
SajidAlamQB: Its the same behaviour as normal i.e it throws a `DatasetError` not found and gives a hint to install `pip install kedro-datasets[<dataset-group>]`

package/kedro_viz/data_access/managers.py
```diff
@@ -173,9 +191,15 @@ def add_pipeline(self, registered_pipeline_id: str, pipeline: KedroPipeline):
                 self.registered_pipelines.add_node(
                     registered_pipeline_id, output_node.id
                 )
-                if isinstance(output_node, TranscodedDataNode):
-                    output_node.original_name = output
-                    output_node.original_version = self.catalog.get_dataset(output)
+
+                # For older catalog
+                if (
```
rashidakanchwala, line 196: quick question; for new catalog we don't need to do this because it is done on Kedro itself right?

package/kedro_viz/models/flowchart/nodes.py
```diff
@@ -103,7 +103,7 @@ def create_data_node(
         dataset_name: str,
         layer: Optional[str],
         tags: Set[str],
-        dataset: AbstractDataset,
+        dataset: Optional[AbstractDataset],
```
rashidakanchwala, line 106: why is this Optional now?

package/kedro_viz/models/flowchart/nodes.py
```diff
@@ -389,11 +397,12 @@ class ParametersNode(GraphNode):
     # The type for Parameters Node
     type: str = GraphNodeType.PARAMETERS.value
 
+    # Remove forced assertion that kedro_obj, name are in values
     @model_validator(mode="before")
     @classmethod
     def check_kedro_obj_and_name_exists(cls, values):
-        assert "kedro_obj" in values
-        assert "name" in values
+        # assert "kedro_obj" in values
```
rashidakanchwala, line 404: can we remove these comments

rashidakanchwala (changes requested):
We need to handle `kedro viz --lite` correctly here. I think we are breaking it. Because when I try to uninstall `kedro-datasets` and do `kedro viz --lite` it breaks. This is not supposed to happen. `kedro viz --lite` should work regardless if dependencies are there or not. If `kedro-datasets` is missing, it runs and just makes all datasets UnavailableDatasets.

package/kedro_viz/data_access/managers.py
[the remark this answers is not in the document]
SajidAlamQB: `kedro viz --lite` is working now, changed the order so we check if we are using `lite` first then patch before getting the catalog.

SajidAlamQB:
> We need to handle `kedro viz --lite` correctly here. I think we are breaking it.

`kedro viz --lite` should be working now we check if `lite` is required first and do the patching before getting the catalog.

package/kedro_viz/data_access/managers.py
```diff
@@ -71,16 +78,22 @@ def _initialize_fields(self):
...
         self._initialize_fields()
 
+
     def add_catalog(self, catalog: DataCatalog):
         """Add the catalog to the CatalogRepository
 
         Args:
             catalog: The DataCatalog instance to add.
         """
+        self._kedro_datacatalog = IS_KEDRODATACATALOG and isinstance(
```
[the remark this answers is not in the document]
SajidAlamQB: I think we still need the `_kedro_datacatalog` check because we are still differentiating between older catalogs and new DataCatalog 2.0?

package/kedro_viz/data_access/managers.py
SajidAlamQB: Yes exactly with DataCatalog 2.0, the built-in lazy logic means we don't have to manually store or fetch the original dataset object (original_version, original_name) in our Viz code.

rashidakanchwala (approved):
Thanks, now it makes sense.

astrojuanlu (commented):
I have some questions about the forward compatibility of the `KedroDataCatalog` https://github.com/kedro-org/kedro/pull/4510#issuecomment-2681991286 not sure they should block this PR, but I think it would be cool to sort them out ahead of the next releases.

SajidAlamQB:
> I have some questions about the forward compatibility of the `KedroDataCatalog` [kedro-org/kedro#4510 (comment)](https://github.com/kedro-org/kedro/pull/4510#issuecomment-2681991286) not sure they should block this PR, but I think it would be cool to sort them out ahead of the next releases.

The current implementation should work with Kedro 0.18.x through 1.0.x (assuming `KedroDataCatalog` is kept as a deprecated alias in 1.0). However, it would break in Kedro 2.0 when `KedroDataCatalog` is completely removed.

We could add version detection logic to handle the transition to Kedro 1.0+ differently. Would you recommend implementing that now, or waiting until Kedro 1.0 is closer to release?

rashidakanchwala:
Good question, i thought Kedro 1.0 will deprecate old DataCatalog but I just saw the notes and it says it will be deprecated in 2.0. I think the plan was for Kedro-viz to remove the old DataCatalog methods when Kedro releases 1.0. 

cc- @ElenaKhaustova , @merelcht

Issue #2273, opened 2025-02-17
Replace DataCatalog private methods with DataCatalog 2.0 API public methods.


rashidakanchwala:
## Description
As part of Polishing API Surface ticket, one last bit of cleanup on Kedro-viz. We need to replace `_get_dataset()` and `_data_set` with Public API of getting dataset using [`catalog.get()`](https://docs.kedro.org/en/stable/data/kedro_data_catalog.html#how-to-access-datasets-in-the-catalog) 

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

## Checklist

- [ ] Include labels so that we can categorise your feature request

PR #2274, opened 2025-02-17
Update DataCatalog private methods with DataCatalog 2.0 API public methods

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro-viz/issues/2273

This PR replaces references to the private `_get_dataset()` with Kedro’s public `catalog.get()` method.

To maintain backward compatibility for older Kedro versions that do not include `KedroDataCatalog`, introduced a version/feature check that falls back to `_get_dataset()` only if necessary.

## Development notes

- Added an `IS_KEDRODATACATALOG` flag to detect whether `KedroDataCatalog` is available.
- Where possible, replaced `_get_dataset()` calls with `catalog.get()`.
- Fallback to `_get_dataset()` is used only for older Kedro versions.

## Checklist


package/kedro_viz/data_access/managers.py
```diff
@@ -6,6 +6,13 @@
 
 from kedro.io import DataCatalog
```
rashidakanchwala, line 7: does it make sense to import this only as a fallback if there's an importError when we are in an older kedro version?

rashidakanchwala (approved):
thanks.

package/kedro_viz/data_access/managers.py
SajidAlamQB: I think we still need `DataCatalog` as the “baseline” import so our code is backward compatible with older Kedro (which doesn’t have `KedroDataCatalog`)

ravi-kumar-pilla: should the line 7 be inside except ImportError at line 13 ? Normally that is how we were [doing](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/data_access/managers.py#L9) bc before. Wouldn't this raise error if DataCatalog is not available in future ?

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -99,8 +110,11 @@ def layers_mapping(self):  # noqa: PLR0912
                         self._layers_mapping[dataset_name] = layer
 
         for dataset_name in datasets:
-            dataset = self._catalog._get_dataset(dataset_name)
-            metadata = getattr(dataset, "metadata", None)
+            if IS_DATACATALOG_2 and isinstance(self._catalog, KedroDataCatalog):
+                dataset = self._catalog.get(dataset_name)
+            else:
+                dataset = self._catalog._get_dataset(dataset_name)
+                metadata = getattr(dataset, "metadata", None)
```
ravi-kumar-pilla, line 117: how are we handling metadata in 2.0 ?

package/kedro_viz/data_access/managers.py
```diff
@@ -91,7 +98,10 @@ def resolve_dataset_factory_patterns(
 
             for dataset_name in datasets:
                 try:
-                    catalog._get_dataset(dataset_name, suggest=False)
+                    if IS_DATACATALOG_2 and isinstance(catalog, KedroDataCatalog):
```
ravi-kumar-pilla, line 101: Not regarding this PR but I remember Elena mentioned `resolve_dataset_factory_patterns` will happen on kedro side. Do we still need to have this function here? Can we call this function only when it is not IS_DATACATALOG_2 ?

package/kedro_viz/data_access/repositories/catalog.py
SajidAlamQB: We handle it in a backward-compatible way. If we detect Kedro 0.19+, then we rely on metadata["kedro-viz"]["layer"] from the dataset config. For older Kedro versions (<0.19), we fall back to `catalog.layers`, if it exists.

package/kedro_viz/data_access/managers.py
SajidAlamQB: Yes, if thats the case then we might not need `resolve_dataset_factory_patterns` anymore but for now we still need to support older Kedro versions that do not do this automatically.

package/kedro_viz/data_access/managers.py
SajidAlamQB: In the future we will need to modify it if we remove `DataCatalog` but as of now it's needed either way (older or newer Kedro). it shouldn't make a difference if its inside the except block this just looks cleaner imo.

ravi-kumar-pilla: Why would you import DataCatalog if KedroDataCatalog is available ?

SajidAlamQB: I see what you are saying if we are definitely also removing this it makes sense.

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -99,8 +110,11 @@ def layers_mapping(self):  # noqa: PLR0912
...
         for dataset_name in datasets:
-            dataset = self._catalog._get_dataset(dataset_name)
-            metadata = getattr(dataset, "metadata", None)
+            if IS_KEDRODATACATALOG and isinstance(self._catalog, KedroDataCatalog):
+                dataset = self._catalog.get(dataset_name)
+            else:
+                dataset = self._catalog._get_dataset(dataset_name)
+                metadata = getattr(dataset, "metadata", None)
             if not metadata:
                 continue
```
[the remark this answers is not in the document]
SajidAlamQB: Oh good catch, we should still do this for `KedroDataCatalog`.

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -99,8 +110,12 @@ def layers_mapping(self):  # noqa: PLR0912
...
 
         for dataset_name in datasets:
-            dataset = self._catalog._get_dataset(dataset_name)
-            metadata = getattr(dataset, "metadata", None)
+            if IS_KEDRODATACATALOG and isinstance(self._catalog, KedroDataCatalog):
+                dataset = self._catalog.get(dataset_name)
+                metadata = getattr(dataset, "metadata", None)
+            else:
+                dataset = self._catalog._get_dataset(dataset_name)
+                metadata = getattr(dataset, "metadata", None)
```
ravi-kumar-pilla, line 118: Suggestion to have metadata independent of if, else as it is the same for both conditions -

```python
         for dataset_name in datasets:
            if IS_KEDRODATACATALOG and isinstance(self._catalog, KedroDataCatalog):
                dataset = self._catalog.get(dataset_name)
            else:
                dataset = self._catalog._get_dataset(dataset_name)
                
            metadata = getattr(dataset, "metadata", None)
```

package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -78,11 +86,14 @@ def layers_mapping(self):  # noqa: PLR0912
...
-        except Exception:  # noqa: BLE001 # pragma: no cover
-            datasets = self._catalog._datasets
+        if IS_KEDRODATACATALOG and isinstance(self._catalog, KedroDataCatalog):
+            datasets = self._catalog.list()
+        else:
+            # Temporary try/except block so the Kedro develop branch can work with Viz.
+            try:
+                datasets = self._catalog._data_sets
+            except Exception:  # noqa: BLE001 # pragma: no cover
+                datasets = self._catalog._datasets
```
merelcht, line 96: Is this still needed? It mentions the kedro develop branch, but I'm guessing that was related to pre-0.19?

SajidAlamQB: Good spot, this branch was released a while ago I will remove this block.

ravi-kumar-pilla (approved):
Thank you @SajidAlamQB

package/kedro_viz/data_access/repositories/catalog.py
SajidAlamQB: Actually we still need that try/except because in older Kedro versions (like 0.18.3), as CI started failing for python >3.10.

PR #2278, opened 2025-02-22
Experiment Tracking CleanUp

rashidakanchwala:
## Description

I did a final clean-up of Experiment Tracking on Kedro-viz 
- mostly minor stuff here and there 
- removed dataset_factory resolver implementation as it only affected Exp. Tracking
- also Experiment Tracking last run data was also originally showed in the flowchart metadata panel, so I removed that
- besides that some plotly references and other stuff. 
- also quite a few of these files are removing experiment tracking data files from demo-project. 
## Checklist


package/kedro_viz/server.py
```diff
@@ -26,7 +26,7 @@ def populate_data(
     if creating an api app from project.
     """
 
-    data_access_manager.add_catalog(catalog, pipelines)
+    data_access_manager.add_catalog(catalog)
```
Huongg, line 29: why don't we need to pass the `pipelines` anymore?  @rashidakanchwala

rashidakanchwala: Yup! Initially, we passed only the catalog. However, we encountered issues with dataset factories not working with Experiment Tracking, so we had to pass pipelines to resolve the dataset factory issue. Now that Experiment Tracking is no longer in use and DataCatalog 2.0 handles this, we can clean up this part and revert to passing just the catalog.

Huongg (approved):
thank you for all the clean up @rashidakanchwala

jitu5 (approved):

package/kedro_viz/data_access/managers.py
```diff
@@ -68,36 +68,13 @@ def __init__(self):
         )
         self.dataset_stats = {}
 
-    def resolve_dataset_factory_patterns(
```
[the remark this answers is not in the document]
ravi-kumar-pilla: Yes it was added to make `tracking_datasets` discoverable

src/utils/data/spaceflights.mock.json
```diff
@@ -195,7 +195,7 @@
       "type": "data",
       "modular_pipelines": [],
       "layer": null,
-      "dataset_type": "tracking.metrics_dataset.MetricsDataset"
+      "dataset_type": null
```
ravi-kumar-pilla, line 198: why is the json still having this dataset ? If it is metrics then with the demo_project clean up this should not be present right ?

src/utils/data/spaceflights_reordered.mock.json
```diff
@@ -195,7 +195,7 @@
       "type": "data",
       "modular_pipelines": [],
       "layer": null,
-      "dataset_type": "tracking.metrics_dataset.MetricsDataset"
+      "dataset_type": null
```
ravi-kumar-pilla, line 198: same as above

src/utils/data/spaceflights.mock.json
rashidakanchwala: @ravi-kumar-pilla , this is not at all related to demo_project. this is just used in jests, and if I remove this, it messes up the entire json because it's a generated one. and starts failing test. so i left the code, but just removed the dataset type. it can be a MemoryDataset

ravi-kumar-pilla: sorry I copied the comment from below which said demo.mock.json. Sure

PR #2308, opened 2025-04-02
Fix missing layer information for dataset factories.

### LINKED ISSUE #2301: Kedro-viz layers not showing in 11.0.0 when using Dataset Factories
## Description
A pipeline with 3 layers defined shows everything under the first-defined layer in the kedro viz view

## Context
Not a blocker but makes it difficult to demo / build my pipeline

## Steps to Reproduce

1. install `kedro-viz==11.0.0`
2. create a project with >1 layer (3 in my case)
3. `kedro viz run`

## Expected Result
I should see my pipeline with 3 unique layers

## Actual Result
Everything in my pipeline falls under the first-defined layer

## Your Environment
* Python 3.11
* `Kedro-viz==11.0.0`
* Chrome

ravi-kumar-pilla:
## Description

Resolves #2301 

## Development notes

* Add `resolve_dataset_factories` to populate layer information in dataset factories
* Updated tests

## QA notes

* All tests should pass
* Manual testing -

1. Modify demo_project's `catalog_02_int.yml` as below -

```yaml
"{namespace}.int_{name}":
  type: pandas.ParquetDataset
  filepath: ${_base_location}/02_intermediate/{name}.pq
  metadata:
    kedro-viz:
      layer: intermediate

ingestion.int_typed_shuttles@pandas1:
  type: pandas.ParquetDataset
  filepath: ${_base_location}/02_intermediate/typed_shuttles.pq
  metadata:
    kedro-viz:
      layer: test

ingestion.int_typed_shuttles@pandas2:
  type: pandas.ParquetDataset
  filepath: ${_base_location}/02_intermediate/typed_shuttles.pq
  metadata:
    kedro-viz:
      layer: test
```
 
 2. And use the factory pattern in your pipeline node as -
 
 ```python
  node(
                func=aggregate_company_data,
                inputs="int_typed_companies",
                outputs="prm_agg_companies",
                name="company_agg",
            ),
 ```

3. Run `kedro viz`, the `intermediate` layer will be missing if you expand the `ingestion` namespace. Also the `api/main` -> layers list does not contain `intermediate` layer. If you check the nodes list, the layer information for `ingestion.int_typed_companies` is null.
4. With this PR, the layers information will be populated

image

image

## Checklist

### DIFF
--- a/package/kedro_viz/data_access/repositories/catalog.py
+++ b/package/kedro_viz/data_access/repositories/catalog.py
@@ -112,10 +112,10 @@ class CatalogRepository:
         for dataset_name in datasets:
             if IS_KEDRODATACATALOG and isinstance(self._catalog, KedroDataCatalog):
                 dataset = self._catalog.get(dataset_name)
-                metadata = dataset.config.metadata
             else:
                 dataset = self._catalog._get_dataset(dataset_name)
-                metadata = getattr(dataset, "metadata", None)
+
+            metadata = getattr(dataset, "metadata", None)
 
             if not metadata:
                 continue
--- a/package/kedro_viz/data_access/repositories/catalog.py
+++ b/package/kedro_viz/data_access/repositories/catalog.py
@@ -86,11 +86,15 @@ class CatalogRepository:
 
         self._layers_mapping = {}
 
+        # Get datasets available in catalog
         if IS_KEDRODATACATALOG and isinstance(self._catalog, KedroDataCatalog):
+            # Returns dataset names
             datasets = self._catalog.list()
         else:
             # try/except block so Viz is backwards compatible with older kedro versions.
             try:
+                # Returns a dictionary where key is dataset name and value
+                # is an instance of AbstractDataset
                 datasets = self._catalog._data_sets
             except Exception:  # noqa: BLE001 # pragma: no cover
                 datasets = self._catalog._datasets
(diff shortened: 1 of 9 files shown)


package/kedro_viz/data_access/repositories/catalog.py
```diff
@@ -116,10 +116,11 @@ def layers_mapping(self):  # noqa: PLR0912
                 dataset = self._catalog._get_dataset(dataset_name)
 
             metadata = getattr(dataset, "metadata", None)
+
             if not metadata:
                 continue
             try:
-                layer = dataset.metadata["kedro-viz"]["layer"]
+                layer = metadata["kedro-viz"]["layer"]
```
rashidakanchwala, line 123: Ravi, have we also tested this with the future KedroDataCatalog. Elena mentioned the way to access the layer information is via `dataset.config.metadata `

ravi-kumar-pilla: Hi Rashida, I did this and there were some tests that are failing. Probably should add skipif checks. Let me see. Also, I am not sure how are we testing KedroDataCatalog on viz. When I debug, everytime `context.catalog` comes out as type `DataCatalog` and not `KedroDataCatalog`. @SajidAlamQB ?

ElenaKhaustova: Hey @rashidakanchwala, @ravi-kumar-pilla 

Some clarifications regarding `KedroDataCatalog`

1.  `dataset.config.metadata` - this is to get metadata from `_LazyDataset `.
2. The datasets are materialized (`_LazyDataset ` replaces with normal dataset) when accessing them, for example calling `catalog.get(ds_name)`
3. `AttributeError: 'PlotlyDataset' object has no attribute 'config'` this error is happening bcs the dataset was materialized already
4. To resolve factory patterns you **DO NOT** need to actually resolve anything or materialize. You can use `catalog_config_resolver.resolve_pattern(ds_name)` - it resolves and returns dataset configuration as a dictionary without materializing the dataset

ravi-kumar-pilla: Hi @ElenaKhaustova , Thanks for the clarification. This makes sense. Since on Viz we will be supporting both DataCatalog and KedroDataCatalog for now (until Kedro 1.0), we need to handle both cases. 

For the below solution I have few questions -


> To resolve factory patterns you DO NOT need to actually resolve anything or materialize. You can use catalog_config_resolver.resolve_pattern(ds_name) - it resolves and returns dataset configuration as a dictionary without materializing the dataset


1. This is only supported when catalog class is `KedroDataCatalog` ?
2. To use `catalog_config_resolver.resolve_pattern(ds_name)` we still need to loop thru all the pipeline datasets and then call this method for resolving ? This is what we are doing [here](https://github.com/kedro-org/kedro-viz/pull/2308/files#diff-3bf934e67fb34ace5795d2b1ce36f96ca36d88a44dde5b9a04e0d4b77ac1db35R86). But for KedroDataCatalog instead of calling `get` we just do  `catalog_config_resolver.resolve_pattern(ds_name)` ?

Thank you

ravi-kumar-pilla: I think the issue is about not resolving the patterns but having the resolved dataset pattern added to the catalog. We were calling `_get_dataset` of DataCatalog which resolves the pattern and also adds it to catalog - https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L348 . The way we build layers on viz is looping through datasets in the catalog after the above step.

ElenaKhaustova: > This is only supported from

This has been supported since earlier versions after we added `catalog_config_resolver` and should be the same for both catalogs

> I think the issue is not about resolving the patterns but having the resolved dataset pattern added to the catalog

Yes, I understand the issue happening bcs you need to support both catalogs. But ideally all this logic should be significantly simplified after we fully move to the new catalog.

PR #2328, opened 2025-05-14
Run status for pipeline run in non realtime

jitu5:
This pull request introduces a new feature to track and display pipeline run events in Kedro Viz. It includes backend changes to generate and serve run event data, as well as frontend updates to process and visualize this data. The key changes are grouped into backend and frontend updates for clarity.

### Backend Updates:
* **Run Events API**:
  - Added a new API endpoint `/run-events` to serve pipeline run event data using the `RunEventAPIResponse` model. (`package/kedro_viz/api/rest/responses/run_events.py`, `package/kedro_viz/api/rest/router.py`) [[1]](diffhunk://#diff-6ba06200dc99bed158c69b8b97e93b230f6e1ecba25b6834acaa2d0d39499bc7R1-R50) [[2]](diffhunk://#diff-10d5667bbc5cdabe3f081db3be113f89b23959c452f03ea0a18d8ae9712a0bb2R26-R29) [[3]](diffhunk://#diff-10d5667bbc5cdabe3f081db3be113f89b23959c452f03ea0a18d8ae9712a0bb2R69-R83)
  
* **Pipeline Hooks**:
  - Implemented `PipelineRunHooks` to capture and log pipeline events such as node execution, dataset loading/saving, and errors. Events are written to a JSON file (`kedro_pipeline_events.json`) for later retrieval. (`package/kedro_viz/integrations/kedro/run_hooks.py`)
  - Registered the new hook in the `pyproject.toml` file. (`package/pyproject.toml`)

### Frontend Updates:
* **State Management**:
  - Added Redux actions, reducers, and utilities to fetch, process, and store run event data. (`src/actions/run-status.js`, `src/reducers/run-status.js`, `src/reducers/index.js`) [[1]](diffhunk://#diff-675ca60a9ecb3a7ef6191596dcb1c51e1aee3b17287e561c02624d89636be601R1-R57) [[2]](diffhunk://#diff-174cd34c99ca2fca1a1e9ee4bb8fb4708f7ff27881189c565c37a5f35be929dfR1-R37) [[3]](diffhunk://#diff-3527887dbf404f0c28be291b83274e5f361cab6b0e6e0bd5ed06f66dbd11c5b0R11) [[4]](diffhunk://#diff-3527887dbf404f0c28be291b83274e5f361cab6b0e6e0bd5ed06f66dbd11c5b0R89)
  
* **Flowchart Integration**:
  - Updated the `FlowChartWrapper` component to load and display run event data on initialization. (`src/components/flowchart-wrapper/flowchart-wrapper.js`) [[1]](diffhunk://#diff-d09a1601f789cbc98134abcedb59c3be34fe3c0d9fc420ef21121d2b1bf499b4R18) [[2]](diffhunk://#diff-d09a1601f789cbc98134abcedb59c3be34fe3c0d9fc420ef21121d2b1bf499b4R61) [[3]](diffhunk://#diff-d09a1601f789cbc98134abcedb59c3be34fe3c0d9fc420ef21121d2b1bf499b4R271-R274) [[4]](diffhunk://#diff-d09a1601f789cbc98134abcedb59c3be34fe3c0d9fc420ef21121d2b1bf499b4R370-R372)
  
* **Utilities**:
  - Created utility functions to fetch and process run event data into a format suitable for visualization. (`src/utils/run-status.js`)

### DIFF
--- a/package/kedro_viz/api/rest/responses/run_events.py
+++ b/package/kedro_viz/api/rest/responses/run_events.py
@@ -1,0 +1,50 @@
+"""Response for run events API endpoint."""
+
+import json
+import logging
+from pathlib import Path
+from typing import Dict, List, Optional, Any, Union
+
+from pydantic import BaseModel, Field
+
+from kedro_viz.constants import VIZ_METADATA_ARGS
+from kedro_viz.launchers.utils import _find_kedro_project
+
+logger = logging.getLogger(__name__)
+
+class RunEventAPIResponse(BaseModel):
+    """Format for run event endpoint response."""
+    events: List[Dict[str, Any]] = Field(default_factory=list, description="List of run events")
+
+
+def get_run_events_response() -> Union[RunEventAPIResponse, Dict[str, List]]:
+    """Get run events data for API endpoint.
+    
+    Returns:
+        API response with run events data or an empty list if not available
+    """
+
+    try:
+        kedro_project_path = _find_kedro_project(Path.cwd())
+
+        if not kedro_project_path:
+            logger.warning("Could not find a Kedro project to load kedro pipeline events file")
+            return {"events": []}
+
+        pipeline_events_file_path = Path(
+            f"{kedro_project_path}/{VIZ_METADATA_ARGS['path']}/kedro_pipeline_events.json"
+        )
+
+        if not pipeline_events_file_path.exists():
+            logger.warning(f"Run events file {pipeline_events_file_path} not found")
+            return {"events": []}        
+
+
+        with pipeline_events_file_path.open("r", encoding="utf8") as file:
+            events = json.load(file)
+        return {"events": events}    
+    except Exception as exc:  # pragma: no cover
+        logger.exception(f"Error loading run events: {exc}")     
+        return {"events": []} 
+    
+    
--- a/package/kedro_viz/integrations/kedro/run_hooks.py
+++ b/package/kedro_viz/integrations/kedro/run_hooks.py
@@ -1,0 +1,174 @@
+"""`kedro_viz.integrations.kedro.run_hooks` defines hooks to add additional
+functionalities for a kedro run."""
+
+import json
+import logging
+from kedro_viz.utils import _hash, _hash_input_output
+import requests
+import threading
+import sys
+from time import perf_counter
+from pathlib import Path
+from typing import Any
+
+from kedro.framework.hooks import hook_impl
+from kedro.pipeline.node import Node as KedroNode
+
+from kedro_viz.constants import VIZ_METADATA_ARGS
+from kedro_viz.launchers.utils import _find_kedro_project
+
+logger = logging.getLogger(__name__)
+
+# Optional: keep a global list to write to file at the end
+pipeline_events = []
+
+class PipelineRunHooks:
+    def __init__(self):
+        self._node_start_times = {}
+        self._dataset_load_times = {}
+        self._dataset_save_times = {}
+        self._dataset_sizes = {}
+
+    def _estimate_size(self, data: Any) -> int:
+        """Estimate size of dataset in bytes."""
+        try:
+            return sys.getsizeof(data)
+        except Exception:
+            return -1
+
+    def _write_events_to_file(self):
+        try:
+            kedro_project_path = _find_kedro_project(Path.cwd())
+
+            if not kedro_project_path:
+                logger.warning("Could not find a Kedro project to create kedro pipeline events file")
+                return
+
+            pipeline_events_file_path = Path(
+                f"{kedro_project_path}/{VIZ_METADATA_ARGS['path']}/kedro_pipeline_events.json"
+            )
+            pipeline_events_file_path.parent.mkdir(parents=True, exist_ok=True)
+
+            with pipeline_events_file_path.open("w", encoding="utf8") as file:
+                json.dump(pipeline_events, file, indent=2)
+
+        except Exception as exc:  # pragma: no cover
+            logger.warning(
+                "Unable to write pipeline run events to file: %s", exc
+            )            
+
+    @hook_impl
+    def before_dataset_loaded(self, dataset_name: str):
+        # Just keep track of the time, but don't add to pipeline_events
+        self._dataset_load_times[dataset_name] = perf_counter()
+
+    @hook_impl
+    def after_dataset_loaded(self, dataset_name: str, data: Any):
+        duration = perf_counter() - self._dataset_load_times.get(dataset_name, 0)
+        size = self._estimate_size(data)
+        self._dataset_sizes[dataset_name] = size
+
+        node_id = _hash_input_output(dataset_name)
+        event = {
+            "event": "after_dataset_loaded",
+            "dataset": dataset_name,
+            "node_id": node_id,
+            "load_time_sec": duration,
+            "size_bytes": size,
+        }
+
+        pipeline_events.append(event)
+
+    @hook_impl
+    def before_node_run(self, node, inputs):
+        # Just keep track of time without storing the event
+        self._node_start_times[node.name] = perf_counter()
+
+    @hook_impl
+    def after_node_run(self, node, inputs, outputs):
+        # Calculate duration
+        duration = perf_counter() - self._node_start_times.get(node.name, 0)
+
+        node_id = (
+            _hash(str(node))
+            if isinstance(node, KedroNode)
+            else _hash_input_output(node)
+        )
+
+        event = {
+            "event": "after_node_run",
+            "node": node.name,
+            "node_id": node_id,
+            "duration_sec": duration,
+            "status": "success"
+        }
+        
+        pipeline_events.append(event)
+
+    @hook_impl
+    def before_dataset_saved(self, dataset_name: str, data: Any):
+        # Just keep timing data without storing the event
+        self._dataset_save_times[dataset_name] = perf_counter()
+        size = self._estimate_size(data)
+        self._dataset_sizes[dataset_name] = size
+
+    @hook_impl
+    def after_dataset_saved(self, dataset_name: str):
+        duration = perf_counter() - self._dataset_save_times.get(dataset_name, 0)
+        size = self._dataset_sizes.get(dataset_name, -1)
+        node_id = _hash_input_output(dataset_name)
+        
+        event = {
+            "event": "after_dataset_saved",
+            "dataset": dataset_name,
+            "save_time_sec": duration,
+            "node_id": node_id,
+            "size_bytes": size,
+        }
+        
+        pipeline_events.append(event)
+
+    @hook_impl
+    def after_pipeline_run(self, run_params, pipeline, catalog):
+        """At the end of the run, write events to a file."""
+        if run_params.get("pipeline_name") is not None:
+            return
+
+        event = {
+            "event": "after_pipeline_run",
+            "timestamp": perf_counter(),
+        }
+        
+        pipeline_events.append(event)
+        self._write_events_to_file()
+
+    @hook_impl
+    def on_node_error(self, error: Exception, node: Any):
+        node_id = (
+            _hash(str(node))
+            if isinstance(node, KedroNode)
+            else _hash_input_output(node)
+        )
+        event = {
+            "event": "on_node_error",
+            "node": node.name,
+            "node_id": node_id,
+            "error": str(error),
+            "timestamp": perf_counter(),
+        }
+
+        pipeline_events.append(event)
+        self._write_events_to_file()
+
+    @hook_impl
+    def on_pipeline_error(self, error: Exception, run_params: dict):
+        event = {
+            "event": "on_pipeline_error",
+            "error": str(error),
+            "timestamp": perf_counter(),
+        }
+
+        pipeline_events.append(event)
+        self._write_events_to_file()    
+
+pipeline_run_hook = PipelineRunHooks()
--- a/demo-project/.viz/kedro_pipeline_events.json
+++ b/demo-project/.viz/kedro_pipeline_events.json
@@ -1,0 +1,608 @@
+[
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "companies",
+    "node_id": "aed46479",
+    "load_time_sec": 0.03810400003567338,
+    "size_bytes": 13998272
+  },
+  {
+    "event": "after_node_run",
+    "node": "ingestion.apply_types_to_companies",
+    "node_id": "69c523b6",
+    "duration_sec": 0.019259124994277954,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "ingestion.int_typed_companies",
+    "save_time_sec": 0.15067720785737038,
+    "node_id": "f23ad217",
+    "size_bytes": 6399158
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "reviews",
+    "node_id": "7b2c6e04",
+    "load_time_sec": 0.037918708054348826,
+    "size_bytes": 6167824
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "params:ingestion.typing.reviews.columns_as_floats",
+    "node_id": "b5609df0",
+    "load_time_sec": 0.0009360839612782001,
+    "size_bytes": 88
+  },
+  {
+    "event": "after_node_run",
+    "node": "ingestion.apply_types_to_reviews",
+    "node_id": "ea604da4",
+    "duration_sec": 0.013807584065943956,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "ingestion.int_typed_reviews",
+    "save_time_sec": 0.02187287504784763,
+    "node_id": "4f7ffa1b",
+    "size_bytes": 5355856
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "shuttles",
+    "node_id": "f1d596c2",
+    "load_time_sec": 5.347216499969363,
+    "size_bytes": 42831629
+  },
+  {
+    "event": "after_node_run",
+    "node": "ingestion.apply_types_to_shuttles",
+    "node_id": "f33b9291",
+    "duration_sec": 0.04458345798775554,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "ingestion.int_typed_shuttles@pandas1",
+    "save_time_sec": 0.07294145785272121,
+    "node_id": "c0ddbcbf",
+    "size_bytes": 29647753
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "ingestion.int_typed_companies",
+    "node_id": "f23ad217",
+    "load_time_sec": 0.22107366705313325,
+    "size_bytes": 6246118
+  },
+  {
+    "event": "after_node_run",
+    "node": "ingestion.company_agg",
+    "node_id": "8de402c1",
+    "duration_sec": 0.6255464998539537,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "ingestion.prm_agg_companies",
+    "save_time_sec": 0.007949833991006017,
+    "node_id": "8f20d98e",
+    "size_bytes": 3920426
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "ingestion.int_typed_shuttles@pandas2",
+    "node_id": "c0ddbcbf",
+    "load_time_sec": 0.013491166988387704,
+    "size_bytes": 29647753
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "ingestion.prm_agg_companies",
+    "node_id": "8f20d98e",
+    "load_time_sec": 0.0038644580636173487,
+    "size_bytes": 3920426
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "ingestion.int_typed_reviews",
+    "node_id": "4f7ffa1b",
+    "load_time_sec": 0.007105833850800991,
+    "size_bytes": 5355856
+  },
+  {
+    "event": "after_node_run",
+    "node": "ingestion.combine_step",
+    "node_id": "cb5166f3",
+    "duration_sec": 0.05565066612325609,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "prm_shuttle_company_reviews",
+    "save_time_sec": 0.054936332860961556,
+    "node_id": "9f266f06",
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "prm_spine_table",
+    "save_time_sec": 0.014652333222329617,
+    "node_id": "f063cc82",
+    "size_bytes": 952592
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_spine_table",
+    "node_id": "f063cc82",
+    "load_time_sec": 0.012671625008806586,
+    "size_bytes": 952592
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_shuttle_company_reviews",
+    "node_id": "9f266f06",
+    "load_time_sec": 0.02148387487977743,
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "params:feature_engineering.feature.derived",
+    "node_id": "abed6a4d",
+    "load_time_sec": 0.0012102499604225159,
+    "size_bytes": 88
+  },
+  {
+    "event": "after_node_run",
+    "node": "feature_engineering.create_derived_features([prm_spine_table;prm_shuttle_company_reviews;params:feature_engineering.feature.derived]) -> [feature_engineering.feat_derived_features]",
+    "node_id": "04ba733a",
+    "duration_sec": 0.018446375150233507,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "feature_engineering.feat_derived_features",
+    "save_time_sec": 0.0023998329415917397,
+    "node_id": "7c92a703",
+    "size_bytes": 714576
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_spine_table",
+    "node_id": "f063cc82",
+    "load_time_sec": 0.003743582870811224,
+    "size_bytes": 952592
+  },
+  {
+    "event": "after_node_run",
+    "node": "feature_engineering.create_feature_importance([prm_spine_table]) -> [feature_importance_output]",
+    "node_id": "7932e672",
+    "duration_sec": 0.0057149159256368876,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "feature_importance_output",
+    "save_time_sec": 0.021344457985833287,
+    "node_id": "1e3cc50a",
+    "size_bytes": 1259
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_shuttle_company_reviews",
+    "node_id": "9f266f06",
+    "load_time_sec": 0.013648375170305371,
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "params:feature_engineering.feature.static",
+    "node_id": "a3627e31",
+    "load_time_sec": 0.0010364169720560312,
+    "size_bytes": 184
+  },
+  {
+    "event": "after_node_run",
+    "node": "feature_engineering.create_static_features([prm_shuttle_company_reviews;params:feature_engineering.feature.static]) -> [feature_engineering.feat_static_features]",
+    "node_id": "e50f81b8",
+    "duration_sec": 0.014500583056360483,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "feature_engineering.feat_static_features",
+    "save_time_sec": 0.005574292037636042,
+    "node_id": "8e4f1015",
+    "size_bytes": 2470760
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_spine_table",
+    "node_id": "f063cc82",
+    "load_time_sec": 0.003906042082235217,
+    "size_bytes": 952592
+  },
+  {
+    "event": "after_node_run",
+    "node": "ingestion.<lambda>([prm_spine_table]) -> [ingestion.prm_spine_table_clone]",
+    "node_id": "9a6ef457",
+    "duration_sec": 0.003747458104044199,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "ingestion.prm_spine_table_clone",
+    "save_time_sec": 0.0025321668945252895,
+    "node_id": "c08c7708",
+    "size_bytes": 952592
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_shuttle_company_reviews",
+    "node_id": "9f266f06",
+    "load_time_sec": 0.013243249850347638,
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_node_run",
+    "node": "reporting.create_matplotlib_chart([prm_shuttle_company_reviews]) -> [reporting.confusion_matrix]",
+    "node_id": "be6b7919",
+    "duration_sec": 0.2613346658181399,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "reporting.confusion_matrix",
+    "save_time_sec": 0.11260966607369483,
+    "node_id": "3b199c6b",
+    "size_bytes": 72
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_shuttle_company_reviews",
+    "node_id": "9f266f06",
+    "load_time_sec": 0.01415120786987245,
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_node_run",
+    "node": "reporting.get_top_shuttles_data([prm_shuttle_company_reviews]) -> [reporting.top_shuttle_data]",
+    "node_id": "44ef9b48",
+    "duration_sec": 0.01980283297598362,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "reporting.top_shuttle_data",
+    "save_time_sec": 0.01234729215502739,
+    "node_id": "c0be8342",
+    "size_bytes": 120
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_shuttle_company_reviews",
+    "node_id": "9f266f06",
+    "load_time_sec": 0.014168624999001622,
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_node_run",
+    "node": "reporting.make_cancel_policy_bar_chart([prm_shuttle_company_reviews]) -> [reporting.cancellation_policy_breakdown]",
+    "node_id": "c7646ea1",
+    "duration_sec": 0.017396959010511637,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "reporting.cancellation_policy_breakdown",
+    "save_time_sec": 2.465412790887058,
+    "node_id": "d0e9b00f",
+    "size_bytes": 3151
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_shuttle_company_reviews",
+    "node_id": "9f266f06",
+    "load_time_sec": 0.02090154099278152,
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_node_run",
+    "node": "reporting.make_price_analysis_image([prm_shuttle_company_reviews]) -> [reporting.cancellation_policy_grid]",
+    "node_id": "3fb71518",
+    "duration_sec": 0.01369445794261992,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "reporting.cancellation_policy_grid",
+    "save_time_sec": 0.013054291950538754,
+    "node_id": "8838ca1f",
+    "size_bytes": 48
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_shuttle_company_reviews",
+    "node_id": "9f266f06",
+    "load_time_sec": 0.01596683287061751,
+    "size_bytes": 16531446
+  },
+  {
+    "event": "after_node_run",
+    "node": "reporting.make_price_histogram([prm_shuttle_company_reviews]) -> [reporting.price_histogram]",
+    "node_id": "40886786",
+    "duration_sec": 0.09546949993818998,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "reporting.price_histogram",
+    "save_time_sec": 0.05006541684269905,
+    "node_id": "c6992660",
+    "size_bytes": 48
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "prm_spine_table",
+    "node_id": "f063cc82",
+    "load_time_sec": 0.004302625078707933,
+    "size_bytes": 952592
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "feature_engineering.feat_static_features",
+    "node_id": "8e4f1015",
+    "load_time_sec": 0.005634374916553497,
+    "size_bytes": 2470760
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "feature_engineering.feat_derived_features",
+    "node_id": "7c92a703",
+    "load_time_sec": 0.0027751659508794546,
+    "size_bytes": 714576
+  },
+  {
+    "event": "after_node_run",
+    "node": "feature_engineering.joiner([prm_spine_table;feature_engineering.feat_static_features;feature_engineering.feat_derived_features]) -> [model_input_table]",
+    "node_id": "6ea2ec2c",
+    "duration_sec": 0.019555332837626338,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "model_input_table",
+    "save_time_sec": 0.017575958045199513,
+    "node_id": "23c94afb",
+    "size_bytes": 2232744
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "feature_importance_output",
+    "node_id": "1e3cc50a",
+    "load_time_sec": 0.004724500002339482,
+    "size_bytes": 1259
+  },
+  {
+    "event": "after_node_run",
+    "node": "reporting.create_feature_importance_plot([feature_importance_output]) -> [reporting.feature_importance]",
+    "node_id": "4adb5c8b",
+    "duration_sec": 0.04741958295926452,
+    "status": "success"
+  },
+  {
+    "event": "after_dataset_saved",
+    "dataset": "reporting.feature_importance",
+    "save_time_sec": 0.07429479202255607,
+    "node_id": "eb7d6d28",
+    "size_bytes": 48
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "model_input_table",
+    "node_id": "23c94afb",
+    "load_time_sec": 0.018298499984666705,
+    "size_bytes": 2232744
+  },
+  {
+    "event": "after_dataset_loaded",
+    "dataset": "params:split_options",
+    "node_id": "22eec376",
+    "load_time_sec": 0.0008834588807076216,
+    "size_bytes": 232
+  },
+  {
+    "event": "after_node_run",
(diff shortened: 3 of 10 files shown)


package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,176 @@
...
+        pipeline_events.append(event)
+        self._write_events_to_file()
+
+
+    @hook_impl
+    def after_dataset_loaded(self, dataset_name: str, data: Any):
+        size = self._estimate_size(data)
+        self._dataset_sizes[dataset_name] = size
+
+        node_id = _hash_input_output(dataset_name)
```
rashidakanchwala, line 78: lets create a `_get_node_id()` here since we call this multiple times.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,233 @@
...
+from typing import Dict, List, Optional, Any, Union, Set, Tuple
+
+from pydantic import BaseModel, Field
+
+from kedro_viz.constants import VIZ_METADATA_ARGS
+from kedro_viz.launchers.utils import _find_kedro_project
+
+logger = logging.getLogger(__name__)
+
+class NodeInfo(BaseModel):
```
rashidakanchwala, line 17: I was thinking maybe we change NodeStatus and DatasetStatus to ENUM 
NodeStatus could be 'Success', 'Fail', 'Skipped' 
DatasetStatus could be 'Available', and 'Missing' for now.

rashidakanchwala: as @Huongg mentioned; let's align on correct states first and then make this change

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+        """Clear dataset I/O context after success."""
+        self._current_dataset = None
+        self._current_op = None
+
+    @hook_impl
+    def after_catalog_created(self, catalog: Union[Any, Any]):  # type: ignore
+        """Grab catalog datasets for size lookups."""
+        try:
+            # prefer new KedroDataCatalog
+            from kedro.io import KedroDataCatalog
```
rashidakanchwala, line 129: We don't need to support DataCatalog but lets keep this for now and remove it a bit later

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+        """Persist events list to the project's .viz JSON file."""
+        write_events(self._events, self.EVENTS_DIR, self.EVENTS_FILE)
+
+    def _add_event(self, event: Dict[str, Any], flush: bool = False) -> None:
+        """Append one event; flush to disk when requested."""
+        self._events.append(event)
+        if flush:
+            self._write_events()
+
+    def _hash_node(self, node: Any) -> str:
```
rashidakanchwala: we can remove this wrapper

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+        when: str,
+        name: str,
+        data: Any = None,
+        status: str = "Available",
+        node: Optional[KedroNode] = None,
+    ) -> Dict[str, Any]:
+        """Generic builder for dataset load/save events."""
+        return make_dataset_event(when, name, data, status, node, self.datasets)
+
+    def _compute_size(self, name: str, data: Any) -> Optional[int]:
```
rashidakanchwala: we don't use this so we can remove this

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+        """Append one event; flush to disk when requested."""
+        self._events.append(event)
+        if flush:
+            self._write_events()
+
+    def _hash_node(self, node: Any) -> str:
+        """Stable ID for KedroNode or I/O reference."""
+        return hash_node(node)
+
+    def _make_dataset_event(
```
rashidakanchwala: can we remove this and called directly the utils func

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+        self._add_event({"event": "after_pipeline_run", "timestamp": datetime.utcnow().isoformat()}, True)
+
+    @hook_impl
+    def on_node_error(self, error: Exception, node: Any) -> None:
+        self._add_event({
+            "event": "on_node_error",
+            "node": getattr(node, "name", str(node)),
+            "node_id": self._hash_node(node),
+            "error": str(error),
+            "timestamp": datetime.utcnow().isoformat()
```
rashidakanchwala: datetime.now(tz=timezone.utc) this is the way kedro does.  [github](https://github.com/kedro-org/kedro/blob/78b5846700f3f71c71e4e56ca3ce710a3b5591bd/kedro/io/core.py#L511)

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    for attr in ("filepath", "_filepath"):
+        fp = getattr(ds, attr, None)
+        if fp:
+            fs, p = fsspec.core.url_to_fs(fp)
+            if fs.exists(p):
+                return fs.size(p)
+    return None
+
+
+def write_events(events: list, events_dir: str = VIZ_METADATA_ARGS["path"], events_file: str = "kedro_pipeline_events.json") -> None:
```
rashidakanchwala:
```suggestion
def write_events(events: List[Dict[str, Any], events_dir: str = VIZ_METADATA_ARGS["path"], events_file: str = "kedro_pipeline_events.json") -> None:
```

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+    """Collect and write pipeline and dataset events during Kedro runs."""
+
+    # Constants to avoid magic strings
+    EVENTS_DIR = VIZ_METADATA_ARGS["path"]
+    EVENTS_FILE = "kedro_pipeline_events.json"
+
+    def __init__(self):
+        # Track times, events, and context for errors
+        self._node_start: Dict[str, float] = {}
+        self._events: list = []
```
rashidakanchwala: ****
```suggestion
        self._events: List = [Dict[str, Any]]
```

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+            pass
+        # fallback older versions
+        self.datasets = getattr(catalog, "_datasets", getattr(catalog, "_data_sets", {}))
+
+    @hook_impl
+    def before_pipeline_run(self, run_params: dict, pipeline, catalog) -> None:
+        """Emit start event unless this is named pipeline."""
+        if run_params.get("pipeline_name"):
+            return
+        self._all_nodes = list(pipeline.nodes)
```
rashidakanchwala, line 153: stead of using list, can we use the more explicit typing: List[]

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+            "node": getattr(node, "name", str(node)),
+            "node_id": self._hash_node(node),
+            "error": str(error),
+            "timestamp": datetime.utcnow().isoformat()
+        }, True)
+
+    @hook_impl
+    def on_pipeline_error(self, error: Exception) -> None:
+        """Emit pipeline errors with last I/O or node context if available."""
+        ev = {"event": "on_pipeline_error", "error": str(error), "timestamp": datetime.utcnow().isoformat()}
```
rashidakanchwala: let's not use small `ev' let's use more meaningful name

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,110 @@
...
+logger = logging.getLogger(__name__)
+
+
+def hash_node(node: Any) -> str:
+    """Stable ID for KedroNode or I/O reference."""
+    return _hash(str(node)) if isinstance(node, KedroNode) else _hash_input_output(node)
+
+
+def make_dataset_event(
+    when: str,
```
rashidakanchwala: I don't think we should call this `when` but rather `event_type`, i think we can better name -- it's confusing when reading code

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,110 @@
...
+    node: Optional[KedroNode] = None,
+    datasets: Dict[str, Any] = None,
+) -> Dict[str, Any]:
+    """Generic builder for dataset load/save events.
+    
+    Args:
+        when: Event type/name
+        name: Dataset name
+        data: Dataset data
+        status: Dataset status
```
rashidakanchwala: i think it's better to give descriptive names 

so we can do dataset_name, dataset_value, dataset_status,

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,110 @@
...
+        datasets: Dictionary of available datasets
+        
+    Returns:
+        Dictionary with event data
+    """
+    ev = {"event": when, "dataset": name, "node_id": _hash_input_output(name), "status": status}
+    if node:
+        ev.update({
+            "node": node.name,
+            "node_id_from": hash_node(node)
```
rashidakanchwala: this part is a bit confusing, cant we just say `node_id`

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def hash_node(node: Any) -> str:
+    """Stable ID for KedroNode or I/O reference."""
+    return _hash(str(node)) if isinstance(node, KedroNode) else _hash_input_output(node)
+
+
+def make_dataset_event(
```
rashidakanchwala:
```suggestion
def create_dataset_event(
```

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,110 @@
...
+            "node_id_from": hash_node(node)
+        })  # include node context
+    if data is not None and status == "Available" and datasets:
+        size = compute_size(name, data, datasets)
+        if size is not None:
+            ev["size_bytes"] = size  # only attach size when available
+    return ev
+
+
+def compute_size(name: str, data: Any, datasets: Dict[str, Any]) -> Optional[int]:
```
rashidakanchwala: again, lets use better names for clarity.

package/kedro_viz/integrations/kedro/hooks_utils.py
rashidakanchwala: don't think we need to hardcode the name here again.  maybe we can define the DEFAULT_EVENT_FILE here and remove it from the other file.

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+from kedro_viz.launchers.utils import _find_kedro_project
+
+logger = logging.getLogger(__name__)
+
+
+class PipelineRunHooks:
+    """Collect and write pipeline and dataset events during Kedro runs."""
+
+    # Constants to avoid magic strings
+    EVENTS_DIR = VIZ_METADATA_ARGS["path"]
```
rashidakanchwala: if you are adding these in defaults already maybe we don't need them here too

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,172 @@
...
+    EVENTS_FILE = "kedro_pipeline_events.json"
+
+    def __init__(self):
+        # Track times, events, and context for errors
+        self._node_start: Dict[str, float] = {}
+        self._events: list = []
+        self.datasets: Dict[str, Any] = {}
+        self._current_node: Optional[Any] = None
+        self._current_dataset: Optional[str] = None
+        self._current_op: Optional[str] = None
```
rashidakanchwala: not sure what `current_op` is ?

jitu5: Its `current_operation`, I updated in the code.

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,193 @@
...
+    create_dataset_event,
+    generate_timestamp,
+    hash_node,
+    write_events,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PipelineRunHooks:
```
ravi-kumar-pilla:
```suggestion
class PipelineRunStatusHook:
```

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def _write_events(self) -> None:
+        """Persist events list to the project's .viz JSON file."""
+        write_events(self._events)
+
+    def _add_event(self, event: Dict[str, Any], flush: bool = False) -> None:
+        """Append one event; flush to disk when requested."""
+        self._events.append(event)
+        if flush:
+            self._write_events()
```
ravi-kumar-pilla: Hi @jitu5 ,

I am not sure how this will write to a file when parallel runner is used. Is MVP only for sequential runner ? Also why do we need to write to a file everytime flush is true even though we have `self._events` ?

It would be nice to have a shared memory / thread safe variable `self._events` which stores all the run events triggered sequentially or parallelly. Then write everything once pipeline run is complete or on error. This would avoid multiple IO operations.

There are pros and cons for either approach. If we are targeting small-medium pipeline runs, writing all at once would avoid multiple IO operations and would be performant. 

In case we do write events like a stream (how you have done now), why do we need `self._events` ?
May be for MVP this is not important but we need to think of parallel runner too (as the DatasetStatsHook had issues which I missed)

Thank you

jitu5: @ravi-kumar-pilla We discussed the parallel runner, and for the MVP, we're proceeding with the sequential runner. I’ve added flush options to control when to write events. We're using `self._events` to collect all events, and previously, we were writing to the file after `before_pipeline_run` and again on completion or error.

Since writing after `before_pipeline_run` doesn’t add value, I’ve removed that write step. Now, events are only written to the file upon pipeline completion or error.

Post-MVP, we’ll explore how to make this work with the parallel runner. Also lets connect on a shared memory / thread safe variable `self._events` if we need to add now or later.

package/features/steps/cli_steps.py
```diff
@@ -88,16 +88,14 @@ def create_project_with_starter(context, starter):
...
+    req_path = f"{context.root_project_dir}/requirements.txt"
+
     if context.kedro_version != "latest":
-        requirements_path = str(context.root_project_dir) + "/src/requirements.txt"
-        # numpy 2.0 breaks with old versions of pandas and this
-        # could be removed when the lowest version supported is updated
-        _add_package_pin(requirements_path, "numpy", "1.26.4")
-    else:
-        requirements_path = str(context.root_project_dir) + "/requirements.txt"
+        _add_package_pin(req_path, "numpy", "1.26.4")
```
rashidakanchwala: I think we can remove these and update the test to the latest Kedro version. Closer to the release, we should also bump the earliest supported Kedro version to 1.0, since this feature won’t be compatible with Kedro 0.19—we’re using only the new DataCatalog.

package/features/viz.feature
```diff
@@ -3,8 +3,8 @@ Feature: Viz plugin in new project
         Given I have prepared a config file with example code
 
     Scenario: Execute viz with the earliest Kedro version that it supports 
-        Given I have installed kedro version "0.18.3"
-        And I have run a non-interactive kedro new with pandas-iris starter
+        Given I have installed kedro version "0.19.11"
```
rashidakanchwala: on second thoughts, for now, just comment these e2e tests; we shouldn't be making those changes in the cli_steps file

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,132 @@
+"""Utility functions for Kedro hooks implementation."""
+
+import json
+import logging
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Dict, List, Optional
```
merelcht, line 7: Same comment here about https://peps.python.org/pep-0585/

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    if dataset_value is not None and datasets:
+        size = compute_size(dataset_name, dataset_value, datasets)
+        if size is not None:
+            event["size_bytes"] = size  # only attach size when available
+    return event
+
+
+def compute_size(
+    dataset_name: str, dataset_value: Any, datasets: Dict[str, Any]
+) -> Optional[int]:
```
merelcht: I'm a little confused about this method. What is `dataset_value` and why is it needed to compute the size if you're still fetching the dataset?

package/features/viz.feature
jitu5: Commented those two e2e test and reverted changes in cli_steps file

package/kedro_viz/integrations/kedro/hooks_utils.py
jitu5: `dataset_value ` This is the actual loaded/saved data need to determine the type of data we're dealing with and `dataset` is the configuration object (e.g., CSVDataSet, ParquetDataSet, MemoryDataSet, etc.)

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class PipelineRunStatusHook:
+    """Collect and write pipeline and dataset events during Kedro runs."""
+
+    def __init__(self):
+        # Track times, events, and context for errors
+        self._node_start: dict[str, float] = {}
+        self._events: list[dict[str, Any]] = []
+        self.datasets: dict[str, Any] = {}
```
merelcht: Why do all values here start with underscore apart from `datasets`?

merelcht: And is there any logic to the order of the values?

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,193 @@
...
+    write_events,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PipelineRunStatusHook:
+    """Collect and write pipeline and dataset events during Kedro runs."""
+
+    def __init__(self):
```
merelcht, line 79: It would be helpful to add some more docstring here with all the values and descriptions of what they do.

package/kedro_viz/integrations/kedro/run_hooks.py
merelcht: Maybe add a todo or create a new ticket so we don't forget to remove this. This code won't work with what's on `develop` on the framework side.

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,193 @@
...
+        except ImportError:
+            pass
+        # fallback older versions
+        self.datasets = getattr(
+            catalog, "_datasets", getattr(catalog, "_data_sets", {})
+        )
+
+    @hook_impl
+    def before_pipeline_run(self, run_params: dict, pipeline) -> None:
+        """Emit start event unless this is named pipeline."""
```
merelcht: Can you elaborate this docstring? Why should the start event not be emitted when it's a named pipeline?

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @hook_impl
+    def before_node_run(self, node: KedroNode) -> None:
+        self._node_start[node.name] = perf_counter()
+        self._current_node = node
+        self._started_nodes.add(node.name)
+
+    @hook_impl
+    def after_node_run(self, node: KedroNode) -> None:
+        start = self._node_start.get(node.name, perf_counter())
```
merelcht: Can you explain this line? Is it accurate here to return `perf_counter()` as default? In what scenario does the node name not exist in `_node_start`?

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,193 @@
...
+            {
+                "event": "after_node_run",
+                "node": node.name,
+                "node_id": hash_node(node),
+                "duration_sec": duration,
+                "status": "success",
+            }
+        )
+        if self._current_node == node:
+            self._current_node = None  # clear node context
```
merelcht, line 214: Why should the node context only be cleared if the current node is the node from the hook?

merelcht (commented):
Some more comments/questions from me.

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,255 @@
...
+            "size_bytes": 1024,
+        }
+
+        assert result == expected
+        mock_compute_size.assert_called_once_with(
+            dataset_name, dataset_value, mock_datasets
+        )
+
+
+class TestComputeSize:
```
merelcht: Why do you need so much mocking for this method? Util functions should usually not need mocking to be tested.

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        dataset_name: Dataset name
+        dataset_value: Dataset data
+        datasets: Dictionary of available datasets
+
+    Returns:
+        File size in bytes, if available
+    """
+    dataset = datasets.get(dataset_name)
+    if not dataset:
+        return None
```
merelcht: It seems odd to me that we'd try to calculate the size of a dataset that doesn't exist in the full dataset dict, in what scenario is that the case?

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    try:
+        import pandas as pd
+
+        if isinstance(dataset_value, pd.DataFrame):
+            file_path = getattr(dataset, "filepath", None) or getattr(
+                dataset, "_filepath", None
+            )
+            if file_path:
+                filesystem, path = fsspec.core.url_to_fs(file_path)
+                return filesystem.size(path) if filesystem.exists(path) else None
```
merelcht: What's the benefit of having separate logic for a pandas dataframe? In the end this is pretty much identical to the below.

package/kedro_viz/integrations/kedro/run_hooks.py
jitu5: With the `run_params` we are trying to figured out if user is trying to run full/default pipeline or a specific named pipeline, For MVP we are going to show run status of only default pipeline and we are skipping in other cases.

package/kedro_viz/integrations/kedro/run_hooks.py
jitu5: Thanks, it was missed, now updated and there is no logic to the order of the values.

package/kedro_viz/integrations/kedro/run_hooks.py
jitu5: Done, Thanks.

package/kedro_viz/integrations/kedro/run_hooks.py
jitu5: Correct, not accurate to use `perf_counter()` as default, this will result in 0 duration. current node name will always exist in `_node_start`.

package/kedro_viz/integrations/kedro/run_hooks.py
jitu5: We set the `_current_node` in `before_node_run` and once same node completed in `after_node_run` we clear the `_current_node`

package/tests/test_integrations/test_hooks_utils.py
jitu5: Agreed, simplified the utils to minimise mocking.

package/kedro_viz/integrations/kedro/hooks_utils.py
jitu5: We are populating datasets in `after_catalog_created`, if some dataset is created during the node run like after saving, that might not be part if `_datasets`.

package/kedro_viz/integrations/kedro/hooks_utils.py
jitu5: Thanks, I simplified the utils.

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,241 @@
...
+            # Verify file was created
+            expected_path = project_path / events_dir / events_file
+            assert expected_path.exists()
+
+            # Verify contents
+            content = expected_path.read_text(encoding="utf8")
+            assert content == test_events_json
+
+
+class TestOnlyWhenMockingIsNecessary:
```
rashidakanchwala: we don't need a seperate class for this. it can be a test

rashidakanchwala (approved):
I will approve this once the logic for calculating data stats stats is similar to how we did using Hooks. I think for now; we do that, and if required we revise later

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,244 @@
...
+        self._current_node = node
+
+    def _clear_context(self) -> None:
+        """Clear dataset I/O context after successful operations."""
+        self._current_dataset = None
+        self._current_operation = None
+
+    @hook_impl
+    def after_catalog_created(self, catalog: Union[Any, Any]):
+        """Grab catalog datasets for size lookups and metadata access."""
```
rashidakanchwala: I think this docstring is a bit confusing. We don't do size lookups here; because most dataset are not created. 

   """
        Triggered after the Kedro DataCatalog is constructed, before pipeline execution.
        Captures references to all datasets for metadata lookups. At this point, no data
        has been loaded or saved yet — so only static dataset configuration (not size/content)
        can be accessed.
        """

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Args:
+        events: List of events to write
+        events_dir: Directory to write events to
+        events_file: Filename for events
+    """
+    try:
+        project = _find_kedro_project(Path.cwd())
+        if not project:
+            logger.warning("No Kedro project found; skipping write.")
```
rashidakanchwala:
```suggestion
            logger.warning("No Kedro project found in %s; skipping write.", Path.cwd())
```

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,244 @@
...
+        self._current_operation = operation
+        self._current_node = node
+
+    def _clear_context(self) -> None:
+        """Clear dataset I/O context after successful operations."""
+        self._current_dataset = None
+        self._current_operation = None
+
+    @hook_impl
+    def after_catalog_created(self, catalog: Union[Any, Any]):
```
rashidakanchwala:
```suggestion
    def after_catalog_created(self, catalog: Any):
```

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,285 @@
...
+    event_type = "after_dataset_loaded"
+    dataset_name = "test_dataset"
+
+    result = create_dataset_event(event_type, dataset_name)
+
+    expected = {
+        "event": event_type,
+        "dataset": dataset_name,
+        "node_id": hash_node(dataset_name),  # We know this works from previous tests
+        "status": "Available",
```
rashidakanchwala: we should use the enum here for Datasets, i think it was in another PR. but maybe not hardcode the values

package/tests/test_integrations/test_run_hooks.py
```diff
@@ -0,0 +1,199 @@
...
+@pytest.fixture()
+def sample_catalog() -> DataCatalog:
+    """Minimal DataCatalog with a single in-memory dataset."""
+    return DataCatalog({"memory": MemoryDataset()})
+
+
+# -----------------------------------------------------------------------------
+# after_catalog_created
+# -----------------------------------------------------------------------------
+def test_after_catalog_created_standard(hooks, sample_catalog):
```
rashidakanchwala: We don't need this test. RunStatus cannot work with old KedroDataCatalog.

package/tests/test_integrations/test_run_hooks.py
```diff
@@ -0,0 +1,199 @@
...
+    hooks.after_catalog_created(sample_catalog)
+    assert hooks._datasets == sample_catalog._datasets
+
+
+def test_after_catalog_created_import_error(hooks, sample_catalog, monkeypatch):
```
rashidakanchwala: I think we need to remove this one as well. because we are not going to be backward compatible. I think we keep it in the code for now; so until Kedro 1.0 is released our CI/CD passes. But let's remove tehse tests and add a note in the code to remove that fallback before we release.

rashidakanchwala (approved):
Just minor comment on the backward compatibility. Let's keep it in the code for now so that it works smoothly with Kedro 0.19 but we can remove the tests and remove the backward compatibility before we do final release. 

cc - @merelcht

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,209 @@
...
+        datasets: Dictionary of available datasets
+
+    Returns:
+        Dictionary with event data
+    """
+    event: Dict[str, Any] = {
+        "event": event_type,
+        "dataset": dataset_name,
+        "node_id": _hash_input_output(dataset_name),
+        "status": "Available",
```
rashidakanchwala: I see that we set the status of datasets later ? So I find it a bit confusing that we say 'Available' here and 'Missing' at another point in the api/responses. 

What's confusing is we have `create_dataset_event` here and `process_dataset_event` in the other PR

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,209 @@
...
+def create_dataset_event(
+    event_type: str,
+    dataset_name: str,
+    dataset_value: Any = None,
+    datasets: Optional[Dict[str, Any]] = None,
+) -> Dict[str, Any]:
+    """Generic builder for dataset load/save events.
+
+    Args:
+        event_type: Event type/name
```
ravi-kumar-pilla: [Nit] why not just `event_name` ?

package/kedro_viz/integrations/kedro/hooks_utils.py
ravi-kumar-pilla: I know this review is a bit late from me. Sorry about that. I am not sure about the discussions that happened prior to this. 

For me, I would rather prefer the status to be simple like "Success/Failed" or "1/0" instead of "Available/Missing" for dataset nodes and "success/fail" for others. Having a consistent status would be great.

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,209 @@
...
+EVENTS_DIR = VIZ_METADATA_ARGS["path"]
+EVENTS_FILE = "kedro_pipeline_events.json"
+
+
+def hash_node(node: Any) -> str:
+    """Stable ID for KedroNode or I/O reference."""
+    return _hash(str(node)) if isinstance(node, KedroNode) else _hash_input_output(node)
+
+
+def create_dataset_event(
```
ravi-kumar-pilla: For non dataset event, do we have create_node_event or something along those lines ? I could not find it here.

rashidakanchwala: I thought the same; and misunderstood that those events are somehow being written in the rest layer but they are none here because they are simple and  they are directly in the run_hooks.py code - maybe some seperation there is a good idea

package/kedro_viz/integrations/kedro/hooks_utils.py
rashidakanchwala: yeah, so datasets might have more statuses in the future, which makes them a bit different from node_status where it's just 'success'/'fail'. right now we have 'available' and 'missing', but we could easily add 'non_persistent' or 'in_memory' for MemoryDataset, and maybe something else for PartitionedDataset—not fully sure yet. if it were just two states, i’d totally prefer using a simple bool too.

and yup, i agree the hardcoded string isn’t great. we already have the enums defined in the next PR, so once both PRs are merged, we’ll update this string to use DatasetStatusEnum. let me know what you think. (https://github.com/kedro-org/kedro-viz/blob/821513a16ef6733d3a51eac11643c18b6f79cb3e/package/kedro_viz/api/rest/responses/utils.py#L33-L37)

merelcht: But "we could easily add 'non_persistent' or 'in_memory' for MemoryDataset" aren't really statuses, they are the type of dataset. I agree with @ravi-kumar-pilla here, let's just aim for simplicity and focus on what's needed _now_. It's never a good idea to design for the future.

package/kedro_viz/integrations/kedro/hooks_utils.py
merelcht: I'd argue that this method isn't really a utility function but quite core to the functioning of the hooks compared to the other methods in this file like `generate_timestamp`. 

I'd suggest only putting more generic utility stuff in this utils.py and the rest together with the hooks.

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,249 @@
...
+        self._datasets: dict[str, Any] = {}
+        self._current_node: Optional[Any] = None
+        self._current_dataset: Optional[str] = None
+        self._current_operation: Optional[str] = None
+        self._all_nodes: list[KedroNode] = []
+        self._started_nodes: set[str] = set()
+
+    def _write_events(self) -> None:
+        """Persist events list to the project's .viz JSON file."""
+        write_events(self._events)
```
merelcht: What's the reasoning here with private and public methods? It seems odd to me to have a private and public method with the same name where one just calls the other. I'd suggest to just move the `write_events` method to here and not have the private wrapper.

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,249 @@
...
+    @hook_impl
+    def after_catalog_created(self, catalog: Any):
+        """
+        Triggered after the Kedro DataCatalog is constructed, before pipeline execution.
+        Captures references to all datasets for metadata lookups. At this point, no data
+        has been loaded or saved yet — so only static dataset configuration (not size/content)
+        can be accessed.
+        """
+        try:
+            # prefer new KedroDataCatalog
```
merelcht, line 128: Flagging this again. @rashidakanchwala this is not going to be in the Viz 11.1.0 right? so then I suggest to just remove it before we forget.

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,175 @@
...
+        filepaths = ["a.txt", "b.txt"]
+
+    class Single:
+        filepath = "single.txt"
+
+    assert extract_file_paths(Multi()) == []
+    assert extract_file_paths(Single()) == ["single.txt"]
+
+
+def test_compute_size_and_get_file_size(tmp_path):
```
merelcht: This should be 2 separate tests.

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,175 @@
...
+    fp = tmp_path / "file.txt"
+    fp.write_text("abcdef")  # 6 bytes
+
+    assert get_file_size(str(fp)) == 6
+
+    datasets = {"ds": type("Local", (), {"filepath": str(fp)})()}
+    assert compute_size("ds", datasets) == 6
+
+
+def test_compute_size_fallbacks():
```
merelcht: Again should be 2 tests.

merelcht (commented):
Finished reviewing this as well and left some small comments. Same comment here on breaking up unit tests into smaller tests.

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,268 @@
...
+
+    @hook_impl
+    def after_catalog_created(self, catalog: Any):
+        """
+        Triggered after the Kedro DataCatalog is constructed, before pipeline execution.
+        Captures references to all datasets for metadata lookups. At this point, no data
+        has been loaded or saved yet — so only static dataset configuration (not size/content)
+        can be accessed.
+        """
+        from kedro.io import KedroDataCatalog
```
ravi-kumar-pilla: Not sure if it is in this PR or the following PR but should we remove this import ?

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,268 @@
...
+        # We add events only for full/default pipeline as for MVP we only support
+        # full/default pipeline.
+        if len(self._all_nodes) == 0:
+            return
+
+        self._events.append(event)
+        if flush:
+            write_events(self._events)
+
+    def _set_context(self, dataset: str, operation: str, node: KedroNode) -> None:
```
ravi-kumar-pilla: The method name might interfere with KedroContext...worth renaming it if possible or add a TODO if planned for refactor post 1.0

package/kedro_viz/integrations/kedro/run_hooks.py
ravi-kumar-pilla: can we do that as part of `self._clear_context()` ?

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -0,0 +1,268 @@
...
+
+    @hook_impl
+    def before_pipeline_run(self, run_params: dict, pipeline) -> None:
+        """
+        Emit start event based on run_params values.
+
+        Records the beginning of a pipeline execution
+        only for full/default pipeline.
+        """
+        if not is_default_run(run_params):
```
ravi-kumar-pilla, line 151: I saw some explanation for this but why do we restrict status recording for only default runs ?
If the user runs specific part of the pipeline as they have huge default pipeline, why dont we log the latest run and show status ? Considering the status is anyway a static reading from a file

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,180 @@
...
+# Fixtures & helpers
+# -----------------------------------------------------------------------------
+
+
+@pytest.fixture
+def sample_node() -> KedroNode:
+    return node(lambda x: x, inputs="a", outputs="b", name="sample")
+
+
+class DummyDataset:
```
ravi-kumar-pilla: cannot find where this is used

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,178 @@
...
+        OSError,
+        ValueError,
+        FileNotFoundError,
+        AttributeError,
+    ) as exc:  # pragma: no cover
+        logger.debug("Unable to get file size for %s: %s", file_path, exc)
+        return None
+
+
+def compute_size(dataset_name: str, datasets: Dict[str, Any]) -> Optional[int]:
```
ravi-kumar-pilla: [Nit] I feel compute_size should return size and returning `None` seems odd. Can we change the definition to `def compute_size(dataset: Any)` . So pass the dataset if exists and compute size returns the actual size instead of returning None and do other checks here ?

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Returns:
+        String representation of the current timestamp.
+
+    """
+    current_ts = datetime.now(tz=timezone.utc).strftime(TIME_FORMAT)
+    return current_ts
+
+
+def is_default_run(run_params: dict) -> bool:
```
ravi-kumar-pilla, line 132: would like to know if this was a design decision for the MVP ? I still feel we should record the runs irrespective of the params passed and show the status how we logged our events. Since we document saying the status view always shows the latest run information

ravi-kumar-pilla: May be I am missing complete picture here.

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,180 @@
...
+    assert compute_size("ds", {"ds": NoPath()}) is None
+
+
+def test_get_file_size_nonexistent(tmp_path):
+    assert get_file_size(str(tmp_path / "no.txt")) is None
+
+
+# -----------------------------------------------------------------------------
+# create_dataset_event
+# -----------------------------------------------------------------------------
```
ravi-kumar-pilla: [Nit] these commenting blocks are good for code separation but I don't think we use this style in our codebase

package/tests/test_integrations/test_hooks_utils.py
```diff
@@ -0,0 +1,180 @@
...
+    events = [{"event": "x"}]
+
+    monkeypatch.setattr(
+        "kedro_viz.integrations.kedro.hooks_utils._find_kedro_project",
+        lambda _: tmp_path,
+    )
+
+    captured = {}
+
+    def _fake(project, events_dir, events_file, events_json):
```
ravi-kumar-pilla: [Nit] worth renaming to understand what are we faking here

package/tests/test_integrations/test_run_hooks.py
```diff
@@ -0,0 +1,201 @@
...
+    return node(lambda x: x, inputs="in", outputs="out", name="test_node")
+
+
+@pytest.fixture()
+def sample_pipeline(sample_node) -> Pipeline:
+    return Pipeline([sample_node])
+
+
+@pytest.fixture()
+def sample_catalog() -> DataCatalog:
```
ravi-kumar-pilla: can we use example catalog from conftest ? - https://github.com/kedro-org/kedro-viz/blob/main/package/tests/conftest.py#L402

package/tests/test_integrations/test_run_hooks.py
```diff
@@ -0,0 +1,201 @@
...
+    return PipelineRunStatusHook()
+
+
+@pytest.fixture()
+def sample_node() -> KedroNode:
+    return node(lambda x: x, inputs="in", outputs="out", name="test_node")
+
+
+@pytest.fixture()
+def sample_pipeline(sample_node) -> Pipeline:
```
ravi-kumar-pilla: can we use example_pipelines['__default__'] from https://github.com/kedro-org/kedro-viz/blob/main/package/tests/conftest.py#L50C5-L50C22 ?

package/tests/test_integrations/test_run_hooks.py
```diff
@@ -0,0 +1,201 @@
...
+
+# Remove below test once we release Kedro 1.0.0
+def test_after_catalog_created_import_error(hooks, sample_catalog, monkeypatch):
+    """Branch where `kedro.io` exists but lacks KedroDataCatalog."""
+    monkeypatch.setitem(sys.modules, "kedro.io", ModuleType("kedro.io"))
+    hooks.after_catalog_created(sample_catalog)
+    assert hooks._datasets == sample_catalog._datasets
+
+
+def test_after_catalog_created_with_kedro_data_catalog(
```
ravi-kumar-pilla: This test is not needed if run_status supports only >= Kedro 1.0

package/tests/test_integrations/test_run_hooks.py
```diff
@@ -0,0 +1,201 @@
...
+    assert hooks._events == []
+
+
+# -----------------------------------------------------------------------------
+# Dataset I/O workflow
+# -----------------------------------------------------------------------------
+def test_dataset_loading_and_saving_workflow(hooks, sample_node):
+    hooks._all_nodes = [sample_node]
+
+    # Loading branch
```
ravi-kumar-pilla:
```suggestion
    # Loading dataset
```

package/tests/test_integrations/test_run_hooks.py
```diff
@@ -0,0 +1,201 @@
...
+    hooks.before_dataset_loaded("dataset", sample_node)
+    hooks.after_dataset_loaded("dataset", {"foo": "bar"})
+    assert hooks._events[-1]["event"].startswith("after_dataset_loaded")
+
+    # Saving branch
```
ravi-kumar-pilla:
```suggestion
    # Saving dataset
```

package/kedro_viz/integrations/kedro/hooks_utils.py
jitu5: @ravi-kumar-pilla Yes we decided to keep for default pipeline run only. Also on frontend in workflow tab we disabled the pipeline list dropdown.

we made this decision to avoid a mismatch between the run data coming from the backend and what the user sees on the flowchart. For example, if a user runs the pipeline from Node A to Node B, but the UI shows the default pipeline view, it would create a broken or confusing experience. Additionally, we don’t have detailed data about which specific part of the pipeline was run.
To avoid this complexity, we decided to support only default pipeline runs in MVP1 and plan to handle custom runs in a future phase.

Correct me if I am wrong @Huongg

package/kedro_viz/integrations/kedro/run_hooks.py
jitu5: Please refer other comment.

Issue #2322, opened 2025-05-06
Backend – surface and save Run Status in .viz/


Huongg:
## Description
Enable the Kedro-Viz backend to expose non-realtime run status updates stored in .viz/kedro_pipeline_events.json, so the frontend can consume and display them.

## Possible Implementation

**MVP 1 Scope suggestion**
* Capture and save run status updates to .viz/kedro_pipeline_events.json:
    * Triggered when the user runs kedro run
    * (Optional/To be confirmed): Also support a dedicated command, e.g., kedro run --debugging
* For each pipeline run, capture:
    * Dataset nodes: size and completion status
    * Function nodes: execution time and completion status

 **To-do**

- [ ] Implement Kedro hook to capture run metadata
- [ ] Write run metadata to .viz/kedro_pipeline_events.json – Almost DONE (needs minor refinements from the current implementation)
- [ ] Read from JSON file and expose data to the frontend – To Do
    * NOTE: JSON file may grow over time — implement trimming, size limits if needed

## Checklist

- [ ] Include labels so that we can categorise your feature request

PR #2350, opened 2025-05-29
Run status: Process Raw data and prepare the response

jitu5:
## Description

Resolves https://github.com/kedro-org/kedro-viz/issues/2322

This pull request introduces a new API endpoint to retrieve structured pipeline run events for visualization purposes. It includes the implementation of the response model, utility functions for data processing, and the integration of the endpoint into the router. The changes are grouped into three themes: API endpoint implementation, data processing utilities, and logging enhancements.

### API endpoint implementation:
* Added `StructuredRunEventAPIResponse` model and related classes (`NodeInfo`, `DatasetInfo`, `PipelineInfo`) to define the structure of the response for the `/run-events` endpoint. 
* Implemented the `/run-events` endpoint in `package/kedro_viz/api/rest/router.py` to return structured pipeline run events using the `get_run_events_response` function. 
* Imported `StructuredRunEventAPIResponse` and `get_run_events_response` in `package/kedro_viz/api/rest/router.py`. 

### Data processing utilities:
* Added utility functions `safe_int`, `convert_status_to_enum`, and `calculate_pipeline_duration` in `package/kedro_viz/utils.py` to handle data transformations and calculations for pipeline events. 

### Logging enhancements:
* Introduced logging in the `package/kedro_viz/api/rest/responses/run_events.py` and `package/kedro_viz/utils.py` files to provide better error handling and debugging information. 

###  JSON structure sent to frontend (example)
```json
{
  "datasets": {
    "dataset_hash_id": {
      "name": "missing_file.csv",
      "size_bytes": 0,
      "status": "Missing",  // "Available" | "Missing"
      "error": {
        "message": "File not found: /path/to/missing_file.csv",
        "error_node": "process_data",
        "error_operation": "loading"  // or "saving"
        "traceback": "traceback( error-traceback )"
      }
    }
  },
  "nodes": {
    "node_hash_id": {
      "status": "Fail",
      "error": {
        "message": "Missinf data argument",
        "traceback": "traceback( error-traceback )"
      }
    }
  },
  "pipeline": {
      "run_id": "749b518d-4402-4c2d-9d0e-689cd8eb8c2f",
      "start_time": "2025-06-04T10.29.08.126723Z",
      "end_time": "2025-06-04T10.29.29.509414Z",
      "total_duration_sec": 11.020523963023152,
      "status": "completed",
      "error": null  
  }
}
```

### Flow
Screenshot 2025-05-23 at 2 41 29 p m

 
 ## QA notes
 
 - Checkout the feature/run-status-nrt and current branch and merge
 - At root of kedro-viz run `pip install -e package`
 - In demo_project run `kedro run`
 - Above step will produce/update `kedro_pipeline_events.json` in `.viz` folder
 - Run `kedro viz run`
- At new end point you can check the response
 
 
 
 **Note:** Only adding test cases is remaining.


package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,264 @@
...
+        Structured API response object
+    """
+    try:
+        kedro_project_path = _find_kedro_project(Path.cwd())
+        if not kedro_project_path:
+            logger.warning("Could not find a Kedro project to load kedro pipeline events file")
+            return StructuredRunEventAPIResponse()
+            
+        pipeline_events_file_path = Path(
+            f"{kedro_project_path}/{VIZ_METADATA_ARGS['path']}/kedro_pipeline_events.json"
```
rashidakanchwala: Maybe we define this as global constants since we use it in hooks and here 
VIZ_METADATA_DIR = VIZ_METADATA_ARGS["path"]
PIPELINE_EVENTS_FILENAME = "kedro_pipeline_events.json"
PIPELINE_EVENT_FULL_PATH = Path(VIZ_METADATA_DIR) / PIPELINE_EVENTS_FILENAME

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,264 @@
...
+    datasets: Dict[str, DatasetInfo],
+    node_id: str,
+    dataset_name: str,
+    size_bytes: Optional[int],
+    status: str,
+    overwrite: bool = False
+) -> None:
+    """Create or update DatasetInfo, controlling size overwrite."""
+    status_enum = convert_status_to_enum(status, DatasetStatus.AVAILABLE)
+    info = datasets.get(node_id)
```
rashidakanchwala: this part is confusing, why are we using node_id here ?

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+    calculate_pipeline_duration,
+    convert_status_to_enum,
+    safe_int,
+)
+
+logger = logging.getLogger(__name__)
+
+
+# Constants for event types
+class EventType:
```
ravi-kumar-pilla:
```suggestion
class EventType(str, Enum):
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+    BEFORE_PIPELINE_RUN = "before_pipeline_run"
+    AFTER_PIPELINE_RUN = "after_pipeline_run"
+    ON_PIPELINE_ERROR = "on_pipeline_error"
+    AFTER_NODE_RUN = "after_node_run"
+    ON_NODE_ERROR = "on_node_error"
+    AFTER_DATASET_LOADED = "after_dataset_loaded"
+    AFTER_DATASET_SAVED = "after_dataset_saved"
+
+
+class PipelineStatus:
```
ravi-kumar-pilla:
```suggestion
class PipelineStatus(str, Enum):
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+
+
+class PipelineInfo(BaseModel):
+    """Information about the pipeline run."""
+
+    run_id: str = "default-run-id"
+    start_time: Optional[str] = None
+    end_time: Optional[str] = None
+    total_duration_sec: float = 0.0
+    status: str = PipelineStatus.COMPLETED
```
ravi-kumar-pilla:
```suggestion
    status: PipelineStatus = PipelineStatus.COMPLETED
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+    error: Optional[Union[str, NodeErrorInfo]] = None
+
+
+class DatasetInfo(BaseModel):
+    """Information about a dataset."""
+
+    name: str
+    size_bytes: int = 0
+    status: DatasetStatus = DatasetStatus.AVAILABLE
+    error: Optional[Union[str, DatasetErrorInfo]] = None
```
ravi-kumar-pilla: I see error is sometimes str and sometimes class. Can we have something like below and always use error to have a type (NodeErrorInfo or DatasetErrorInfo) and may be PipelineErrorInfo too ?

```python
class BaseErrorInfo(BaseModel):
    message: str
    traceback: Optional[str] = None

class NodeErrorInfo(BaseErrorInfo):
    pass

class DatasetErrorInfo(BaseErrorInfo):
    error_node: Optional[str] = None
    error_operation: Optional[str] = None
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
+"""Response for run status API endpoint."""
+
+import json
+import logging
+import uuid
+from enum import Enum
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
```
merelcht: This is the old way of type annotations, see https://peps.python.org/pep-0585/. In the framework we've already moved to using the in-built types available since Python 3.9. Let's start doing that in the Viz repo too.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+
+    run_id: str = "default-run-id"
+    start_time: Optional[str] = None
+    end_time: Optional[str] = None
+    total_duration_sec: float = 0.0
+    status: str = PipelineStatus.COMPLETED
+    error: Optional[str] = None
+
+
+class StructuredRunStatusAPIResponse(BaseModel):
```
merelcht: This is quite a long class name, could it just be `RunStatusAPIResponse`? As far as I can tell there's no unstructured API response.

package/kedro_viz/utils.py
```diff
@@ -134,3 +139,54 @@ def run_spinner():
...
+
+def safe_int(value: Any) -> int:
+    """Safely parse an integer value."""
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return 0
+
+
+def convert_status_to_enum(status: Optional[str], default: T) -> T:
```
merelcht: What is this for?

merelcht: And shouldn't this go in the `responses/utils.py`?

package/kedro_viz/utils.py
```diff
@@ -134,3 +139,54 @@ def run_spinner():
     def stop(self):
         self.stop_running = True
         self._spinner_thread.join()
+
+
+def safe_int(value: Any) -> int:
```
merelcht: This method seems like overkill to me. It's only used in one place, where we could easily just do a try/except block. It also looks like `size_bytes` is set to 0 by default in most places already, so it's not necessary to be so overly cautious.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+    # Find pipeline end event (last pipeline-related event)
+    end_events = [
+        event
+        for event in reversed(events)
+        if event.get("event")
+        in {EventType.AFTER_PIPELINE_RUN, EventType.ON_PIPELINE_ERROR}
+    ]
+
+    if end_events:
+        end_event = end_events[0]
```
merelcht: Can there ever be more than 1 event? And why store it in a list if that's not possible?

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+        if "timestamp" in end_event:
+            pipeline_info.end_time = end_event["timestamp"]
```
merelcht: If `end_event` is a dictionary, you could do `end_event.get("timestamp", None)`. It's generally better to fetch things from a dictionary with the `get()` method.

merelcht: And I see you're using `get()` in the lines below too, so it's also better for consistency in the codebase.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,447 @@
...
+def _process_node_completion_event(
+    event: Dict[str, Any], nodes: Dict[str, NodeInfo]
+) -> None:
+    """Process after_node_run event to update node information.
+
+    Args:
+        event: Event dictionary containing node completion data
+        nodes: Dictionary of node info objects to update
+    """
+    node_id = event.get("node_id")
```
merelcht: In what case is there no `node_id`?

package/kedro_viz/api/rest/router.py
```diff
@@ -62,6 +66,53 @@ async def get_version():
     return get_version_response()
 
 
+@router.get("/run-status", response_model=StructuredRunStatusAPIResponse)
+async def get_run_status():
```
merelcht: Will this return all run statuses available for a project?

package/kedro_viz/api/rest/responses/run_events.py
jitu5: This is not relevant anymore.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Thanks, updated.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Thanks, updated.

package/kedro_viz/utils.py
jitu5: Yes, correct looks overkill, moved to try/except block.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Thanks updated.

package/kedro_viz/api/rest/router.py
jitu5: It will return only the latest or last run status.

package/kedro_viz/api/rest/responses/utils.py
```diff
@@ -2,14 +2,18 @@
...
+from typing import Any, Dict, Optional, TypeVar
 
 import orjson
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import ORJSONResponse
 
 logger = logging.getLogger(__name__)
 
+T = TypeVar("T", bound=Enum)
+
```
rashidakanchwala, line 23: what's this? not very clear can we use descriptive names

package/kedro_viz/api/rest/router.py
rashidakanchwala: make sense to rename it get_last_run_status()

package/kedro_viz/utils.py
jitu5: This is to converts string statuses from events (like "success", "SUCCESS", "fail") into clean Python enum objects, with case-insensitive matching. Also moved to `responses/utils.py` Thanks.

package/kedro_viz/api/rest/responses/utils.py
jitu5: Updated, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Here we are capturing the timestamp either from end event `AFTER_PIPELINE_RUN` or `ON_PIPELINE_ERROR`. In the case if  we have both the events then I will take the latest one. But yes I have updated for better clarity and moved from list.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: The `node_id` comes from each event captured by the hooks during the run. Initially, I added checks in case something went wrong while creating it. However, I realised that `node_id` will always be present, so I removed those checks.

package/kedro_viz/api/rest/router.py
ravi-kumar-pilla: or `get_latest_run_status`

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,461 @@
...
+    COMPLETED = "completed"
+    FAILED = "failed"
+
+
+class NodeStatus(str, Enum):
+    """Enum representing the possible statuses of a node."""
+
+    SUCCESS = "Success"
+    FAIL = "Fail"
+
```
merelcht: Why use completed & failed for pipeline status but success & fail for node status? It makes more sense to me to use consistent terminology for these.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,461 @@
...
+    error: Optional[DatasetErrorInfo] = None
+
+
+class PipelineInfo(BaseModel):
+    """Information about the pipeline run."""
+
+    run_id: str = "default-run-id"
+    start_time: Optional[str] = None
+    end_time: Optional[str] = None
+    total_duration_sec: float = 0.0
```
merelcht: In NodeInfo this is named `duration_sec`, let's make it consistent and give them both the same name.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,461 @@
...
+    )
+
+    if end_event:
+        if "timestamp" in end_event:
+            pipeline_info.end_time = end_event.get("timestamp", None)
+
+        if end_event.get("event") == EventType.ON_PIPELINE_ERROR:
+            pipeline_info.status = PipelineStatus.FAILED
+            error_message = end_event.get("error", "Unknown pipeline error")
+            traceback_message = end_event.get("traceback")
```
merelcht: In what case is there no error (so you need to provide a default) but always a traceback? 

There's a lot of dict accessing in the code and I think we should be consistent in how we handle this in terms of providing a default or not. On the one hand you shouldn't overdo it with defaults, because that might then hide a real issue in data flow if it's missing, but on the other if a value can be optional it might be better to have default.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Thanks, kept  Failed/Successful as status for both node and pipeline.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Update, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Thanks, Kept it consistent for both error & trackback, and Tired to keep consistent for dict accessing for other places as well.

package/tests/test_api/test_rest/test_responses/test_run_events.py
```diff
@@ -0,0 +1,345 @@
+from __future__ import annotations
+
+import json
+from datetime import datetime, timedelta
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+from kedro_viz.api.rest.responses import run_events as rv
```
rashidakanchwala: lets call it run_events. rv is not understandable

package/tests/test_api/test_rest/test_responses/test_run_events.py
```diff
@@ -0,0 +1,345 @@
...
+from kedro_viz.utils import _hash_input_output
+
+
+# --------------------------------------------------------------------------- #
+# Helper to make building events more readable
+# --------------------------------------------------------------------------- #
+def _event(kind: rv.EventType, **extra: Any) -> dict[str, Any]:
```
rashidakanchwala: I think we can consider changing the function name here—_event doesn’t clearly convey what the function actually does. It might be more intuitive if the name reflected the action or purpose of the function, since right now it’s a bit ambiguous.

package/tests/test_api/test_rest/test_responses/test_run_events.py
```diff
@@ -0,0 +1,345 @@
...
+    err = datasets[hashed].error
+    assert err is not None and err.message == "second"
+
+
+# --------------------------------------------------------------------------- #
+# _finalize_pipeline_info
+# --------------------------------------------------------------------------- #
+def test_pipeline_finalisation_populates_run_id_and_duration() -> None:
+    pipeline = rv.PipelineInfo()
+    nodes = {"a": rv.NodeInfo(duration_sec=2), "b": rv.NodeInfo(duration_sec=3)}
```
rashidakanchwala: Just a small suggestion—do you think we could use more descriptive variable names instead of "a" and "b" here? I get that they’re just test values, but even in mocks or tests, meaningful names can really help with readability for future contributors.

package/tests/test_api/test_rest/test_responses/test_run_events.py
```diff
@@ -0,0 +1,345 @@
...
+            node_id="ds",
+            dataset="d",
+            size_bytes=1,
+            status="Available",
+        ),
+        _event(rv.EventType.ON_NODE_ERROR, node_id="n_bad", error="x", traceback="t"),
+        _event(
+            rv.EventType.ON_PIPELINE_ERROR,
+            timestamp=(now + timedelta(seconds=3)).isoformat(),
+            dataset="d",
```
rashidakanchwala: Let’s update the test code to avoid using single- or two-letter variables

rashidakanchwala:
```
   "dataset": "customers.csv",
    "node": "load_customers_node",
    "node_id": "load_customers_node",
    "error": "DatasetLoadError: Failed to read customers.csv",
    "traceback": "Traceback (most recent call last): ...",
    "operation": "load",
```

see like above example

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,453 @@
...
+    """Constants for pipeline statuses."""
+
+    SUCCESSFUL = "Successful"
+    FAILED = "Failed"
+
+
+class NodeStatus(str, Enum):
+    """Enum representing the possible statuses of a node."""
+
+    SUCCESSFUL = "Successful"
```
rashidakanchwala: we should export these classes, maybe move it to utils so we can use these ENUMS also in the hooks instead of hardcoding the values.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,453 @@
...
+class NodeErrorInfo(BaseErrorInfo):
+    """Information about a node error."""
+
+    pass
+
+
+class DatasetErrorInfo(BaseErrorInfo):
+    """Information about a dataset error."""
+
+    error_node: Optional[str] = None
```
rashidakanchwala, line 62: i am not sure what error_node is it? is this a KedroNode, because it's a str. If it's the name of the dataset, then i would call it dataset_name here.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,453 @@
...
+    """Information about a pipeline error."""
+
+    pass
+
+
+class NodeInfo(BaseModel):
+    """Information about a node."""
+
+    status: NodeStatus = NodeStatus.SUCCESSFUL
+    duration_sec: float = 0.0
```
rashidakanchwala:
```suggestion
    duration: float = 0.0 #duration in seconds
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,453 @@
...
+    error: Optional[NodeErrorInfo] = None
+
+
+class DatasetInfo(BaseModel):
+    """Information about a dataset."""
+
+    name: str
+    size_bytes: int = 0
```
rashidakanchwala:
```suggestion
    size: int = 0 #size in bytes
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,453 @@
...
+
+class RunStatusAPIResponse(BaseModel):
+    """Format for structured run status endpoint response."""
+
+    nodes: dict[str, NodeInfo] = Field(default_factory=dict)
+    datasets: dict[str, DatasetInfo] = Field(default_factory=dict)
+    pipeline: PipelineInfo = Field(default_factory=PipelineInfo)
+
+
+def _create_or_update_dataset_info(
```
rashidakanchwala: I think we should have the common if logic, and based on that split this into two funcs 

`create_dataset_info`, and `update_dataset_info`

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,453 @@
...
+    if not dataset_info.name and dataset_name:
+        dataset_info.name = dataset_name
+
+    if size_bytes is not None and (overwrite_size or dataset_info.size_bytes == 0):
+        dataset_info.size_bytes = size_bytes
+
+    dataset_info.status = status_enum
+
+
+def _extract_pipeline_timing_and_status(
```
rashidakanchwala:
```suggestion
def _update_pipeline_info_from_events
```

package/tests/test_api/test_rest/test_responses/test_run_events.py
jitu5: Done, Thanks.

package/tests/test_api/test_rest/test_responses/test_run_events.py
jitu5: Done, Thanks.

package/tests/test_api/test_rest/test_responses/test_run_events.py
jitu5: Done, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Done, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Done, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Done, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Done, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Done, Thanks.

package/tests/test_api/test_rest/test_responses/test_run_events.py
jitu5: Done, Thanks.

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Here `error_node` is the name of the node who tried to load/save the failed dataset, thats the reason why I kept `error_node`.

rashidakanchwala: ok it was a bit confusing maybe we can add this information in the docstring so that it's clear

jitu5: Sure, let me add that.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    event: dict[str, Any], nodes: dict[str, NodeInfo]
+) -> None:
+    """Process after_node_run event to update node information.
+
+    Args:
+        event: Event dictionary containing node completion data
+        nodes: Dictionary of node info objects to update
+    """
+    node_id = event.get("node_id", "unknown_node")
+    status = event.get("status", "Successful")
```
rashidakanchwala, line 218: can we make this lower case `successful`

package/kedro_viz/api/rest/responses/utils.py
```diff
@@ -2,14 +2,40 @@
...
 
+EnumType = TypeVar("EnumType", bound=Enum)
+
+
+# Shared status enums for pipeline, nodes, and datasets
+class PipelineStatus(str, Enum):
+    """Constants for pipeline statuses."""
+
+    SUCCESSFUL = "Successful"
+    FAILED = "Failed"
```
rashidakanchwala:
```suggestion
    SUCCESSFUL = "successful"
    FAILED = "failed"
```

nit

package/kedro_viz/api/rest/responses/run_events.py
rashidakanchwala:
```suggestion
    status = event.get("status", "successful")
```

nit - so that's its easier in the FE

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    if node_id not in datasets:
+        _create_dataset_info(
+            datasets,
+            node_id,
+            dataset_name,
+            size,
+            status_enum,
+        )
+    else:
+        _update_dataset_info(
```
rashidakanchwala, line 320: I see now why you kept them together. Initially I thought separating them would avoid the extra boolean flag, since a function should ideally do one thing. In this case, combining them into an `_upsert_dataset_info` with an `overwrite_size` parameter makes perfect sense. Thanks for your patience, and sorry for the confusion!

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    """Information about a node error."""
+
+    pass
+
+
+class DatasetErrorInfo(BaseErrorInfo):
+    """Information about a dataset error.
+
+    Attributes:
+        error_node: Optional; the name of the node who tried to load/save the failed dataset
```
merelcht:
```suggestion
        error_node: Optional; the name of the node that tried to load/save the failed dataset
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+
+    pass
+
+
+class NodeInfo(BaseModel):
+    """Information about a node."""
+
+    status: NodeStatus = NodeStatus.SUCCESSFUL
+    duration: float = 0.0  # duration in seconds
+    error: Optional[NodeErrorInfo] = None
```
merelcht: For consistency with the dataset and pipeline info models, let's stick to the same argument order as much as possible, so `status` and `error` as the last two.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    nodes: dict[str, NodeInfo] = Field(default_factory=dict)
+    datasets: dict[str, DatasetInfo] = Field(default_factory=dict)
+    pipeline: PipelineInfo = Field(default_factory=PipelineInfo)
+
+
+def _create_dataset_info(
+    datasets: dict[str, DatasetInfo],
+    dataset_id: str,
+    dataset_name: str,
+    size: Optional[int],  # size in bytes
```
merelcht: The doc string already explains this is in bytes, so the comment seems a bit unnecessary here and in other places.

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    )
+
+
+def _update_dataset_info(
+    datasets: dict[str, DatasetInfo],
+    dataset_id: str,
+    dataset_name: str,
+    size: Optional[int],  # size in bytes
+    status: DatasetStatus,
+    overwrite_size: bool = False,
```
merelcht, line 165: Why is this argument needed?

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    """Process dataset error events to update dataset and node error information.
+
+    Args:
+        event: Event dictionary containing dataset error data
+        datasets: Dictionary of dataset info objects to update
+        nodes: Dictionary of node info objects to update
+        pipeline_info: Pipeline info object to potentially update
+    """
+    dataset_name = event.get("dataset", "")
+    node_id = event.get("node_id", "")
```
merelcht, line 345: Above in `_process_dataset_event` and in the similar functions handling nodes there's a default value here. I'm also wondering why dataset_name and node_name don't have a default name but just empty string?

package/tests/test_api/test_rest/test_responses/test_run_events.py
```diff
@@ -0,0 +1,389 @@
...
+# --------------------------------------------------------------------------- #
+def _make_event(kind: run_events.EventType, **extra: Any) -> dict[str, Any]:
+    data: dict[str, Any] = {"event": kind}
+    data.update(extra)
+    return data
+
+
+# --------------------------------------------------------------------------- #
+# _create_or_update_dataset_info
+# --------------------------------------------------------------------------- #
```
merelcht: Instead of these blocks you can add test classes: see e.g. https://github.com/kedro-org/kedro/blob/main/tests/io/test_core.py#L210

package/tests/test_api/test_rest/test_responses/test_run_events.py
```diff
@@ -0,0 +1,389 @@
...
+
+    # second call: update existing entry with new name and size
+    run_events._update_dataset_info(
+        datasets,
+        dataset_id="sale_dataset",
+        dataset_name="sales",
+        size=111,
+        status=run_events.DatasetStatus.AVAILABLE,
+    )
+    assert (
```
merelcht: Ideally unit tests only test one method at a time, so let's split the creation and update into separate tests.

package/tests/test_api/test_rest/test_responses/test_run_events.py
```diff
@@ -0,0 +1,389 @@
...
+        datasets["sale_dataset"].name == "sales"
+        and datasets["sale_dataset"].size == 111
+    )
+
+
+# --------------------------------------------------------------------------- #
+# _update_pipeline_info_from_events
+# --------------------------------------------------------------------------- #
+def test_pipeline_timing_success_and_failure() -> None:
```
merelcht: Again let's split success and failure into two tests

package/tests/test_api/test_rest/test_responses/test_utils.py
```diff
@@ -41,3 +47,32 @@ def test_get_encoded_response(mocker):
...
+
+
+def test_convert_status_to_enum_cases():
+    assert (
+        convert_status_to_enum("successful", NodeStatus.FAILED) is NodeStatus.SUCCESSFUL
+    )
+    assert convert_status_to_enum("FAILED", NodeStatus.SUCCESSFUL) is NodeStatus.FAILED
+    # Unknown / None → default
+    assert (
+        convert_status_to_enum("does-not-exist", NodeStatus.SUCCESSFUL)
```
merelcht: Shouldn't it error out if the status doesn't exist?

package/tests/test_api/test_rest/test_responses/test_utils.py
```diff
@@ -41,3 +47,32 @@ def test_get_encoded_response(mocker):
     mock_jsonable_encoder.assert_called_once_with(mock_response)
     mock_encode_to_human_readable.assert_called_once_with(mock_response)
     assert result == mock_encoded_response
+
+
+def test_convert_status_to_enum_cases():
+    assert (
+        convert_status_to_enum("successful", NodeStatus.FAILED) is NodeStatus.SUCCESSFUL
+    )
+    assert convert_status_to_enum("FAILED", NodeStatus.SUCCESSFUL) is NodeStatus.FAILED
```
merelcht: You can parameterise this instead of having all these asserts.

package/kedro_viz/utils.py
```diff
@@ -4,9 +4,10 @@
 import sys
 import threading
 import time
+from enum import Enum
```
ravi-kumar-pilla: if not used anywhere, you can remove this ?

package/kedro_viz/utils.py
```diff
@@ -4,9 +4,10 @@
...
 from itertools import cycle
 from pathlib import Path
-from typing import Any, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple, TypeVar
```
ravi-kumar-pilla: if not used anywhere, you can remove this ?

package/kedro_viz/api/rest/responses/utils.py
```diff
@@ -2,14 +2,40 @@
...
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import ORJSONResponse
 
 logger = logging.getLogger(__name__)
 
+EnumType = TypeVar("EnumType", bound=Enum)
+
+
+# Shared status enums for pipeline, nodes, and datasets
+class PipelineStatus(str, Enum):
```
ravi-kumar-pilla:
```suggestion
class RunEventStatus(str, Enum):
```
OR
```suggestion
class EventStatus(str, Enum):
```

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+
+    pass
+
+
+class DatasetErrorInfo(BaseErrorInfo):
+    """Information about a dataset error.
+
+    Attributes:
+        error_node: Optional; the name of the node who tried to load/save the failed dataset
+        error_operation: Optional; the operation that caused the error"""
```
ravi-kumar-pilla:
```suggestion
        error_operation (Optional[str]): The dataset operation (i.e., loading/saving) resulting in an error"""
```

ravi-kumar-pilla: [Nit] Some classes have docstrings with attributes and some don't. Also the convention on viz for docstrings has been like - https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/api/rest/responses/nodes.py#L27

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    size: int = 0  # size in bytes
+    status: DatasetStatus = DatasetStatus.AVAILABLE
+    error: Optional[DatasetErrorInfo] = None
+
+
+class PipelineInfo(BaseModel):
+    """Information about the pipeline run."""
+
+    run_id: str = "default-run-id"
+    start_time: Optional[str] = None
```
ravi-kumar-pilla, line 115: why is it that NodeInfo does not have start_time/end_time but just duration ?

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+    if not dataset_info.name and dataset_name:
+        dataset_info.name = dataset_name
+
+    if size is not None and (overwrite_size or dataset_info.size == 0):
+        dataset_info.size = size
+
+    dataset_info.status = status
+
+
+def _update_pipeline_info_from_events(
```
ravi-kumar-pilla:
```suggestion
def _update_pipeline_info(
```

ravi-kumar-pilla: to be consistent with https://github.com/kedro-org/kedro-viz/pull/2350/files#diff-6ba06200dc99bed158c69b8b97e93b230f6e1ecba25b6834acaa2d0d39499bc7R128

package/kedro_viz/api/rest/responses/run_events.py
```diff
@@ -0,0 +1,464 @@
...
+        if not kedro_project_path:
+            logger.warning(
+                "Could not find a Kedro project to load pipeline events file"
+            )
+            return RunStatusAPIResponse()
+
+        pipeline_events_file_path = PIPELINE_EVENT_FULL_PATH
+
+        if not pipeline_events_file_path.exists():
+            logger.warning(f"Run events file {pipeline_events_file_path} not found")
```
ravi-kumar-pilla:
```suggestion
            logger.warning(f"Run events file {pipeline_events_file_path} not found. It may be due to missing `kedro run`")
```
Not sure of the wording but should we hint the user what might be the reason ?

package/tests/test_api/test_rest/test_responses/test_utils.py
rashidakanchwala: i think it's because it's an unknown string -- according to the func it fallsback to default which in our case is SUCCESS.

package/kedro_viz/api/rest/responses/utils.py
rashidakanchwala: I just changed everything to `Status` because it's the same for all three of them now

package/kedro_viz/api/rest/responses/run_events.py
rashidakanchwala: we don't calculate it. PipelineInfo has it because it's part of the design to show timestamp.

package/kedro_viz/api/rest/responses/utils.py
```diff
@@ -16,27 +16,13 @@
 
 
 # Shared status enums for pipeline, nodes, and datasets
-class PipelineStatus(str, Enum):
+class Status(str, Enum):
```
ravi-kumar-pilla:
```suggestion
class RunEventStatus(str, Enum):
```

Status seems very generic. May be something related to the events should provide clear information.

package/kedro_viz/api/rest/responses/utils.py
```diff
@@ -16,27 +16,13 @@
...
     """Constants for pipeline statuses."""
```
ravi-kumar-pilla:
```suggestion
    """Enum representing pipeline run events status."""
```

Issue #2323, opened 2025-05-06
Frontend – render run status on the flowchart


Huongg:
## Description
Display non-realtime pipeline run status on the Kedro-Viz flowchart by consuming metadata from .viz/kedro_pipeline_events.json.

## Possible Implementation
 **MVP 1 Scope suggestion**
* Show static (latest-run) pipeline status with:
    * Dataset nodes: show size and completion status
    * Function nodes: show execution time and completion status
* Integrate with backend API that surfaces the JSON contents
* No live updates — render once on load

**Todo**

- [ ] Define when and how the frontend fetches run status data (e.g., on app load)
- [ ] Parse and map metadata to relevant nodes in the flowchart
- [ ] Implement UI for displaying node statuses
- [ ] Handle loading/error

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

## Checklist

- [ ] Include labels so that we can categorise your feature request

PR #2351, opened 2025-05-29
Run status: Set run data into redux store

jitu5:
## Description

Resolve https://github.com/kedro-org/kedro-viz/issues/2323

This pull request introduces functionality to manage and display pipeline run status data in a structured format. Key changes include adding actions, reducers, selectors, and normalisation logic for handling run status data, as well as integrating this functionality into the application state and lifecycle.

### New functionality for handling run status data:

* **Actions and API integration**: Added actions (`updateRunStatusData`, `loadRunStatusData`) and a function (`fetchRunStatus`) to fetch, process, and dispatch pipeline run status data from an API endpoint (`src/actions/run-status.js`).
* **Reducer**: Introduced a `runStatusReducer` to manage the `runStatus` slice of the Redux state (`src/reducers/run-status.js`).
* **Selectors**: Added selectors to access nodes, datasets, pipeline metadata, and specific statuses or errors from the `runStatus` state (`src/selectors/run-status.js`).

### Integration into the application:

* **State initialization**: Enhanced the initial state preparation to include normalized run status data (`src/store/initial-state.js`). [[1]](diffhunk://#diff-8aec283ba10f2d8f1aa71cf147f0b016e3491e28087e437043e921d4956b1da9R4) [[2]](diffhunk://#diff-8aec283ba10f2d8f1aa71cf147f0b016e3491e28087e437043e921d4956b1da9R216-R224) [[3]](diffhunk://#diff-8aec283ba10f2d8f1aa71cf147f0b016e3491e28087e437043e921d4956b1da9R269-R274)
* **Redux store**: Combined the new `runStatus` reducer into the application's root reducer (`src/reducers/index.js`). [[1]](diffhunk://#diff-3527887dbf404f0c28be291b83274e5f361cab6b0e6e0bd5ed06f66dbd11c5b0R11) [[2]](diffhunk://#diff-3527887dbf404f0c28be291b83274e5f361cab6b0e6e0bd5ed06f66dbd11c5b0R89)
* **Component updates**: Updated the `App` component to fetch and process run status data during initialization and updates (`src/components/app/app.js`). [[1]](diffhunk://#diff-38ffafc4109640a95acd39ca5d04edf2a090353c07b7c822c16ccdc5ddc412c0R12-R21) [[2]](diffhunk://#diff-38ffafc4109640a95acd39ca5d04edf2a090353c07b7c822c16ccdc5ddc412c0R47-R62) [[3]](diffhunk://#diff-38ffafc4109640a95acd39ca5d04edf2a090353c07b7c822c16ccdc5ddc412c0R121-R127)

### Data normalisation:

* **Normalisation logic**: Added utility functions to validate and process raw run status data into a structured format, including handling nodes, datasets, and pipeline metadata (`src/store/normalize-run-data.js`).

### State 
```
runStatus: {
    nodes: {},
    datasets: {},
    pipeline: {},
  }
```

### Data flow for run status:
Screenshot 2025-06-24 at 11 53 03 a m

Please refer this [Miro board](https://miro.com/app/board/uXjVImNahoo=/?share_link_id=875671523641 ) to compare it with pipeline.
 

 ## QA notes
 
 - Checkout the `feature/run-status-nrt`, `feature/process-json-run-data` and current branch and merge
 - At root of kedro-viz run `pip install -e package`
 - In demo_project run `kedro run`
 - Above step will produce/update `kedro_pipeline_events.json` in `.viz` folder
 - Run `kedro viz run`
- At new end point you can check the response


src/utils/run-status.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * Fetch pipeline run events from the API endpoint
```
rashidakanchwala: does it make sense to call this file `normalizeRunStatus`

src/components/app/app.js
```diff
@@ -39,13 +40,23 @@ class App extends React.Component {
...
+    }
     this.announceFlags(this.store.getState().flags);
   }
 
   componentDidUpdate(prevProps) {
     if (prevProps.data !== this.props.data) {
       this.updatePipelineData();
     }
+    if (prevProps.runData !== this.props.runData) {
+      this.updateRunData();
```
Huongg: why can't we dispatch an action directly here? 

```
 if (this.props.runData) {
      this.store.dispatch(updateRunStatusData(this.props.runData));
    } else {
      this.store.dispatch(loadRunStatusData());
    }
```


As in it feels like `updateRunData` is the same as `updateRunStatusData()`

src/components/app/app.js
```diff
@@ -104,6 +127,13 @@ App.propTypes = {
       tags: PropTypes.array,
     }),
   ]),
+  /**
+   * Determines what run status data will be displayed on the chart.
+   * You can supply an object with run status information -
+   * Alternatively, the string 'json' indicates that data is being
+   * loaded asynchronously from /api/run-events
+   */
+  runData: PropTypes.oneOfType([PropTypes.oneOf(['json']), PropTypes.object]),
```
Huongg, line 127: hmm how does it work when it comes to other application like vscode? Do we need a flag in option to disable/enable this feature?

src/selectors/run-status.js
```diff
@@ -0,0 +1,87 @@
...
+ * This file provides selectors for accessing the new node-ID based structured format
+ * for pipeline run events data.
+ */
+
+/**
+ * Get all node data grouped by node ID
+ * @param {Object} state Redux state
+ * @returns {Object} Nodes data grouped by node ID
+ */
+export const getNodesById = (state) => {
```
Huongg, line 13: love how you split all different selectors here ❤️

src/utils/run-status.js
```diff
@@ -0,0 +1,71 @@
...
+ * @returns {Promise<Object>} Pipeline run events data in structured format
+ */
+export const fetchRunEvents = async () => {
+  try {
+    const response = await fetch(`/api/run-events`);
```
Huongg: I found the name a bit confusing—when I first saw `runEvents`, I assumed it was related to a different kind of run, like Experiment Tracking. But it looks like it's specifically about showing the status of various nodes, datasets, and pipelines.

src/utils/run-status.js
```diff
@@ -0,0 +1,71 @@
...
+    const data = await response.json();
+    return data;
+  } catch (error) {
+    console.error('Failed to load run events:', error);
+    return { nodes: {}, datasets: {}, pipeline: {} };
+  }
+};
+
+/**
+ * Process structured run events into a format usable by the flowchart
```
Huongg: hmm but then down here the description says "usable by the flowchart"...I'm not very sure what this `processRunEvents` prefers to?

src/components/app/app.js
jitu5: We can dispatch directly from there but I kept this in `updateRunData` function as it is being used at 2 places.

src/components/app/app.js
jitu5: `runData` props is optional so whenever we use Kedro-Viz as a react component, we will not have any issues and we don't need flag.

src/utils/run-status.js
jitu5: I see, run events is not exact match here to what it does. Initially we were thinking to send run events to frontend and process but later we change it to send the only relevant info to frontend and not all the run events. 

I am thinking to rename this api end point to `/run-details` or `/run-status` @Huongg @rashidakanchwala

src/utils/run-status.js
jitu5: As I mentioned in previous comment, we are not sending every run events like `after_node_run` or `after_dataset_load` now. So in `processRunEvents` we are just formatting the response in below structure.
```
const runStatusData = {
    nodes: {},
    datasets: {},
    pipeline: {},
  }
```

src/utils/run-status.js
Huongg: I'd lean towards using /run-status since it's already referenced in many places. Also, /run-details makes me think of Experiment Tracking.

src/utils/run-status.js
Huongg: Hey, my comment is about the function name `processRunEvents` and its description that says a format usable by the flowchart. I think both should be updated to better reflect their actual purpose

src/components/app/app.js
Huongg: It seems like `updateRunStatusData()` is doing the same thing as `updateRunData`, unless I'm missing that there's something additional happening inside `updateRunData` that `updateRunStatusData()` doesn't cover?

jitu5: Okay I got it `updateRunStatusData()` same as `updateRunData`. I updated the code.

src/utils/normalizeRunStatus.js
rashidakanchwala, line 1: firstly, we should call this normalise-run-status.js; all our files should be kebab-case. And i feel like this should be in the store where normalize.data.js is

rashidakanchwala: also think the current design isn’t very extensible. For example, if we want to support URLs for run status in the future — like we do for pipelines — it might be worth following the same approach we use to load pipeline info into the Redux store.

That way, we maintain uniformity across the codebase, and it’ll give us more flexibility down the line — especially in MVP2 when we want to support additional use cases on the frontend.

What do you think about aligning the structure now to make this easier later?

jitu5: @rashidakanchwala I aligned the run status data flow with pipeline data flow. I added dataflow diagram and also Miro board link in description to compare it with pipeline dataflow.

src/actions/run-status.js
```diff
@@ -0,0 +1,44 @@
...
+    console.error('Failed to load run status:', error);
+    return { nodes: {}, datasets: {}, pipeline: {} };
+  }
+};
+
+/**
+ * Load run status data
+ * @returns {Function} Thunk that loads run status data
+ */
+export function loadRunStatusData() {
```
rashidakanchwala: is this only for initalRun and we do updateRunStatus later; if so i would call this loadInitialRunStatusData()

src/store/normalize-run-data.js
```diff
@@ -0,0 +1,104 @@
...
+ * Check whether data is in expected format
+ * @param {Object} data - The parsed data input
+ * @return {Boolean} True if valid for formatting
+ */
+const validateRunStatusInput = (data) => {
+  if (!data) {
+    return;
+  }
+
+  const isValidNodes =
```
rashidakanchwala, line 21: Why do we need this validation step?

src/actions/run-status.js
jitu5: Updated.

src/store/normalize-run-data.js
jitu5: For run data coming from parent React component.

PR #2342, opened 2025-05-22
Feat/run status draw nodes

Huongg:
## Description

Fixes #2331 
This PR introduces an implementation of the new run status/workflow UI using mock data. Please ignore unrelated changes—focus on reviewing the structural and design updates outlined below.

## Development notes
 Key Changes
* New main component:
    * src/pages/workflow.js – Main entry point that imports and renders the Workflow UI components and connects to the Redux store.
* Styling:
    * src/styles/pages/workflow.scss – Styles specific to the Workflow page.
* Utility functions:
    * src/utils/workflow-utils.js – Handles data formatting and key mapping for the Workflow page.
* Shared loading component:
    * src/components/loading-pipeline.js – Extracted and reused from the flowchart component for consistency across pages.

## How to QA this PR
You can test this either in Codespaces or locally.
Focus area:
* Verify the design and styling of nodes with different statuses: ✅ Success ❌ Failed ⏭️ Skipped
* Check that the styles are correctly applied and that the UI renders as expected.

## Checklist

- [ ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #2353, opened 2025-05-31
Run status: Add ability to switch tabs to view workflow

jitu5:
## Description

This pull request introduces a feature that allows toggling between two views: "flowchart" and "workflow." It includes changes to the Redux state, UI components, and routing logic to support this functionality. The most important changes are grouped by theme below.

### Redux State Management:

* Added a new action `TOGGLE_VIEW` and corresponding reducer to manage the `view` state, defaulting to "flowchart." (`src/actions/index.js` - [[1]](diffhunk://#diff-0a872dd59c943ab9b150cf03020c63d439bf77e6a447de2605145a7a46d2709dR365-R377) `src/reducers/index.js` - [[2]](diffhunk://#diff-3527887dbf404f0c28be291b83274e5f361cab6b0e6e0bd5ed06f66dbd11c5b0R29) [[3]](diffhunk://#diff-3527887dbf404f0c28be291b83274e5f361cab6b0e6e0bd5ed06f66dbd11c5b0R131)
* Updated the `preparePipelineState` function to set the active pipeline to `__default__` when the "workflow" view is selected. (`src/store/initial-state.js` - [src/store/initial-state.jsR212-R216](diffhunk://#diff-8aec283ba10f2d8f1aa71cf147f0b016e3491e28087e437043e921d4956b1da9R212-R216))

### UI Components:

* Updated `FlowchartPrimaryToolbar` to conditionally render certain buttons based on the selected view. (`src/components/flowchart-primary-toolbar/flowchart-primary-toolbar.js` - [[1]](diffhunk://#diff-64abcf88e16f9579d962c705c113192517a4dcf6c7f7e2af9bc90a8fd788702bR43-R46) [[2]](diffhunk://#diff-64abcf88e16f9579d962c705c113192517a4dcf6c7f7e2af9bc90a8fd788702bR62) [[3]](diffhunk://#diff-64abcf88e16f9579d962c705c113192517a4dcf6c7f7e2af9bc90a8fd788702bR73) [[4]](diffhunk://#diff-64abcf88e16f9579d962c705c113192517a4dcf6c7f7e2af9bc90a8fd788702bR98) [[5]](diffhunk://#diff-64abcf88e16f9579d962c705c113192517a4dcf6c7f7e2af9bc90a8fd788702bL103-R117) [[6]](diffhunk://#diff-64abcf88e16f9579d962c705c113192517a4dcf6c7f7e2af9bc90a8fd788702bR140)
* Modified `PipelineList` to disable the dropdown when the "workflow" view is active. (`src/components/pipeline-list/pipeline-list.js` - [[1]](diffhunk://#diff-4ecc87eea25ce524efdbc77d4f0b7cf4162b9bdac24caf09bbec32fb2fc9419cR23) [[2]](diffhunk://#diff-4ecc87eea25ce524efdbc77d4f0b7cf4162b9bdac24caf09bbec32fb2fc9419cL32-R33) [[3]](diffhunk://#diff-4ecc87eea25ce524efdbc77d4f0b7cf4162b9bdac24caf09bbec32fb2fc9419cR68)
* Added a new `WorkflowIcon` component for the "workflow" view button in the global toolbar. (`src/components/icons/workflow.js` - [src/components/icons/workflow.jsR1-R32](diffhunk://#diff-16d9a9d37b6dc6de0f66f0520c24580f9ccc99e5ffa0938fe20c1b1390dbc816R1-R32))

### Routing Logic:

* Updated the `Wrapper` component to handle routing between "flowchart" and "workflow" views and trigger the `onToggleView` action accordingly. (`src/components/wrapper/wrapper.js` - [[1]](diffhunk://#diff-e8ee05eb82bf5e9d915146317a649e352436fdc5506eb6f845d36d33f63663e0R7) [[2]](diffhunk://#diff-e8ee05eb82bf5e9d915146317a649e352436fdc5506eb6f845d36d33f63663e0L19-R20) [[3]](diffhunk://#diff-e8ee05eb82bf5e9d915146317a649e352436fdc5506eb6f845d36d33f63663e0L64-R84) [[4]](diffhunk://#diff-e8ee05eb82bf5e9d915146317a649e352436fdc5506eb6f845d36d33f63663e0L83-R106)

### Global Toolbar:

* Added buttons for toggling between "flowchart" and "workflow" views in the global toolbar, with corresponding actions to update the active pipeline and expand all pipelines. (`src/components/global-toolbar/global-toolbar.js` - [[1]](diffhunk://#diff-69e4872bdcef36fb8acc299c3b4612f114e7dfdcae3346f8121c21ad03c7f821R8-R21) [[2]](diffhunk://#diff-69e4872bdcef36fb8acc299c3b4612f114e7dfdcae3346f8121c21ad03c7f821R35-R38) [[3]](diffhunk://#diff-69e4872bdcef36fb8acc299c3b4612f114e7dfdcae3346f8121c21ad03c7f821R61-R78) [[4]](diffhunk://#diff-69e4872bdcef36fb8acc299c3b4612f114e7dfdcae3346f8121c21ad03c7f821R143-R153)

## QA notes

This branch depends on other code and can't be tested in isolation.

**Note:** Please ignore workflow icon change, I took it from https://github.com/kedro-org/kedro-viz/pull/2342


src/actions/index.js
```diff
@@ -362,3 +362,16 @@ export const updateStateFromOptions = (updatedOptions) => {
     payload: updatedOptions,
   };
 };
+
+export const TOGGLE_VIEW = 'TOGGLE_VIEW';
```
rashidakanchwala: let's say SET_VIEW since there's no toggle action involved.

src/components/flowchart-primary-toolbar/flowchart-primary-toolbar.js
```diff
@@ -57,16 +59,18 @@ export const FlowchartPrimaryToolbar = ({
...
+        {isFlowchartView && (
+          <IconButton
+            active={textLabels}
+            ariaLabel={`${textLabels ? 'Hide' : 'Show'} text labels`}
+            className={'pipeline-menu-button--labels'}
+            dataTest={`sidebar-flowchart-labels-btn-${textLabels}`}
+            icon={LabelIcon}
+            labelText={`${textLabels ? 'Hide' : 'Show'} text labels`}
+            onClick={() => onToggleTextLabels(!textLabels)}
+            visible={display.labelBtn}
```
Huongg: i think you can pass `isFlowchartView &&..` directly in this `visible` prop

src/components/flowchart-primary-toolbar/flowchart-primary-toolbar.js
```diff
@@ -91,22 +95,26 @@ export const FlowchartPrimaryToolbar = ({
...
-          className={'pipeline-menu-button--pipeline'}
-          dataTest={`sidebar-flowchart-expand-pipeline-btn-${expandedPipelines}`}
-          icon={expandedPipelines ? CollapsePipelinesIcon : ExpandPipelinesIcon}
-          labelText={
-            expandedPipelines ? 'Collapse pipelines' : 'Expand pipelines'
-          }
-          onClick={handleToggleExpandAllPipelines}
-          visible={display.expandPipelinesBtn}
-        />
+        {isFlowchartView && (
```
Huongg: same here

src/components/global-toolbar/global-toolbar.js
```diff
@@ -50,8 +59,32 @@ export const GlobalToolbar = ({
...
+              dataTest={'global-toolbar-workflow-btn'}
+              className={
+                'pipeline-menu-button--large pipeline-menu-button--link'
+              }
+              disabled={false}
+              icon={WorkflowIcon}
+              labelText="Workflow"
+              onClick={() => {
+                onSetView(VIEW.WORKFLOW);
+                onUpdateActivePipeline(PIPELINE.DEFAULT);
```
Huongg: can we leave a comment here to say that we're doing expanding all the pipeline and only show the default pipeline as a first MVP of run status? So just if we come back here again we remember the context?

src/components/icons/workflow.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+const WorkflowIcon = ({ className }) => (
```
Huongg: I think this is still the old icon, Steph has updated the new icon now I think. 

https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=3908-33753&m=dev

src/components/flowchart-primary-toolbar/flowchart-primary-toolbar.js
```diff
@@ -40,8 +40,10 @@ export const FlowchartPrimaryToolbar = ({
   onToggleExpandAllPipelines,
   orientation,
   onToggleOrientation,
+  view,
 }) => {
   const { toSetQueryParam } = useGeneratePathname();
+  const isFlowchartView = view === 'flowchart';
```
Huongg: why don't we use similar config for Flowchart like Workflow?

eg:
```
const isWorkflowView = view === VIEW.WORKFLOW;`
```

Huongg (commented):
Hey @jitu5, thanks for the PR — I've left a few comments below.

One general suggestion: I noticed that in multiple child components, we're checking the logic `view === VIEW.WORKFLOW` or `view === "flowchart"` It might be cleaner to compute this once in the parent component and pass it down as a prop, like `isWorkflowView.`

We could do this directly in `sidebar.js`:

```

export const Sidebar = ({
  displayGlobalNavigation,
  visible,
  displayFilterBtn,
}) => {
  const [pipelineIsOpen, togglePipeline] = useState(false);

  const isWorkflowView = view === VIEW.WORKFLOW;

  return (
    <>
      <div
        className={classnames('pipeline-sidebar', {
          'pipeline-sidebar--visible': visible,
          'pipeline-sidebar--no-global-toolbar': !displayGlobalNavigation,
        })}
      >
        <div className="pipeline-ui">
          <PipelineList onToggleOpen={togglePipeline} disable={isWorkflowView}/>
          <NodesPanel faded={pipelineIsOpen} />
        </div>
        <nav className="pipeline-toolbar">
          {displayFilterBtn && <ToolbarFilterButton />}
          <FlowchartPrimaryToolbar isWorkFlowView={isWorkflowView} />
          <MiniMapToolbar />
        </nav>
        <MiniMap />
      </div>
    </>
  );
```

Then, in `PipelineList` and `FlowchartPrimaryToolbar`, we can just read the prop and handle logic accordingly.

This way, the logic is centralised, making it easier to manage and add context through comments if needed.

What do you think?

Huongg:
when testing the `temp/run-status`, I also noticed a few things

1. The workflow tab generates the url as `workflow/workflow` if you double click on it
Screenshot 2025-06-09 at 10 20 56

2. Should the `workflow` tab has the active css styling? and also as mentioned above, it's not the correct icon at the moment
3. The pipeline list is hidden/disabled on the `workflow` view now`, however i'm still able to expand or collapse the pipelines
Screenshot 2025-06-09 at 10 21 30

jitu5:
> Hey @jitu5, thanks for the PR — I've left a few comments below.
> 
> One general suggestion: I noticed that in multiple child components, we're checking the logic `view === VIEW.WORKFLOW` or `view === "flowchart"` It might be cleaner to compute this once in the parent component and pass it down as a prop, like `isWorkflowView.`
> 
> We could do this directly in `sidebar.js`:
> 
> ```
> 
> export const Sidebar = ({
>   displayGlobalNavigation,
>   visible,
>   displayFilterBtn,
> }) => {
>   const [pipelineIsOpen, togglePipeline] = useState(false);
> 
>   const isWorkflowView = view === VIEW.WORKFLOW;
> 
>   return (
>     <>
>       <div
>         className={classnames('pipeline-sidebar', {
>           'pipeline-sidebar--visible': visible,
>           'pipeline-sidebar--no-global-toolbar': !displayGlobalNavigation,
>         })}
>       >
>         <div className="pipeline-ui">
>           <PipelineList onToggleOpen={togglePipeline} disable={isWorkflowView}/>
>           <NodesPanel faded={pipelineIsOpen} />
>         </div>
>         <nav className="pipeline-toolbar">
>           {displayFilterBtn && <ToolbarFilterButton />}
>           <FlowchartPrimaryToolbar isWorkFlowView={isWorkflowView} />
>           <MiniMapToolbar />
>         </nav>
>         <MiniMap />
>       </div>
>     </>
>   );
> ```
> 
> Then, in `PipelineList` and `FlowchartPrimaryToolbar`, we can just read the prop and handle logic accordingly.
> 
> This way, the logic is centralised, making it easier to manage and add context through comments if needed.
> 
> What do you think?

Done.

jitu5:
> when testing the `temp/run-status`, I also noticed a few things
> 
> 1. The workflow tab generates the url as `workflow/workflow` if you double click on it
> 
> Screenshot 2025-06-09 at 10 20 56
> 2. Should the `workflow` tab has the active css styling? and also as mentioned above, it's not the correct icon at the moment
> 3. The pipeline list is hidden/disabled on the `workflow` view now`, however i'm still able to expand or collapse the pipelines
> 
> Screenshot 2025-06-09 at 10 21 30

I am looking into point 1 & 2. point 3 is fixed.

Huongg (approved):
code wise looks good to me, thank you @jitu5 

Just 2 small notes, you can either address in this PR or we can tackle them in a smaller separate PR

1. The gap between the yellow dot and icon are too close
QA branch
Screenshot 2025-06-10 at 12 36 00

Design

Screenshot 2025-06-10 at 12 36 38


2. Currently the workflow icon doesn't work on the light theme
Screenshot 2025-06-10 at 12 36 59

src/components/global-toolbar/global-toolbar.js
```diff
@@ -50,8 +59,35 @@ export const GlobalToolbar = ({
...
+              }
+              disabled={false}
+              icon={WorkflowIcon}
+              labelText="Workflow"
+              onClick={() => {
+                onSetView(VIEW.WORKFLOW);
+
+                // When switching to workflow view, we're expanding all the pipeline
+                // and only show the default pipeline as a first MVP of run status
+                onUpdateActivePipeline(PIPELINE.DEFAULT);
```
rashidakanchwala: heyy I noticed a bug here when we were interviewing with @antonymilne , let's say I toggle 'Text Labels' then this impacts my workflow tab as well when I switch I think we need to handle tht properly.

rashidakanchwala: I think we should also include an on load behavior for the Workflow view — to reset the chart state independently, so that changes in the Flowchart view don’t affect it. We could also consider moving these functions there and keep this js file clean.

jitu5:
> code wise looks good to me, thank you @jitu5
> 
> Just 2 small notes, you can either address in this PR or we can tackle them in a smaller separate PR
> 
> 1. The gap between the yellow dot and icon are too close
>    QA branch
> 
> Screenshot 2025-06-10 at 12 36 00
> Design
> 
> Screenshot 2025-06-10 at 12 36 38
> 
> 2. Currently the workflow icon doesn't work on the light theme
> 
> Screenshot 2025-06-10 at 12 36 59

@Huongg This is fixed in https://github.com/kedro-org/kedro-viz/pull/2383

src/components/flowchart-route-wrapper/flowchart-route-wrapper.js
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect } from 'react';
```
rashidakanchwala: these seems weird coz we have wrapper around wrapper; can we remove this.

src/components/workflow-route-wrapper/workflow-route-wrapper.js
```diff
@@ -0,0 +1,60 @@
+import React, { useEffect } from 'react';
```
rashidakanchwala: I see now why you have added this view; but can we not do these things in the initial_pipeline_state, similar to how we do expand pipelines for flowchart view based on urls?

rashidakanchwala: i see why; u might need this if the user sets some setting in flowchart view then we don't want these settings to be passed to workflow view.

jitu5: I was thinking along the same lines and implemented a similar approach in initial_pipeline_state. But, there's an issue , it only works on the first load.

In the current implementation (e.g., for expanding pipelines), the state is set during the initial load, and there is also a dedicated update action to handle changes on user interaction.

But if we go with similar approach as "expand pipelines" the we have to set in the initial_pipeline_state and also when user click on any tab, we need to reset relevant redux state through `onClick`.

So, instead of setting the state on initial load and resetting it in the `onClick` handler for the tab, I created a wrapper for both routes and handled the state setting/resetting there in one place.

Let me know if there's a better approach. I'm happy to update it.

rashidakanchwala: i will have to think through this but regardless, we will have to revert a lot of stuff here including textLabels right? li

jitu5: Yes we need to reset textLabels and there are lot more may be like some filters.  So if we look at this overall, we created a separate view and separate flowchart to show run status, we might need to think on if need to maintain separate state for the workflow's flowchart. I see it as big work and we need to make a call. @rashidakanchwala wdyt ?

Huongg: I think we can address this as part of [this ticket](https://github.com/kedro-org/kedro-viz/issues/2386) after the release?

Specifically, the scope would include looking at how to simplify the FlowchartWrapper/FlowchartRouteWrapper and WorkflowWrapper/WorkflowRouteWrapper

rashidakanchwala: I think we can create a single redux action or use SET_VIEW action to do these state changes. first can you list what are the state changes we will need to do here. 
for instance; we don't need to change theme; theme can persist.

rashidakanchwala: But this change affects the bug around textLabels.

src/components/global-toolbar/global-toolbar.scss
```diff
@@ -68,3 +68,7 @@
   top: 31px;
   width: 8px;
 }
+
+.run-status-nav-wrapper {
```
Huongg: Do we need this styling?

Huongg (commented):
Hey @jitu5 , everything looks good aside from one small comment about the CSS, and the notification isn't working as expected. 





I think it's because after merging normalize-run-data, the data structure has changed slightly compared to the mock data. So this part probably needs to be updated now.


```
<RunStatusNotification
              ....
              timestamp={formatTimestamp(pipelineStatus.endTime)}
            />
```

src/components/global-toolbar/global-toolbar.scss
jitu5: Not needed here, removed.

jitu5:
> Hey @jitu5 , everything looks good aside from one small comment about the CSS, and the notification isn't working as expected.
> 
> Screenshot 2025-06-30 at 22 15 10
> I think it's because after merging normalize-run-data, the data structure has changed slightly compared to the mock data. So this part probably needs to be updated now.
> 
> ```
> <RunStatusNotification
>               ....
>               timestamp={formatTimestamp(pipelineStatus.endTime)}
>             />
> ```

Synced staging branch with missing fixed which we did not apply to notification branch but directly to QA branch.

Issue #2344, opened 2025-05-27
Refactor draw.js


Huongg:
## Description

Refactor draw.js so that we can re-use those component both in Flowchart and Workflow

PR #2358, opened 2025-06-02
Refactor draw.js

Huongg:
## Description

Fixes https://github.com/kedro-org/kedro-viz/issues/2344
This update refactors the draw.js component to make it more functional and reusable. Previously, it was tightly coupled with the Flowchart component, limiting its flexibility. The new structure allows us to import and reuse it in different contexts across the codebase.

## Development notes

- The logic from draw.js has been fully abstracted into a standalone Draw component.

Screenshot 2025-05-29 at 12 21 18

- Within the Draw component, there are separate subcomponents responsible for rendering specific elements, including:
    - draw-nodes
    - draw-layers
    - draw-layers-name
    - graph-svg (a wrapper component)

- The component also includes its own styling, configuration, and utility functions.

Screenshot 2025-05-29 at 12 21 29

- As a result, the Draw component is now more independent. In the Flowchart component, we can simply import it and pass in the necessary callbacks, props, and data to render each part appropriately.

For example, anywhere you can now just import those components as below

```

   <GraphSVG
          ...props
    >
        <DrawLayersGroup
              ...props
        />

        <DrawEdges
              ...props
        />

        <DrawNodes
               ....props
        />
    </GraphSVG>

     <DrawLayerNamesGroup
         ...props
     />
```

## QA notes

There are two ways to QA this PR:
1. Technical review – Dive into the code changes to ensure nothing affects the UI behavior after the refactor.
2. Non-technical/UI review – If you're less familiar with the React/frontend code, you can simply test the UI in the browser to make sure everything still looks and works the same.

To test the UI:
- You can run the app locally or test it via our Codespace environment. Instructions for using Codespace are available [here](https://github.com/kedro-org/kedro-viz/blob/main/CONTRIBUTING.md#testing-ui-using-gitpod).
- There are no expected UI changes—this is just to confirm that everything functions as before after this refactor.

## Checklist


src/components/flowchart/flowchart.test.js
```diff
@@ -399,7 +399,7 @@ describe('FlowChart', () => {
       .find('.pipeline-node__parameter-indicator--visible');
     nodeRects.each((i, el) => {
       const y = parseFloat(select(el).attr('y'));
-      expect(y).toBeLessThan(nodeY);
+      expect(y).toEqual(-56);
```
rashidakanchwala: This test is a bit flaky now because we moved the draw methods into their own functional component. So I’ve hardcoded this value since it’s always the same. Essentially, I just want to check whether the parameter indicator is positioned to the left of the node—and this does that.

src/components/draw/utils/config.js
```diff
@@ -0,0 +1,5 @@
+export const DURATION = 700;
+export const LAYER_NAME_DURATION = 0.05;
+export const MARGIN = 500;
+export const MIN_SCALE = 0.8;
+export const MAX_SCALE = 2;
```
jitu5, line 5: Are we using this? same config is also found un flowchart.js

jitu5 (approved):
Awesome work Huong! 🎉

src/components/draw/utils/config.js
Huongg: ah yes good point, I've now removed them and use values from the config file instead

PR #2365, opened 2025-06-03
Bye Enzyme: fully migrated to RTL 

rashidakanchwala:
## Description

This was long long long due; enzyme was deprecated in 2022. So finally; let's put an end to it. 

There r still some tests I need to update which requires some code update. But I will do that either part of this ticket, or follow up tickets. 

## Development notes

The main change in this migration is the shift from Enzyme’s `setup.mount` to React Testing Library’s `setup.render`. The `state.mock.js` file has been updated accordingly to support RTL rendering.

Some of the tests have been rewritten to use RTL methods, while others remain largely the same in behavior — they still test what they’re supposed to, just using the new rendering utility.

- I’ve removed most of the tests from app.test.js and a couple from wrapper.js because they relied on Enzyme’s `shallow` rendering. `shallow` was useful for unit testing the wrapper or top-level component without diving into its children, but React Testing Library doesn’t support or encourage shallow testing.
Since RTL promotes integration testing for higher-level components, I’ve removed the `shallow` tests with the intention of adding meaningful integration tests later.

## Checklist


Huongg (approved):
hey @rashidakanchwala thank you for all these big changes. I tested it locally and everything is working without errors. As you mentioned, there are some missing tests—let’s create a backlog ticket to track them so we don’t forget. For now, let's go ahead and merge this to unblock the other PRs.

Issue #2332, opened 2025-05-19
Improve flowchart layout to ensure clear spacing, layering, and full pipeline visibility with run status


Huongg:
## Description
As a Kedro user, I want to be able to see the entire flowchart view of my pipeline laid out clearly, with evenly spaced nodes and defined layers, so I can understand the full structure without nodes overlapping or crowding each other.

Figma design: https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=3631-40098&m=dev

## Acceptance Criteria

<img width="1081" alt="Image" src="https://github.com/user-attachments/assets/c46b3d7c-7528-4b2c-9a12-9a62d1f3de45" />

- The flowchart view automatically adjusts to show the entire pipeline within the visible canvas or viewport.
- Nodes are evenly spaced horizontally and vertically, avoiding overlap.
- Layers or stages in the pipeline are visually grouped or aligned to show a logical flow.
- The layout adapts to different pipeline sizes while maintaining clear separation between nodes and layers.
- There is no overlap between node labels, icons, or connections.
- Both for horizontal and vertical view

astrojuanlu:
I'm confused - the wording of the task seems to imply that at the moment there is overlap between nodes, layers aren't visually grouped etc. Or is this specifically about the run status view?

astrojuanlu:
Ugh interesting, those screenshots clarify everything. Thanks @Huongg

Issue #2048, opened 2024-08-20
URL Access to Experiment Tracking Causes 404 Error on viz demo


SajidAlamQB:
## Description
When accessing the experiment tracking page directly via the URL https://demo.kedro.org/experiment-tracking, a `404` error is returned. This issue only occurs when accessing the page directly via the URL, while navigation through the site’s tabs works as expected.

## Context
This bug affects our ability to access the experiment tracking page directly via a URL, which is important for user experience and ease of access.

## Steps to Reproduce

- Attempt to access the experiment tracking page directly by visiting https://demo.kedro.org/experiment-tracking.
- See the `404` error.

## Expected Result
The experiment tracking page should load.

## Actual Result
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/9acef439-8f5f-4fa7-bc79-d8cdf03a7a80">

## Checklist
- [x] Include labels so that we can categorise your issue

ravi-kumar-pilla:
Some Context:

GitHub Pages doesn’t support routers that use the HTML5 pushState history API under the hood (for example, React Router using browserHistory). This is because when there is a fresh page load for a url like http://user.github.io/todomvc/todos/42, where /todos/42 is a frontend route, the GitHub Pages server returns 404 because it knows nothing of /todos/42.

Possible solution:

More information and possible solutions are mentioned here - https://create-react-app.dev/docs/deployment/#github-pages

Thank you

PR #2387, opened 2025-06-24
Run status : improve layout

Huongg:
## Description
To address this https://github.com/kedro-org/kedro-viz/issues/2332

This is to merge this change into the `staging-run-status-frontend`

## Checklist

### DIFF
--- a/src/components/draw/utils/renderNodeDetailsContainer.js
+++ b/src/components/draw/utils/renderNodeDetailsContainer.js
@@ -60,14 +60,16 @@ export function renderNodeDetailsContainer(
     })
     .style('fill', 'none');
 
-  // Details background
+  // Draw the background rectangle for the node details section
+  // The height is calculated as half the node height plus the workflow node details height
+  // While the width remains the same as the node width, which is calculated inside graph/index.js
   detailsContainer
     .append('rect')
     .attr('class', 'pipeline-node__details-bg')
     .attr('width', nodeWidth)
-    .attr('height', nodeHeight)
+    .attr('height', nodeHeight / 2 + workflowNodeDetailsHeight)
     .attr('x', nodeWidth / -2)
-    .attr('y', node.type === 'task' ? nodeHeight / 2 + 1 : 0)
+    .attr('y', 0)
     .attr('rx', 0);
 
   // Details outline (bottom part only)


src/reducers/status.js
```diff
@@ -0,0 +1,8 @@
+import { statusMockData } from '../utils/status-mock-data';
```
rashidakanchwala: this is not related to my PR

src/selectors/status.js
```diff
@@ -0,0 +1,23 @@
+import { createSelector } from 'reselect';
```
rashidakanchwala: this is not part of this PR

src/reducers/status.js
Huongg: this is prob merged by mistake. I can remove it

src/selectors/status.js
Huongg: sorry this was here by mistake. I removed it now

src/config.js
```diff
@@ -33,7 +33,7 @@ export const codeSidebarWidth = {
   open: 480,
 };
 
-export const workFlowStatuses = ['success', 'failed', 'skipped'];
+export const workFlowStatuses = ['success', 'failed'];
```
Huongg, line 39: please ignore this part, I'll address this after merging this PR to the draw nodes

src/components/draw/utils/renderNodeDetailsContainer.js
```diff
@@ -19,7 +19,7 @@ export function renderNodeDetailsContainer(
   tasksStatus,
   dataSetsStatus
 ) {
-  const nodeWidth = Math.max(node.width || 0, MINIMUM_WIDTH);
+  const nodeWidth = Math.max(node.width || 0, NODE_DETAILS_WIDTH);
```
Huongg: this was done to keep the variable name consistent

Huongg:
> Tried Codespaces, but couldn't do anything with it... gets stuck in a "loading" status forever.

Oh let me look into the codespace issue, It was fine when we did the QA testing last week. Maybe something needs to be reset

> 
> Also, directly navigating to `/workflow` doesn't work? Similar to #2048, I get 404 if I refresh or type the URL directly.

This issue should be resolved after merging this PR in I think https://github.com/kedro-org/kedro-viz/pull/2353

src/utils/index.js
```diff
@@ -192,6 +192,9 @@ export const formatNumberWithCommas = (number) => {
  * @returns {Boolean} True if the app is running locally.
  */
 export const isRunningLocally = () => {
+  if (typeof window === 'undefined' || !window.location) {
```
Huongg, line 195: This resolves the recent issue that has been causing failures both locally and in CI.

Screenshot 2025-06-27 at 15 08 30

src/utils/index.js
```diff
@@ -227,6 +230,9 @@ export const isRunningLocally = () => {
  * @returns {string} Sanitized pathname
  */
 export const sanitizedPathname = () => {
+  if (typeof window === 'undefined' || !window.location) {
```
Huongg, line 233: same as above, to resolve this window issue

Screenshot 2025-06-27 at 15 08 30

src/components/draw/utils/renderNodeDetailsContainer.js
```diff
@@ -68,7 +67,9 @@ export function renderNodeDetailsContainer(
     .attr('width', nodeWidth)
     .attr(
       'height',
-      node.type === 'task' ? NODE_DETAILS_HEIGHT : NODE_DETAILS_HEIGHT + 20
+      node.type === 'task'
+        ? workflowNodeDetailsHeight
```
rashidakanchwala: @Huongg , we don't need to calculate this here; this should be calculated directly in utils/graph the same way as width

Huongg: Hey @rashida, there's a difference between the width and height handling here. Specifically, this part is responsible for drawing the details background, not the node itself.

As shown in the screenshot, the red area represents the details background. It doesn’t share the same height as the top part of the node.
Screenshot 2025-06-29 at 22 17 43


- Width: It stays consistent for both the top (node title) and the bottom (details).
- Height: They are different between the top and bottom. So if we try to calculate this in graph.js, the result isn't quite accurate.

The screenshot shows what happens if we share the same height between 2 sections—the top section ends up looking too tall.

Screenshot 2025-06-29 at 22 12 56


I might be missing some context from graph.js, but at the moment I’m not sure how the height calculation would work reliably there. Shall we pair on this tomorrow to figure it out together?

PR #2388, opened 2025-06-24
Staging/4aef9717c8659d455454ba4d3161e98e7d6eef2c

Huongg:
## Checklist


src/components/no-run-status/no-run-status.js
```diff
@@ -0,0 +1,46 @@
...
+
+import './no-run-status.scss';
+
+export const NoRunStatus = ({
+  displaySidebar,
+  displayMetadataPanel,
+  displayExportBtn,
+}) => {
+  return (
+    <div className="kedro-pipeline">
```
Huongg: I think rather than writing the whole component from scratch, we just re-use the existing component here. For example `PipelineWarningContent`

Huongg: you just need to pass the required title and sub-title, then you can add an icon component on top. that way we can re-use the styling etc


```
<PipelineWarningContent
        isVisible={}
        title={}
        subtitle={}
        sidebarVisible={sidebarVisible}
      />
```

Huongg: also for other Warning, Not found, we call them `PipelineWarning`, perhaps we should follow the same naming convention `NoRunStatusWarning` or `WorkFlowWarning` in case we want to extend more errors/warning in the future

src/components/nodes-panel/nodes-panel.js
```diff
@@ -59,6 +59,10 @@ const NodesPanel = ({ faded }) => {
     return () => window.removeEventListener('keydown', handleKeyDown);
   });
 
+  // Show the node panel data only if we are not in flowchart view or if we are in workflow view but run status is not available
```
Huongg: this comment is a bit confusing for me. Did you mean 
```
Show the node panel data only if we are in flowchart view or if we are in workflow view but run status is not available
```

i think the wording `not in flowchart view` is confusing

src/components/nodes-panel/nodes-panel.js
```diff
@@ -59,6 +59,10 @@ const NodesPanel = ({ faded }) => {
...
+  const showNodePanelData =
```
Huongg: if you bring this to the parent component, you only need to pass down 1 prop rather than 2

perhaps somethings like this in the `sidebar.js`

```
// comment here
const showNodePanelData = !isWorkflowView || (isWorkflowView && isRunStatusAvailable);

<NodesPanel
        visible={showNodePanelData}
        faded={faded}
/>
```

src/components/container/container.js
```diff
@@ -1,16 +1,17 @@
 import React from 'react';
 import App from '../app';
 import getPipelineData from '../../utils/data-source';
-import { funNodeError } from '../../utils/run-status-mock-data/fun-node-error';
+// import { funNodeError } from '../../utils/run-status-mock-data/fun-node-error';
```
Huongg: to be removed before merging the PR

src/components/container/run-data.json
```diff
@@ -0,0 +1,12 @@
+{
```
Huongg: to be removed before merging the PR, this is mock data to test this PR specifically. It should not be merged to the staging branch

src/components/sidebar/sidebar.js
```diff
@@ -30,8 +35,17 @@ export const Sidebar = ({
         })}
       >
         <div className="pipeline-ui">
-          <PipelineList onToggleOpen={togglePipeline} />
-          <NodesPanel faded={pipelineIsOpen} />
+          <PipelineList
+            onToggleOpen={togglePipeline}
+            isWorkflowView={isWorkflowView}
```
Huongg, line 41: @jitu5 do we need this check here if you handle this in the Redux side?

src/components/no-run-status/no-run-status.js
Huongg: DONE

src/components/nodes-panel/nodes-panel.js
Huongg: @jitu5 to confirm

src/components/nodes-panel/nodes-panel.js
Huongg: DONE

src/components/sidebar/sidebar.js
jitu5: @Huongg  Yes this check will be needed in PipelineList component to disable pipeline dropdown if its workflow view.

src/components/run-not-found-warning/run-not-found-warning.js
```diff
@@ -0,0 +1,25 @@
+import { connect } from 'react-redux';
```
rashidakanchwala, line 1: I would keep this within workflow folder instead of having a new component folder

Huongg: Hmm, in that case, it’s not very consistent with other workflow-related components—for example, run-status-notification currently lives outside the workflow folder. If we move this one, should we also consider moving run-status-notification for consistency?

Screenshot 2025-07-01 at 12 11 49

Huongg: also should it follow the same pattern as “Flowchart,” which doesn’t have any smaller components nested inside it.

jitu5 (approved):

src/components/run-not-found-warning/run-not-found-warning.js
rashidakanchwala: makes sense maybe after this we should decide a consistent approach for all I would prefer very small components to be nested if they belong to a single component

Issue #2346, opened 2025-05-27
Show error log in meta data panel when clicking on failed node or failed dataset


Huongg:
@iamelijahko is currently working on the design, please update the description once you have the details

PR #2407, opened 2025-07-02
Run status: Show run status errors on metadata panel on click

jitu5:
Resolves https://github.com/kedro-org/kedro-viz/issues/2346

This pull request introduces a new feature to display error tracebacks in the metadata panel for workflow views.

### New Feature: Error Traceback Panel
* [`src/actions/index.js`](diffhunk://#diff-0a872dd59c943ab9b150cf03020c63d439bf77e6a447de2605145a7a46d2709dR298-R310): Added a new action `TOGGLE_TRACEBACK` and its corresponding function `toggleTraceback` to manage the visibility of the traceback panel.
* [`src/reducers/visible.js`](diffhunk://#diff-f4ea6d549c9ad6c89a4fc13c21d3f67bb1630dd81938ff48845903c979ef517bR11): Updated the `visibleReducer` to handle the new `TOGGLE_TRACEBACK` action, enabling state management for traceback visibility. [[1]](diffhunk://#diff-f4ea6d549c9ad6c89a4fc13c21d3f67bb1630dd81938ff48845903c979ef517bR11) [[2]](diffhunk://#diff-f4ea6d549c9ad6c89a4fc13c21d3f67bb1630dd81938ff48845903c979ef517bR64-R69)
* [`src/components/error-log/error-log.js`](diffhunk://#diff-3881b1c97f768045ed497c1c6604168a02bb2338241d4fecfd78043240b92e0cR1-R64): Introduced the `ErrorLog` component to display error information with a toggleable traceback panel.
* [`src/store/initial-state.js`](diffhunk://#diff-8aec283ba10f2d8f1aa71cf147f0b016e3491e28087e437043e921d4956b1da9R53): Added `traceback` to the initial state to track the visibility of the traceback panel.

### Metadata Panel Enhancements
* [`src/components/metadata/metadata.js`](diffhunk://#diff-75f06a5a17b3eeb6bde03d6cfffc97b833de28b38c08a36928d432d2754c8492R38-R57): Integrated the `ErrorLog` component into the metadata panel, added logic to handle error details and traceback visibility, and refined metadata display based on the current view (workflow or dataset). [[1]](diffhunk://#diff-75f06a5a17b3eeb6bde03d6cfffc97b833de28b38c08a36928d432d2754c8492R38-R57) [[2]](diffhunk://#diff-75f06a5a17b3eeb6bde03d6cfffc97b833de28b38c08a36928d432d2754c8492R131-R181) [[3]](diffhunk://#diff-75f06a5a17b3eeb6bde03d6cfffc97b833de28b38c08a36928d432d2754c8492R232-R245)
* [`src/components/metadata/metadata-code.js`](diffhunk://#diff-0c4f6e1448faae502655b722545b6a2a7979e12951479ef9993c79b466da6305R21): Modified the `MetaDataCode` component to dynamically update the panel title and content based on the current view (e.g., "Error traceback" for workflow views). [[1]](diffhunk://#diff-0c4f6e1448faae502655b722545b6a2a7979e12951479ef9993c79b466da6305R21) [[2]](diffhunk://#diff-0c4f6e1448faae502655b722545b6a2a7979e12951479ef9993c79b466da6305L38-R39)

### Selector Updates for Error Handling
* [`src/selectors/run-status.js`](diffhunk://#diff-d1bf9905ab4b9ca53015321575c51b154a50340bbe9071a48b0c086391fe9f6eL34-R34): Replaced redundant selectors with new ones (`getNodeError` and `getDatasetError`) to retrieve error details for nodes and datasets, simplifying error handling logic. [[1]](diffhunk://#diff-d1bf9905ab4b9ca53015321575c51b154a50340bbe9071a48b0c086391fe9f6eL34-R34) [[2]](diffhunk://#diff-d1bf9905ab4b9ca53015321575c51b154a50340bbe9071a48b0c086391fe9f6eL73-R74)

### Styling Updates
* [`src/components/error-log/error-log.scss`](diffhunk://#diff-7804a32ed587ecf18f994e1d8ab784e1fd4f56d73b2f5a8dec3a5adca4aace3dR1-R43): Added styles for the `ErrorLog` component, including theme-specific background colors and scrollbar customization for better user experience.

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -295,6 +295,19 @@ export function toggleCode(visible) {
   };
 }
 
+export const TOGGLE_TRACEBACK = 'TOGGLE_TRACEBACK';
+
+/**
+ * Toggle traceback panel
+ * @param {String} visible Visibility status
+ */
+export function toggleTraceback(visible) {
+  return {
+    type: TOGGLE_TRACEBACK,
+    visible,
+  };
+}
+
 export const TOGGLE_PARAMETERS_HOVERED = 'TOGGLE_PARAMETERS_HOVERED';
 
 /**
--- a/src/components/error-log/error-log.js
+++ b/src/components/error-log/error-log.js
@@ -1,0 +1,64 @@
+import React from 'react';
+import classNames from 'classnames';
+import Toggle from '../ui/toggle';
+
+import './error-log.scss';
+
+/**
+ * ErrorLog component displays error information with a toggleable traceback
+ */
+export default function ErrorLog({
+  errorDetails,
+  className = '',
+  onToggleCode,
+  dataTest = 'error-log',
+  visibleTraceback = false,
+  isDataNode = false,
+  nodeName = '',
+}) {
+  // Early return if no error details provided
+  if (!errorDetails) {
+    return (
+      <div className={classNames('error-log--wrapper', className)}>
+        <div className="error-log--message">No error details available.</div>
+      </div>
+    );
+  }
+
+  const errorMessage = errorDetails.message || 'An unknown error occurred.';
+
+  const getErrorHeader = () => {
+    if (isDataNode) {
+      const operation = errorDetails.error_operation;
+      if (operation) {
+        const toOrFrom = operation === 'loading' ? 'from' : 'to';
+        return `Failed while ${operation} data ${toOrFrom} dataset.`;
+      }
+      return 'Failed while loading/saving data to/from dataset.';
+    }
+
+    return `Failed while performing function: ${nodeName || 'Unknown'}`;
+  };
+
+  return (
+    <div className={classNames('error-log--wrapper', className)}>
+      <Toggle
+        id="code"
+        dataTest={dataTest}
+        title="Show traceback"
+        onChange={onToggleCode}
+        checked={visibleTraceback}
+      />
+
+      <div className="error-log--header">{getErrorHeader()}</div>
+
+      <div className="error-log--details">
+        <pre dangerouslySetInnerHTML={{ __html: errorMessage }} />
+      </div>
+
+      <div className="error-log--footer">
+        Please refer to the CLI for the full error log and details.
+      </div>
+    </div>
+  );
+}
--- a/src/components/error-log/error-log.js
+++ b/src/components/error-log/error-log.js
@@ -4,6 +4,19 @@ import Toggle from '../ui/toggle';
 
 import './error-log.scss';
 
+const ERROR_MESSAGES = {
+  noError: 'No error details available.',
+  unknownError: 'An unknown error occurred.',
+  genericDataError: 'Failed while loading/saving data to/from dataset.',
+  genericFunctionError: (name) =>
+    `Failed while performing function: ${name || 'Unknown'}`,
+  loadingOrSaving: (operation) =>
+    `Failed while ${operation} data ${
+      operation === 'loading' ? 'from' : 'to'
+    } dataset.`,
+  footer: 'Please refer to the CLI for the full error log and details.',
+};
+
 /**
  * ErrorLog component displays error information with a toggleable traceback
  */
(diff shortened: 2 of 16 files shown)


src/actions/index.js
```diff
@@ -295,6 +295,19 @@ export function toggleCode(visible) {
   };
 }
 
+export const TOGGLE_TRACEBACK = 'TOGGLE_TRACEBACK';
+
+/**
+ * Toggle traceback panel
+ * @param {String} visible Visibility status
+ */
+export function toggleTraceback(visible) {
```
Huongg, line 304: I'm wondering if we really need a separate action for this?  I’m thinking we could just reuse `toggleCode`, even for showing the traceback, since it follows the same flow—we’d just pass different data to display in the code panel.

src/components/metadata/metadata.js
```diff
@@ -31,17 +35,26 @@ const MetaData = ({
...
+  getNodeError,
+  view,
 }) => {
   const { toSelectedPipeline } = useGeneratePathname();
 
   // Hide code panel when selected metadata changes
   useEffect(() => onToggleCode(false), [metadata, onToggleCode]);
+
+  // Hide traceback panel when selected metadata changes
+  useEffect(() => onToggleTraceback(false), [metadata, onToggleTraceback]);
```
Huongg, line 56: you can see this one is basically the same one as the one above
```
// Hide code panel when selected metadata changes
  useEffect(() => onToggleCode(false), [metadata, onToggleCode]);
```

src/actions/index.js
rashidakanchwala: I think we should keep it separate and later on get Toggle Code back if possible. Also in general we are using metadata panel for both workflow and flowchart view but later we should have parents component and two child components to keep some separation of concern

Huongg: hey @rashidakanchwala from what I understand in the last chat with Juanlu and Elijah, it sounds like we probably won’t bring back the toggle code button while show traceback button is visible on the workflow page. If we do revisit it, it would likely be in the context of helping users do that kind of thing directly in VS Code. 

So to me, it feels like we're adding extra code for a future scenario that might not happen.

Screenshot 2025-07-02 at 12 50 06

rashidakanchwala: I know we decided not to go for it in 1.0 because of time constraints, but I’m not sure we won’t revisit it. Both Antony and Maxim were looking for the `Show Code` option when they encountered the node error, so we shouldn’t disregard it — especially since we already surface everything else in the metadata panel. It feels like a gap.

Once Steph is back, she can design this properly — both for live debugging in VSCode and in Kedro-Viz.

That said, from a coding perspective, while both are toggles that open the same div with different content, they’re actually two distinct features, so I think keeping the separation makes sense.

src/components/error-log/error-log.js
```diff
@@ -0,0 +1,64 @@
...
+  // Early return if no error details provided
+  if (!errorDetails) {
+    return (
+      <div className={classNames('error-log--wrapper', className)}>
+        <div className="error-log--message">No error details available.</div>
+      </div>
+    );
+  }
+
+  const errorMessage = errorDetails.message || 'An unknown error occurred.';
```
rashidakanchwala: can we do something like this 
```
const ERROR_MESSAGES = {
  noError: 'No error details available.',
  unknownError: 'An unknown error occurred.',
  genericDataError: 'Failed while loading/saving data to/from dataset.',
  genericFunctionError: (name) => `Failed while performing function: ${name || 'Unknown'}`,
  loadingOrSaving: (operation) =>
    `Failed while ${operation} data ${operation === 'loading' ? 'from' : 'to'} dataset.`,
  footer: 'Please refer to the CLI for the full error log and details.',
};
```

src/components/metadata/metadata.js
```diff
@@ -166,6 +226,20 @@ const MetaData = ({
...
                 />
+                <MetaDataRow
+                  label="Error Log:"
+                  visible={isWorkflowView && hasError}
+                >
+                  <ErrorLog
+                    errorDetails={currentErrorDetails}
+                    className="pipeline-metadata__error-log"
+                    onToggleCode={handleErrorLogToggle}
+                    dataTest={getDataTestAttribute('metadata', 'error-log')}
```
rashidakanchwala, line 237: I just noticed this. It's a good change, but I noticed we don't do this everywhere so pls let's create a follow up ticket to apply this across codebase. @jitu5 ?

src/components/metadata/metadata.js
```diff
@@ -31,17 +35,26 @@ const MetaData = ({
...
   onToggleMetadataModal,
   onToggleNodeSelected,
   theme,
   visible = true,
   visibleCode,
+  visibleTraceback,
   showDatasetPreviews,
+  getDatasetError,
+  getNodeError,
+  view,
```
rashidakanchwala, line 48: I think this Metadata file is too big and can be refactored. Can we please create a follow up ticket to refactor this component @jitu5 ?

rashidakanchwala (approved):
Thanks @jitu5 <3 — I’ve left a few minor comments.

Overall, this works well. I do think the metadata panel component is getting quite large, and there’s definitely room to refactor and improve it. Let’s raise a follow-up ticket to tackle that right after the 1.0 release.

src/actions/index.js
Huongg: okay lets go with this for now, we can always re-visit it later 👍

src/components/error-log/error-log.js
jitu5: Done

src/components/metadata/metadata.js
jitu5: Issue created https://github.com/kedro-org/kedro-viz/issues/2410

src/components/metadata/metadata.js
jitu5: Issue created https://github.com/kedro-org/kedro-viz/issues/2411

PR #2427, opened 2025-07-08
Fix/run status fe bugs

Huongg:
## Description
To address these 2 FE bugs:

1. https://github.com/kedro-org/kedro-viz/pull/2414#issuecomment-3036543350
 the fix => the metadata panel width is currently not matching with the design, max-width should 458px rather than 400px

2. https://github.com/kedro-org/kedro-viz/pull/2414#issuecomment-3036657640
the fix => change the left value for the notification bar to be more centered

## Checklist


src/components/container/container.js
```diff
@@ -9,7 +10,7 @@ import './container.scss';
  */
 const Container = () => (
   <>
-    <App data={getPipelineData()} />
+    <App data={getPipelineData()} runData={nodeError} />
```
Huongg: This is just for testing purpose, will be removed before merging this PR

src/components/container/container.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import App from '../app';
 import getPipelineData from '../../utils/data-source';
+import { nodeError } from './node-error';
```
Huongg: This is just for testing purpose, will be removed before merging this PR

src/components/container/node-error.js
```diff
@@ -0,0 +1,98 @@
+/* eslint-disable camelcase, no-unused-vars */
```
Huongg: This is just for testing purpose, will be removed before merging this PR

src/styles/_variables.scss
```diff
@@ -18,7 +18,7 @@ $sidebar-padding: 1.8em;
 $sidebar-width-open: 400px;
 $sidebar-width-closed: 56px;
 $sidebar-width-breakpoint: 700px;
-$metadata-sidebar-width-open: 400px;
+$metadata-sidebar-width-open: 458px;
```
jitu5, line 21: @Huongg Any reason why we increased `metadata-sidebar-width-open` to `458px` from `400px`

Huongg: hey @jitu5 it is to match the design, https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=3928-153690&m=dev

The meta data panel should always be 458px instead 400px. We set the number wrong the whole time i think

jitu5: I am asking because it might affect other layout calculation like when some component is hidden or when we centred the full flowchart when metadata panel is open. We might need to test some of the scenarios.

Huongg: ah okay, I've checked locally when i made this change and everything seems fine to me, but would be great if you can test it as well. So far this `$metadata-sidebar-width-open` only uses in side `metadata` and `metadata-code` and `metadata-panel`, not in any other layout

jitu5: Tested and looks fine.

PR #2390, opened 2025-06-25
Feature Branch(FE and BE) - Run Status

rashidakanchwala:
## Description

**Run status parent ticket:** https://github.com/kedro-org/kedro-viz/issues/2310

**All BE features to test under this PR:**
- Process Raw data and prepare the response: https://github.com/kedro-org/kedro-viz/pull/2350
- Hook implementation to generate and store run data: https://github.com/kedro-org/kedro-viz/pull/2328
- Saving api response for run status on kedro viz build/deploy: https://github.com/kedro-org/kedro-viz/pull/2409

**All FE features to test under this PR:**
- Switch tabs to view workflow: https://github.com/kedro-org/kedro-viz/issues/2323
- Set run data into redux store: https://github.com/kedro-org/kedro-viz/pull/2351
- Update flowchart design: https://github.com/kedro-org/kedro-viz/issues/2331
- Improve flowchart layout: https://github.com/kedro-org/kedro-viz/issues/2332
- Refactor draw.js: https://github.com/kedro-org/kedro-viz/issues/2344
- Show error log in metadata panel: https://github.com/kedro-org/kedro-viz/issues/2346
- Run status notification: https://github.com/kedro-org/kedro-viz/issues/2330
- No kedro run found: https://github.com/kedro-org/kedro-viz/issues/2367

### To test locally:
 - Checkout the current branch
 - Create a new virtual environment and install required dependencies
     - `pip install -r package/requirements.txt -r demo-project/src/demo_project/requirements.in`
     - `pip install kedro==1.0.0rc1`
 - At root of kedro-viz run `pip install -e package`
 - In demo_project run `kedro run`
 - In new terminal (use same virtual environment) and In demo_project run `kedro viz run`

#### A. Generating and processing run data

- Running kedro pipeline  
  - [x] When "kedro run" completes—either successfully or with failure a `kedro_pipeline_events.json` file is automatically created in the .viz directory.  
  - [x] The file should contain relevant pipeline run events (e.g., before_pipeline_run, before_node_run).

- Front-end accessibility  
  - [ ] The front-end application must be able to access run status response through `/run-status` API.

- Error handling  
  - [ ] If the pipeline run fails, the .json file must still be created.  
  - [ ] The file must accurately reflect the failure state, including:  
  - [ ] The status set to “failed”  
  - [ ] A relevant error message

- Backward Compatibility  
  - [ ] Existing projects that do not support or enable this feature must continue to function normally.

[...]

PR #2437, opened 2025-07-10
Bug fixes and code review suggestions for run status

jitu5:
## Description

Resolves some of the comments https://github.com/kedro-org/kedro-viz/pull/2390#issuecomment-3054163041 and https://github.com/kedro-org/kedro-viz/pull/2390#issuecomment-3057317179 

- Now if duration is less then a second then its show in ms
- Logging file path if something goes wrong.  
- Ignore some mock data for testing.
- Fix for incorrect usage of label for attribute

## Checklist


src/components/container/container.js
```diff
@@ -9,7 +10,7 @@ import './container.scss';
  */
 const Container = () => (
   <>
-    <App data={getPipelineData()} />
+    <App data={getPipelineData()} runData={runData} />
```
jitu5: Ignore: mock data for frontend testing

src/components/container/run-data.js
```diff
@@ -0,0 +1,336 @@
+/* eslint-disable camelcase, no-unused-vars */
```
jitu5: Ignore: mock data for frontend testing

src/components/workflow/workflow-utils/format.js
```diff
@@ -7,14 +7,23 @@ export function formatNumber(num) {
...
+    return `${milliseconds}ms`;
+  }
+
+  const totalSeconds = Math.floor(numSeconds);
   const minutes = Math.floor(totalSeconds / 60);
-  const secs = Number(seconds) - minutes * 60;
+  const secs = numSeconds - minutes * 60;
   const secsStr = formatNumber(secs);
   if (minutes > 0) {
     return `${minutes}m ${secsStr}s`;
```
ravi-kumar-pilla: if it is exactly 1min, then we will show `1m 0s` ? Can we instead show `1m` and show seconds only when there is something ?

jitu5: Done

jitu5:
> I think the orientation was intended; if we change it to horizontal then we should keep it consistent across views.

@rashidakanchwala I reseted orientation to "Vertical" whenever switch to workflow. Do you want me to revert ?

rashidakanchwala:
Yeah because i am thinking orientation change is thier preferred of viewing thier pipeline. So let's say they change it to horizontal because it makes sense to them then upon changing to workflow - we should keep it horizontal.

ravi-kumar-pilla (approved):
LGTM 👍

PR #2440, opened 2025-07-11
Log when user do a partial or non sequential pipeline run.

jitu5:
## Description

Fix for https://github.com/kedro-org/kedro-viz/pull/2390#issuecomment-3061561668

When use do a partial pipeline run like` kedro run -p "Data ingestion"` we will warn user with 
"Workflow run data tracking was skipped. `.viz/kedro_pipeline_events.json` was not generated because this was a partial pipeline run. Run plain `kedro run` to enable tracking."

## Checklist


package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -135,6 +135,9 @@ def before_pipeline_run(self, run_params: dict, pipeline) -> None:
         only for full/default pipeline.
         """
         if not is_default_run(run_params):
+            logger.warning(
+                "Workflow run data tracking was skipped. `.viz/kedro_pipeline_events.json` was not generated because this was a partial pipeline run. Run plain `kedro run` to enable tracking."
```
rashidakanchwala: I am not sure if i understood plain `kedro run` 
```suggestion
      "Workflow tracking skipped: `.viz/kedro_pipeline_events.json` not created due to partial run. Use full `kedro run` to enable tracking."
```

jitu5: @rashidakanchwala plain `kedro run` is same as full `kedro run` or without any other flags like `--pipeline` or `--to-outputs`.

rashidakanchwala: yeah but i am not sure if that terminology makes sense to users unless it's being used other places. Let's also confirm with @astrojuanlu because he used `plain` word as well. If you are both ok, then it's fine.

rashidakanchwala: although i am not sure abt full `kedro run` too

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -135,6 +135,9 @@ def before_pipeline_run(self, run_params: dict, pipeline) -> None:
         only for full/default pipeline.
         """
         if not is_default_run(run_params):
+            logger.warning(
+                "Workflow tracking skipped: `.viz/kedro_pipeline_events.json` not created due to partial run. Use full `kedro run` to enable tracking."
+            )
```
ravi-kumar-pilla, line 143: Workflow tracking is disabled during partial pipeline runs (executed using -p, --from-nodes, --to-nodes). `.viz/kedro_pipeline_events.json` will be created only during a full `kedro run`.

wdyt ?

ravi-kumar-pilla: The Workflow view is available for full pipeline runs using kedro run. Partial pipeline runs (using --from-nodes, --to-nodes, --tags, --pipeline, and more) are not supported. - This is what we have in the docs. May be we need similar warning here. 

Also if we want to show these warnings should we also handle the case of a different runner (Parallel or Thread) ?

package/kedro_viz/integrations/kedro/run_hooks.py
astrojuanlu: I concur with @ravi-kumar-pilla , we should call them "full pipeline runs" to be consistent with our docs https://kedro--2436.org.readthedocs.build/projects/kedro-viz/en/2436/workflow-view/#run-a-pipeline

package/kedro_viz/integrations/kedro/run_hooks.py
jitu5: @ravi-kumar-pilla 

> Also if we want to show these warnings should we also handle the case of a different runner (Parallel or Thread) ?

I added the check and logger for a different runner.

rashidakanchwala (approved):
lgtm

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -166,3 +166,27 @@ def is_default_run(run_params: dict) -> bool:
...
+
+    # If no runner specified, default is SequentialRunner
+    if runner is None:
+        return True
+
+    # If runner is a string, check if it contains "SequentialRunner"
+    if isinstance(runner, str):
+        return "SequentialRunner" in runner
+
+    # For non-string cases, return False (not SequentialRunner)
```
ravi-kumar-pilla:
```suggestion
    # For non-string cases (Custom Runner)
```

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -135,7 +136,17 @@ def before_pipeline_run(self, run_params: dict, pipeline) -> None:
         only for full/default pipeline.
         """
         if not is_default_run(run_params):
+            logger.warning(
+                "Workflow tracking is disabled during partial pipeline runs (executed using -p, --from-nodes, --to-nodes). `.viz/kedro_pipeline_events.json` will be created only during a full kedro run. See issue https://github.com/kedro-org/kedro-viz/issues/2443 for more details."
```
ravi-kumar-pilla:
```suggestion
                "Workflow tracking is disabled during partial pipeline runs (executed using --from-nodes, --to-nodes, --tags, --pipeline, and more). `.viz/kedro_pipeline_events.json` will be created only during a full kedro run. See issue https://github.com/kedro-org/kedro-viz/issues/2443 for more details."
```

package/kedro_viz/integrations/kedro/run_hooks.py
ravi-kumar-pilla: We have `if not is_default_run(run_params)` and `if not is_sequential_runner(run_params)`
checks only in `before_pipeline_run` but we are collecting events in other hook implementations like `after_dataset_loaded` etc. How are we skipping these triggers ?

jitu5: @ravi-kumar-pilla We have those checks in `after_pipeline_run` as well where we finally write all the events to file. If we have partial run or non sequential runner we skip writing to the file.

ravi-kumar-pilla: yes I saw that, but what about other hook implementations which still try to create events which are later discarded in `after_pipeline_run`. There seems to be no need to create those events if we are anyway not writing it to file ?

jitu5: Here I just wanted to avoid duplicating same checks in all the hooks.

479BetaCoder: it does feel odd to do checks in all the hook_impl...may be we can have a field in the hook class like `should_collect_events` which gets set in `before_pipeline_run`. This can be used in all other hook_impl. This will allow us not to do the event creation if we are not using it. wdyt ?

ravi-kumar-pilla: it does feel odd to do checks in all the hook_impl...may be we can have a field in the hook class like `should_collect_events` which gets set in `before_pipeline_run`. This can be used in all other hook_impl. This will allow us not to do the event creation if we are not using it. wdyt ?

jitu5: Yes, idea we can do that but we have to do this in all the hooks like this at least.

```
    @hook_impl
    def before_dataset_loaded(self, dataset_name: str, node: KedroNode) -> None:
        """Set context before a dataset is loaded by a node."""
        if not self._should_collect_events:
            return
        self._set_event_context(dataset_name, "loading", node)
```

ravi-kumar-pilla: I guess that would be the correct way to do until we support other runners.

ravi-kumar-pilla: I would leave it to the team, what they think of this. Thank you

jitu5: Okay perfect let me do that.

package/kedro_viz/integrations/kedro/hooks_utils.py
```diff
@@ -166,3 +166,27 @@ def is_default_run(run_params: dict) -> bool:
...
+    """
+    runner = run_params.get("runner")
+
+    # If no runner specified, default is SequentialRunner
+    if runner is None:
+        return True
+
+    # If runner is a string, check if it contains "SequentialRunner"
+    if isinstance(runner, str):
+        return "SequentialRunner" in runner
```
astrojuanlu: This logic is a bit weird 😅 what if I have a custom runner called `AwesomeSequentialRunner` or even `my_library.SequentialRunner` that happens to be incompatible with Kedro-Viz in some way? We have accepted that Kedro Viz has issues with the `ParallelRunner` but I think this is a bit of overshooting. @merelcht @ElenaKhaustova what do you think?

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -92,13 +93,11 @@ def __init__(self):
...
         self._started_nodes: set[str] = set()
+        self._should_collect_events: bool = False
 
     def _add_event(self, event: dict[str, Any], flush: bool = False) -> None:
         """Append one event to the events list and optionally flush to disk."""
-
-        # We add events only for full/default pipeline as for MVP we only support
-        # full/default pipeline.
-        if len(self._all_nodes) == 0:
+        if not self._should_collect_events:
```
ElenaKhaustova: Do we need `_add_event`? After the logic has changed it only returns the state of `_should_collect_events`. Can't we use it directly?

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -132,10 +131,24 @@ def before_pipeline_run(self, run_params: dict, pipeline) -> None:
...
         Records the beginning of a pipeline execution
-        only for full/default pipeline.
+        only for full/default pipeline runs with sequential runner.
         """
+        # Determine if we should collect events based on run parameters
         if not is_default_run(run_params):
+            logger.warning(
+                "Workflow tracking is disabled during partial pipeline runs (executed using --from-nodes, --to-nodes, --tags, --pipeline, and more). `.viz/kedro_pipeline_events.json` will be created only during a full kedro run. See issue https://github.com/kedro-org/kedro-viz/issues/2443 for more details."
+            )
+            self._should_collect_events = False
```
ElenaKhaustova: Do we need to set this to `False` here if it is a default value? I think it's safer to change the flags one way only.

package/kedro_viz/integrations/kedro/run_hooks.py
```diff
@@ -132,10 +131,24 @@ def before_pipeline_run(self, run_params: dict, pipeline) -> None:
...
             return
+
+        if not is_sequential_runner(run_params):
+            logger.warning(
+                "Workflow tracking is disabled for non-sequential runners. `.viz/kedro_pipeline_events.json` will be created only during a sequential run. See issue https://github.com/kedro-org/kedro-viz/issues/2443 for more details."
+            )
+            self._should_collect_events = False
```
ElenaKhaustova: Same about setting `False` here

package/kedro_viz/integrations/kedro/hooks_utils.py
ElenaKhaustova: I'm a bit confused by @astrojuanlu's examples. 🤔
With the current implementation, we return `True`, so they **will be compatible**  even though they might have nothing in common with the original sequential runner.

So I'm more concerned about the opposite problem - if `runner` name has `SequentialRunner` substring in it but the implementation is different from what we expect. 

If we want to keep this check, we should probably rely on the full name and do `==` check. Otherwise, this check doesn't seem to be working properly.

package/kedro_viz/integrations/kedro/run_hooks.py
Huongg: I think @jitu5 wanted to explicitly ensure the flag remains false in these conditions. That said, I see your point—since it's false by default, it may not be necessary. We could remove it, unless @jitu5 thinks there’s a specific risk in doing so?

package/kedro_viz/integrations/kedro/run_hooks.py
Huongg: Hey I think it’s clearer to keep everything within `_add_event` rather than calling append(event) and write_events(self._events) directly. I’ve updated the comment to better explain the purpose of this.

ravi-kumar-pilla (approved):

rashidakanchwala (approved):

Issue #2125, opened 2024-10-03
Enhance warning message for circular dependency detection in layers


tdhooghe:
## Description
When running Kedro Viz, the following warning rightfully appears when having circular dependencies in the data catalog: `WARNING Layers visualisation is disabled as circular dependency detected among layers`. This message does not provide sufficient information to diagnose which datasets in the catalog are causing the circular dependency.

## Context
Currently, the warning message makes it difficult to identify where the issue exactly. Providing more detailed information about the datasets (or layers) causing the circular dependency would greatly enhance the debugging experience.

## Possible Implementation
A very simple fix (which I am very happy to make a PR for), would be to log the `CycleError` exception and possibly also the `layer_dependencies`. 

## Possible Alternatives
A more advanced option would be to even show the exact datasets that are causing the circular dependency, but that might introduce too much complexity (haven't looked at it this deep yet).

## Checklist
?

rashidakanchwala:
This would be very useful. Please do go ahead and submit your PR. thanks

PR #2467, opened 2025-08-19
log the CycleError exception and layer_dependencies

Huongg:
## Description
Fixes https://github.com/kedro-org/kedro-viz/issues/2125

To provide clearer details about the datasets (or layers) that cause a circular dependency, the warning message now appears as follows:

Screenshot 2025-08-19 at 14 51 54

## Development notes
Update `catalog_03_prm.yml` by setting any of the layers to `raw`, then run `kedro viz` again. You should see the warning in the CLI (as shown in the screenshot above), and no layers will be displayed in the Kedro-Viz local host.
## Checklist


package/kedro_viz/services/layers.py
```diff
@@ -122,8 +122,13 @@ def find_child_layers(node_id: str) -> Set[str]:
         return sorted(
             sorted_layers, key=lambda layer: (sorted_layers.index(layer), layer)
         )
-    except CycleError:
+    except CycleError as e:
         logger.warning(
-            "Layers visualisation is disabled as circular dependency detected among layers."
+            "Layers visualisation is disabled as circular dependency detected among layers. "
+            "Circular dependency detected: %s. "
+            "Please check your pipeline configuration to ensure layers follow a proper hierarchy. "
```
ravi-kumar-pilla: may be we need to get design help for wording. What does proper hierarchy means ?

**Suggestion:**

```bash
"Layers visualisation is disabled as circular dependency is detected among {actual layers causing issue} layers. Please check the `layer` configuration in your catalog for the datasets: {actual dataset names}"
```

Huongg:
hey @ravi-kumar-pilla I like your suggestion! That said, I think it might need a bit more investigation to figure out how to get the {actual layers causing the issue} and {actual dataset names}. How about starting with something a bit simpler instead?

```
"Layers visualisation is disabled as circular dependency detected among layers. "
"Circular dependency detected: %s. "
"Please check the `layer` configuration in your catalog for the datasets to avoid circular references. "
"Layer dependencies found: %s",
```
Screenshot 2025-08-19 at 14 51 54

Huongg:
@stephkaiser @iamelijahko any thoughts on how to make this copy simpler and easier to understand?

rashidakanchwala:
So i am finding the CycleError message makes sense.

package/kedro_viz/services/layers.py
```diff
@@ -122,8 +122,13 @@ def find_child_layers(node_id: str) -> Set[str]:
...
-    except CycleError:
+    except CycleError as e:
         logger.warning(
-            "Layers visualisation is disabled as circular dependency detected among layers."
+            "Layers visualisation is disabled as circular dependency detected among layers. "
+            "Circular dependency detected: %s. "
+            "Please check the `layer` configuration in your catalog for the datasets to avoid circular references. "
+            "Layer dependencies found: %s",
+            str(e),
+            dict(layer_dependencies),
```
ravi-kumar-pilla: "Layers visualisation is disabled as circular dependency detected among layers: %s"
"Please check the `layer` configuration in your catalog for the datasets to avoid circular references. "
str(e)

ravi-kumar-pilla: Having `dict(layer_dependencies)` in the message is a bit confusing for me. This will be a huge message considering a bigger pipeline. Instead we can just pass the error caught. To be more clear if the error `e` is a tuple, the e[1] actually has the circular layer information. But we can avoid that for now. wdyt ?

ravi-kumar-pilla (approved):
Left my suggestion but everything else looks good...Thank you @Huongg

package/kedro_viz/services/layers.py
Huongg: yeah i see your point about the bigger pipeline. I think just having the layers is already good enough. Let me update it

PR #2487, opened 2025-09-16
Node Color Backend

ravi-kumar-pilla:
## Description

Partially resolves #2480 

## Development notes

- Added `styles.json` to `.viz` folder (In future this can be created via a hook, for now keeping it a manual edit)
- Created `NodeExtras` model which contains `stats` and `styles` for now.
- Updated REST API response to include `NodeExtrasAPIResponse`. This will also help in expanding any additional metadata we want to show for each node in the flowchart for future usecases.
- Removed `stats` key from DataNodeAPIResponse as it will be part of `NodeExtrasAPIResponse`
- Updated `data_access/managers.py` to include node_extras and removed `dataset_stats`
- Added utility functions to add_node_extras and get_extras_for_node
- Updated `add_node` and `add_dataset` to include `node_extras`
- Created helper functions `_get_node_styles` and `_create_node_extras_mapping` in `data_loader.py` to help with data loading from json and mapping node_names to node_extras which contain stats and styles loaded from 2 different files for now. In future, we can see how to combine both files and make it `node_extras.json`
- Updated all factory methods `create_task_node`, `create_data_node` etc to include `node_extras`\
- Updated few tests to fix format and lint issues

**NOTE:** 
1. I will add pytests in a separte PR (for easy reviews) once this approach looks good to the team as it contains a bit of refactor + styles addition
2. There is a separate [PR](https://github.com/kedro-org/kedro-viz/pull/2488) for frontend changes. I will update the frontend PR once the API response is finalized here.

## QA notes

- Checkout the current branch
- Install an editable package `pip install -e package`
- `cd demo project`
- Execute `kedro viz`
- Verify the `main` API response will now contain `node_extras` within each node response (since we have node metadata, viz metadata, I used the term extras to avoid confusion. Happy to hear other suggestions)

image

## Checklist

- [x] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes

PR #2488, opened 2025-09-16
Node Color Frontend

### LINKED ISSUE #2480: Custom colours for nodes and datasets in Kedro Viz
## Description
Currently, Kedro-Viz renders all nodes and datasets with the same default color scheme. This makes it difficult to visually distinguish between normal data pipelines and agent setup/agentic workflow components.

To improve clarity when working with agentic flows, we should allow custom colors for nodes and datasets in Kedro-Viz. This would make it easier to separate:

Agent setup nodes (e.g., intent detection, response generation context).

Regular ETL/ML pipeline nodes and datasets.

## Context

In the customer support automation use case, the graph mixes data ingestion pipelines with agent orchestration pipelines. Colouring agent-related nodes differently from standard datasets will make the visualization much more readable and help explain workflows to developers and non-technical stakeholders.

Image

ravi-kumar-pilla:
## Description

Partially resolves #2480 

## Development notes

- Implemented frontend support for `node_extras.styles` so node-level appearance can be driven by data from the backend.
- Added `processNodeStyles` helper in `src/components/draw/utils/draw-utils.js`
- Added a useEffect that iterates node DOM groups and applies processed styles when `node.extras.styles` exists:
  - Converts JS-style camelCase keys into CSS property names (strokeWidth → stroke-width) before applying.
  - Applies background-related properties (keys that include background or keys in ['fill','stroke','strokeWidth','opacity']) to the `.pipeline-node__bg` element using selection.style(cssProperty, value, 'important').
  - Applies text/font/color-style properties (keys including text, font, or the 'color' key) to the `.pipeline-node__text` and `.pipeline-node__icon` elements. For color specifically, sets fill instead of the property name.
  - Marks nodes with applied custom styles by adding the `kedro-viz-custom-styled` class to the node group.
  - The effect runs on changes to theme and nodes.
- Added a lightweight selector in `src/selectors/nodes.js` to surface node extras to the frontend
- `getNodeData` now accepts `nodeExtras` and uses it when producing node data - enabling the renderer to read `node.extras.styles`.
- Updated store normalization `src/store/normalize-data.js` to include extras in the initial pipeline state

## QA notes

- To QA along with backend, follow the instructions for backend setup in #2487 
- For frontend, checkout this branch
- Start both backend server (`make run`) and frontend (`npm start`)
- Load a graph where nodes include `node_extras.styles` in the `main` API. For this make sure your `demo_project` has `.viz/styles.json`

**NOTE:** If you do not want backend changes and still want to see color effects in the frontend, you can initialize the store with a default like below -

At line 165 in `src/store/normalize-data.js` provide a default node_extras which applies to all nodes.

```js
state.node.extras[id] = node.node_extras || {
    stats: {
      rows: 77096,
      columns: 5,
      file_size: 1810602
    },
    styles: {
      color: "white",
      themes: {
        dark: {
          fill: "red"
        },
        light: {
          fill: "green"
        }
      }
    }
  }
```
- Alternatively, I created a combined PR for testing here - #2489 

**Verify:**
- Nodes with `node_extras.styles` show the expected background, stroke, opacity, fill and text/icon color/font changes.
- Theme-specific overrides (if node.extras.styles.themes) apply when switching theme (dark/light).
- Nodes with styles get the `kedro-viz-custom-styled` class on their node group in the DOM.
- color style applies to text/icon fill as intended.
- No JS console errors are thrown when nodes without styles are rendered.

## Checklist


src/components/draw/draw-nodes.jsx
```diff
@@ -296,7 +298,64 @@ export function DrawNodes({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [nodes, orientation]);
 
+  // --- Apply theme-based node styling ---
```
Huongg, line 299: Hey @ravi-kumar-pilla, I was wondering if this logic really belongs here. From the description, I expected the custom styling to be passed down as a prop (like `nodeActive`, etc.). Something like `nodeStyleOverrides` or `nodeStyle` etc

In this component, it feels like we only need to know the nodeId and the colour code to apply styling. My concern with keeping this heavier logic here is performance — since each time <DrawNodes /> is called, it would have to run through all of this before realising it doesn’t actually need to. 

And also how about in the Workflow component? Does the custom styling gets applied there too? or just in Flowchart only?

ravi-kumar-pilla: Hi @Huongg , I felt DrawNodes is where the flowchart nodes are being rendered. The styles are applied for each node, so the loop is needed. Please suggest an alternative as I haven't worked on React for a while. I see this is how the animations and other things work on flowchart in this component. 

I separated the useEffect so this logic is not triggered everytime but just when there is a theme change or nodes change. I could not think of any other efficient way to do this.

> And also how about in the Workflow component? Does the custom styling gets applied there too? or just in Flowchart only?

Good question. I just saw we are using DrawNodes in workflow component too. So this might affect workflow component which should not be the case. Let me see what I can do. In the meantime, please suggest a better way if any. Thank you

Huongg: Hey Im thinking of something like this

flowchart.jsx

```

 /**
   * Create node style overrides map for theme-based styling
   */
  createNodeStyleOverrides = (nodes, theme) => {
    const nodeStyleOverrides = {};
    
    nodes.forEach(node => {
      if (node.extras && node.extras.styles && Object.keys(node.extras.styles).length > 0) {
        const processedStyles = processNodeStyles(node.extras.styles, theme);
        if (processedStyles) {
          nodeStyleOverrides[node.id] = processedStyles;
        }
      }
    });
    
    return nodeStyleOverrides;
  };


 return {
	……

    // Create node style overrides for theme-based styling
    const nodeStyleOverrides = this.createNodeStyleOverrides(nodes, theme);

   <DrawNodes
	…..
        nodeStyleOverrides={nodeStyleOverrides}
   />

}


```


Then in draw-nodes.jsx

```
export function DrawNodes({
  ….
 nodeStyleOverrides = {},
   
}) => {
   // --- Apply theme-based node styling ---
  useEffect(() => {
    const selections = getNodeSelections(groupRef, nodes);
    if (!selections) {
      return;
    }
    const { allNodes } = selections;
    
    // Apply custom styling to nodes with style overrides
    allNodes.each(function (node) {
      const nodeGroup = select(this);
      const nodeStyles = nodeStyleOverrides[node.id];
      
      if (nodeStyles) {
        Object.entries(nodeStyles).forEach(([key, value]) => {
          const cssProperty = key.replace(/([A-Z])/g, '-$1').toLowerCase();

          // Background properties
          if (
            key.toLowerCase().includes('background') ||
            ['fill', 'stroke', 'strokeWidth', 'opacity'].includes(key)
          ) {
            const bgElement = nodeGroup.select('.pipeline-node__bg');
            bgElement.style(cssProperty, value, 'important');
          }

          // Text properties
          else if (
            key.toLowerCase().includes('text') ||
            key.toLowerCase().includes('font') ||
            ['color'].includes(key)
          ) {
            const textElement = nodeGroup.select('.pipeline-node__text');
            const iconElement = nodeGroup.select('.pipeline-node__icon');

            [textElement, iconElement].forEach((element) => {
              if (key === 'color') {
                element.style('fill', value, 'important');
              } else {
                element.style(cssProperty, value, 'important');
              }
            });
          }
        });
        nodeGroup.classed('kedro-viz-custom-styled', true);
      }
    });
  }, [nodeStyleOverrides]);

}

Comment (Huongg) [reply src/components/draw/draw-nodes.jsx:299]: so with this approach, this specific useEffect for styling in render-nodes will only get called when `nodeStyleOverrides` is changed. If don't pass it through `Workflow` component, the styling never gets applied

Comment (Huongg) [reply src/components/draw/draw-nodes.jsx:299]: and i expect nodeStyleOverrides to return this format

```
{
  "node-id-1": {
    "color": "#ffffff",
    // ... other processed CSS properties
  },
  "node-id-2": {
    "fill": "#00ff00",
    // ... other processed CSS properties
  },
  // ... more node IDs
}
```

Comment (ravi-kumar-pilla) [reply src/components/draw/draw-nodes.jsx:299]: okay, so the advantage of this is workflow component will not get affected. Is there anything else I am missing ?
nodeStyleOverrides will be changed with theme or nodes change as before.

Comment (Huongg) [reply src/components/draw/draw-nodes.jsx:299]: Other benefit would be:
- Before: processNodeStyles() was called for each node on every render, now it called once in the parent only. And maybe the parent can potentially memoize nodeStyleOverrides() with `useMemo()` (i havent tried this approach locally yet but i have a feeling it might work)
- Simplify the draw-nodes.jsx component and make it more robust/re-usable?

Comment (ravi-kumar-pilla) [reply src/components/draw/draw-nodes.jsx:299]: Thanks for the suggestion @Huongg , I have implemented it. Please review and let me know if there are further changes required. I could not use `useMemo` as Flowchart is not a functional component.

**Comment by jitu5** [comment]
@ravi-kumar-pilla I observed one small thing when testing it locally,
With below style setting,
```
  state.node.extras[id] = node.node_extras || {
    stats: {
      rows: 77096,
      columns: 5,
      file_size: 1810602
    },
    styles: {
      themes: {
        dark: {
          fill: "red",
          color: "blue",
        }
      }
    }
  }
```

In dark theme it correctly applies red/blue colours, but if you check theme from button to light, colour remain same instead of default. But it works perfectly on reload. I am not sure if its a big problem to solve.  

lsp-validqqqq1

ravi-kumar-pilla:
Hi @rashidakanchwala and @jitu5, 

Thank you for pointing this out. There was a bug in the FE logic where I was not resetting the styles applied before reapplying when the theme is switched. I fixed it here - https://github.com/kedro-org/kedro-viz/pull/2488/commits/74b8eb07572d5c743137f1ef46b21853b0d3dd01

When I was debugging the above issue, I found the `createNodeStyleOverrides` method [here](https://github.com/kedro-org/kedro-viz/pull/2488/commits/74b8eb07572d5c743137f1ef46b21853b0d3dd01#diff-7c7aba87a2372c7ae12bcfdbdacad000cbe1b688dceb472107df95f00dbf941fR775) was called multiple times during rendering. I fixed it using a conditional check to call the method when the styles change - https://github.com/kedro-org/kedro-viz/pull/2488/commits/74b8eb07572d5c743137f1ef46b21853b0d3dd01#diff-7c7aba87a2372c7ae12bcfdbdacad000cbe1b688dceb472107df95f00dbf941fR797

Please have a look and let me know if there are any issues. Thank you

rashidakanchwala:
It works for me now! are we going to merge these two do one last QA before we merge to main?

rashidakanchwala (approved):

jitu5 (approved):
Thanks

ravi-kumar-pilla:
Hi Team, Now that the backend is merged, this branch contains both the changes. I tested it locally and it works. If you want to give a try please do so. Another question I have is, since there is no way to disable colors for now, 

1. I renamed `styles.json` to `styles.example.json` so our demo project will not have styles by default. But if we want to demo custom colors, we can rename it to `styles.json`
2. The other option would be to have styles only in light theme which is rarely used, and we can have no colors in dark theme so our regular experience in dark theme stays the same.

Please let me know your thoughts. @jitu5 @rashidakanchwala . Thank you

Issue #2343, opened 2025-05-26
Preview Text/Markdown Datasets


andrader:
## Description
Hi, it would be awesome to preview plaintext/markdown datasets!

## Context
Use case: especially for IA experiments, we keep all of our prompts and instructions as markdown files.

## Possible Implementation

I think a possible implementation would be to create a general HTMLPreview class on kedro viz and preview methods on appropriate dataset classes such as TextDataset.

## Possible Alternatives

Currrently, I am using a custom class of `TextDataset` with a preview method returning a `TablePreview` type. And using that class on the `catalog.yml`.  It shows the text, but the markdown syntax is obviously not applied.

```python
import logging

from kedro_datasets._typing import TablePreview
from kedro_datasets.text.text_dataset import TextDataset

logger = logging.getLogger(__name__)


class CustomTextDataset(TextDataset):
    def preview(self) -> TablePreview:
        str_data = self.load()

        logger.info(f"Previewing data from {self._filepath}")
        preview_data = {
            "index": [0],
            "columns": ["data"],
            "data": [[str_data]],
        }
        return TablePreview(preview_data)
```

## Checklist

- [ ] Include labels so that we can categorise your feature request

rashidakanchwala:
Thanks for the idea — I think that’s a great suggestion and definitely a valuable use case. I’ll add it to our backlog to be groomed.

On the Kedro-Viz side, we currently support `JSONPreview`, `ImagePreview`, `PlotlyPreview`, and `TablePreview`. We could extend this by adding support for an `HTMLPreview`, which would allow you to render formatted markdown or plaintext in a more readable way.

Once that’s in place, you’d be able to create your own CustomTextDataset that returns an HTMLPreview for markdown files.


cc - @astrojuanlu

andrader:
That would be great!!

ankatiyar:
A similar request https://github.com/kedro-org/kedro-viz/issues/2476 by @patcon  (I've closed that issue but would like to copy over the request)


## Description
I would like to be able to render markdown outputs in kedro-viz (even though it would require custom datasets to implement).

## Context
[...]

PR #2490, opened 2025-09-16
Feat/text markdown datasets

Huongg:
## Description
Fixes https://github.com/kedro-org/kedro-viz/issues/2343
This PR adds HTML preview functionality to Kedro-Viz, enabling users to preview text and markdown datasets in the meta-data panel

## Development notes

Changes include:
- CustomTextDataset class extending TextDataset
- HTMLPreview type annotation for automatic detection
- Preview method that returns text content for Kedro-Viz
- html-renderer component to render in meta-data panel
- using `react-markdown` to render the component and `isomorphic-dompurify` to sanitise the content

https://github.com/user-attachments/assets/7b72d1f3-2158-4f7b-8d95-8dd96cb62de6

### In catalog.yml
```
ingestion.readme:
  type: demo_project.extras.datasets.markdown_text_dataset.CustomTextDataset
  filepath: ${_base_location}/01_raw/sample_markdown.md
  metadata:
    kedro-viz:
      layer: raw
```

=> The dataset will automatically show HTMLPreview in Kedro-Viz

## Checklist


demo-project/src/demo_project/extras/datasets/markdown_text_dataset.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class HTMLPreview:
+    """Preview type for HTML content."""
+
+    def __init__(self, data: str):
+        self.data = data
+
+
+class CustomTextDataset(TextDataset):
```
rashidakanchwala: let's just call this MarkdownTextDataset

demo-project/src/demo_project/pipelines/data_ingestion/nodes.py
```diff
@@ -29,6 +30,7 @@ def apply_types_to_shuttles(shuttles: pd.DataFrame) -> pd.DataFrame:
 
     Args:
         shuttles: Raw data.
+        config: Configuration data from YAML file.
```
rashidakanchwala: what's this?

src/components/html-renderer/html-renderer.jsx
```diff
@@ -0,0 +1,19 @@
...
+import './html-renderer.scss';
+
+const HTMLRenderer = ({ content, className = '', fontSize }) => {
+  const style = fontSize ? { fontSize } : {};
+
+  return (
+    <div
+      className={`html-renderer ${className}`}
+      style={style}
+      dangerouslySetInnerHTML={{
```
rashidakanchwala: Since we can’t guarantee what user-authored content will contain, using dangerouslySetInnerHTML without sanitization is risky. We could switch to react-markdown or consider another lighter alternative  (https://www.npmjs.com/package/isomorphic-dompurify) to sanitize the content?

demo-project/src/demo_project/pipelines/data_ingestion/nodes.py
Huongg: oh this was for the old config file when i did testing, will remove it

demo-project/src/demo_project/extras/datasets/markdown_text_dataset.py
Huongg: sounds good

package.json
```diff
@@ -182,7 +185,9 @@
     ],
     "moduleNameMapper": {
       "\\.(css|scss)$": "identity-obj-proxy",
-      "^(.*)\\?worker&inline$": "$1"
+      "^(.*)\\?worker&inline$": "$1",
```
Huongg, line 188: When Jest tried to run the tests, it couldn't understand the ES module syntax, especially from react-markdown. So the work around is to use these mocks, which seems to provide enough functionality to test component's props, rendering, and basic functionality

package.json
```diff
@@ -182,7 +185,9 @@
...
+      "^react-markdown$": "<rootDir>/src/components/html-renderer/__mocks__/react-markdown.js",
```
Huongg: When Jest tried to run the tests, it couldn't understand the ES module syntax, especially from react-markdown. So the work around is to use these mocks, which seems to provide enough functionality to test component's props, rendering, and basic functionality

demo-project/conf/base/catalog_01_raw.yml
```diff
@@ -27,3 +27,11 @@ shuttles:
       layer: raw
       preview_args:
         nrows: 15
+
+# Test markdown dataset for Kedro-Viz preview functionality
```
Huongg: we can remove this before merging

demo-project/src/demo_project/pipelines/data_ingestion/pipeline.py
```diff
@@ -24,14 +24,14 @@ def create_pipeline(**kwargs) -> Pipeline:
         [
             node(
                 func=apply_types_to_companies,
-                inputs="companies",
+                inputs=["companies", "readme"],
```
Huongg: TBR before merging

demo-project/src/demo_project/pipelines/data_ingestion/nodes.py
```diff
@@ -8,11 +8,12 @@ def _is_true(column: pd.Series) -> pd.Series:
     return column == "t"
 
 
-def apply_types_to_companies(companies: pd.DataFrame) -> pd.DataFrame:
+def apply_types_to_companies(companies: pd.DataFrame, readme: str) -> pd.DataFrame:
```
Huongg: TBR before merging

src/components/html-renderer/utils.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * Simple markdown-to-HTML converter for basic markdown features
```
jitu5, line 2: @Huongg In this utils we are converting markdown to HTML with the help of some regex and with some specific order of conversion, Have you consider using existing popular npm library like [marked](https://www.npmjs.com/package/marked) or [markdown-it](https://github.com/markdown-it/markdown-it) ? 

In the case of complex markdown below conversion code might or might not handle all the cases wdyt ?

Huongg: hey @jitu5 thanks for the suggestion, i havent used those myself yet. Would you recommend? Originally we were going to use the react-markdown library to render the whole thing then I reverted it back as we didnt want to use extra package for this.

jitu5: @Huongg I’ve never used any of those before, but from looking at their GitHub pages, they seem popular. The benefit of using a library like react-markdown or marked is that we don’t have to handle or worry about edge cases or complex Markdown conversion ourselves. At the same time, I understand that it adds an extra dependency to kedro-viz.

Huongg: ah okay cool, if so then maybe we can go back to use react-markdown so we dont need extra util file to handle this. What do you think @rashidakanchwala ?

rashidakanchwala: We need to decide whether to add a dependency like `react-markdown` (heavier, but handles sanitisation and other features out of the box) or use a lighter option like `isomorphic-dompurify` and handle the markdown rendering ourselves. Since HTML/Markdown preview is only for datasets and we don’t expect very heavy markdown files, I’m leaning toward the lighter option. If we see user demand or edge cases later, we can always switch to react-markdown. What do you all think?

Huongg: that sounds very sensible to me @rashidakanchwala 👍 

if so is this PR good to merge? @jitu5 @rashidakanchwala ?

jitu5 (approved):

src/components/html-renderer/utils.js
jitu5: @Huongg All good from my side.

PR #2510, opened 2025-10-27
fix(viz): support bound methods in node metadata

MalikAly:
Handle bound methods in TaskNodeMetadata.set_code() and set_filepath() by unwrapping with .__func__. This enables kedro-viz to display source and filepath for nodes implemented as class methods. Backward compatible.

## Description

Kedro-Viz currently extracts node metadata (source, filepath) only when `kedro_node.func` is a plain function (`inspect.isfunction`). For **bound methods**, `inspect.isfunction()` returns `False`, so metadata is skipped even though `inspect.getsource()`/`inspect.getfile()` work when unwrapping via `.__func__`.

This PR adds support for **class/bound methods** by detecting `inspect.ismethod()` and unwrapping to the underlying function with `.__func__` in both validators.

**Effect:** Source code and file path are now shown for nodes implemented as class methods (in addition to regular functions).
**Compatibility:** Minimal and backward-compatible.

## Development notes

* Updated `package/kedro_viz/models/flowchart/node_metadata.py`:

  * `TaskNodeMetadata.set_code()`:

    * If `inspect.isfunction(func)`: unchanged (use `_extract_wrapped_func(func)`).
    * If `inspect.ismethod(func)`: unwrap with `func.__func__` then pass to `_extract_wrapped_func(...)` before `inspect.getsource(...)`.

  * `TaskNodeMetadata.set_filepath()`:

    * Mirrored logic using `inspect.getfile(...)` on `func` vs `func.__func__`.

  * Preserved existing relative-path resolution and error handling.

### Unit tests

* Added unit test to cover a **bound method** node to ensure both `code` and `filepath` are populated.

  * Asserts `set_code(None)` returns the method’s source (contains `def ...(`).
  * Asserts `set_filepath(None)` returns a plausible `.py` path (existing file or endswith `.py`).

### Risk / Perf

* Low risk: conditional branch on `inspect.ismethod` only.
* Negligible performance impact.

## Checklist

- [] Opened this PR as a 'Draft Pull Request' if it is work-in-progress

### DIFF
--- a/.github/workflows/deploy-demo-with-kedro-builder.yml
+++ b/.github/workflows/deploy-demo-with-kedro-builder.yml
@@ -1,0 +1,96 @@
+# Simple workflow for deploying static content to GitHub Pages with Kedro Builder
+name: Deploy static content to Pages with Kedro Builder
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+env:
+  PYTHON_VERSION: '3.10'
+
+jobs:
+  # Single deploy job since we're just deploying
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: demo
+
+      - name: Set up Python ${{ env.PYTHON_VERSION }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Cache python packages for Linux
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/pip
+          key: ubuntu-latest-python-3.10
+
+      - name: Install Kedro and other Python Dependencies
+        uses: "./.github/actions/install_kedro_and_python_dependencies"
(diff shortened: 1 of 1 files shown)


package/kedro_viz/models/flowchart/node_metadata.py
```diff
@@ -83,8 +83,13 @@ def set_task_and_kedro_node(cls, task_node):
     @classmethod
     def set_code(cls, code):
         # this is required to handle partial, curry functions
-        if inspect.isfunction(cls.kedro_node.func):
-            code = inspect.getsource(_extract_wrapped_func(cls.kedro_node.func))
+        func = cls.kedro_node.func
+
+        if inspect.ismethod(func):
+            func = func.__func__
```
ravi-kumar-pilla, line 89: Hi @MalikAly , The implementation looks good. I am not sure if accessing internal `__func__` is a recommended way. Have you tried `inspect.unwrap(func)` ?

RELEASE.md
```diff
@@ -13,6 +13,7 @@ Please follow the established format:
 
 ## Bug fixes and other changes
  - Add HTML preview support for text/markdown datasets. (#2490)
+ - Display source metadata for task nodes defined as class methods.
```
ravi-kumar-pilla:
```suggestion
 - Display source metadata for task nodes defined as bound methods.
```

ravi-kumar-pilla: I think bound methods are specific to instance. Can you adjust the title of the PR to be consistent with the release comment ? Thank you

package/kedro_viz/models/flowchart/node_metadata.py
MalikAly: Hi, thanks for the review! I believe that accessing `.__func__` on bound methods is the standard way to retrieve the underlying function (see the “Instance methods” section of the Python data model: [Python documentation][1])
`inspect.unwrap()` is still helpful for removing decorators, but from what I understand, it doesn’t replace the need for `.__func__` in this context.

[1]: https://docs.python.org/3/reference/datamodel.html#instance-methods "3. Data model — Python 3.14.0 documentation"

rashidakanchwala (approved):
Thanks. LGTM :)

Issue #2411, opened 2025-07-02
Refactor metadata component


jitu5:
## Description
Current metadata component is is getting bigger and it needs refactor after adding run status/workflow code in it. Its already 450+ lines for code.

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

## Checklist

- [ ] Include labels so that we can categorise your feature request

PR #2551, opened 2026-01-16
Support TaskNode Previews

ravi-kumar-pilla:
## Description

Part of https://github.com/kedro-org/kedro/issues/5317

**NOTE:** 

1. Added a `preview_tests` pipeline for testing preview functionality within demo-project. This pipeline will be removed before merging the PR. It is added for testing purposes
2. The file changes seem big but there are some file renames and refactoring which is much needed
3. Pushed `html` folder for testing. This will be reverted back after testing is completed

### Overview
This PR standardizes the naming convention for preview renderer components and refactors the preview rendering logic to eliminate code duplication between DataNode and TaskNode previews.

---

### 1.1 Renamed `preview-table` → `table-renderer`
**Rationale**: Standardized naming to follow the `componentName-renderer` pattern used by other renderers.

**Changes**:
- **Directory**: `src/components/preview-table/` → `src/components/table-renderer/`
- **Component**: `PreviewTable` → `TableRenderer`
- **CSS Classes**: `preview-table` → `pipeline-table-renderer`
- **SCSS Variable**: `$zindex-preview-table-header` → `$zindex-table-renderer-header`

**Files Modified**:
- `src/components/table-renderer/table-renderer.jsx` (new)
- `src/components/table-renderer/table-renderer.scss` (new)
- `src/components/table-renderer/table-renderer.test.jsx` (new)
- `src/components/metadata/metadata.jsx` (imports updated)
- `src/components/metadata-modal/metadata-modal.jsx` (imports updated)
- `src/styles/_variables.scss` (variable renamed)

**Files Deleted**:
- `src/components/preview-table/` (entire directory)

### 1.2 Renamed `json-object` → `json-renderer`
**Rationale**: Consistent with other renderer components.

**Changes**:
- **Directory**: `src/components/json-object/` → `src/components/json-renderer/`
- **Component**: `JSONObject` → `JsonRenderer`

**Files Modified**:
- `src/components/json-renderer/json-renderer.jsx` (new)
- `src/components/json-renderer/json-renderer.scss` (new)
- `src/components/json-renderer/json-theme.js` (moved)
- `src/components/metadata/metadata.jsx` (imports updated)
- `src/components/metadata-modal/metadata-modal.jsx` (imports updated)
- `src/components/metadata/metadata-row.jsx` (imports updated)

**Files Deleted**:
- `src/components/json-object/` (entire directory)

---

### 2.1 Created `PreviewWrapper` Component
**File**: `src/components/metadata/preview-wrapper.jsx`

**Purpose**: Eliminates duplication by providing a reusable wrapper for all preview renderers with:
- Consistent scrollable container with shadow boxes
- Expand button functionality
- Configurable className and shadow visibility

**Props**:
- `children`: The preview renderer component
- `onExpand`: Callback for expand button
- `className`: Custom CSS class (default: `pipeline-metadata__preview`)
- `showShadows`: Whether to show scrollable shadow hints (default: `true`)
- `onClick`: Optional click handler for the container

### 2.2 Unified Preview Data Normalization
**Function**: `getNormalizedPreview()`

**Purpose**: Normalizes DataNode and TaskNode preview data into a unified format, eliminating the need for separate rendering logic.

**Data Structure**:
```javascript
{
  kind: 'plotly' | 'image' | 'table' | 'json' | 'html' | 'text' | 'mermaid',
  content: <preview content>,
  meta: <metadata object>,
  isDataNode: boolean
}
```

**Type Mapping**:
- DataNode: `PlotlyPreview` → `plotly`, `ImagePreview` → `image`, etc.
- TaskNode: Uses `kind` directly from preview payload

### 2.3 Unified Preview Rendering Function
**Function**: `renderPreview()`

**Purpose**: Single rendering function that handles all preview types for both DataNode and TaskNode using normalized data.

**Supported Preview Types**:
- `plotly`: PlotlyRenderer
- `image`: Image with expand functionality
- `table`: TableRenderer
- `json`: JsonRenderer
- `html`: HTMLRenderer
- `text`: TextRenderer
- `mermaid`: MermaidRenderer

**Benefits**:
- Eliminated ~160 lines of duplicate conditional rendering code
- Single source of truth for preview rendering logic
- Easier to add new preview types

## QA notes

- Install `kedro-viz`

```bash
pip install git+https://github.com/kedro-org/kedro-viz.git@feat/node_preview#subdirectory=package
```
- If you want to test against the `preview_tests` pipeline in the `demo-project`, pull the branch `feat/node_preview`
- `cd demo-project`
- Run `kedro viz`
- Test for all DataNode and TaskNode previews
- Alternately you can test for any kedro project implementing DataNode previews as before and TaskNode previews as documented [here](https://github.com/kedro-org/kedro/issues/5317) 

### Components to Test

**Unified Preview Rendering**: Verify all preview types work for both DataNode and TaskNode

### Test Scenarios
- [ ] DataNode previews (PlotlyPreview, ImagePreview, TablePreview, JSONPreview, HTMLPreview)
- [ ] TaskNode previews (plotly, image, table, json, text, mermaid)
- [ ] Code previews render inline without positioning issues
- [ ] Expand button works for all preview types
- [ ] Modal view renders correctly for all preview types

## Checklist

### DIFF
--- a/src/components/metadata/metadata.jsx
+++ b/src/components/metadata/metadata.jsx
@@ -94,18 +94,10 @@ const MetaData = ({
       };
     }
 
-    // Check for TaskNode preview - only supports mermaid, text, and image
+    // Check for TaskNode preview
     if (metadata?.preview && metadata.preview.kind) {
-      const supportedTaskNodePreviews = ['mermaid', 'text', 'image'];
-      const previewKind = metadata.preview.kind;
-
-      // Filter out unsupported preview types for TaskNode
-      if (!supportedTaskNodePreviews.includes(previewKind)) {
-        return null;
-      }
-
       return {
-        kind: previewKind,
+        kind: metadata.preview.kind,
         content: metadata.preview.content,
         meta: metadata.preview.meta || {},
         isDataNode: false,


demo-project/src/demo_project/pipelines/preview_tests/nodes.py
```diff
@@ -0,0 +1,301 @@
...
+def make_preview_fn(data_sample):
+    """Create a preview function with captured context."""
+    def preview_fn() -> TablePreview:
+        return TablePreview(content=data_sample)
+    return preview_fn
+
+from kedro.framework.session import KedroSession
+
+
+def preview_with_data_access() -> TablePreview:
```
rashidakanchwala, line 26: is this example jus to show that this way is possible, I wouldn't add it to the demo.kedro.org; maybe just in the kedro docs, because ideally the better way to show this is directly using dataset previews.

demo-project/src/demo_project/pipelines/preview_tests/nodes.py
```diff
@@ -0,0 +1,301 @@
...
+
+It can contain multiple lines of text.
+Useful for displaying log outputs, reports, or any textual information.
+
+Line 1: Sample data
+Line 2: More information
+Line 3: Additional details
+This is a plain text preview.
+
+It can contain multiple lines of text.
```
rashidakanchwala, line 56: why is this repeated so many times?

package/kedro_viz/models/flowchart/model_utils.py
```diff
@@ -32,6 +38,149 @@ def _extract_wrapped_func(func: FunctionType) -> FunctionType:
     # Return the original function if no wrapping detected
     return func
 
+# Preview limiter helpers
+
+# Constants
+DEFAULT_ROW_LIMIT = 5
```
rashidakanchwala: Can we follow the same logic as dataset previews for this part. I don't think we need all this.

src/components/metadata/metadata.jsx
```diff
@@ -3,17 +3,19 @@ import { connect } from 'react-redux';
 import modifiers from '../../utils/modifiers';
 import NodeIcon from '../../components/icons/node-icon';
 import IconButton from '../../components/ui/icon-button';
-import PreviewTable from '../../components/preview-table';
-import JSONObject from '../../components/json-object';
+import TableRenderer from '../../components/table-renderer';
```
rashidakanchwala, line 6: This component and metadata-model is getting too long. I think we should do a refactoring.

package/kedro_viz/models/flowchart/model_utils.py
```diff
@@ -32,6 +38,149 @@ def _extract_wrapped_func(func: FunctionType) -> FunctionType:
...
+    if isinstance(content, dict):
+        limited = _limit_json_nested(content, limit)
+        if limited != content:
+            return limited, _update_meta_with_truncation(meta)
+
+    return content, meta
+
+
+# limiter api
+def limit_preview_data(preview: dict) -> dict:
```
rashidakanchwala: I’m getting a bit confused here. If users want to preview data, they should use dataset previews — we already have good support for that, so node previews for tables/JSON feel unnecessary.

I know Kedro core allows node previews to return tables, JSON, etc., but do we actually know when a user would want to do that? Since these are outputs, not the node logic itself, why wouldn’t users just rely on dataset previews instead?

Since Node previews are primarily for GenAI-style use cases not for data inspection, could we restrict node previews to only support Mermaid (and maybe text) for now (similar to how dataset previews were intentionally limited to just tables at first)? We can always extend this later if a real use case emerges.

At the moment, the additional limiting logic feels like over-engineering, driven by the fact that Kedro technically supports table/JSON previews on nodes — and because there’s no clean way to restrict this at the project level, we’re adding complexity we may not actually need.

cc - @ElenaKhaustova

demo-project/src/demo_project/pipelines/preview_tests/nodes.py
ravi-kumar-pilla: Hi @rashidakanchwala , Thank you for reviewing. As mentioned in the PR description, I will revert all the changes in demo-project. I added this pipeline just to make testing easy, so we can see how each preview looks and also tweak a little to test any other scenarios

demo-project/src/demo_project/pipelines/preview_tests/nodes.py
ravi-kumar-pilla: This is just for testing and to see how long texts work. This will be removed as mentioned in the description

package/kedro_viz/models/flowchart/model_utils.py
ravi-kumar-pilla: We can discuss on this but I think we should have a limit for dataset previews as well to stop users breaking Kedro-Viz. I tested this and Kedro-Viz breaks if the data is huge (which is a bad dev experience). 

We can improve our docs saying users to use the preview feature for just summaries but we should also have guardrails (atleast some) to not break the UI.

src/components/metadata/metadata.jsx
ravi-kumar-pilla: I agree on this. I tried separating the preview using a preview wrapper and I also removed around 160 lines of repeated code. For this PR, I do not want to refactor the whole file but just the preview part of it.

package/kedro_viz/models/flowchart/model_utils.py
ravi-kumar-pilla: Regarding preview types, I think we had this discussion initially and during the KEP we thought of supporting all preview types that datasets support and also additional preview types like Mermaid and Text. cc: @ElenaKhaustova 

But if we think this is too much, we can add later as well. I would say we can keep it as most of the work is done which follows the KEP.

---

> At the moment, the additional limiting logic feels like over-engineering

I feel we should have limiting logic even for dataset previews. We can discuss on this. My reasoning being this - https://github.com/kedro-org/kedro-viz/pull/2551#discussion_r2706510428

This might seem over-engineering but if someone can break the app by returning entire data (which we allow to do so) does not look good to me. wdyt ?

ravi-kumar-pilla: > I know Kedro core allows node previews to return tables, JSON, etc., but do we actually know when a user would want to do that? Since these are outputs, not the node logic itself, why wouldn’t users just rely on dataset previews instead?

I think this is a fair point. Even I do not have a clear picture on when someone will use table or json or any previews that dataset already supports. But I think the reasoning was - this is a user-injectable preview function and users are free to return whatever they want to view on `kedro-viz`. Since we have support for some preview types via dataset previews, lets reuse them and add additional preview types. 

@ElenaKhaustova do you know of any specific use case ? 

Thank you

ElenaKhaustova: That’s a fair point - dataset previews should remain the primary way to inspect data. I see node previews as complementary: more about explaining node behaviour or decisions (Mermaid, text, GenAI outputs) than raw data inspection.

I agree that we don’t currently have strong, concrete use cases for returning tables or JSON from node previews. The bare minimum for node previews is probably Mermaid, text and maybe images, especially since dataset previews already cover data inspection well. Given the current implementation on the Kedro side, it’s also unlikely users would preview large tables anyway, since the content has to be constructed inside the node.

That said, I don’t see a strong reason to explicitly restrict the contract unless it adds real overhead on the Viz side. Supporting additional types in the contract doesn’t mean Viz has to render them now similar to how dataset previews started small and evolved over time. Also, the additional preview types aren’t intended for the dataset inspection. They’re meant for lightweight, user-defined outputs that help explain node behaviour.

I think part of the concern may be that this PR bundles a lot together. Some of the refactoring and renaming makes sense on its own and would be useful regardless of node previews. If we split this into multiple PRs (core refactors first, node preview support after), it might reduce the impression that node previews themselves are adding too much complexity.

But anyway, I’d be happy if Viz initially only supported `mermaid` / `images` / `text`, and that would be a good way to start.

src/components/metadata/metadata.jsx
jitu5: Yes, this need refactor, we already have ticket for that. https://github.com/kedro-org/kedro-viz/issues/2411

src/components/mermaid-renderer/mermaid-renderer.jsx
```diff
@@ -0,0 +1,137 @@
...
+        setError(null);
+        // Small delay to ensure container is sized
+        await new Promise((resolve) => setTimeout(resolve, 0));
+
+        // Generate unique ID for each diagram
+        const id = `mermaid-${Math.random().toString(36).substring(2, 9)}`;
+        const { svg } = await mermaid.render(id, content);
+
+        // Insert SVG into container
+        containerRef.current.innerHTML = svg;
```
jitu5, line 84: We might need cleanup on unmount

```
return () => {
  if (containerRef.current) {
    containerRef.current.innerHTML = '';
  }
}
```

src/components/metadata-modal/metadata-modal.jsx
```diff
@@ -96,6 +129,67 @@ const MetadataModal = ({ metadata, onToggle, visible, theme }) => {
           <HTMLRenderer content={metadata.preview} fontSize="15px" />
         </div>
       )}
+      {/* TaskNode PreviewPayload renderers */}
```
jitu5, line 132: There's no handler for `previewKind === 'html'` for TaskNode previews. Is it expected?

src/components/metadata-modal/metadata-modal.jsx
```diff
@@ -12,13 +14,39 @@ import { getClickedNodeMetaData } from '../../selectors/metadata';
...
+      !tableContent ||
+      !Array.isArray(tableContent) ||
+      tableContent.length === 0
+    ) {
+      return { columns: [], data: [] };
+    }
+    const columns = Object.keys(tableContent[0]);
+    const data = tableContent.map((row) => columns.map((col) => row[col]));
+    return { columns, data };
+  };
```
jitu5, line 46: [nit] May be we can move this `transformTableData` function to some where in util. I see `transformTableData` in `metadata.jsx` as well.

package/kedro_viz/models/flowchart/model_utils.py
rashidakanchwala: Yeah, i think let's do that; let's support mermaid, images, and text for now; and then later if there's a user ask we can support more.

package/kedro_viz/models/flowchart/model_utils.py
rashidakanchwala: From a dataset preview perspective, most previews already have a default limit set in kedro-datasets (typically 5 rows). If a user explicitly specifies a very large limit (e.g. 10000) in catalog.yml and that ends up breaking Kedro-Viz, that’s a valid problem. However, silently truncating the preview back to 5 rows is not the right way to solve it.

In this case, we should make it explicit to the user that their requested preview size is too large and guide them accordingly, rather than altering the preview output without any signal.

package/kedro_viz/models/flowchart/model_utils.py
ElenaKhaustova: One of the example of why we might need other formats can be our `llm_context_node`. We could, for example, easily add tools to be seen at the preview.

package/kedro_viz/models/flowchart/model_utils.py
ravi-kumar-pilla: > If a user explicitly specifies a very large limit (e.g. 10000) in catalog.yml and that ends up breaking Kedro-Viz, that’s a valid problem. However, silently truncating the preview back to 5 rows is not the right way to solve it.

Yes I agree on that. We should document this behavior and may be add a warning so the user is aware. But not implementing the limit itself might cause the app to break. 

> In this case, we should make it explicit to the user that their requested preview size is too large and guide them accordingly, rather than altering the preview output without any signal.

Yes. Let me know if you have any suggestions apart from documenting and placing a warning. Thank you

rashidakanchwala: For now, since we are only supporting mermaid, text, and image previews. The limiting logic applies to json and table, so it isn’t required for node previews.

For dataset previews, we can create a separate ticket and address this properly -- adding documentation and warnings seems like the best approach. We should also investigate at which stage the app breaks so we can introduce appropriate error boundaries. But let's do that seperately not as part of this

ravi-kumar-pilla: Sure, removing this from the PR. Thank you

package/kedro_viz/models/flowchart/model_utils.py
ravi-kumar-pilla: So I will keep only mermaid / images / text for TaskNode previews. 

> I think part of the concern may be that this PR bundles a lot together.

I should have refactored it in a different PR but I was doing it along the way developing. I thought the PR description would make it clear. If it is hard to follow, I can split this work or create a new PR without any refactors. Let me know 
@rashidakanchwala @ElenaKhaustova @jitu5

ElenaKhaustova: +1 for splitting

ravi-kumar-pilla: I will split the PR.

src/components/metadata-modal/metadata-modal.jsx
ravi-kumar-pilla: yes

PR #2583, opened 2026-02-27
fix: URL tag parameters not reflected in filter panel

### LINKED ISSUE #2186: Mismatch Between Tags in URL and Tags Displayed in UI
## Description
When a URL is entered with default tags, the specified value appears in the URL. However, in the tags filter panel of the UI, this tag option is missing.

For example, as shown in the demo screenshot, the URL includes tags=feature_engineering, but the filters panel in the UI does not display an option for feature_engineering.
Screenshot 2024-11-12 at 15 31 21

## Root cause
This issue may be due to the URL-stored tags being saved in localStorage even if they aren’t existed in the Redux store, or the opposite. Further investigation is needed to confirm the root cause.

## Checklist

priya-gitTest:
When navigating to a URL with tag parameters (e.g. ?tags=preprocessing), the tags appeared in the URL but were not reflected in the UI filter panel.

Root cause: `getInitialState` parsed URL params but never passed them to `preparePipelineState`. This meant `mergeLocalStorage` would override the URL-applied tag state, and `applyUrlParametersToPipelineState` (which re-applies URL params after the localStorage merge) was never called.

Fix:
- Pass `urlParams` to `preparePipelineState` in `getInitialState` so that URL tag parameters correctly override localStorage settings on initial load
- Add project-specific localStorage key helpers (`getProjectStorageKey`, `buildStorageKey`) to prevent tag state conflicts between different projects
- Use project-specific keys in `store.js` and `filters-context.jsx`
- Fix `filters-context.jsx` where `storedState` was removed but its usage in `useState` and `handleToggleGroupCollapsed` was not updated
- Remove orphaned dead code (`return state; };`) in `initial-state.js`
- Add test verifying URL tag params override localStorage on initial load

Fixes #2186

## Checklist


ravi-kumar-pilla:
Thank you for contributing @priya-gitTest . We will be reviewing it shortly. Could you please sign off commits following - https://github.com/kedro-org/kedro-viz/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

rashidakanchwala:
Closing this PR for now. We will look and revise this solution later on

PR #2609, opened 2026-03-31
NodeContextMenu support for VSCode

jitu5:
## Description                                                                                                                          
                                                                                                                                        
  Adds a right-click context menu integration point to Kedro-Viz, enabling embedded consumers (e.g. the Kedro VS Code extension) to respond to node right-click events without resorting to DOM scraping.
                                                                                                                                          
  ### Context                                                                                                                          
   
  The [Kedro VS Code extension](https://github.com/kedro-org/vscode-kedro/pull/257) embeds Kedro-Viz inside a webview to provide a one-click debug experience — users right-click a pipeline node and select it as a debug target. Without a first-class API for right-click events, the extension was forced to implement ~80 lines of fragile DOM scraping: capture-phase event listeners, SVG class name traversal, D3 `__data__` property access, and `composedPath` fallbacks. This approach breaks whenever Kedro-Viz refactors its internals.

  These changes expose right-click as a proper integration point through the existing `onActionCallback` middleware pattern, eliminating the need for any DOM introspection by consumers.
                                                                                                                                          
  ### Changes                                                          

  #### `src/actions/nodes.js`
  - Added `NODE_CONTEXT_MENU` action constant
  - Added `nodeContextMenu(nodeId)` action creator                                                                                        
   
  #### `src/components/draw/draw-nodes.jsx`                                                                                               
  - Added `onNodeContextMenu` prop                                     
  - Added `data-node-type` and `data-node-fullname` attributes to each node element in the D3 enter chain, consistent with the existing
  `data-id` attribute                                                                                                                     
  - Wired `.on('contextmenu', onNodeContextMenu)` in the D3 enter chain
  - Added `onNodeContextMenu` to the `useEffect` dependency array                                                                         
                                                                                                                                          
  #### `src/components/flowchart/flowchart.jsx`                                                                                           
  - Imported `nodeContextMenu` action                                                                                                     
  - Added `handleNodeContextMenu` handler — intentionally does **not** call `preventDefault()` or `stopPropagation()` so the native
  browser context menu still fires                                                                                                        
  - Passed `onNodeContextMenu={this.handleNodeContextMenu}` to `<DrawNodes>`
  - Added `onNodeContextMenu` dispatch in `mapDispatchToProps`                                                                            
                                                                                                                                          
  #### `src/store/middleware.js`
  - Imported `NODE_CONTEXT_MENU`                                                                                                          
  - Added `NODE_CONTEXT_MENU` case to the callback middleware switch, following the same pattern as `TOGGLE_NODE_CLICKED` — looks up the
  full node object from Redux state and passes it as `payload` to the callback                                                            
   
  #### Tests                                                                                                                              
  - `draw-nodes.test.jsx` — added tests for `data-node-type` and `data-node-fullname` attribute rendering
  - `middleware.test.js` — added tests for the `NODE_CONTEXT_MENU` middleware case (node found and node not found)                        
                                                                                                                                          
  ### Backward compatibility
                                                                                                                                          
  All changes are purely additive:                                                                                                        
   
  - `onNodeContextMenu` is an optional prop — omitting it means D3 registers no handler (no-op)                                           
  - The middleware guard (`if (!callback) return next(action)`) ensures the new case is never reached in standalone Kedro-Viz where no
  `onActionCallback` is provided                                                                                                          
  - `handleNodeContextMenu` does not suppress the native context menu, so right-click behaviour in the browser is unchanged
  - No existing reducer handles `NODE_CONTEXT_MENU`, so Redux state is unaffected                                                         
  - `data-*` attributes on SVG elements are inert metadata with no functional side effects                                                
                                                                                                                                          
  ### How consumers use this                                                                                                              
                                                                                                                                          
  Consumers that pass `onActionCallback` to `configureStore` will now receive:                                                            
   
  ```js                                                                                                                                   
  {                                                                    
    type: 'NODE_CONTEXT_MENU',
    payload: {                                                                                                                            
      id: 'abc123',
      name: 'split_data',                                                                                                                 
      fullName: 'uk.data_processing.split_data',                       
      type: 'task',                                                                                                                       
      // ...full node object from Redux state
    }                                                                                                                                     
  }                                                                    
  ```                                                                                                                                        
  The data-node-type and data-node-fullname DOM attributes are also available for consumers that need node metadata synchronously during the contextmenu event (e.g. VS Code's data-vscode-context which must be set before the browser reads it).
                                                                                                                                          
### Related                                                              

  - kedro-org/vscode-kedro#257
  
  ### Test
  There is no visible or functional changes to Kedro-Viz when used out side of VSCode extension.

### DIFF
--- a/src/store/middleware.js
+++ b/src/store/middleware.js
@@ -1,4 +1,4 @@
-import { TOGGLE_NODE_CLICKED } from '../actions/nodes';
+import { TOGGLE_NODE_CLICKED, NODE_CONTEXT_MENU } from '../actions/nodes';
 import { SHOW_PIPELINE_FILTER } from '../actions';
 
 /**
@@ -39,14 +39,21 @@ const createCallbackMiddleware =
         };
         callback(showPipelineFilterAction);
         break;
-      /** 
-      * Add additional cases here to handle other action types.
-      * Ensure on whatever action you want to trigger, It should be a Redux action. 
-      * And payload should be in Redux state.
-      * Example:
-      * const action = { type: SLICE_PIPELINE, payload: runCommand };
-        callback(action);
-      */
+      case NODE_CONTEXT_MENU:
+        if (action.nodeId) {
+          const state = store.getState();
+          const node = state?.graph?.nodes?.find(
+            (node) => node.id === action.nodeId
+          );
+          if (node) {
+            const contextMenuAction = {
+              type: NODE_CONTEXT_MENU,
+              payload: node,
+            };
+            callback(contextMenuAction);
+          }
+        }
+        break;
       default:
         break;
     }
--- a/src/store/middleware.js
+++ b/src/store/middleware.js
@@ -39,6 +39,9 @@ const createCallbackMiddleware =
         };
         callback(showPipelineFilterAction);
         break;
+      // Forwards node right-click events to the host app for embedding cases
+      // (e.g. Kedro VS Code extension) to show a custom context menu without
+      // DOM scraping.
       case NODE_CONTEXT_MENU:
         if (action.nodeId) {
           const state = store.getState();
--- a/src/actions/nodes.js
+++ b/src/actions/nodes.js
@@ -14,6 +14,19 @@ export function toggleNodeClicked(nodeClicked) {
   };
 }
 
+export const NODE_CONTEXT_MENU = 'NODE_CONTEXT_MENU';
+
+/**
+ * Dispatch when a node is right-clicked (context menu)
+ * @param {String} nodeId The node's unique identifier
+ */
+export function nodeContextMenu(nodeId) {
+  return {
+    type: NODE_CONTEXT_MENU,
+    nodeId,
+  };
+}
+
 export const TOGGLE_NODES_DISABLED = 'TOGGLE_NODES_DISABLED';
 
 /**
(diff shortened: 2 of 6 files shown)


jitu5:
> What is needed to get this reviewed and merged so we can merge the linked VSCode PR?

@rashidakanchwala @AliceCima10 Are we considering taking this PR and https://github.com/kedro-org/kedro-viz/pull/2621 (this is optional for now) to current sprint to unblock [One-click Debug Experience](https://github.com/kedro-org/vscode-kedro/pull/257)

rashidakanchwala (approved):
LGTM.

src/store/middleware.js
```diff
@@ -39,14 +39,21 @@ const createCallbackMiddleware =
...
         break;
-      /** 
-      * Add additional cases here to handle other action types.
-      * Ensure on whatever action you want to trigger, It should be a Redux action. 
-      * And payload should be in Redux state.
-      * Example:
-      * const action = { type: SLICE_PIPELINE, payload: runCommand };
-        callback(action);
-      */
+      case NODE_CONTEXT_MENU:
```
ravi-kumar-pilla, line 45: [nit] It would be nice to add some comments for future reference to understand this usecase is for vscode or embedding cases like the vscode plugin thing.

ravi-kumar-pilla (approved):
Code looks good to me. Left a nit comment. Thank you @jitu5

SajidAlamQB (approved):
Awesome work @jitu5!

PR #2618, opened 2026-04-04
Security: Potential DOM XSS via unsafe HTML insertion in row labels

tuanaiseo:
## Problem

The component renders `label` with `dangerouslySetInnerHTML` after passing it through `replaceAngleBracketMatches(label)`. If that helper does not fully HTML-escape untrusted input (including quotes, entities, and event-handler payloads), attacker-controlled labels can inject scriptable markup.

**Severity**: `high`
**File**: `src/components/ui/row-text/row-text.jsx`

## Solution

Avoid `dangerouslySetInnerHTML` for labels. Render as plain text (`{label}`) or sanitize with a strict, well-tested sanitizer immediately before insertion, with an allowlist that excludes scripts/events/unsafe URLs.

## Changes

- `src/components/ui/row-text/row-text.jsx` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced


src/components/ui/row-text/row-text.jsx
```diff
@@ -42,10 +41,9 @@ export const RowText = ({
             'row-text__label--disabled': disabled,
           }
         )}
-        dangerouslySetInnerHTML={{
-          __html: replaceAngleBracketMatches(label),
-        }}
-      />
+      >
+        {label}
```
jitu5, line 45: This fixes the security related issue but this breaks the current bold-highlight behavior.

Image

rashidakanchwala:
@tuanaiseo, please let us know if you would like to fix the issue above. If not, someone from the team will take over this PR and make the fix. Many thanks for your contribution.

rashidakanchwala:
closing this as we opened #2669

PR #2619, opened 2026-04-04
Security: Template variable injection into inline JavaScript (telemetry bootstrap)

tuanaiseo:
## Problem

Template variables (`heap_app_id`, `heap_user_identity`, `kedro_viz_version`) are embedded directly inside inline JavaScript string literals. If any value contains quotes or script-breaking characters and is not context-escaped by the template engine, this can lead to reflected/stored XSS.

**Severity**: `high`
**File**: `public/telemetry.html`

## Solution

Apply JavaScript-context escaping for all injected variables, or move dynamic data into `data-*` attributes/JSON script tags and parse safely. Enforce CSP (`script-src` with nonce/hash) to reduce impact.

## Changes

- `public/telemetry.html` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced


rashidakanchwala (approved):
LGTM, thanks a lot for the contribution!

jitu5 (approved):

PR #2621, opened 2026-04-07
Expose full internal Kedro node name as `kedro_node_name` in pipeline API response.

jitu5:
Related: [kedro-org/vscode-kedro#60](https://github.com/kedro-org/vscode-kedro/issues/60)     

## Description                                                                                                                          
                                                                                                                                                                                        
When a Kedro node is defined without an explicit `name=` parameter, Kedro internally generates a unique name with a hash suffix (e.g., `reporting.make_cancel_policy_bar_chart__7c6d9304`). Kedro-viz was only surfacing the bare function name (e.g., `make_cancel_policy_bar_chart`) as the display name, and the full internal name was only accessible via the `/api/nodes/<node_id>` endpoint — which is problematic because `node_id` is not a stable identifier.                                                           
                                                                                                                                          
This meant the VS Code Kedro extension (`vscode-kedro`) had no reliable way to retrieve the full internal node name from the main pipeline JSON, requiring either a separate unstable API call or loading the Kedro project independently (risking inconsistency with the viz graph).                                                                                                                                                                                                                               
                                                                                                                                          
  ### Solution                                                                                                                            
  Expose `node.name` (the full internal Kedro node name) as a new `kedro_node_name` field on `TaskNode` and `TaskNodeAPIResponse`. This field is now included in the response from `/api/pipelines/<id>` and `get_kedro_project_json_data()`.                                                                              
                                                                                                                                          
  ### Behaviour                                                                                                                           
                                                                                                                                          
  | Scenario | `name` (existing display name) | `kedro_node_name` (new) |                                                                 
  |---|---|---|                                             
  | With explicit `name="my_node"` | `"my_node"` | `"my_node"` |                                                                          
  | Without name | `"identity"` | `"identity__784efba9"` |  
  | Without name + namespace | `"identity"` | `"reporting.identity__784efba9"` |                                                          
  | With name + namespace | `"my_node"` | `"reporting.my_node"` |                                                                         
                                                                                                                                          
  ### Changes                                                                                                                             
  - `models/flowchart/nodes.py` — added `kedro_node_name` field to `TaskNode`; populated via `node.name` in `create_task_node()`
  - `api/rest/responses/pipelines.py` — added `kedro_node_name` field to `TaskNodeAPIResponse` including OpenAPI schema example           
  - `tests/test_models/test_flowchart/test_nodes.py` — asserts `kedro_node_name` in existing test; adds new test covering the             
  no-explicit-name case         
  
  
  ## How to Test                       
                                                                                                                                                                                                                                                          
  - This branch checked out: `vscode/api-new-entry`                    
  - Dependencies installed: `make install` / `pip install -e package`                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                            
### 1. Start the backend                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                     
  ```bash                                                                                                                                                                                                                                                                                                                            
  make build                                                           
  make run
  ```

  ### 2. Start the frontend (in another terminal)

  ```bash                                                                                                                                                                                                                                                                                                                            
  npm run start:dev
  ```                                                                                                                                                                                                                                                                                                                                
                                                                       
  ### 3. Inspect the `/api/main` response

  - Open the app in the browser.                                                                                                                                                                                                                                                                                                     
  - Open DevTools → **Network** tab → find the `main` request.
  - In the **Response** tab, locate the `nodes` array.                                                                                                                                                                                                                                                                               
  - For each **task** node, confirm a new `kedro_node_name` field is present and holds the full internal Kedro name (e.g. `reporting.make_cancel_policy_bar_chart` or `<func_name>__<8-char-hash>` when no explicit name was set).                                                                                                   
  - For **data** / **parameter** nodes, the field should be absent / `null`. 

                                                                       
  ### 4. Verify the static-file export path

  ```bash
  kedro viz build
  cat build/api/main | python -m json.tool | grep kedro_node_name
  ```                                                                                                                                                                                                                                                                                                                                
   
  **Expected:** same `kedro_node_name` values appear in the exported static JSON.

### DIFF
--- a/package/kedro_viz/api/rest/responses/pipelines.py
+++ b/package/kedro_viz/api/rest/responses/pipelines.py
@@ -54,9 +54,11 @@ class TaskNodeAPIResponse(BaseGraphNodeAPIResponse):
 
     Attributes:
         parameters (Dict): A dictionary containing the parameters for the task node.
+        kedro_node_name (Optional[str]): The full internal Kedro node name including namespace and hash suffix.
     """
 
     parameters: Dict
+    kedro_node_name: Optional[str] = None
     model_config = ConfigDict(
         json_schema_extra={
             "example": {
@@ -80,6 +82,7 @@ class TaskNodeAPIResponse(BaseGraphNodeAPIResponse):
                         "review_scores_rating",
                     ],
                 },
+                "kedro_node_name": "data_science.split_data_node",
             }
         }
     )
--- a/package/kedro_viz/api/rest/responses/pipelines.py
+++ b/package/kedro_viz/api/rest/responses/pipelines.py
@@ -54,16 +54,17 @@ class TaskNodeAPIResponse(BaseGraphNodeAPIResponse):
 
     Attributes:
         parameters (Dict): A dictionary containing the parameters for the task node.
-        kedro_node_name (Optional[str]): The full internal Kedro node name including namespace and hash suffix.
+        full_name (str): The full internal Kedro node name including namespace and hash suffix.
     """
 
     parameters: Dict
-    kedro_node_name: Optional[str] = None
+    full_name: str
     model_config = ConfigDict(
         json_schema_extra={
             "example": {
                 "id": "6ab908b8",
                 "name": "split_data_node",
+                "full_name": "data_science.split_data_node",
                 "tags": [],
                 "pipelines": ["__default__", "ds"],
                 "modular_pipelines": [],
@@ -82,7 +83,6 @@ class TaskNodeAPIResponse(BaseGraphNodeAPIResponse):
                         "review_scores_rating",
                     ],
                 },
-                "kedro_node_name": "data_science.split_data_node",
             }
         }
     )
--- a/package/kedro_viz/models/flowchart/nodes.py
+++ b/package/kedro_viz/models/flowchart/nodes.py
@@ -95,6 +95,7 @@ class GraphNode(BaseModel, ABC):
             kedro_obj=node,
             modular_pipelines=modular_pipelines,
             node_extras=node_extras,
+            kedro_node_name=node.name,
         )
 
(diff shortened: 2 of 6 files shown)


noklam (approved):

package/kedro_viz/api/rest/responses/pipelines.py
```diff
@@ -54,9 +54,11 @@ class TaskNodeAPIResponse(BaseGraphNodeAPIResponse):
 
     Attributes:
         parameters (Dict): A dictionary containing the parameters for the task node.
+        kedro_node_name (Optional[str]): The full internal Kedro node name including namespace and hash suffix.
```
rashidakanchwala: nit - should we call this full_name; earlier we used to have `full_name` and `name` that's why

also why is it optional? node.name is there for every node right?

package/kedro_viz/api/rest/responses/pipelines.py
```diff
@@ -80,6 +82,7 @@ class TaskNodeAPIResponse(BaseGraphNodeAPIResponse):
                         "review_scores_rating",
                     ],
                 },
+                "kedro_node_name": "data_science.split_data_node",
```
rashidakanchwala: can u move this on top under "name"?

rashidakanchwala (approved):
LGTM. few nit comments.

package/kedro_viz/api/rest/responses/pipelines.py
jitu5: Done!

package/kedro_viz/api/rest/responses/pipelines.py
jitu5: Done!

PR #2629, opened 2026-04-21
Use `kedro_session` directly to avoid `AbstractSession` typing error

SajidAlamQB:
## Description

We recently merged https://github.com/kedro-org/kedro/pull/5441 which introduced an `AbstractSession` base class and removed `KedroSession.__enter__`. `KedroSession` now inherits `__enter__` from `AbstractSession`, whose return type is `AbstractSession` which has no `load_context` method.

mypy infers `session: AbstractSession` and fails on `load_context`.

## Checklist

### DIFF
--- a/package/kedro_viz/integrations/kedro/data_loader.py
+++ b/package/kedro_viz/integrations/kedro/data_loader.py
@@ -187,12 +187,12 @@ def _load_data_helper(
         runtime_params=extra_params,
     )
 
-    with kedro_session as session:
+    with kedro_session:
         # check for --include-hooks option
         if not include_hooks:
-            session._hook_manager = _VizNullPluginManager()  # type: ignore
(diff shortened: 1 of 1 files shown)


jitu5 (approved):

ravi-kumar-pilla (approved):

ravi-kumar-pilla:
Thank you @SajidAlamQB for the quick fix 💯

ankatiyar:
Just checking if this fix is because of mypy only - lmk if we need to add `__enter__` back to session?

SajidAlamQB:
> Just checking if this fix is because of mypy only - lmk if we need to add `__enter__` back to session?

Yeah it's just a mypy typing issue, the behaviour is unchanged since `KedroSession` still inherits `__enter__` from AbstractSession and returns self. The `as session` just gets typed as AbstractSession now, which doesn't expose `load_context`.

No strong need to add `__enter__` back imo

PR #2664, opened 2026-05-27
Publish Kedro-Viz 12.4.0 release to Demo site

jitu5:
Publish Kedro-Viz 12.4.0 release to Demo site


ravi-kumar-pilla:
Hi @jitu5 , I see lot of files changed. Since the demo branch was updated for 12.2.0, I think there is some issue with the local copy. I expect fewer file changes

jitu5:
> Hi @jitu5 , I see lot of files changed. Since the demo branch was updated for 12.2.0, I think there is some issue with the local copy. I expect fewer file changes

@ravi-kumar-pilla I compare the first commit after `12.2.0` till last commit up to 12.4.0 here https://github.com/kedro-org/kedro-viz/compare/2a0f85bfbd47fdf88f5a70264d1a81573894f5d2...1be84e99893956f2cec2995a7174269ea35e3803

and the file diff is around `129` files. Also demo branch currently at v12.2.0 and this PR updates it to main/v12.4.0, so it includes all changes since v12.2.0. wdyt?

ravi-kumar-pilla (approved):

ankatiyar:
Should this not be automated, something like we do with `main` and `develop` on Kedro? https://github.com/kedro-org/kedro/blob/main/.github/workflows/sync.yml

jitu5:
> Should this not be automated, something like we do with `main` and `develop` on Kedro? https://github.com/kedro-org/kedro/blob/main/.github/workflows/sync.yml

@ankatiyar 

Yes, this should already be automated by `.github/workflows/auto-merge-demo.yml`, which merges `main` into `demo` after the `Release to NPM and PyPI` workflow succeeds.

In this case it likely didn’t run because the release workflow failed at the npm publish step due to the OTP error, so the auto-merge job was skipped. This is also the first release where we’re using that auto-merge-to-demo workflow since the last release, so I wasn’t aware it was already intended to handle this path. 

And as soon as I ran `.github/workflows/auto-merge-demo.yml` workflow manually, some how this PR merged automatically by Github Actions.

PR #2684, opened 2026-06-26
Fix node boxes shrinking when the flowchart loads while hidden

### LINKED ISSUE #2686: Flowchart: layer labels overlap and node text overflows on first load
## Description

  When the flowchart first loads, two things can render incorrectly. Both show up most often when Kedro-Viz is embedded in another page (for example an iframe, a Jupyter notebook cell, or a dashboard panel). The layer label problem can also appear on a normal page load.

  ### 1. Layer labels stack in the top-left corner

  The layer names (for example `raw`, `intermediate`, `primary`) pile up on top of each other in the top-left of the chart, instead of sitting next to the band they belong to. Resizing the window, or panning/zooming the chart, snaps them back to the correct place.

  ### 2. Node boxes shrink to the icon and the text spills out

  Each node's box is sized so small that it only fits the icon, and the full label overflows the box and overlaps the nodes around it. This happens when the chart is loaded while it isn't visible yet, for example inside a hidden tab or an iframe that becomes visible a moment later. Unlike the labels, resizing the window does not fix this one.

  ## How to reproduce

  1. Embed Kedro-Viz in a page where it starts hidden, or inside an iframe that is shown/sized slightly after load (a hidden tab, an accordion, or a Streamlit/Jupyter iframe all work).
  2. Open the page.
  3. On first load you see the layer labels stacked in the top-left, and/or node boxes that are too small with the text overflowing.

  ## What we expect

  On first load, layer labels should sit next to their bands, and node boxes should be sized to fit their full text, without needing a resize.

## Screenshots

Image

Image

## Checklist

jitu5:
Fixes https://github.com/kedro-org/kedro-viz/issues/2686

## The problem

When the flowchart is loaded while it isn't visible yet (for example inside a hidden tab, or an iframe that becomes visible shortly after), each node's box is sized to fit only the icon. The full label then overflows the box and overlaps the surrounding nodes. Resizing the window does not fix it.

## Why it happens

A node's box width is based on the measured width of its label text. The text is measured once, early on, by briefly drawing it and reading its size. That only works if the chart is actually on screen. If the chart is hidden, the measurement comes back as zero, the box gets sized for the icon only, and it is never measured again.

## What this PR does

When the chart is mounted while hidden, re-measure the label widths once it becomes visible, then re-run the graph layout with the correct sizes. A normal, already-visible load is left untouched (it measured correctly the first time), so its first render and animation are unchanged.

This applies to both the flowchart view and the workflow view.

## How to test

1. **Run Kedro-Viz locally** to get a URL to point at. From the repo root:
   ```bash
   make build
   make run
   npm run start (In separate terminal) Note the local URL it serves, e.g. `http://localhost:4141`.
   ```

2. **Save the test page.** Copy the HTML posted in this PR [comment](https://github.com/kedro-org/kedro-viz/pull/2684#issuecomment-4865124358) into a new file named `kedro-viz-test.html`.

3. **Open `kedro-viz-test.html` in a browser** (double-click it, or drag it into a tab).

4. **Paste your local URL** into the box at the top.

Screenshot 2026-07-02 at 12 27 51 p m
Screenshot 2026-07-02 at 12 27 41 p m


jitu5:
```
<!DOCTYPE html>
<!--
  Kedro-Viz first-render test

  Paste a running Kedro-Viz URL and use the two buttons to check that the
  flowchart renders correctly on first load:
    - "Load normally"          : the baseline (chart is visible from the start).
    - "Load hidden, then reveal": recreates the embedded/hidden case (a tab or
                                  iframe shown after load) that used to break the
                                  layer labels and node boxes.
  With the fix, both loads should look correct without resizing the window.
-->
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <title>Kedro-Viz first-render test</title>
    <style>
      :root {
        color-scheme: light;
        font-family: system-ui, -apple-system, 'Segoe UI', Roboto, sans-serif;
      }
      body { margin: 0; padding: 24px; max-width: 1040px; color: #1c1c1c; }
      h1 { font-size: 20px; margin: 0 0 4px; }
      p.lead { margin: 0 0 20px; color: #555; }
      .controls {
        display: flex; flex-wrap: wrap; gap: 8px;
        align-items: center; margin-bottom: 16px;
      }
      #url {
        flex: 1 1 320px; min-width: 240px; padding: 8px 10px;
        font-size: 14px; border: 1px solid #bbb; border-radius: 6px;
      }
      button {
        padding: 8px 14px; font-size: 14px; border: 1px solid #999;
        border-radius: 6px; background: #f4f4f4; cursor: pointer;
      }
      button:hover { background: #e9e9e9; }
      #status { color: #555; font-size: 13px; }
      .checklist {
        background: #f7f7f7; border: 1px solid #e2e2e2; border-radius: 8px;
        padding: 12px 16px; margin-bottom: 16px; font-size: 14px;
      }
      .checklist ul { margin: 8px 0 0; padding-left: 20px; }
      .checklist li { margin-bottom: 6px; }
      #stage {
        border: 1px solid #ccc; border-radius: 8px;
        overflow: hidden; background: #fff;
      }
      #stage iframe { width: 100%; height: 640px; border: 0; display: block; }
    </style>
  </head>
  <body>
    <h1>Kedro-Viz first-render test</h1>
    <p class="lead">
      Paste a running Kedro-Viz URL, then use the buttons to check the first
      render of the flowchart.
    </p>

    <div class="controls">
      <input id="url" type="text" value="http://localhost:4141" />
      <button id="normal">Load normally</button>
      <button id="hidden">Load hidden, then reveal</button>
      <span id="status"></span>
    </div>

    <div class="checklist">
      <strong>After each load, check the flowchart (without resizing the window):</strong>
      <ul>
        <li>
          Layer names (for example <code>raw</code>, <code>intermediate</code>,
          <code>primary</code>) line up down the left edge next to their bands.
          They should not be stacked on top of each other in the top-left
          corner.
        </li>
        <li>
          Each node box fits its full label. The text should not overflow a box
          that has shrunk to just the icon.
        </li>
      </ul>
      <p style="margin: 8px 0 0; color: #555">
        <b>Load normally</b> is the baseline. <b>Load hidden, then reveal</b>
        recreates the embedded/hidden case that used to break both. With the fix,
        both should look correct.
      </p>
    </div>

    <div id="stage"></div>

    <script>
      // How long to keep the chart hidden before revealing it, so it mounts and
      // measures while off-screen. Increase this if the chart loads slowly.
      const REVEAL_DELAY_MS = 1200;

      const urlInput = document.getElementById('url');
      const stage = document.getElementById('stage');
      const statusEl = document.getElementById('status');
      let runCount = 0;

      // Add a unique query param so each run is a fresh load, not a cached one.
      function withCacheBust(url) {
        const separator = url.includes('?') ? '&' : '?';
        return url + separator + 'kv_test=' + ++runCount;
      }

      function newIframe(url) {
        const iframe = document.createElement('iframe');
        iframe.title = 'Kedro-Viz';
        iframe.src = withCacheBust(url);
        return iframe;
      }

      function loadNormally() {
        const url = urlInput.value.trim();
        if (!url) return;
        stage.style.display = 'block';
        stage.innerHTML = '';
        stage.appendChild(newIframe(url));
        statusEl.textContent = 'Loaded normally.';
      }

      function loadHiddenThenReveal() {
        const url = urlInput.value.trim();
        if (!url) return;
        stage.innerHTML = '';
        stage.style.display = 'none'; // chart mounts and measures while hidden
        const iframe = newIframe(url);
        stage.appendChild(iframe);
        statusEl.textContent = 'Loading while hidden…';
        iframe.addEventListener('load', () => {
          setTimeout(() => {
            stage.style.display = 'block';
            statusEl.textContent = 'Revealed — now check the flowchart above.';
          }, REVEAL_DELAY_MS);
        });
      }

      document.getElementById('normal').addEventListener('click', loadNormally);
      document
        .getElementById('hidden')
        .addEventListener('click', loadHiddenThenReveal);
    </script>
  </body>
</html>
```

SajidAlamQB (approved):
Thank you @jitu5 lgtm!

ravi-kumar-pilla:
Hi @jitu5 ,

I can confirm it works pretty well. Great fix !!

A few minor thoughts (non-blocking):

1. The remeasure trigger is really the container `ResizeObserver` → `handleWindowResize`, not just `window.resize`. Worth updating the comment in `node-remeasure.js` so it's clear why this fires when a hidden iframe/tab is revealed.

2. On the no-`ResizeObserver` fallback path, a hidden-tab reveal may not trigger remeasure, worth a brief comment if that's intentional.

3. Small nit: `nodeMeasureToken` never resets on pipeline change, but `getPipelineNodeIDs` changing should bust the selector anyway - a comment in the reducer would help future readers.

Thank you

ravi-kumar-pilla (approved):
Left some suggestions but the fix works well. Thank you @jitu5

PR #2685, opened 2026-06-26
Fix layer labels overlapping in the top-left corner on first load.

jitu5:
## Description

On first load, the layer names (`raw`, `intermediate`, `primary`, and so on) stack on top of each other in the top-left of the chart instead of lining up next to their bands. Resizing the window or panning the chart puts them back in the right place.

  ## Why it happens

  The layer label elements are created after the chart paints, but their positions were only ever set during a view change (zoom, pan, or resize). On first load the positioning ran before the labels existed, so they never got placed and fell back to the default top-left spot. Anything that caused a view change
  afterwards (such as a resize) is what moved them into place.

  ## What this PR does

  Position the layer labels right after they are created, using the current view, instead of waiting for the next view change. They now land in the correct place on first paint, with no resize needed. Live updates on pan/zoom are unchanged.

  This applies to both the flowchart view and the workflow view.
  
  ## Gif
  
  ### Before 
  
viz-f-resize-11

  
  ### After
  
viz-f-resize-22

## How to test (via kedro-viz-standalone)

These changes are published to a test npm package: **`kedro-vi122@1.21.0-rc37`**. To try them in a real embedding app:

1. **Clone the standalone app:**
   ```bash
   git clone https://github.com/kedro-org/kedro-viz-standalone
   cd kedro-viz-standalone
   ```

2. **Point it at the test package** — in `package.json`:
   ```diff
   - "@quantumblack/kedro-viz": "^12.4.0",
   + "kedro-vi122": "1.21.0-rc37",
   ```

3. **Update the imports** in `app.js` (or `app.ts`):
   ```diff
   - import "@quantumblack/kedro-viz/lib/styles/styles.min.css";
   - import spaceflight from "@quantumblack/kedro-viz/lib/utils/data/spaceflights.mock.json";
   - import KedroViz from "@quantumblack/kedro-viz";
   + import "kedro-vi122/lib/styles/styles.min.css";
   + import demo from "kedro-vi122/lib/utils/data/demo.mock.json";
   + import KedroViz from "kedro-vi122";
   ```
   Also update where the data is passed to the component to match the renamed variable, e.g. `data={spaceflight}` → `data={demo}`.

4. **Install and run:**
   ```bash
   npm install
   npm start
   ```
   Open the local URL it prints (e.g. `http://localhost:3000`).

5. **Check the flowchart on first load** (no resizing):
   - Layer names (`raw`, `intermediate`, `primary`, …) line up down the left edge next to their bands — **not** stacked on top of each other in the top-left corner.
   - Node boxes fit their full labels (not shrunk to just the icon).

   Reload once or twice if needed — the layer-label case is a first-render race.

### DIFF
--- a/src/components/draw/utils/position-layer-names.js
+++ b/src/components/draw/utils/position-layer-names.js
@@ -1,0 +1,33 @@
+/**
+ * Position the layer-name labels next to their bands for the given view
+ * transform. Reads the label elements from `container` and offsets each one.
+ * @param {HTMLElement} container The layer-names list element, or null.
+ * @param {Array} layers The layers, each with x/y/width/height.
+ * @param {Object} transform The current view transform ({ k, x, y }), or null.
+ * @param {String} orientation 'vertical' or 'horizontal'.
+ */
+export const positionLayerNames = (
+  container,
+  layers,
+  transform,
+  orientation
+) => {
+  if (!container || !transform) {
+    return;
+  }
+  const { k: scale, x, y } = transform;
+  const layerNames = container.querySelectorAll('.pipeline-layer-name');
+  layers.forEach((layer, i) => {
+    const el = layerNames[i];
+    if (!el) {
+      return;
+    }
+    if (orientation === 'vertical') {
+      const updateY = y + (layer.y + (layer.height || 0) / 2) * scale;
+      el.style.transform = `translateY(${updateY}px)`;
+    } else {
+      const updateX = x + (layer.x + (layer.width || 0) / 2) * scale;
+      el.style.transform = `translateX(${updateX}px) translateX(-50%)`;
+    }
+  });
+};
--- a/src/components/workflow/workflow.jsx
+++ b/src/components/workflow/workflow.jsx
@@ -57,6 +57,7 @@ import {
   GraphSVG,
 } from '../draw';
 import { DURATION, MARGIN, MIN_SCALE, MAX_SCALE } from '../draw/utils/config';
+import { positionLayerNames } from '../draw/utils/position-layer-names';
 import { getNodeStatusKey } from './workflow-utils/getNodeStatusKey';
 
 import ExportModal from '../export-modal';
@@ -85,6 +86,8 @@ export class Workflow extends Component {
     this.wrapperRef = React.createRef();
     this.layersRef = React.createRef();
     this.layerNamesRef = React.createRef();
+    // Holds the latest view transform so the layer labels can self-position
+    this.viewTransformRef = React.createRef();
   }
 
   componentDidMount() {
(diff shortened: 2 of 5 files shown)


SajidAlamQB (approved):
@jitu5 the fix looks good I've tested and seems to be working, thank you!

ravi-kumar-pilla:
Hi @jitu5 ,

I could not see the issue. If possible, can you add a gif of before and after the fix ?

jitu5:
> Hi @jitu5 ,
> 
> I could not see the issue. If possible, can you add a gif of before and after the fix ?

@ravi-kumar-pilla  Done.

src/components/workflow/workflow.jsx
```diff
@@ -739,6 +756,7 @@ export class Workflow extends Component {
               chartSize={chartSize}
               orientation={orientation}
               layerNamesRef={this.layerNamesRef}
+              onLayersRendered={this.handleLayerNamesRendered}
```
ravi-kumar-pilla: Nice fix - the `onLayersRendered` callback works and is definitely better than the setTimeout retry.

One suggestion: you could avoid the callback prop drilling with a viewTransformRef instead.

**Idea:** keep a ref on the parent that always holds the latest zoom transform ({ k, x, y }), updated in onViewChange. Pass it into DrawLayerNames, and at the end of the D3 useEffect (after labels are created), read viewTransformRef.current and position them there.

```
// flowchart.jsx / workflow.jsx
onViewChange(transform) {
  this.viewTransformRef.current = transform;
  positionLayerNames(this.layerNamesRef.current, this.props.layers, transform, this.props.orientation);
  // ...
}
// draw-layer-names.jsx — end of useEffect, after D3 join
if (viewTransformRef?.current) {
  positionLayerNames(layerNamesRef.current, layers, viewTransformRef.current, orientation);
}
```

**Why this might be slightly cleaner:**

No onLayersRendered callback through DrawLayerNamesGroup
No callback in the useEffect dependency array
Child self-positions on mount; parent still handles zoom-time updates imperatively (no extra re-renders per pan/zoom)
Either way, worth extracting positionLayerNames to a shared util - it's duplicated identically in flowchart.jsx and workflow.jsx.

Not blocking - the current approach is correct and shippable. This would just reduce coupling a bit.

jitu5: I have implemented the suggestions, please review again whenever you get time.

src/components/draw/utils/position-layer-names.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Position the layer-name labels next to their bands for the given view
+ * transform. Reads the label elements from `container` and offsets each one.
+ * @param {HTMLElement} container The layer-names list element, or null.
+ * @param {Array} layers The layers, each with x/y/width/height.
+ * @param {Object} transform The current view transform ({ k, x, y }), or null.
+ * @param {String} orientation 'vertical' or 'horizontal'.
+ */
+export const positionLayerNames = (
```
ravi-kumar-pilla, line 9: Nice extraction to `position-layer-names.js`.  💯 

I do not see if this util is unit-tested. `draw-layer-names.test.jsx` doesn't pass `viewTransformRef`, so the positioning path isn't exercised. A small `position-layer-names.test.js` covering vertical/horizontal transforms and the null early-return would make this regression harder to reintroduce.

ravi-kumar-pilla (approved):
Hi @jitu5 , left a minor non-blocking comment to add a small unit test. Other than that, the fix looks good. Thank you

Issue #2655, opened 2026-05-26
Foundations: snapshot loader, ID scheme, parity harness


jitu5:
**Parent issue:** #2265

## Summary

Stand up the adapter package, a thin loader for Kedro's inspection snapshot, the new viz-side node-ID functions, and a harness that captures the current live-backend output as the "answer key" everything else is checked against.

> **Note (updated after the live backend was removed):** the harness captured the live-backend output as the parity oracle *while the adapter was being built*. The live backend has since been deleted (see the runtime-integration ticket #2659), so the captured baseline now serves as the adapter's own regression fixture and `capture_baseline.py` reads from the adapter rather than the live path. `snapshot_source.py` is also extended by later phases (it gains `load_parameters` for `--params` value resolution and lite-mode import stubs). The loader, ID scheme and fixture defined here are otherwise unchanged.

## Why this exists

Everything in tickets 2–7 either builds on or is verified against this work. It's the floor.

## What's in scope

- A new package at `kedro_viz/integrations/kedro/inspection/`.
- `load_snapshot()` and `is_inspection_available()` around `kedro.inspection.get_project_snapshot`.
- `task_node_id()` / `dataset_node_id()` in a shared module both the adapter and the run-status hook can import (the hook switches in ticket 6).
- A capture script that records `/api/main` and `/api/pipelines/{id}` for the demo project as a golden baseline.
- A pytest fixture that restores Kedro's global state after each test module so bootstrapping the demo doesn't leak into other tests.

## Decisions locked in

- Adapter location: `package/kedro_viz/integrations/kedro/inspection/`.
- Minimum Kedro = **1.4.0**. The runtime adapter path will be auto-disabled below that in Phase 6 wiring; today, `load_snapshot()` raises if the inspection API is unavailable.
- Node IDs are a hash of `[name, inputs, outputs]` (JSON-encoded). Tags are deliberately excluded so re-tagging a node never changes its ID.
- ID functions live at `kedro_viz/integrations/kedro/node_ids.py` so the run-status hook can share the same implementation in ticket 6.

## Trade-offs taken

- ID reconstruction from the snapshot is impossible for nodes with an explicit `name=` that differs from the function name (the snapshot doesn't carry the function name). Rather than ask Kedro for a viz-specific field, we chose a new ID scheme and accepted the one-time break.

## Files

```
package/kedro_viz/integrations/kedro/
├── node_ids.py                       (task_node_id, dataset_node_id)
└── inspection/
[...]

Issue #2265, opened 2025-02-03
[Spike] Modularity of Kedro Viz


astrojuanlu:
> Hi @ravi-kumar-pilla, @astrojuanlu 
> 
> This is amazing spike work @ravi-kumar-pilla —it has given us a lot to think about. I really appreciate the effort you’ve put into it
> 
> We need to evaluate maintainability, effort vs. impact, and alignment with future Kedro-Viz developments. Given that the second half of 2025 will focus on the Pipeline Editor, which will be a major architectural shift, I think we should take a step back and plan PS sessions to align on the best direction for Kedro-Viz.
> 
> Next Steps, can we do a PS session:
> -  Original problem - Kedro-Viz in Notebooks
>    - Reviewing the user feedback we’ve received so far on the high fidelity prototypes 
>    - Understand what we can do to release an MVP soon 
> - UMD Bundling
>    - What additional benefits would UMD bundling bring beyond notebook integration? cons to this? 
> 
> and also another session on :- 
> - Broader Kedro-Viz Architecture
>   - Flowchart Rendering as a separate Library
>   -   Separate the logic of Kedro project --> structure json pipeline in another package (https://github.com/kedro-org/kedro/issues/4363)
>      - Pros/cons of the above two? 
>      - Would this help future-proof Kedro-Viz, especially for the Pipeline Editor?

_Originally posted by @rashidakanchwala in https://github.com/kedro-org/kedro-viz/issues/2256#issuecomment-2630374896_

astrojuanlu:
Also https://github.com/kedro-org/kedro-viz/issues/1857

ravi-kumar-pilla:
# Kedro-Viz Modularisation Proposal (Part 1 - Problem and possible solutions)

## What this is about

Three related issues point to the same underlying problem: Kedro-Viz is doing too much in one package.

| Issue | Problem |
|---|---|
| [#2265](https://github.com/kedro-org/kedro-viz/issues/2265) | Can't use Kedro-Viz in notebooks without installing a full web server |
| [#1857](https://github.com/kedro-org/kedro-viz/issues/1857) | Can't export pipeline JSON without `kedro viz run` |
| [kedro#4363](https://github.com/kedro-org/kedro/issues/4363) | Every deployment plugin re-implements pipeline inspection |

The [kedro-inspect proof-of-concept](https://github.com/AlpAribal/kedro-inspect) shows what's missing: a way to extract pipeline structure without pulling in the entire viz toolchain.

[...]

PR #2689, opened 2026-06-29
Foundations: snapshot loader, node-ID scheme, parity harness

jitu5:
**Sub-task:** #2655 · **Parent:** #2265

First slice of the inspection-adapter backend, split out of the big work-in-progress branch into a reviewable phase-PR. It targets the **`feat/backend_v2`** integration branch (not `main`); later phases stack on top.

## What's in this PR (the floor)
Purely **additive** on top of `feat/backend_v2`: no existing behaviour changes, and the live backend is untouched.

- **`integrations/kedro/inspection/snapshot_source.py`**: `_InspectionSession`, a small per-request class over `kedro.inspection.get_project_snapshot` plus the project's Kedro config loader (bootstrapped and built once, then cached). It exposes `snapshot()`, `catalog_config()`, and `parameters()` (parameter values with `--params` overrides already merged by the loader). Also `lite_import_stubs`, a context manager that mocks missing project imports for kedro-viz `--lite` mode (used by later phases).
- **`integrations/kedro/inspection/__init__.py`**: package marker (module docstring only). The session class and the node-ID helpers are internal (underscore-prefixed), so nothing is re-exported; the graph builder, modular pipelines and layers arrive in later phases.
- **`integrations/kedro/node_ids.py`**: `_create_task_node_id()` / `_create_dataset_node_id()`, the shared new-scheme IDs that both the adapter graph and the run-status hook use. New scheme = hash of `[name, inputs, outputs]` (tags excluded). **Breaking ID change**, taken deliberately (see #2655).
- **`tests/test_inspection_adapter/`**: `conftest.py` (restores Kedro global state per module so bootstrapping the demo doesn't leak), `capture_baseline.py` + `baseline/` (golden answer key captured from the live backend, used by later phases for parity), `test_snapshot_source.py`, `test_ids.py`.

## Notes
- Requires **`kedro>=1.4.0`** for the inspection API. This is now a hard floor in `requirements.txt` (pyproject uses `hatch-requirements-txt`, so that IS the package floor). The earlier `is_inspection_available()` runtime guard was removed, since the floor guarantees the API is present.
- `capture_baseline.py` is shipped in its live-capturing form (its Phase-0 purpose: record the live backend output as the parity oracle).

## Verification
In a `kedro 1.4.0` env (conda `viz-3-14`):
- `ruff` and `ruff format`: clean
- `mypy`: clean (prod + tests, config `package/mypy.ini`)
- `pytest`: the inspection-adapter suite this PR adds passes (**20 passed**). The rest of the unit suite is unaffected (this PR is additive), and the unit legs are green in CI. A few launcher, deployer and static-asset tests fail only in a local sandbox (they need free ports and a built frontend); they are pre-existing and not related to this change.

## Review note
`baseline/*.json` are generated fixtures (produced by `capture_baseline.py`) so please don't line-review them; skim/collapse them. Focus the review on the code: `snapshot_source.py`, `node_ids.py`, `capture_baseline.py`, `test_snapshot_source.py`, `test_ids.py`, and `__init__.py`.


package/kedro_viz/integrations/kedro/inspection/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""Kedro inspection snapshot adapter for Kedro-Viz.
+
+Converts a Kedro project inspection snapshot (``kedro>=1.4.0``) into a graph response.
+"""
+
+from kedro_viz.integrations.kedro.inspection.snapshot_source import (
+    is_inspection_available,
```
ravi-kumar-pilla: lets make this a breaking release supporting the one which does have inspection. So this check is not required or `is_inspection_available` is not needed

package/kedro_viz/integrations/kedro/inspection/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    load_snapshot,
+)
+from kedro_viz.integrations.kedro.node_ids import dataset_node_id, task_node_id
+
+__all__ = [
+    "dataset_node_id",
+    "is_inspection_available",
+    "load_snapshot",
+    "task_node_id",
+]
```
ravi-kumar-pilla: Do we expect these to be used somewhere other than via the API calls ?

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -0,0 +1,175 @@
...
+
+@contextmanager
+def lite_import_stubs(
+    project_path: str | Path, package_name: str | None = None
+) -> Iterator[None]:
+    """Temporarily mock missing project imports in ``sys.modules``.
+
+    ``get_project_snapshot`` imports pipeline modules, which can import optional node-function
+    dependencies that are absent in lite mode. This context uses ``LiteParser`` to register mock
+    modules only while the snapshot is being built.
```
ravi-kumar-pilla: This is a bit confusing to me. Lets remove this docstring. We can keep - Temporarily mock missing project imports for kedro-viz lite mode

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -0,0 +1,175 @@
...
+            "mocked. Install them for full functionality:\n%s",
+            len(modules_to_mock),
+            sorted(modules_to_mock),
+        )
+
+    with patch.dict("sys.modules", sys_modules_patch):
+        yield
+
+
+def is_inspection_available() -> bool:
```
ravi-kumar-pilla: I do not think we need to be backward compat

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -0,0 +1,175 @@
...
+    except ImportError as exc:
+        raise RuntimeError(
+            "Kedro inspection API is unavailable; the inspection adapter path "
+            "requires kedro>=1.4.0."
+        ) from exc
+
+    return get_project_snapshot(project_path=Path(project_path), env=env)
+
+
+def _config_loader(
```
ravi-kumar-pilla:
```suggestion
def _create_config_loader(
```

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -0,0 +1,175 @@
...
+        env: Optional Kedro environment override; ``None`` uses the project default.
+        runtime_params: Parsed ``--params`` overrides for ``${runtime_params:...}`` templating.
+
+    Returns:
+        The raw catalog config, or an empty dict if the project has none.
+    """
+    from kedro.config import MissingConfigException
+
+    try:
+        return _config_loader(project_path, env, runtime_params)["catalog"]
```
ravi-kumar-pilla:
```suggestion
        return _create_config_loader(project_path, env, runtime_params)["catalog"]
```

package/kedro_viz/integrations/kedro/node_ids.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import annotations
+
+import json
+from collections.abc import Sequence
+from typing import Any
+
+from kedro_viz.utils import _hash, _hash_input_output
+
+
+def dataset_node_id(dataset_name: str) -> str:
```
ravi-kumar-pilla:
```suggestion
def _create_dataset_node_id(dataset_name: str) -> str:
```

package/kedro_viz/integrations/kedro/node_ids.py
```diff
@@ -0,0 +1,44 @@
...
+    """Return the Viz graph ID for a data or parameter node.
+
+    Transcoded names (``name@suffix``) hash on the base name.
+    """
+    return _hash_input_output(dataset_name)
+
+
+def task_node_id(node_name: str, inputs: Sequence[str], outputs: Sequence[str]) -> str:
```
ravi-kumar-pilla:
```suggestion
def _create_task_node_id(node_name: str, inputs: Sequence[str], outputs: Sequence[str]) -> str:
```

package/kedro_viz/integrations/kedro/node_ids.py
```diff
@@ -0,0 +1,44 @@
...
+        outputs: Node output names, in declaration order.
+
+    Returns:
+        The task node's Viz graph ID.
+    """
+    signature = json.dumps([node_name, list(inputs), list(outputs)])
+    return _hash(signature)
+
+
+def task_node_id_for(node: Any) -> str:
```
ravi-kumar-pilla: can we remove this and use only the above 2 functions to extract node ids for datasets and task nodes ?

package/tests/test_inspection_adapter/capture_baseline.py
```diff
@@ -0,0 +1,191 @@
...
+Runs the live-object backend against ``demo-project`` and writes normalized JSON for ``/api/main``
+and every ``/api/pipelines/{id}``, plus a per task-node ID report.
+
+Run in the ``viz-3-14`` env (Python 3.14, kedro 1.4.0):
+
+    conda run -n viz-3-14 python package/tests/test_inspection_adapter/capture_baseline.py
+
+The generated files are committed under ``baseline/`` and used by inspection adapter tests.
+"""
+
```
ravi-kumar-pilla, line 12: The json files above i.e., default.json to main.json are for testing and temporary I believe, including this capture_baseline.py ?

Can we add a TODO here or in each ticket or parent that these will be removed before merging to main ?

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Generated parity-baseline fixtures (produced by
+# package/tests/test_inspection_adapter/capture_baseline.py from the live backend). Mark as
+# generated so GitHub collapses them in diffs and excludes them from language stats — review the
+# generator script and the tests, not the JSON.
+package/tests/test_inspection_adapter/baseline/** linguist-generated=true
```
ravi-kumar-pilla, line 5: what is this for ?

package/tests/test_inspection_adapter/test_snapshot_source.py
```diff
@@ -0,0 +1,110 @@
...
+
+@pytest.mark.skipif(
+    not snapshot_source.is_inspection_available(),
+    reason="kedro inspection API unavailable (requires kedro>=1.4.0)",
+)
+def test_load_snapshot_returns_demo_pipelines() -> None:
+    """Test that loading the demo project's snapshot yields the ``__default__`` pipeline."""
+    snapshot = snapshot_source.load_snapshot(DEMO_PROJECT)
+    pipeline_names = {pipeline.name for pipeline in snapshot.pipelines}
+    assert "__default__" in pipeline_names
```
ravi-kumar-pilla, line 63: can we add tests to check if the snapshot returned has the fields we need in kedro-viz ?

ravi-kumar-pilla:
Hi @jitu5 ,

Most of the changes look good to me. I left some minor comments. I will re-review and test it tomorrow. In the meantime, I feel it is better to involve @SajidAlamQB  or anyone interested. It would be good to have another reviewer apart from us.

Thank you

package/kedro_viz/integrations/kedro/inspection/__init__.py
jitu5: `load_snapshot` is used by the provider/server; the node-id helpers are imported directly from `node_ids` by their consumers (the graph builder and the run-status hook in later phases). I've removed the node-id re-exports from `__init__`

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: Done, simplified.

package/kedro_viz/integrations/kedro/node_ids.py
jitu5: Done (b7e0e6c7). It was an unused wrapper here; I'll re-add it in the phase that needs it (the metadata bridge).

.gitattributes
jitu5: It marks the generated baseline JSON as generated, so GitHub collapses those ~9k lines in the diff, the intent is that reviewers look at the generator + tests, not JSON.

package/tests/test_inspection_adapter/test_snapshot_source.py
jitu5: Added.

package/kedro_viz/integrations/kedro/node_ids.py
jitu5: These are deliberately shared across modules, the graph builder and the run-status hook both import `task_node_id`/`dataset_node_id` so the graph and run-status use one identical ID scheme (that's why they live in their own node_ids module).

package/kedro_viz/integrations/kedro/node_ids.py
jitu5: These are deliberately shared across modules, the graph builder and the run-status hook both import `task_node_id`/`dataset_node_id` so the graph and run-status use one identical ID scheme (that's why they live in their own node_ids module). I've instead removed them from the package `__init__` re-exports so they're no longer public surface.

package/tests/test_inspection_adapter/capture_baseline.py
jitu5: TODO added

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: Removed.

package/kedro_viz/integrations/kedro/inspection/__init__.py
jitu5: removed `is_inspection_available()` and Bumped `kedro>=1.4.0` in requirements.txt

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -0,0 +1,152 @@
...
+        env: Optional Kedro environment override; ``None`` uses the project default.
+        runtime_params: Parsed ``--params`` overrides, merged on top of the base values.
+
+    Returns:
+        The resolved parameter values.
+    """
+    from kedro.config import MissingConfigException
+
+    try:
+        params = _create_config_loader(project_path, env, runtime_params)["parameters"]
```
ravi-kumar-pilla: `_create_config_loader` is called here and also in `load_catalog_config`. Creating multiple times is not required. May be encapsulate this entire file within a class so we can create reusable fields across functions ?

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -0,0 +1,152 @@
...
+    bootstrap_project(project_path)
+    return settings.CONFIG_LOADER_CLASS(
+        conf_source=str(project_path / settings.CONF_SOURCE),
+        env=env,
+        runtime_params=runtime_params or {},
+        **settings.CONFIG_LOADER_ARGS,
+    )
+
+
+def load_catalog_config(
```
ravi-kumar-pilla: `load_catalog_config` and `_create_config_loader` are doing almost the same thing. May be we can combine them ?

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
ravi-kumar-pilla:
```python
class InspectionSession:
    """One bootstrap + one config loader per Viz request."""

    project_path: Path
    env: str | None = None
    runtime_params: dict[str, Any] | None = None
    _metadata: ProjectMetadata | None = field(default=None, init=False, repr=False)
    _config_loader: AbstractConfigLoader | None = field(default=None, init=False, repr=False)

    @property
    def metadata(self) -> ProjectMetadata:
        if self._metadata is None:
            self._metadata = bootstrap_project(self.project_path)
        return self._metadata

    @property
    def config_loader(self) -> AbstractConfigLoader:
        if self._config_loader is None:
            self._config_loader = settings.CONFIG_LOADER_CLASS(
                conf_source=str(self.project_path / settings.CONF_SOURCE),
                env=self.env,
                runtime_params=self.runtime_params or {},
                **settings.CONFIG_LOADER_ARGS,
            )
        return self._config_loader

    def snapshot(self) -> ProjectSnapshot:
        return get_project_snapshot(env=self.env, metadata=self.metadata)

    def catalog_config(self) -> dict[str, Any]: ...
    def parameters(self) -> dict[str, Any]: ...
```

package/kedro_viz/integrations/kedro/node_ids.py
```diff
@@ -0,0 +1,35 @@
...
+
+def dataset_node_id(dataset_name: str) -> str:
+    """Return the Viz graph ID for a data or parameter node.
+
+    Transcoded names (``name@suffix``) hash on the base name.
+    """
+    return _hash_input_output(dataset_name)
+
+
+def task_node_id(node_name: str, inputs: Sequence[str], outputs: Sequence[str]) -> str:
```
ravi-kumar-pilla: can we rename this and make it private `_create_task_node_id`, since its used only within kedro-viz ? Similar to how kedro does ?....same goes to `dataset_node_id`

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: Done. Added InspectionSession, which bootstraps and builds the config loader once and reuses it for both catalog and parameters, so we no longer rebuild it.

package/kedro_viz/integrations/kedro/node_ids.py
jitu5: Renamed to _create_task_node_id and _create_dataset_node_id. Agreed they are internal to kedro-viz.

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: Done

package/kedro_viz/integrations/kedro/node_ids.py
```diff
@@ -0,0 +1,37 @@
...
+
+def _create_dataset_node_id(dataset_name: str) -> str:
+    """Return the Viz graph ID for a data or parameter node.
+
+    Transcoded names (``name@suffix``) hash on the base name.
+    """
+    return _hash_input_output(dataset_name)
+
+
+def _create_task_node_id(
```
ankatiyar, line 19: This is what I mentioned - https://github.com/kedro-org/kedro/blob/59c7285962cb9b88ee0227dd20370dd9010a8a11/kedro/pipeline/node.py#L333

The node should already have a unique ID on Kedro side - could we somehow use this and return it in the project snapshot from Kedro side instead of creating a new hash here?

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -0,0 +1,115 @@
...
+        yield
+
+
+class InspectionSession:
+    """Read a project's snapshot and config, bootstrapping and building the loader once.
+
+    Create one session per request. The project is bootstrapped and the Kedro config loader is built
+    lazily on first use and then cached, so the catalog config and parameters reuse a single loader
+    instead of rebuilding it.
+    """
```
ankatiyar: Could we make this a private class for now? I would like to work together and see if we can make `KedroServiceSession` fit for this purpose if possible

package/kedro_viz/integrations/kedro/node_ids.py
jitu5: Thanks @ankatiyar, this connects directly to what we discussed in the tech design.

You're right that we're hashing `name + inputs + outputs` here. The background: the old Viz ID is `hash(str(node))`, and `str(node)` is made of four things, node name, function name, inputs and outputs. The snapshot already carries three of them; only the function name is missing. That one missing field is the whole reason we had to create a new hash and accept a breaking ID change.

So if the snapshot exposes the function name (e.g. a `func_name` field on `NodeSnapshot`), we can rebuild `str(node)` exactly and keep the old IDs unchanged and this becomes non-breaking. 

On reusing the generated name from `_set_unique_name()`: it would almost work, because the function name is embedded in it `({func}__{hash})`. But we'd have to parse the function name back out of that string. So exposing `func_name` directly (or even the ready-made `str(node)` value) would be simpler and serves the same purpose.

The only cost on our side is waiting for the next Kedro release. I'll discuss feasibility with Ravi tomorrow and get back to you on the Kedro-side ask.

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: Sure.

tools/test-lib/react-app/jest.setup.js
```diff
@@ -4,4 +4,15 @@ if (typeof window.URL.createObjectURL === 'undefined') {
   window.URL.createObjectURL = () => {};
 }
 
-global.fetch = require('node-fetch');
\ No newline at end of file
+global.fetch = require('node-fetch');
```
ravi-kumar-pilla, line 7: these are not needed ? or may be we need to update the base branch feat/backend_v2 with main before merging

ravi-kumar-pilla (approved):
LGTM and I think e2e tests are expected to fail for now ? Also please make a note that the json files for testing will be deleted before we merge this to main branch. Thank you @jitu5

ankatiyar (approved):

PR #2697, opened 2026-07-09
Graph builder: main graph from the inspection snapshot

jitu5:
**Sub-task:** #2656 · **Parent:** #2265

Second slice of the inspection-adapter backend: the main graph. Stacks on top of the merged Foundation PR (#2689) and targets the **`feat/backend_v2`** integration branch, not `main`.

## Description

Purely **additive**: this PR adds the snapshot graph builder and its parity tests. There is no runtime wiring yet; the builder is validated against the captured legacy-backend baseline but does not yet serve `/api/main`.

- **`integrations/kedro/inspection/graph_builder.py`**: adds `GraphBuilder`, which turns a `ProjectSnapshot` into the main graph response: task, data, and parameter nodes; edges; tag and pipeline lists; and per-node registered-pipeline membership and tags. Membership and tags are aggregated across all registered pipelines, matching the legacy backend. Node IDs use the foundation helpers `_create_task_node_id` and `_create_dataset_node_id`.
- **`integrations/kedro/inspection/__init__.py`**: re-exports `GraphBuilder`.
- **`tests/test_inspection_adapter/test_graph_builder.py`**: validates structural parity against the captured baseline across every registered pipeline view, including node sets by name and type, edge connectivity, top-level tags and pipelines, task display names, and per-node tags and pipeline membership. Modular-pipeline nodes and edges are excluded because they are implemented in a later slice.
- **`tests/test_inspection_adapter/test_graph_builder_edge_cases.py`**: provides focused hermetic coverage for task identity, global tag aggregation, pipeline selection and ordering, source and sink nodes, transcoding, unregistered in-memory datasets, and empty catalog types.

## Deferred to later phases

The required response fields remain valid using empty placeholders:

- **Modular pipelines** (phase 3): `modular_pipelines={}` on the response and `modular_pipelines=None` on nodes.
- **Layers** (phase 4): `layers=[]` on the response and `layer=None` on datasets.
- **Resolved parameter values and node metadata** (phase 5/6): `parameters={}` on task nodes.
- **Runtime wiring**: the builder is not yet routed to serve `/api/main`; that is handled in a later phase.

## Notes

- Parity is compared **structurally, not by literal ID**. Node IDs deliberately changed in D9, so tests compare node names, types, and connectivity rather than hashes.
[...]

Issue #2658, opened 2026-05-26
Build Layers info from the Snapshot 


jitu5:
**Parent issue:** #2265
**Depends on:** 01, 02

## Summary

Read `metadata.kedro-viz.layer` from the catalog config (the snapshot drops it), set each data node's `layer`, and produce the sorted `layers` list by reusing the existing `sort_layers` service.

## Why this exists

Layers are the horizontal bands the frontend draws (`raw → intermediate → primary → …`). The snapshot does not expose the layer because it is a viz concept (it lives under the `metadata.kedro-viz.*` namespace), not Kedro's to interpret.

## Decisions locked in

- Layer reading stays **viz-side**. Kedro doesn't interpret it; we won't ask them to.
- The live backend has a quirk where the `layers` *list* shows every project layer in every pipeline view (only the ordering is per-view). We reproduce that by seeding the sort with all layered datasets project-wide.

## Trade-offs taken

- The snapshot does not expose the viz metadata, so the adapter reads the catalog config itself. Living with it. The only conceivable Kedro ask would be a *generic* `metadata` dict on `DatasetSnapshot`, not a viz-specific `layer`.

## Files

```
package/kedro_viz/integrations/kedro/inspection/
├── layers.py                         (extract_layers)
└── snapshot_source.py                (load_catalog_config added)

package/tests/test_inspection_adapter/
├── test_layers.py                    (hermetic extractor tests)
└── test_graph_builder.py             (per-node layer + layers-list parity)
```

## Acceptance

- Per-node `layer` matches the baseline by dataset name.
- The `layers` list matches the baseline for the default pipeline **and** for at least one sub-view (we use `reporting_stage`, which exercises the global-presence + per-view ordering quirk).

Issue #2657, opened 2026-05-26
Modular pipelines from the snapshot


jitu5:
**Parent issue:** #2265 
**Built on:** #2655, #2656, #2658, #2715

## Decisions locked in

- A task belongs only to its own (deepest) namespace, not to every ancestor namespace.
- A non-parameter dataset belongs to a modular pipeline when it is used by one of that pipeline's direct tasks or is a boundary input/output of the whole namespace subtree. Parameter nodes have no modular-pipeline membership.
- A produced dataset that is consumed both inside and outside a namespace is still a boundary output.
- Boundary calculations and transcoded dataset names must match Kedro's `Pipeline` behaviour before the names are converted to shared Viz dataset IDs.
- Global membership must preserve distinct task snapshots; their I/O must not be discarded by deduplicating tasks only by `node.name` across registered pipelines.
- Modular-pipeline tags are the union of tags in the namespace subtree.
- Modular edges and cycle removal match the legacy backend.

## Trade-off

The legacy root tree contains one known duplicate parameter child with the same ID represented once as `data` and once as `parameters`. The adapter emits that child once with the correct `parameters` type. Tests should canonicalise only that known duplicate and continue comparing every other root child by both ID and type.

## Files

```text
package/kedro_viz/integrations/kedro/inspection/
└── modular_pipelines.py

package/tests/test_inspection_adapter/
├── test_graph_builder.py
└── test_modular_pipelines.py
```

## Acceptance criteria

- Modular nodes, membership, tree inputs/outputs/children, and modular edges match the captured baseline by exact ID for all demo pipelines.
- Modular-node tags and registered-pipeline membership match the baseline.
- A dataset produced and consumed inside a namespace, then consumed outside it, remains a boundary output and has the correct membership.
- Transcoded boundary datasets match the legacy `Pipeline` behaviour.
- Two registered pipelines may contain same-named but distinct task snapshots without either task's I/O being dropped from global dataset-membership calculations.
- Membership parity comparisons key nodes by type and ID, so a same-named task and dataset cannot hide one another.
- Root children retain their correct types after canonicalising only the known legacy duplicate.
- Focused tests cover nested namespaces, parameters, source/sink nodes, modular-edge cycle removal, deterministic edge serialisation, and pipelines with no namespaces.
- A real namespace named `__root__` cannot silently collide with the synthetic root entry.

PR #2700, opened 2026-07-16
Layers: dataset layers from the inspection snapshot

jitu5:
**Sub-task:** #2658 · **Parent:** #2265

Third slice of the inspection-adapter backend: dataset layers. This stacks on the merged graph-builder PR (#2697) and foundation PR (#2689), and targets **`feat/backend_v2`**, not `main`.

## Description

This PR adds dataset-layer support to the snapshot graph builder. It is additive: nothing is wired to `/api/main` yet, and parity is checked against the captured legacy-backend baseline.

What changed:

- **`layers.py` (new):** extracts `metadata.kedro-viz.layer` from raw catalog configuration, resolves dataset-factory patterns, normalises transcoded names, and rejects conflicting layers for transcoded variants. It also contains an inspection-native sorter that works with a node-ID-to-layer mapping and preserves the legacy ordering and cycle behaviour.
- **`graph_builder.py`:** `GraphBuilder` accepts optional raw catalog configuration and resolves its layer mapping internally. Data nodes receive their layer, and each graph response contains the project-wide set of used layers ordered by the selected pipeline's dependencies. Parameter references and unused catalog entries do not create layers.
- **Tests:** add catalog extraction and sorting unit tests, layer parity for every registered pipeline, and focused coverage for factory patterns, parameters, transcoding, absent configuration, cross-pipeline layers, and unused catalog entries.

## Deferred to later phases

The required response fields remain valid using empty placeholders:

- **Modular pipelines** (#2657): `modular_pipelines={}` on the response and `modular_pipelines=None` on nodes.
- **Resolved parameter values and node metadata** (later slice): `parameters={}` on task nodes.
- **Runtime wiring:** the builder is not yet routed to serve `/api/main`; that is handled in a later phase.

## Notes

- Parity is structural rather than based on literal IDs because node IDs changed in the foundation work. Layer parity is compared by dataset name.
- The `layers` list is order-sensitive and is asserted against every pipeline baseline.
- Layer extraction and sorting remain private inspection-adapter details and are not re-exported from the package.
- The inspection-native sorter replaces the temporary dependency on the legacy `GraphNode`-based layer service. The legacy implementation can be removed with the old backend.
- `dataset_type` is unchanged and continues to use the raw catalog string from the snapshot.
- `RELEASE.md` is intentionally deferred until the `backend_v2` integration line merges to `main`, consistent with #2689 and #2697.

## Follow-up when the legacy backend is removed

- Move the dictionary-based `sort_layers` from `integrations/kedro/inspection/layers.py` to `services/layers.py`.
- Remove the old `GraphNode`-based `sort_layers`.
- Remove `CatalogRepository.layers_mapping` and its related legacy layer helpers and tests.

## Verification

Validated with Python 3.14.4 and Kedro 1.4.0:

- `ruff check`: clean
- `mypy --config-file package/mypy.ini`: clean for package and tests
- `pytest package/tests/test_inspection_adapter/`: **117 passed**
- `graph_builder.py` and `layers.py`: **100% line coverage**

## Checklist

- [ ] Opened this PR as a draft if it is work in progress
- [ ] Updated user-facing documentation (not applicable: internal adapter only)
- [ ] Added a `RELEASE.md` entry (deferred; see Notes)
- [x] Added tests for the changes

### DIFF
--- a/package/kedro_viz/integrations/kedro/inspection/graph_builder.py
+++ b/package/kedro_viz/integrations/kedro/inspection/graph_builder.py
@@ -9,8 +9,7 @@ from __future__ import annotations
 
 import re
 from collections import defaultdict
-from dataclasses import dataclass
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Any
 
 from kedro_viz.api.rest.responses.pipelines import (
     DataNodeAPIResponse,
@@ -20,12 +19,15 @@ from kedro_viz.api.rest.responses.pipelines import (
     TaskNodeAPIResponse,
 )
 from kedro_viz.constants import DEFAULT_REGISTERED_PIPELINE_ID
+from kedro_viz.integrations.kedro.inspection.layers import (
+    _extract_layers,
+    sort_layers,
+)
 from kedro_viz.integrations.kedro.node_ids import (
     _create_dataset_node_id,
     _create_task_node_id,
 )
 from kedro_viz.models.flowchart.model_utils import GraphNodeType
-from kedro_viz.services.layers import sort_layers
 from kedro_viz.utils import _strip_transcoding, is_dataset_param
 
 if TYPE_CHECKING:
@@ -35,20 +37,11 @@ if TYPE_CHECKING:
         ProjectSnapshot,
     )
 
-    from kedro_viz.models.flowchart.nodes import GraphNode
-
 _AUTO_NAME_RE = re.compile(r"^(?P<func>.+)__[0-9a-f]{8}$")
 
 # Match the live backend's get_dataset_type(MemoryDataset()) output for datasets
 # absent from the snapshot catalog.
 MEMORY_DATASET_TYPE = "io.memory_dataset.MemoryDataset"
-
-
-@dataclass(frozen=True)
-class _LayerNode:
-    """Minimal stand-in carrying just ``.layer`` for the layer sorter."""
-
-    layer: str
 
 
 class _SnapshotGraphIndex:
@@ -102,21 +95,16 @@ class GraphBuilder:
     def __init__(
         self,
         snapshot: ProjectSnapshot,
-        layer_mapping: dict[str, str] | None = None,
+        catalog_config: dict[str, Any] | None = None,
     ) -> None:
         self._snapshot = snapshot
-        self._layers = layer_mapping or {}
+        self._layer_by_dataset = _extract_layers(
+            catalog_config or {}, _dataset_names_from_snapshot(snapshot)
+        )
         self._pipelines_by_id = {
             pipeline.name: pipeline for pipeline in snapshot.pipelines
         }
         self._index = _SnapshotGraphIndex(self._pipelines_by_id)
-        # Seed the sort with every layered dataset in the graph so the layer order stays
-        # consistent across pipeline views.
-        self._global_layer_nodes: dict[str, _LayerNode] = {
-            _create_dataset_node_id(name): _LayerNode(layer)
-            for name, layer in self._layers.items()
-            if self._index.get_pipelines_for_dataset_name(name)
-        }
 
     def default_pipeline_id(self) -> str:
         """Return the default pipeline ID to render.
@@ -196,7 +184,7 @@ class GraphBuilder:
         return GraphAPIResponse(
             nodes=nodes,
             edges=list(edges.values()),
-            layers=self._sorted_layers(nodes, edges),
+            layers=self._sorted_layers_for_pipeline(nodes, edges),
             tags=[
                 NamedEntityAPIResponse(id=tag, name=tag)
                 for tag in self._index.get_all_tags()
@@ -251,27 +239,33 @@ class GraphBuilder:
                 else GraphNodeType.DATA.value
             ),
             modular_pipelines=None,
-            layer=None if is_parameter else self._layers.get(stripped_name),
+            layer=(None if is_parameter else self._layer_by_dataset.get(stripped_name)),
             dataset_type=dataset_type,
         )
 
-    def _sorted_layers(
+    def _sorted_layers_for_pipeline(
         self,
         nodes: list[TaskNodeAPIResponse | DataNodeAPIResponse],
         edges: dict[tuple[str, str], GraphEdgeAPIResponse],
     ) -> list[str]:
-        """Topologically sort the data-node layers, reusing the existing layers service."""
-        if not self._layers:
+        """Sort the project-wide layer set using the selected pipeline's edges."""
+        if not self._layer_by_dataset:
             return []
         dependencies: dict[str, set[str]] = defaultdict(set)
         for source, target in edges:
             dependencies[source].add(target)
-        # Global layered datasets give layer presence; this view's nodes give the ordering.
-        nodes_by_id: dict[str, object] = dict(self._global_layer_nodes)
-        nodes_by_id.update({node.id: node for node in nodes})
-        # sort_layers only reads ``.layer`` off each node, so these stand in for the domain
-        # ``GraphNode`` it is typed against.
-        return sort_layers(cast("dict[str, GraphNode]", nodes_by_id), dependencies)
+        layer_by_node_id: dict[str, str | None] = {
+            node.id: node.layer if isinstance(node, DataNodeAPIResponse) else None
+            for node in nodes
+        }
+        # Include layered datasets used by any pipeline so every view exposes the
+        # project-wide layer set; this view's edges determine their order.
+        for dataset_name, layer in self._layer_by_dataset.items():
+            dataset_pipelines = self._index.get_pipelines_for_dataset_name(dataset_name)
+            if is_dataset_param(dataset_name) or not dataset_pipelines:
+                continue
+            layer_by_node_id.setdefault(_create_dataset_node_id(dataset_name), layer)
+        return sort_layers(layer_by_node_id, dependencies)
 
     @staticmethod
     def _register_dataset(
@@ -300,3 +294,14 @@ def _display_name(snapshot_name: str, namespace: str | None) -> str:
         local = local[len(prefix) :]
     auto = _AUTO_NAME_RE.match(local)
     return auto.group("func") if auto else local
+
+
+def _dataset_names_from_snapshot(snapshot: ProjectSnapshot) -> set[str]:
+    """Return non-parameter dataset names referenced by registered pipelines."""
+    return {
+        name
+        for pipeline in snapshot.pipelines
+        for node in pipeline.nodes
+        for name in (*node.inputs, *node.outputs)
+        if not is_dataset_param(name)
+    }
--- a/package/kedro_viz/integrations/kedro/inspection/graph_builder.py
+++ b/package/kedro_viz/integrations/kedro/inspection/graph_builder.py
@@ -99,7 +99,8 @@ class GraphBuilder:
     ) -> None:
         self._snapshot = snapshot
         self._layer_by_dataset = _extract_layers(
-            catalog_config or {}, _dataset_names_from_snapshot(snapshot)
+            catalog_config=catalog_config or {},
+            dataset_names=_dataset_names_from_snapshot(snapshot),
         )
         self._pipelines_by_id = {
             pipeline.name: pipeline for pipeline in snapshot.pipelines
--- a/package/kedro_viz/integrations/kedro/inspection/graph_builder.py
+++ b/package/kedro_viz/integrations/kedro/inspection/graph_builder.py
@@ -9,7 +9,8 @@ from __future__ import annotations
 
 import re
 from collections import defaultdict
-from typing import TYPE_CHECKING
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, cast
 
 from kedro_viz.api.rest.responses.pipelines import (
     DataNodeAPIResponse,
@@ -24,6 +25,7 @@ from kedro_viz.integrations.kedro.node_ids import (
     _create_task_node_id,
 )
 from kedro_viz.models.flowchart.model_utils import GraphNodeType
+from kedro_viz.services.layers import sort_layers
 from kedro_viz.utils import _strip_transcoding, is_dataset_param
 
 if TYPE_CHECKING:
@@ -33,11 +35,20 @@ if TYPE_CHECKING:
         ProjectSnapshot,
     )
 
+    from kedro_viz.models.flowchart.nodes import GraphNode
+
 _AUTO_NAME_RE = re.compile(r"^(?P<func>.+)__[0-9a-f]{8}$")
 
 # Match the live backend's get_dataset_type(MemoryDataset()) output for datasets
 # absent from the snapshot catalog.
 MEMORY_DATASET_TYPE = "io.memory_dataset.MemoryDataset"
+
+
+@dataclass(frozen=True)
+class _LayerNode:
+    """Minimal stand-in carrying just ``.layer`` for the layer sorter."""
+
+    layer: str
 
 
 class _SnapshotGraphIndex:
@@ -88,12 +99,24 @@ class GraphBuilder:
     while rendered nodes and edges are scoped to the selected pipeline.
     """
 
-    def __init__(self, snapshot: ProjectSnapshot) -> None:
+    def __init__(
+        self,
+        snapshot: ProjectSnapshot,
+        layer_mapping: dict[str, str] | None = None,
+    ) -> None:
         self._snapshot = snapshot
+        self._layers = layer_mapping or {}
         self._pipelines_by_id = {
             pipeline.name: pipeline for pipeline in snapshot.pipelines
         }
         self._index = _SnapshotGraphIndex(self._pipelines_by_id)
+        # Seed the sort with every layered dataset in the graph so the layer order stays
+        # consistent across pipeline views.
+        self._global_layer_nodes: dict[str, _LayerNode] = {
+            _create_dataset_node_id(name): _LayerNode(layer)
+            for name, layer in self._layers.items()
+            if self._index.get_pipelines_for_dataset_name(name)
+        }
 
     def default_pipeline_id(self) -> str:
         """Return the default pipeline ID to render.
@@ -173,7 +196,7 @@ class GraphBuilder:
         return GraphAPIResponse(
             nodes=nodes,
             edges=list(edges.values()),
-            layers=[],
+            layers=self._sorted_layers(nodes, edges),
             tags=[
                 NamedEntityAPIResponse(id=tag, name=tag)
                 for tag in self._index.get_all_tags()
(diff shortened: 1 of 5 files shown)


package/kedro_viz/integrations/kedro/inspection/layers.py
```diff
@@ -0,0 +1,75 @@
+"""Extract dataset layers from the raw catalog config.
+
+Layer is a Kedro-Viz concept (under ``metadata.kedro-viz.layer``); Kedro stores the ``metadata``
```
ravi-kumar-pilla: we do not need to explain this here in the doc strings, can we remove it but just keep what the file does

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -228,10 +251,28 @@ def _build_dataset_node(
...
+            return []
+        dependencies: dict[str, set[str]] = defaultdict(set)
+        for source, target in edges:
+            dependencies[source].add(target)
+        # Global layered datasets give layer presence; this view's nodes give the ordering.
+        nodes_by_id: dict[str, object] = dict(self._global_layer_nodes)
+        nodes_by_id.update({node.id: node for node in nodes})
+        # sort_layers only reads ``.layer`` off each node, so these stand in for the domain
+        # ``GraphNode`` it is typed against.
+        return sort_layers(cast("dict[str, GraphNode]", nodes_by_id), dependencies)
```
ravi-kumar-pilla: This feels a bit complex. Anyway we are removing legacy code, not sure if GraphNode will be around after removal. I would recommend an easy to follow version if the parity tests pass and nothing is missed -

Suggestion: replace `_LayerNode`, `_global_layer_nodes`, and `_sorted_layers` with a dict-based `sort_layers` and a single private helper on `GraphBuilder`.

Legacy `services/layers.py` is left untouched for now; when the old backend is removed, this inspection-local `sort_layers` can move into `services/layers.py` and replace the `GraphNode`-based version.

---

## 1. Add `sort_layers` to `inspection/layers.py`

Place alongside `_extract_layers`. Same algorithm as legacy `sort_layers`, but takes `node_id → layer` instead of `GraphNode` objects. (why do we need an entire object when we have a node_id and sorting just the layers ? The code has docstrings explaining the behavior but remove those when you feel this code works well @jitu5 )

```python
import logging
from collections import defaultdict
from graphlib import CycleError, TopologicalSorter

def sort_layers(
    layers_by_node_id: dict[str, str | None],
    dependencies: dict[str, set[str]],
) -> list[str]:
    """Return layer labels topologically sorted by data-flow dependencies.

    Same behaviour as legacy ``sort_layers(nodes, dependencies)``, but takes a plain
    ``node_id -> layer`` map instead of ``GraphNode`` objects.

    Args:
        layers_by_node_id: Node IDs mapped to a layer label, or ``None`` if the node has
            no layer (e.g. task nodes). Must include every node ID that appears in
            ``dependencies`` or is reachable from those nodes during traversal.
        dependencies: ``{node_id: {child_node_id, ...}}`` for the selected pipeline.

    Returns:
        Ordered layer names, or ``[]`` if there are no layers or a cycle is detected.
    """
    if not any(layer is not None for layer in layers_by_node_id.values()):
        return []

    node_layers: dict[str, set[str]] = {}

    def find_child_layers(node_id: str) -> set[str]:
        if node_id in node_layers:
            return node_layers[node_id]

        node_layers[node_id] = set()

        node_layer = layers_by_node_id.get(node_id)
        if node_layer is not None:
            node_layers[node_id].add(node_layer)

        for child_node_id in dependencies.get(node_id, set()):
            child_layer = layers_by_node_id.get(child_node_id)
            if child_layer is not None:
                node_layers[node_id].add(child_layer)
            node_layers[node_id].update(find_child_layers(child_node_id))

        return node_layers[node_id]

    for node_id in sorted(layers_by_node_id):
        find_child_layers(node_id)

    layer_dependencies: dict[str, set[str]] = defaultdict(set)
    all_layers: set[str] = set()

    for node_id, child_layers in node_layers.items():
        node_layer = layers_by_node_id.get(node_id)
        if node_layer is not None:
            all_layers.add(node_layer)
            for layer in child_layers:
                all_layers.add(layer)
                if layer != node_layer:
                    layer_dependencies[layer].add(node_layer)

    for layer in all_layers:
        layer_dependencies.setdefault(layer, set())

    layer_dependencies = defaultdict(
        set, {key: layer_dependencies[key] for key in sorted(layer_dependencies)}
    )

    try:
        sorter = TopologicalSorter(layer_dependencies)
        sorted_layers = list(sorter.static_order())
        return sorted(
            sorted_layers,
            key=lambda layer: (sorted_layers.index(layer), layer),
        )
    except CycleError as exc:
        logger.warning(
            "Layers visualisation is disabled as circular dependency detected among layers. "
            "Circular dependency detected: %s. "
            "Please check the `layer` configuration in your catalog for the datasets to avoid circular references. ",
            str(exc),
        )
        return []
```

---

## 2. Simplify `graph_builder.py`

### Remove

- `@dataclass _LayerNode`
- `self._global_layer_nodes` in `__init__`
- `_sorted_layers` method
- `from dataclasses import dataclass`
- `cast` and the `GraphNode` TYPE_CHECKING import
- `from kedro_viz.services.layers import sort_layers`

### Add / change

```python
from collections import defaultdict

from kedro_viz.integrations.kedro.inspection.layers import sort_layers
```

Rename `self._layers` → `self._layer_by_dataset` for clarity.

```python
def __init__(
    self,
    snapshot: ProjectSnapshot,
    layer_by_dataset: dict[str, str] | None = None,
) -> None:
    self._snapshot = snapshot
    self._layer_by_dataset = layer_by_dataset or {}
    self._pipelines_by_id = {
        pipeline.name: pipeline for pipeline in snapshot.pipelines
    }
    self._index = _SnapshotGraphIndex(self._pipelines_by_id)
```

In `_build_dataset_node`:

```python
layer=None if is_parameter else self._layer_by_dataset.get(stripped_name),
```

In `build()`:

```python
return GraphAPIResponse(
    nodes=nodes,
    edges=list(edges.values()),
    layers=self._sorted_layers_for_pipeline(nodes, edges),
    # ...
)
```

### Private helper

```python
def _sorted_layers_for_pipeline(
    self,
    nodes: list[TaskNodeAPIResponse | DataNodeAPIResponse],
    edges: dict[tuple[str, str], GraphEdgeAPIResponse],
) -> list[str]:
    """Return topologically sorted layer labels for the selected pipeline."""
    if not self._layer_by_dataset:
        return []

    layers_by_node_id: dict[str, str | None] = {
        node.id: node.layer if isinstance(node, DataNodeAPIResponse) else None
        for node in nodes
    }
    # Mirror legacy sort_layers(nodes.as_dict(), pipeline_edges): include layered
    # datasets from all pipelines so layer order stays stable across pipeline views.
    for name, layer in self._layer_by_dataset.items():
        if self._index.get_pipelines_for_dataset_name(name):
            layers_by_node_id.setdefault(_create_dataset_node_id(name), layer)

    dependencies: dict[str, set[str]] = defaultdict(set)
    for source, target in edges:
        dependencies[source].add(target)

    return sort_layers(layers_by_node_id, dependencies)
```

---

## Why this is simpler

| Current PR | Proposed |
|------------|----------|
| `_layers` + `_global_layer_nodes` | `_layer_by_dataset` only |
| `_LayerNode` dataclass stub | plain `dict[str, str \| None]` |
| `cast("dict[str, GraphNode]", ...)` | no cast |
| Reuses legacy `GraphNode` API | dict-native sort, same algorithm |

The helper builds two inputs for `sort_layers`:

1. **`layers_by_node_id`** — pipeline nodes (tasks as `None`, data nodes with layer) plus project-wide layered dataset IDs via `setdefault` (legacy `nodes.as_dict()` seeding).
2. **`dependencies`** — pipeline-scoped edges only (same as legacy).

Orphan datasets (layer mapped but unused by any node) stay excluded via `get_pipelines_for_dataset_name(name)`.

---

## Tests

- Existing graph builder edge-case tests should pass unchanged.
- Add unit tests for `sort_layers` in `test_layers.py` using plain dicts (can port cases from `test_services/test_layers.py` without `GraphNode.create_data_node`).

---

## Follow-up (when legacy backend is removed)

- Move `sort_layers` from `inspection/layers.py` to `services/layers.py`.
- Delete the old `GraphNode`-based `sort_layers` and `CatalogRepository.layers_mapping`.

package/kedro_viz/integrations/kedro/inspection/layers.py
```diff
@@ -0,0 +1,75 @@
...
+    """Map concrete dataset names to layers from ``metadata.kedro-viz.layer``.
+
+    Dataset factory patterns are resolved against ``dataset_names`` using Kedro's catalog
+    resolver.
+
+    Raises:
+        ValueError: If transcoded variants of one dataset (``name@a``, ``name@b``) declare
+            different layers, matching the legacy backend's validation.
+    """
+    mapping: dict[str, str] = {}
```
ravi-kumar-pilla:
```suggestion
    layer_by_dataset: dict[str, str] = {}
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -88,12 +99,24 @@ class GraphBuilder:
     while rendered nodes and edges are scoped to the selected pipeline.
     """
 
-    def __init__(self, snapshot: ProjectSnapshot) -> None:
+    def __init__(
+        self,
+        snapshot: ProjectSnapshot,
+        layer_mapping: dict[str, str] | None = None,
```
ravi-kumar-pilla: layer_by_dataset or layer_mapping on GraphBuilder.__init__ exposes catalog wiring to callers. Since layers are dataset metadata from the catalog (not in the snapshot), lets do GraphBuilder(snapshot, catalog_config) and resolve _extract_layers internally, so init doesn’t take a pre-built mapping.

ravi-kumar-pilla:
```python
class GraphBuilder:
    def __init__(
        self,
        snapshot: ProjectSnapshot,
        catalog_config: dict[str, Any] | None = None,
    ) -> None:
        self._snapshot = snapshot
        self._layer_by_dataset = _extract_layers(
            catalog_config or {},
            _dataset_names_from_snapshot(snapshot),
        )
```

ravi-kumar-pilla:
```python
def _dataset_names_from_snapshot(snapshot: ProjectSnapshot) -> set[str]:
    """Return every dataset name referenced across all registered pipelines."""
    return {
        name
        for pipeline in snapshot.pipelines
        for node in pipeline.nodes
        for name in [*node.inputs, *node.outputs]
    }
```

package/kedro_viz/integrations/kedro/inspection/layers.py
jitu5: Thanks updated.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated this. `sort_layers()` now uses a node-to-layer mapping, removing `_LayerNode`, `GraphNode`, and the cast. The existing behaviour is preserved and covered by tests.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated this. GraphBuilder now accepts the catalog config and resolves layers internally.

package/kedro_viz/integrations/kedro/inspection/layers.py
```diff
@@ -0,0 +1,131 @@
...
+        ValueError: If transcoded variants of one dataset (``name@a``, ``name@b``) declare
+            different layers, matching the legacy backend's validation.
+    """
+    layer_by_dataset: dict[str, str] = {}
+    resolver_config: dict[str, dict[str, Any]] = {}
+    has_patterns = False
+    for name, config in catalog_config.items():
+        if not isinstance(config, dict):
+            continue
+        resolver_entry: dict[str, Any] = {"type": "kedro.io.MemoryDataset"}
```
ravi-kumar-pilla, line 37: [Nit] Add some comments for readability
 
```py
# Include every catalog entry so pattern precedence matches Kedro's resolver
# (e.g. a more specific pattern without a layer blocks a broader one with a layer).
resolver_entry: dict[str, Any] = {"type": "kedro.io.MemoryDataset"}
```

package/tests/test_inspection_adapter/test_graph_builder_edge_cases.py
```diff
@@ -271,3 +273,102 @@ def test_build_rejects_unknown_pipeline_id() -> None:
...
+    catalog_config = {
+        "raw_data": {"metadata": {"kedro-viz": {"layer": "raw"}}},
+        "model": {"metadata": {"kedro-viz": {"layer": "model"}}},
+        "external": {"metadata": {"kedro-viz": {"layer": "external"}}},
+        "report": {"metadata": {"kedro-viz": {"layer": "reporting"}}},
+        "orphan": {"metadata": {"kedro-viz": {"layer": "unused"}}},
+    }
+
+    graph = _builder(snapshot, catalog_config).build("pipe_a")
+    assert set(graph.layers) == {"raw", "model", "external", "reporting"}
```
ravi-kumar-pilla: Set comparison won't catch ordering regressions. Baseline tests already cover order on the demo project; this test could pin expected order for the two-pipeline fixture, e.g.:
```py
assert graph.layers == ["raw", "model", "external", "reporting"]  # if that's the intended order
```

package/kedro_viz/integrations/kedro/inspection/layers.py
```diff
@@ -0,0 +1,131 @@
...
+) -> list[str]:
+    """Return dataset layers in graph dependency order."""
+    child_layers_by_node: dict[str, set[str]] = {}
+
+    def find_child_layers(node_id: str) -> set[str]:
+        if node_id in child_layers_by_node:
+            return child_layers_by_node[node_id]
+
+        child_layers = child_layers_by_node[node_id] = set()
+        node_layer = layer_by_node_id[node_id]
```
ravi-kumar-pilla:
```suggestion
        node_layer = layer_by_node_id.get(node_id)
```

package/kedro_viz/integrations/kedro/inspection/layers.py
```diff
@@ -0,0 +1,131 @@
...
+        if node_layer is not None:
+            child_layers.add(node_layer)
+
+        for child_id in dependencies.get(node_id, set()):
+            child_layer = layer_by_node_id[child_id]
```
ravi-kumar-pilla:
```suggestion
            child_layer = layer_by_node_id.get(child_id)
```

package/kedro_viz/integrations/kedro/inspection/layers.py
```diff
@@ -0,0 +1,131 @@
...
+            "Please ensure consistent layering in your Kedro catalog. "
+            f"Mismatch found for: {stripped}"
+        )
+    layer_by_dataset[stripped] = layer
+
+
+def sort_layers(
+    layer_by_node_id: dict[str, str | None],
+    dependencies: dict[str, set[str]],
+) -> list[str]:
```
ravi-kumar-pilla, line 81: Legacy had a final sorted(..., key=lambda layer: (sorted_layers.index(layer), layer)) for alphabetical tie-breaking. The new version relies on TopologicalSorter + sorted dependency keys. can we restore the alphabetical tie-break ?

ravi-kumar-pilla:
Hi @jitu5 , left minor comments but overall looks good. Thank you

package/kedro_viz/integrations/kedro/inspection/layers.py
jitu5: Thanks, I checked this. The legacy final `sorted()` does not change the order because each layer has a unique index. Stable ordering comes from the sorted dependency keys, which the new code retains and the tests cover.

package/kedro_viz/integrations/kedro/inspection/layers.py
```diff
@@ -0,0 +1,132 @@
...
+    resolver_config: dict[str, dict[str, Any]] = {}
+    has_patterns = False
+    for name, config in catalog_config.items():
+        if not isinstance(config, dict):
+            continue
+        # Include entries without layers to preserve Kedro's resolver precedence.
+        resolver_entry: dict[str, Any] = {"type": "kedro.io.MemoryDataset"}
+        resolver_config[name] = resolver_entry
+        try:
+            layer = config["metadata"]["kedro-viz"]["layer"]
```
ankatiyar, line 40: nit: You can do `config.get()` here so it returns `None` instead of erroring if the key doesn't exist and then maybe can get rid of the `try` bit

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -88,8 +92,15 @@ class GraphBuilder:
...
     """
 
-    def __init__(self, snapshot: ProjectSnapshot) -> None:
+    def __init__(
+        self,
+        snapshot: ProjectSnapshot,
+        catalog_config: dict[str, Any] | None = None,
+    ) -> None:
         self._snapshot = snapshot
+        self._layer_by_dataset = _extract_layers(
```
ankatiyar, line 101: nit: Would be nice to specify argument name also like `_extract_layers(snapshot=...)` just so if something changes in the private method (like argument ordering), it doesn't break

ankatiyar (commented):
Some minor comments otherwise looks good! 
I have a general question regarding the `catalog_config`: Where is this coming from? Does it come from the Kedro snapshot API - if so do you resolve the catalog before sending it to the snapshot or do you receive the raw config? I would assume that catalog resolution to find "layers" wouldn't happen on the viz side

package/kedro_viz/integrations/kedro/inspection/layers.py
jitu5: Thanks. A chained `.get()` could still fail when nested metadata is `None` or not a mapping. The current try/except safely handles both missing and malformed metadata.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Done.

jitu5:
> I have a general question regarding the catalog_config: Where is this coming from? Does it come from the Kedro snapshot API - if so do you resolve the catalog before sending it to the snapshot or do you receive the raw config? I would assume that catalog resolution to find "layers" wouldn't happen on the viz side

@ankatiyar  `catalog_config` does not come from the snapshot. `_InspectionSession` reads the loaded catalog configuration separately through Kedro’s config loader, without creating a `DataCatalog`. The snapshot does not currently include `metadata.kedro-viz.layer`, so Viz extracts that metadata, while factory matching is delegated to Kedro’s `CatalogConfigResolver`.

ravi-kumar-pilla (approved):
Thank you @jitu5

ankatiyar:
In that case, would it make sense to include the layer data in the `DatasetSnapshot` part of the API so you don't have to do this Viz side? If the catalog config is needed only for the layer information, since the snapshot includes all resolved datasets? WDYT @ravi-kumar-pilla

ravi-kumar-pilla:
> In that case, would it make sense to include the layer data in the `DatasetSnapshot` part of the API so you don't have to do this Viz side? If the catalog config is needed only for the layer information, since the snapshot includes all resolved datasets? WDYT @ravi-kumar-pilla

It make sense for kedro-viz specific requirement. But during our discussions on snapshot api, we wanted to keep the api surface specific to Kedro. The layer information is something specific to Kedro-Viz metadata. I feel deriving it here as in the PR looks good to me. Having said that if majority of you feel we can include this on Kedro side, happy to change that. I personally would leave this as is now. 

Do you have any concerns on doing this way ?

ankatiyar:
@ravi-kumar-pilla @jitu5 No specific concerns, just wondering. I don't see the harm in including `metadata:` key in the `DatasetSnapshot` as the inspection API is meant to be consumed by plugins and so is the stuff under `metadata:` in the catalog. We don't have to necessarily extract layers on Kedro side, that can still be done here. But if you are going to need the catalog config anyway, for example to create a catalog to load the dataset stats etc - then I suppose it is fine! This is just a design suggestion, the implementation here looks good to me!

jitu5:
@ravi-kumar-pilla @ankatiyar 
Thanks both. Viz already needs access to the config loader and catalog for parameter resolution, dataset previews, and statistics, so keeping layer extraction on the Viz side is reasonable for this PR. We can revisit exposing generic metadata through `DatasetSnapshot` separately if broader plugin use cases emerge.

Issue #2687, opened 2026-06-29
Populate snapshot graph nodes with resolved parameter values


jitu5:
**Parent issue:** #2265
**Depends on:** #2655, #2656, #2657, #2658, #2715

> **Update:** The foundation already provides `_InspectionSession.parameters()` and a cached config loader that applies runtime parameter overrides. This ticket reuses it; it does not add another loader or manually merge `--params` values.

## Summary

Pass the resolved parameter mapping into `GraphBuilder` and populate each task node's `parameters` field in the same shape as the legacy backend.

## Scope

- Let `GraphBuilder` receive resolved parameter values.
- For an input named `parameters`, attach the complete parameter mapping.
- For `params:x`, attach `{x: value}`.
- Resolve dotted names such as `params:model_options.test_size` from nested configuration.
- Preserve value types and handle a missing value without failing graph construction.
- Add focused parity and edge-case tests.

Runtime provider installation belongs to #2659. Parameter-node `/api/nodes/{id}` metadata belongs to #2660.

## Acceptance criteria

- Task-node parameter mappings match the captured baseline for every demo pipeline.
- Root `parameters`, individual `params:x`, dotted keys, and missing keys are covered.
- Resolved values passed to `GraphBuilder`, including typed runtime overrides returned by `_InspectionSession.parameters()`, appear in task-node parameter mappings.
- Supplying runtime parameters does not change any node or edge ID.

jitu5:
@ravi-kumar-pilla 
You can use this branch `feature/snapshot-adapter-synced` as reference as this has all the changes including latest modular pipeline PR. and its working end to end.

Issue #2659, opened 2026-05-26
Serve graph endpoints through the inspection adapter


jitu5:
**Parent issue:** #2265
**Depends on:** #2657, #2658, #2687, #2715

> **Update:** The original ticket described an environment flag, two graph providers, live fallback, and a special live path for `--params`. Those decisions are obsolete. The inspection adapter will be the only engine serving runtime HTTP graph responses, and resolved runtime parameters come from the existing config loader.

## Summary

Serve `/api/main` and `/api/pipelines/{id}` from the snapshot `GraphBuilder`. The adapter is already implemented and tested, but these routes still use the old `data_access_manager` graph path.

## Scope

- Add one small inspection-adapter provider around `GraphBuilder`.
- Build and install the provider once during server startup using the selected project, environment, pipeline, catalog configuration, and resolved parameters.
- Preserve the full graph response contract before cutover: attach `.viz` stats/styles as `node_extras`, and in full mode use a same-ID live lookup to supply the resolved `dataset_type` where the snapshot contains only a raw catalog type string. This enrichment must not change graph topology or IDs.
- Route `/api/main` and `/api/pipelines/{id}` through that provider.
- Honour `--env`, `--pipeline`, and `--params`.
- Fail clearly if the adapter cannot be built; do not silently switch to a second graph engine.

At this point the adapter is authoritative only for runtime HTTP graph endpoints. The current live load may remain temporarily for full-mode graph enrichment, node metadata/previews, static export, and notebook/VSCode consumers. Moving `/api/nodes/{id}`, `/api/run-status`, and static export is #2660; skipping the live load is #2661; deleting old graph construction is a later cleanup task.

## Out of scope

- An adapter feature flag.
- `LiveDataProvider` or automatic live-graph fallback.
- Node metadata, static export, notebook/VSCode migration, and lite startup.

## Acceptance criteria

- `/api/main` and every valid `/api/pipelines/{id}` response come from the inspection adapter.
- Responses match the captured baseline, including exact IDs, task parameters, `node_extras`, and resolved dataset types.
- Existing graph styles, stats, and dataset-type icons do not regress.
- The requested pipeline and environment are honoured.
- Typed runtime parameter overrides appear in task-node parameters without changing graph IDs.
- An unknown pipeline ID returns 404.
- Provider construction occurs once per server load, not once per request.

ravi-kumar-pilla:
Hi @jitu5 ,

[...]

PR #2726, opened 2026-08-06
Build modular pipelines from the inspection snapshot

jitu5:
**Sub-task:** #2657 
**Parent:** #2265

## Description

Adds modular-pipeline rendering to the snapshot graph builder. It builds group nodes, the namespace tree, boundary connections, dataset ownership and subtree tags directly from inspection snapshots, without live Kedro `Pipeline` objects.

This PR targets `feat/backend_v2` and remains additive. The inspection adapter is not yet serving `/api/main`.

## Implementation

- `modular_pipeline_algebra.py` contains the snapshot-based namespace set algebra used to calculate nested namespaces, free inputs and outputs, and datasets owned by each modular pipeline.
- `ModularPipelineIndex` builds project-wide dataset ownership once from all registered pipelines.
- `ModularPipelineView` builds the selected pipeline's tree, group nodes and boundary connections for each render.
- `GraphBuilder` depends only on `ModularPipelineIndex` and `ModularPipelineView` for modular-pipeline behavior.
- Each registered pipeline is calculated independently before dataset ownership is combined across the project. This preserves same-ID tasks in different namespaces and prevents unrelated pipelines from cancelling valid boundaries.
- Task nodes retain their direct namespace. Parameter references have no modular-pipeline owners, transcoded dataset names share ownership, subtree tags are combined, and cyclic boundary connections are removed.
- `__root__` is reserved for the synthetic Viz tree root. Conflicting namespaces raise an actionable error instead of producing an invalid tree.

## QA notes

- Structural baseline comparisons cover modular-pipeline nodes, each node's `modular_pipelines` value, tree inputs, outputs and children, boundary connections, tags and registered-pipeline lists for every captured demo pipeline.
- The captured baseline contains one duplicate root parameter child. Tests normalise only that known duplicate; every other child is compared by ID and type.
- Focused tests cover the set algebra, public index and view interfaces, nested namespaces, boundary datasets, parameters, transcoding, source and sink nodes, pipelines without namespaces, cycle removal, deterministic edge ordering, cross-pipeline identity collisions and the reserved root namespace.

## Deferred work

- Resolved parameter values remain part of #2687.
- Runtime endpoint wiring remains part of #2659.
- `RELEASE.md` will be updated when the complete `backend_v2` work is prepared for `main`.

## Verification

- `pytest package/tests/test_inspection_adapter/`: 184 passed
- `ruff check`: clean

## Checklist

- [x] Opened this PR as a draft while it is in progress
- [ ] Updated the documentation (not applicable: internal adapter work with no runtime wiring)
- [ ] Added a `RELEASE.md` entry (deferred until the complete backend work targets `main`)
- [x] Added tests to cover the changes


package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,330 @@
...
+        self.ids: list[str] = sorted(_modular_pipeline_ids(nodes))
+
+    def build(self) -> dict[str, _ModularTreeEntry]:
+        """Return the tree keyed by modular pipeline id, including ``__root__``."""
+        tree = {ROOT_MODULAR_PIPELINE_ID: _ModularTreeEntry(ROOT_MODULAR_PIPELINE_ID)}
+        params: set[str] = set()
+
+        for mp_id in self.ids:
+            entry = tree.setdefault(mp_id, _ModularTreeEntry(mp_id))
+            free_inputs, free_outputs = _free_io(self._nodes, mp_id)
```
ravi-kumar-pilla: `_ModularPipelineIndex.__init__` and `_ModularTreeBuilder.build()` both call `_free_io()` for every modular pipeline on the same node set. On bigger pipelines that means doing the same work twice on every `GraphBuilder.build()` call. You already compute boundaries in the index, so caching the `(free_inputs, free_outputs)` pairs there and reusing them when filling in the tree could save some work.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,330 @@
...
+                for node in nodes
+                if node.namespace == mp_id
+                for io in [*node.inputs, *node.outputs]
+            }
+            free_inputs, free_outputs = _free_io(nodes, mp_id)
+            boundary = {_strip_transcoding(d) for d in free_inputs | free_outputs}
+            self._datasets_by_modular_pipeline[mp_id] = direct | boundary
+
+    @classmethod
+    def from_registered_pipelines(
```
ravi-kumar-pilla: `from_registered_pipelines` starts with `cls([])` and then patches `_ids` and `_datasets_by_modular_pipeline` on the instance. It works fine today, but a factory that builds up local dicts and returns a fully constructed object would be easier to read and less fragile if `__init__` picks up more logic later.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -1,8 +1,9 @@
 """Build the main Kedro-Viz graph response from an inspection snapshot.
 
-Creates task, data, parameter nodes, edges, global tags and pipeline dictionaries. Node IDs come
-from ``kedro_viz.integrations.kedro.node_ids``; registered non-transcoded datasets use raw catalog
-type strings from the snapshot.
+Creates task, data, parameter and modular-pipeline nodes; the edges between them, including
+modular edges; the modular-pipeline tree; and global tags and pipeline dictionaries. Node IDs
+come from ``kedro_viz.integrations.kedro.node_ids``; registered non-transcoded datasets use raw
+catalog type strings from the snapshot.
 """
```
ravi-kumar-pilla, line 9: Builds the main Kedro Viz graph response from an inspection snapshot.

It assembles task, data, parameter, and modular pipeline nodes, their connecting edges, the modular pipeline tree, the global tag list, and the registered pipeline list.

Node IDs come from ``kedro_viz.integrations.kedro.node_ids``. Registered datasets without transcoding use the raw catalog type strings stored in the snapshot.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -22,9 +25,16 @@
     _extract_layers,
     sort_layers,
 )
+from kedro_viz.integrations.kedro.inspection.modular_pipelines import (
+    _add_modular_edges,
```
ravi-kumar-pilla: `_add_modular_edges` sounds like a different kind of edge, but it only wires boundary datasets to the Viz group nodes. Something like `_add_modular_pipeline_boundary_edges` would better reflect what the code does. wdyt ? If you feel _add_modular_edges make sense, ignore it

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -22,9 +25,16 @@
...
+    _ModularPipelineIndex,
+    _ModularTreeBuilder,
+    _ModularTreeEntry,
+    _remove_cyclic_modular_edges,
+)
 from kedro_viz.integrations.kedro.node_ids import (
     _create_dataset_node_id,
-    _create_task_node_id,
+    _create_task_node_id_from_snapshot,
```
ravi-kumar-pilla: is this needed ? can we have the same as before like `_create_task_node_id` ?

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla, line 58: may be update the docstring, so it is clear why `_SnapshotGraphIndex` class is required

```
    """Index tags and registered pipelines by task and dataset.
    Built once from every pipeline in the snapshot and reused when constructing
    node responses.
    """
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -53,13 +63,7 @@ def __init__(self, pipelines: dict[str, PipelineSnapshot]) -> None:
...
         for pipeline_id, pipeline in pipelines.items():
             for node in pipeline.nodes:
-                task_id = _create_task_node_id(
-                    node_name=node.name,
-                    func_name=node.func_name,
-                    namespace=node.namespace,
-                    inputs=node.inputs,
-                    outputs=node.outputs,
-                )
+                task_id = _create_task_node_id_from_snapshot(node)
```
ravi-kumar-pilla: `_create_task_node_id_from_snapshot` - this name is misleading as the input is a `node` and not a snapshot. I would prefer having `_create_task_node_id`. But if there is any specific reason to not have this, it would make sense to name it as `_create_task_node_id_from_node`

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla: suggested docstring as this is a bit hard to understand for me -

```
"""Build ``GraphAPIResponse`` objects from a project inspection snapshot.
    
    Renders nodes and edges for one selected pipeline. Tags and registered
    pipelines on each node are collected across every pipeline in the project.
"""
```

or something like -

```
"""Renders one pipeline at a time, but each node's tags and pipeline list
reflect every pipeline that node appears in."""
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -110,6 +114,9 @@ def __init__(
             pipeline.name: pipeline for pipeline in snapshot.pipelines
         }
         self._index = _SnapshotGraphIndex(self._pipelines_by_id)
+        self._modular = _ModularPipelineIndex.from_registered_pipelines(
```
ravi-kumar-pilla: `_modular` is a bit vague next to `_index`. Since this object is really a lookup table for datasets and namespaces, something like `_modular_pipeline_index` would read more clearly and match the class name.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -216,7 +224,7 @@ def _build_task_node(self, node: NodeSnapshot, task_id: str) -> TaskNodeAPIRespo
             tags=self._index.get_tags_for_task_id(task_id),
             pipelines=self._index.get_pipelines_for_task_id(task_id),
             type=GraphNodeType.TASK.value,
-            modular_pipelines=None,
+            modular_pipelines=self._modular.get_modular_pipelines_for_task(node),
```
ravi-kumar-pilla: This index is built for datasets. For tasks it just returns `node.namespace`, so it never actually looks anything up. You could set `modular_pipelines` inline here (for example `[node.namespace]` when the namespace is set, otherwise `None`) and drop `get_modular_pipelines_for_task` from the index. That would make it clearer the index exists for datasets, not tasks.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -150,8 +157,9 @@ def build(self, pipeline_id: str | None = None) -> GraphAPIResponse:
                 pipeline (see :meth:`default_pipeline_id`) is used.
 
         Returns:
-            The ``GraphAPIResponse`` for the selected pipeline: its task, data and parameter
-            nodes and edges, plus the globally-aggregated tag and pipeline lists.
+            The ``GraphAPIResponse`` for the selected pipeline: its task, data, parameter and
+            modular-pipeline nodes; the edges between them, including modular edges; the
+            modular-pipeline tree; and the globally-aggregated tag and pipeline lists.
```
ravi-kumar-pilla, line 167: **Suggestion:**

Builds the ``GraphAPIResponse`` for the selected pipeline.

The response includes task, data, parameter, and modular pipeline nodes, their connecting edges, the modular pipeline tree, the global tag list, and the registered pipeline list. Tags and pipeline lists on each node are taken from the whole project, not only the selected pipeline.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -22,9 +25,16 @@
     _extract_layers,
     sort_layers,
 )
+from kedro_viz.integrations.kedro.inspection.modular_pipelines import (
+    _add_modular_edges,
+    _ModularPipelineIndex,
+    _ModularTreeBuilder,
+    _ModularTreeEntry,
```
ravi-kumar-pilla:
```suggestion
    _ModularPipelineTreeEntry,
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -22,9 +25,16 @@
     _extract_layers,
     sort_layers,
 )
+from kedro_viz.integrations.kedro.inspection.modular_pipelines import (
+    _add_modular_edges,
+    _ModularPipelineIndex,
+    _ModularTreeBuilder,
```
ravi-kumar-pilla:
```suggestion
    _ModularPipelineTreeBuilder,
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -192,6 +194,12 @@ def build(self, pipeline_id: str | None = None) -> GraphAPIResponse:
                 )
             )
 
+        tree_builder = _ModularTreeBuilder(pipeline.nodes)
+        tree = tree_builder.build()
+        nodes.extend(self._build_modular_pipeline_nodes(tree_builder, selected))
+        _add_modular_edges(edges, tree)
```
ravi-kumar-pilla:
```suggestion
        _add_modular_pipeline_boundary_edges(edges, tree)
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -192,6 +194,12 @@ def build(self, pipeline_id: str | None = None) -> GraphAPIResponse:
                 )
             )
 
+        tree_builder = _ModularTreeBuilder(pipeline.nodes)
```
ravi-kumar-pilla: The task and dataset loops read clearly, but this block does something different (tree, group nodes, boundary connections). Pulling it into a helper like `_add_modular_pipelines_to_graph(...)` would keep `build()` as a short outline and make it easier to see where the core graph ends and the modular pipeline work begins.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -297,6 +327,33 @@ def _add_edge(
         )
 
 
+def _to_tree_response(
```
ravi-kumar-pilla: This does not build the tree. It converts the internal tree into the API shape for the `modular_pipelines` field. Something like `_build_modular_pipeline_tree_response` would make that clearer, especially next to `_build_modular_pipeline_nodes`.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla:
```python
# tasks and datasets
...

self._add_modular_pipelines_to_graph(nodes, edges, pipeline.nodes, selected)

return GraphAPIResponse(...)
```

The helper `_add_modular_pipelines_to_graph` could own -

```python
tree_builder = _ModularPipelineTreeBuilder(pipeline.nodes)
tree = tree_builder.build()
nodes.extend(self._build_modular_pipeline_nodes(tree_builder, selected))
_add_modular_pipeline_boundary_edges(edges, tree)
_remove_cyclic_modular_pipeline_boundary_edges(edges, tree)
return tree  # if the caller needs it for modular_pipelines=
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla, line 242: Side note on naming in _build_dataset_node: stripped_name and original_name took me a moment to parse. These are really two views of the same dataset. The base name for the graph node (with any @pandas suffix removed) and the name as it appears on the node for catalog lookup.

Renaming to something like base_name and referenced_name would help, and a short docstring with one transcoded example would make the intent obvious to someone reading this cold. Worth aligning the loop and _register_dataset naming too so it reads the same throughout.


```python
def _build_dataset_node(
    self,
    base_name: str,
    referenced_name: str,
    *,
    is_transcoded: bool,
) -> DataNodeAPIResponse:
    """Build a data or parameter node for the graph.

    Args:
        base_name: Logical dataset name used for the graph node ID and display
            name. Transcoding suffixes (``@pandas``, etc.) are removed.
        referenced_name: Dataset name as it appears on the pipeline node,
            used to look up the catalog entry. May match ``base_name`` or
            include a transcoding suffix.
        is_transcoded: Whether any node in this pipeline referenced the
            dataset with a transcoding suffix.
    """
```

ravi-kumar-pilla: One line in `_register_dataset` helps too:

```python
# Group transcoded names (sales_data@pandas) under one graph node (sales_data).
# Keep the pipeline reference for catalog lookup.
```

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Good catch. `_free_io()` is now calculated once per namespace for each pipeline view, and the same boundary data is reused for dataset ownership and the tree.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Updated. `from_registered_pipelines()` now builds the aggregate state first and returns a fully initialised index. Ownership is still calculated separately for each pipeline before being combined.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated the module docstring with clearer wording and the correct response-list terminology.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Agreed. Renamed both helpers to `_add_modular_pipeline_boundary_edges` and `_remove_cyclic_modular_pipeline_boundary_edges`.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Good point. `_create_task_node_id` accepts five separate fields, while these call sites already have a `NodeSnapshot`. I kept the small wrapper to avoid repeating that extraction, but renamed it to `_create_task_node_id_from_node_snapshot` so its purpose is clear.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated the docstring to explain that the index is built once across all registered pipelines and reused for each rendered response.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Agreed. Renamed it to `_modular_pipeline_index` to match the object’s role and class name.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Agreed. Task membership now comes directly from `node.namespace`, and the unnecessary index method and its direct tests were removed.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, extracted the modular-pipeline tree, nodes, and boundary-edge handling into `_add_modular_pipelines_to_graph()`, keeping `build()` as the high-level flow.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Updated to `_build_modular_pipeline_tree_response` to clarify that it converts the internal tree into the API response.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated. I renamed `stripped_name` and `original_name` to `base_name` and `referenced_name` throughout, and added a short docstring explaining transcoded names and catalog lookup.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Good point. I renamed it to `_create_task_node_id_from_node_snapshot` so it is clear that the helper accepts a `NodeSnapshot`, rather than a live Kedro Node.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: done

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated. The `build()` docstring now lists the response contents and clarifies the selected-pipeline versus project-wide behavior.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla: The doc string still uses lot of `and`, modular-pipeline hyphens. Better than before but I feel it can be further improved. I will leave it to you and the team 👍

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla: ` pipeline membership ` , I think we thought of avoiding membership terminology.

ravi-kumar-pilla: Small follow-up: line 162 still says "pipeline membership". Something like "tags and registered pipelines on each node are collected across the whole project" would match what we discussed earlier.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,379 @@
...
+
+    return (
+        modular_pipeline_ids,
+        datasets_by_modular_pipeline,
+        boundary_io_by_modular_pipeline,
+    )
+
+
+class _ModularPipelineIndex:
+    """Resolve which modular pipelines each task and dataset belongs to."""
```
ravi-kumar-pilla: Small follow-up: this docstring still mentions tasks, but the index is dataset-only now that task namespaces are set inline in `_build_task_node`. Something like "Look up which modular pipelines own each dataset" would match the current behaviour.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -216,44 +232,93 @@ def _build_task_node(self, node: NodeSnapshot, task_id: str) -> TaskNodeAPIRespo
...
+        nodes.extend(
+            self._build_modular_pipeline_nodes(tree_builder, selected_pipeline)
+        )
+        _add_modular_pipeline_boundary_edges(edges, tree)
+        _remove_cyclic_modular_pipeline_boundary_edges(edges, tree)
+        return tree
+
+    @staticmethod
+    def _build_modular_pipeline_nodes(
+        tree_builder: _ModularPipelineTreeBuilder, selected: str
```
ravi-kumar-pilla: `_add_modular_pipelines_to_graph` already uses `selected_pipeline`, but this helper still takes `selected`. Renaming to `selected_pipeline_id` here too would keep the naming consistent.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla: Concrete suggestion if helpful:

```python
"""Build the main Kedro Viz graph response from an inspection snapshot.

Builds task, data, parameter and modular pipeline nodes for one selected pipeline,
along with their edges and the modular pipeline tree. Also includes the global tag,
layer and registered pipeline lists.
"""
```

Splits it into two sentences and drops the repeated "and" list.

package/tests/test_inspection_adapter/test_graph_builder.py
```diff
@@ -174,3 +172,105 @@ def test_layers_list_matches_baseline(builder: GraphBuilder, pipeline_id: str) -
...
+        (node["type"], node.get("full_name", node["name"])): node.get(
+            "modular_pipelines"
+        )
+        for node in graph["nodes"]
+        if node["type"] in {"task", "data", "parameters"}
+    }
+
+
+@pytest.mark.parametrize("pipeline_id", ALL_PIPELINES)
+def test_modular_pipeline_membership_matches_baseline(
```
ravi-kumar-pilla: There is still quite a bit of "membership" wording across this PR, in tests and the `build()` docstring. We avoided that term elsewhere, so it would be good to align these too.

The API field `modular_pipelines` should stay as-is. Everything else could use clearer language:

**`build()` docstring (line 162):**
"tags and registered pipelines on each node are collected across the whole project"

**Test names:**
- `test_modular_pipeline_membership_matches_baseline` → `test_node_modular_pipelines_match_baseline`
- `test_same_named_tasks_in_two_pipelines_keep_their_modular_membership` → `...keep_dataset_modular_pipelines`
- `test_same_id_tasks_in_different_namespaces_keep_dataset_membership` → `...keep_dataset_modular_pipelines`
- `test_dataset_membership_unions_boundaries_from_each_pipeline` → `test_dataset_modular_pipelines_union_boundaries_from_each_pipeline`

**Test helpers and variables:**
- `_membership()` → `_modular_pipeline_index()`
- local `membership = {...}` → `modular_pipelines_by_dataset`

**Test docstrings:** prefer "each dataset's `modular_pipelines` list" or "which namespaces own each dataset" instead of "membership".

Using `modular_pipelines` consistently (matching the API field) would read more naturally throughout.

jitu5: Thanks for the detailed feedback, updated. The test names, helpers, variables and docstrings now use modular_pipelines or dataset ownership instead of “membership”. The API field and behaviour are unchanged.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Agreed. I renamed the internal variables and parameters to selected_pipeline_id consistently. The API field remains selected_pipeline.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Good catch. The docstring now describes the index as dataset-only.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated. I split the description into shorter sentences and used “modular pipeline” and “registered pipeline” consistently.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks, updated. The docstring now refers directly to tags and registered pipelines on each node.

SajidAlamQB (approved):
LGTM, thank you @jitu5!

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -216,44 +234,93 @@ def _build_task_node(self, node: NodeSnapshot, task_id: str) -> TaskNodeAPIRespo
...
+        )
+        _add_modular_pipeline_boundary_edges(edges, tree)
+        _remove_cyclic_modular_pipeline_boundary_edges(edges, tree)
+        return tree
+
+    @staticmethod
+    def _build_modular_pipeline_nodes(
+        tree_builder: _ModularPipelineTreeBuilder, selected_pipeline_id: str
+    ) -> list[DataNodeAPIResponse]:
+        """Build one ``modularPipeline`` node per modular pipeline in the rendered view."""
```
ravi-kumar-pilla: The method name and return type does not match for readability. This is the same issue in our previous implementation as well. Either we can update the docstring as below and rename method name to - `_build_modular_pipeline_group_nodes`

"""Build flowchart group nodes for each modular pipeline namespace.

Returns ``DataNodeAPIResponse`` objects with ``type="modularPipeline"``.
The API reuses the data-node model for all non-task graph nodes.
"""

OR

a better long term solution could be to have a class for `ModularPipelineNodeAPIResponse`. 

The former is a low effort solution. wdyt @jitu5

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,379 @@
+"""Build modular pipelines, their tree and their edges from snapshot namespaces.
```
ravi-kumar-pilla: Since the old backend will be removed after this lands, I think we should drop the "legacy backend" framing from docs and tests. A few spots still use it:

**`modular_pipelines.py`**
- Module docstring (here): replace with something like:
  ```
  Build modular pipeline groups, their tree and boundary connections from snapshot namespaces.

  Derives namespace ownership, boundary inputs and outputs, and tree structure from
  NodeSnapshot data using Kedro's modular pipeline set algebra.
  ```
- `_free_io` docstring: "Follows Kedro modular pipeline set algebra" instead of "legacy backend"
- `_remove_intermediates` reference to `Pipeline._remove_intermediates` is fine to keep

**`test_graph_builder.py`**
- `_edge_keys`: "Node IDs match the established Viz format" instead of "legacy-compatible"
- `test_node_modular_pipelines_match_baseline`: "same modular pipelines as the baseline"
- `test_modular_tree_children_match_baseline`: "The baseline lists a parameter..." and consider renaming `drop_legacy_duplicate` to `drop_baseline_duplicate`

**`test_modular_pipelines.py`**
- `test_transcoded_boundary_matches_legacy_set_algebra` → `test_transcoded_boundary_follows_pipeline_set_algebra`

"Baseline" as a test fixture name is fine. The issue is wording that treats the old backend as the reference implementation going forward.

ravi-kumar-pilla: This file carries a lot of responsibility and I had to read it twice before the pieces clicked. I think a short concepts block at the top would help a lot. Something that names the four ideas upfront: namespace, boundary I/O, tree, and boundary edges. Then the set algebra below has context to land in.

Also still worth dropping the legacy backend wording here once the old path is gone.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,379 @@
...
+    counts as intermediate, and the surviving names are returned in their original
+    (possibly transcoded) form so later set operations compare like for like.
+    """
+    intermediate = {_strip_transcoding(i) for i in all_inputs} & {
+        _strip_transcoding(o) for o in all_outputs
+    }
+    return {d for d in datasets if _strip_transcoding(d) not in intermediate}
+
+
+def _free_io(nodes: list[NodeSnapshot], mp_id: str) -> tuple[set[str], set[str]]:
```
ravi-kumar-pilla: This feels like the core of the file and I think it sits in the right place. The formula in the docstring helped me follow along. One small addition that might help: a pointer to Kedro's `Pipeline.inputs()` and `Pipeline.outputs()` for anyone trying to verify the behaviour against the source.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,379 @@
...
+    rest_outputs = {o for node in rest for o in node.outputs}
+
+    free_inputs = _remove_intermediates(sub_inputs, sub_inputs, sub_outputs)
+    free_outputs = _remove_intermediates(sub_outputs, sub_inputs, sub_outputs) | (
+        _remove_intermediates(rest_inputs, rest_inputs, rest_outputs) & sub_outputs
+    )
+    return free_inputs, free_outputs
+
+
+def _build_modular_pipeline_data(
```
ravi-kumar-pilla: I kept coming back to this name and expecting API objects. It is really computing boundaries and ownership. Something like `_compute_namespace_boundaries` might read clearer.

Same area: the function returns a three tuple and callers unpack by position. A small frozen dataclass (something like `_NamespaceBoundaries`) would make that harder to get wrong as this grows.

```python
@dataclass(frozen=True)
class _NamespaceBoundaries:
    modular_pipeline_ids: set[str]
    datasets_by_modular_pipeline: dict[str, set[str]]
    boundary_io_by_modular_pipeline: _BoundaryIOByModularPipeline
```

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,379 @@
...
+        # modular owner *in this pipeline*, which can differ from its owners across the project.
+        ids, datasets, boundary_io = _build_modular_pipeline_data(nodes)
+        self._boundary_io_by_modular_pipeline = boundary_io
+        self._index = _ModularPipelineIndex(
+            modular_pipeline_ids=ids,
+            datasets_by_modular_pipeline=datasets,
+        )
+        self.ids = sorted(ids)
+
+    def build(self) -> dict[str, _ModularPipelineTreeEntry]:
```
ravi-kumar-pilla: The `params` set threads through a few methods here and it took me a bit to see why. A short note at the top of `build()` explaining that parameters are kept out of internal dataset children would save the next reader some jumping around.

The per mp_id loop also does three distinct steps. Extracting something like `_populate_entry` might make the flow easier to scan.

ravi-kumar-pilla (commented):
Nice work on this. The modular pipeline behaviour matches what we expect and the tests cover the tricky cases well (ancestor boundaries, transcoding, cycles).

The main feedback is about readability in `modular_pipelines.py`. The set algebra is inherently dense but it is mostly well isolated. A concepts docstring at the top, clearer naming for `_build_modular_pipeline_data`, and a worked example on the cycle removal would go a long way for the next person.

I left inline notes on those spots. Nothing blocking from my side. Happy to resolve threads once addressed.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,379 @@
...
+                GraphEdgeAPIResponse(source=input_id, target=mp_id),
+            )
+        for output_id in sorted(entry.outputs):
+            edges.setdefault(
+                (mp_id, output_id),
+                GraphEdgeAPIResponse(source=mp_id, target=output_id),
+            )
+
+
+def _remove_cyclic_modular_pipeline_boundary_edges(
```
ravi-kumar-pilla: This was the hardest part for me to follow on first read. The test docstring has a really clear worked example (the case where `a` feeds into `ns` but is also produced downstream). Worth lifting that into the function docstring so you do not need to open the test file to get the intent.

```
"""Drop input→mp edges when the input is reachable downstream of mp.

Example: ns.inner produces ``a`` from ``b``; an outer task consumes ``b`` and
produces ``a``. ``a`` is a boundary input to ``ns``, but also reachable via
ns → b → outer → a, so the inbound edge would cycle the graph.
"""
```

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -22,9 +27,16 @@
     _extract_layers,
     sort_layers,
 )
+from kedro_viz.integrations.kedro.inspection.modular_pipelines import (
```
ravi-kumar-pilla, line 28: Not urgent, but we pull in quite a few private symbols from modular_pipelines here. Longer term a small facade class on that side might keep graph_builder from needing to know about the internal types. Fine to leave as a follow up.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -216,44 +234,93 @@ def _build_task_node(self, node: NodeSnapshot, task_id: str) -> TaskNodeAPIRespo
...
+        tree = tree_builder.build()
+        nodes.extend(
+            self._build_modular_pipeline_nodes(tree_builder, selected_pipeline_id)
+        )
+        _add_modular_pipeline_boundary_edges(edges, tree)
+        _remove_cyclic_modular_pipeline_boundary_edges(edges, tree)
+        return tree
+
+    @staticmethod
+    def _build_modular_pipeline_nodes(
```
ravi-kumar-pilla: Returning `DataNodeAPIResponse` with type `modularPipeline` works but reads a bit odd on first pass. A one line comment noting that Viz has no separate response model for modular pipelines would help the next person.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
ravi-kumar-pilla: Example:

"""Build modular pipeline groups, tree, and boundary edges from snapshot namespaces.

Concepts:
- Namespace: dotted path like ``data_eng.cleaning``; each prefix is a parent folder.
- Boundary I/O: datasets that enter or leave a namespace subtree (Kedro set algebra).
- Tree: Viz folder hierarchy; children are tasks, internal datasets, nested namespaces.
- Boundary edges: flowchart links between boundary datasets and group nodes.
"""

ravi-kumar-pilla:
Hi @jitu5 ,

This is definitely the most complex part of the refactor. Thanks for being patient and addressing the review comments. Lets take some time on this to get this right as modular pipelines has been an issue in earlier implementation too. Below are some suggestions that I felt worth investing time in restructuring this PR -

## Summary of suggested changes

Three changes before merge. No intended behaviour change, structural only.

| # | Change | Why |
|---|--------|-----|
| 1 | Extract `modular_pipeline_algebra.py` | Separates Kedro pipeline math from Viz rendering. Keeps the door open to delete this file later. |
| 2 | Add `ModularPipelineIndex` and `ModularPipelineView` | Two public classes match two lifecycles. `graph_builder.py` imports two names instead of five private symbols. |
| 3 | Docstring cleanup | Drop legacy backend wording. Add a worked example on cycle removal. |

---

The problem is structure. One file does three jobs:

1. **Kedro set algebra** on flat `NodeSnapshot` lists (boundary inputs/outputs per namespace)
2. **Project wide lookup** (which modular pipelines own each dataset)
3. **Per pipeline rendering** (tree, group nodes, boundary edges, cycle cleanup)

Meanwhile `graph_builder.py` imports five private symbols and orchestrates steps that belong inside the modular pipeline layer.

---

## Change 1: Extract `modular_pipeline_algebra.py`

The hardest part of this file is not Viz rendering. It is Kedro pipeline set algebra reimplemented on snapshots. Can we pull that into `modular_pipeline_algebra.py`?

Benefits:

* Reader sees the boundary clearly: algebra first, Viz rendering second
* No Viz API imports in the algebra file, easier to unit test
* Makes the duplication visible and easy to remove when Kedro precomputes boundaries on the snapshot

Move `_ancestor_namespaces`, `_in_subtree`, `_modular_pipeline_ids`, `_remove_intermediates`, `_free_io`, and rename `_build_modular_pipeline_data` to `compute_namespace_boundaries`. Replace the three tuple return with a `_NamespaceBoundaries` dataclass.

This file might not need to exist long term. When `kedro inspect` runs, Kedro still has live `Pipeline` objects. `_build_pipeline_snapshots` could compute per namespace boundaries using `only_nodes_with_namespaces()` and store them on `PipelineSnapshot`. Viz would just read them. See the Kedro follow up section at the end of this doc.

### New file: `modular_pipeline_algebra.py`

**Module docstring (suggested):**

```python
"""Kedro modular pipeline set algebra on inspection snapshots.

Kedro defines namespace boundaries via Pipeline.inputs() and Pipeline.outputs().
The inspection API exposes flat NodeSnapshot lists, not live Pipeline objects,
so this module reimplements that logic here.

See kedro.pipeline.pipeline for the source of truth.
"""
```

**Move these symbols from `modular_pipelines.py`:**

| Symbol | Notes |
|--------|-------|
| `_ancestor_namespaces()` | `"a.b.c"` → `["a", "a.b", "a.b.c"]` |
| `_in_subtree()` | Whether a node is under a namespace |
| `_modular_pipeline_ids()` | All namespace IDs including ancestors |
| `_remove_intermediates()` | Mirrors `Pipeline._remove_intermediates` |
| `_free_io()` | Boundary inputs/outputs for one namespace subtree |
| `compute_namespace_boundaries()` | Renamed from `_build_modular_pipeline_data` |

**Add dataclass (replace positional tuple unpacking):**

```python
@dataclass(frozen=True)
class _NamespaceBoundaries:
    modular_pipeline_ids: set[str]
    datasets_by_modular_pipeline: dict[str, set[str]]
    boundary_io_by_modular_pipeline: dict[str, tuple[set[str], set[str]]]


def compute_namespace_boundaries(
    nodes: list[NodeSnapshot],
) -> _NamespaceBoundaries:
    """Compute namespace IDs, dataset ownership, and boundary I/O for one node list."""
    ...
```

**Dependencies:** only `NodeSnapshot`, `_strip_transcoding`, and stdlib. No Viz API response models.

**Transcoding rule (must not change):** strip `@pandas` style suffixes only when detecting intermediates. Keep original dataset names for set membership everywhere else.

### Tests to move

From `test_modular_pipelines.py` into `test_modular_pipeline_algebra.py`:

* `test_ancestor_namespaces_expands_every_level`
* `test_transcoded_boundary_matches_legacy_set_algebra` (rename to `test_transcoded_boundary_follows_pipeline_set_algebra`)
* Any other tests that assert boundary I/O or intermediate removal directly

Keep tree, edge, and index ownership tests in `test_modular_pipelines.py`.

### Implementation steps

1. Create `modular_pipeline_algebra.py`, move functions unchanged
2. Add `_NamespaceBoundaries`, update callers to use attribute access instead of tuple unpacking
3. Update imports in `modular_pipelines.py`
4. Create `test_modular_pipeline_algebra.py`, move algebra tests
5. Run `pytest package/tests/test_inspection_adapter/`. All tests pass, same assertions

---

## Change 2: Two public classes

`graph_builder.py` imports five private symbols from modular_pipelines. GraphBuilder should not orchestrate tree building, boundary edges, and cycle cleanup.

There are two lifecycles tangled together:

* **Project scope** (built once in `__init__`): which modular pipelines own each dataset
* **Pipeline scope** (built on every `build()`): tree, group nodes, boundary edges

Split into two public classes: `ModularPipelineIndex` and `ModularPipelineView`. After this, graph_builder imports two names.

### Before and after: `graph_builder.py` imports

**Before:**

```python
from kedro_viz.integrations.kedro.inspection.modular_pipelines import (
    _add_modular_pipeline_boundary_edges,
    _ModularPipelineIndex,
    _ModularPipelineTreeBuilder,
    _ModularPipelineTreeEntry,
    _remove_cyclic_modular_pipeline_boundary_edges,
)
```

**After:**

```python
from kedro_viz.integrations.kedro.inspection.modular_pipelines import (
    ModularPipelineIndex,
    ModularPipelineView,
)
```

### Class 1: `ModularPipelineIndex` (project scope)

Rename and promote `_ModularPipelineIndex`. Built once from all registered pipelines.

**Suggested API:**

```python
class ModularPipelineIndex:
    @classmethod
    def from_registered_pipelines(
        cls, pipelines: Iterable[PipelineSnapshot]
    ) -> ModularPipelineIndex:
        """Union ownership calculated independently per registered pipeline."""

    @classmethod
    def from_nodes(cls, nodes: list[NodeSnapshot]) -> ModularPipelineIndex:
        """Build index for one pipeline's nodes. Keep for tests."""

    def owners_for_dataset(self, name: str) -> list[str] | None:
        """Sorted modular pipeline IDs that own this dataset, or None."""
```

**Changes from today:**

| Today | After |
|-------|-------|
| `_ModularPipelineIndex` | `ModularPipelineIndex` (public) |
| `get_modular_pipelines_for_dataset()` | `owners_for_dataset()` |
| calls `_build_modular_pipeline_data()` | calls `compute_namespace_boundaries()` from algebra module |

**`graph_builder.py` call sites:**

```python
# GraphBuilder.__init__
self._modular_pipeline_index = ModularPipelineIndex.from_registered_pipelines(
    snapshot.pipelines
)

# GraphBuilder._build_dataset_node
modular_pipelines=self._modular_pipeline_index.owners_for_dataset(base_name)
```

Task nodes are unchanged. They still get `modular_pipelines=[node.namespace]` directly in `_build_task_node`.

### Class 2: `ModularPipelineView` (pipeline scope)

New public class. Wraps everything `GraphBuilder` currently does in `_add_modular_pipelines_to_graph`.

**Suggested API:**

```python
class ModularPipelineView:
    def __init__(self, nodes: list[NodeSnapshot]) -> None:
        """Prepare tree and boundary data for one pipeline view."""

    def extend_graph(
        self,
        nodes: list[TaskNodeAPIResponse | DataNodeAPIResponse],
        edges: dict[tuple[str, str], GraphEdgeAPIResponse],
        selected_pipeline_id: str,
    ) -> dict[str, ModularPipelinesTreeNodeAPIResponse]:
        """Append group nodes, wire boundary edges, return API tree."""
```

**What `extend_graph` does internally (all private):**

| Step | Today (in GraphBuilder or modular_pipelines) |
|------|---------------------------------------------|
| 1 | `_ModularPipelineTreeBuilder.build()` |
| 2 | `_build_modular_pipeline_nodes()` → append group nodes |
| 3 | `_add_modular_pipeline_boundary_edges()` |
| 4 | `_remove_cyclic_modular_pipeline_boundary_edges()` |
| 5 | `_build_modular_pipeline_tree_response()` → return API tree |

**Private symbols that stay inside `modular_pipelines.py`:**

* `_ModularPipelineTreeEntry`
* `_ModularPipelineTreeBuilder`
* `_add_modular_pipeline_boundary_edges`
* `_remove_cyclic_modular_pipeline_boundary_edges`
* `_build_modular_pipeline_tree_response` (private helper called by `extend_graph`)

**`graph_builder.py` call site:**

```python
# GraphBuilder.build() — replace _add_modular_pipelines_to_graph
modular_pipelines_tree = ModularPipelineView(pipeline.nodes).extend_graph(
    nodes,
    edges,
    selected_pipeline_id,
)

return GraphAPIResponse(
    ...
    modular_pipelines=modular_pipelines_tree,
    ...
)
```

**Delete from `GraphBuilder`:**

* `_add_modular_pipelines_to_graph`
* `_build_modular_pipeline_nodes`
* `_build_modular_pipeline_tree_response`

### `modular_pipelines.py` module docstring (suggested)

```python
"""Build modular pipeline groups, tree, and boundary connections for Viz.

Public entry points:
  ModularPipelineIndex — project-wide dataset ownership (built once)
  ModularPipelineView — per-pipeline tree, group nodes, and edges (built per render)

Set algebra lives in modular_pipeline_algebra.py.
"""
```

### Implementation steps

1. Rename `_ModularPipelineIndex` → `ModularPipelineIndex`
2. Rename `get_modular_pipelines_for_dataset` → `owners_for_dataset`
3. Create `ModularPipelineView` with `extend_graph`
4. Move `_build_modular_pipeline_nodes` and `_build_modular_pipeline_tree_response` logic into `ModularPipelineView`
5. Update `GraphBuilder.build()`, remove three methods from `GraphBuilder`
6. Update test helpers in `test_modular_pipelines.py` (`_modular_pipeline_index` helper imports public class)
7. Run baseline tests in `test_graph_builder.py`

---

## Change 3: Docstring and naming cleanup

Small fixes while restructuring. Can land in the same PR.

| Location | Suggested change |
|----------|-----------------|
| `modular_pipeline_algebra.py` module docstring | Explain reimplementation, link to `kedro.pipeline.pipeline` |
| `modular_pipelines.py` module docstring | Drop "legacy backend" wording |
| `_free_io` docstring | "Follows Kedro modular pipeline set algebra" instead of "legacy backend" |
| `_ModularPipelineTreeBuilder.build()` | Short note explaining why `params` are excluded from internal dataset children |
| `_remove_cyclic_modular_pipeline_boundary_edges` | Lift worked example from `test_cyclic_modular_edge_is_removed` into docstring |
| `test_graph_builder.py` | Replace "legacy backend" / "legacy-compatible" with "baseline" |
| PR description | Update dev/QA notes that mention legacy backend |

---

## Data flow after refactor

```
ProjectSnapshot
       │
       ├─► ModularPipelineIndex.from_registered_pipelines()
       │         │
       │         └─► owners_for_dataset()  ──► dataset node.modular_pipelines
       │
       └─► GraphBuilder.build(selected_pipeline)
                 │
                 └─► ModularPipelineView(pipeline.nodes).extend_graph()
                           │
                           ├─► compute_namespace_boundaries()     [algebra module]
                           ├─► _ModularPipelineTreeBuilder.build() [private]
                           ├─► append modularPipeline group nodes
                           ├─► boundary edges + cycle cleanup
                           └─► return ModularPipelinesTreeNodeAPIResponse dict
```

---

## Kedro follow up: precompute boundaries on the snapshot (TBD)

This is the path to removing `modular_pipeline_algebra.py` from Viz entirely. Track as a separate Kedro issue.

### Why

When `kedro inspect` runs, Kedro has live `Pipeline` objects. `_build_pipeline_snapshots` in `kedro/inspection/snapshot.py` already calls `pipeline.inputs()` and `pipeline.outputs()` for the top level. The same pass can compute per namespace boundaries and store them on `PipelineSnapshot`.

Viz then reads precomputed data. No reimplementation, no drift risk.

### Suggested schema addition (Kedro)

```python
@dataclass
class ModularPipelineBoundarySnapshot:
    """Boundary I/O for one modular pipeline namespace."""
    namespace: str
    inputs: list[str]   # free inputs crossing into the namespace
    outputs: list[str]  # free outputs crossing out of the namespace


@dataclass
class PipelineSnapshot:
    name: str
    nodes: list[NodeSnapshot]
    inputs: list[str]
    outputs: list[str]
    modular_pipeline_boundaries: list[ModularPipelineBoundarySnapshot] = field(default_factory=list)
```

### Suggested computation (Kedro `snapshot.py`)

```python
def _compute_modular_pipeline_boundaries(pipeline: Pipeline) -> list[ModularPipelineBoundarySnapshot]:
    namespace_ids = {
        mp_id
        for node in pipeline.nodes
        if node.namespace
        for mp_id in _ancestor_namespaces(node.namespace)
    }
    boundaries = []
    for mp_id in sorted(namespace_ids):
        sub = pipeline.only_nodes_with_namespaces([mp_id])
        sub_node_names = {n.name for n in sub.nodes}
        rest_nodes = [n for n in pipeline.nodes if n.name not in sub_node_names]
        rest = Pipeline(rest_nodes)

        free_inputs = sub.inputs()
        free_outputs = sub.outputs() | (rest.inputs() & sub.all_outputs())

        boundaries.append(
            ModularPipelineBoundarySnapshot(
                namespace=mp_id,
                inputs=sorted(free_inputs),
                outputs=sorted(free_outputs),
            )
        )
    return boundaries
```

Exact rest pipeline construction may need refinement, but the key point is Kedro has the live object at inspect time.

### What changes in Viz after that lands

* Delete `modular_pipeline_algebra.py`
* `ModularPipelineIndex` reads ownership from snapshot boundaries instead of recomputing
* `ModularPipelineView` reads boundary I/O from snapshot instead of calling `_free_io`
* Tree building and edge wiring stay in Viz (that is genuine rendering concern)

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Agreed. I used the smaller-scoped option: the helper is now `_build_modular_pipeline_group_nodes`, and its docstring explains that modular-pipeline groups reuse `DataNodeAPIResponse` with `type="modularPipeline"`.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Agreed. I updated every location listed here to use Kedro set algebra, established Viz format, or baseline terminology. The behavior and test assertions are unchanged.

jitu5: Thanks, added a short concepts block covering namespaces, boundary I/O, the tree, and boundary edges. It also identifies the project-wide Index and per-render View entry points.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Added. The boundary-algebra docstring now points to `Pipeline.inputs()`, `Pipeline.outputs()`, `Pipeline.all_outputs()`, and `kedro.pipeline.pipeline` as the source of truth.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Agreed. The calculation is now `compute_namespace_boundaries()` in `modular_pipeline_algebra.py`, and it returns a frozen `_NamespaceBoundaries` dataclass instead of a positional tuple.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Updated. `build()` now explains why parameters remain under `__root__` instead of becoming internal data children, and the per-namespace work is extracted into `_populate_entry()`.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Added. The cycle-removal docstring now includes the worked example and explains why the inbound dataset-to-group edge must be removed.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Implemented as part of the restructuring. `GraphBuilder` now imports only `ModularPipelineIndex` and `ModularPipelineView`; tree construction, group nodes, boundary edges, cycle cleanup, and response conversion stay inside the modular-pipeline module.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Added to `_build_modular_pipeline_group_nodes()`. It now explains that the API has no separate modular-pipeline node model, so group nodes reuse `DataNodeAPIResponse`.

jitu5:
> suggested change 1, 2 and 3

Thanks a lot for the detailed restructuring proposal. All three changes are now applied:

**1. Algebra extraction:** The namespace set algebra now lives in `modular_pipeline_algebra.py`, returns a frozen `_NamespaceBoundaries`, and has focused algebra tests. It has no Viz response-model dependencies; the existing root-ID constant is retained only for reserved-namespace validation.

**2. Index and View:** `ModularPipelineIndex` handles project-wide dataset ownership, while `ModularPipelineView` handles the selected pipeline’s tree, group nodes, and boundary edges. `GraphBuilder` now depends only on those two classes.

**3. Documentation:** The requested concepts, parameter explanation, cycle example, baseline terminology, and PR-description updates are in place.

The proposed Kedro-side snapshot boundary work remains a separate follow-up.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,375 @@
...
+            boundaries = compute_namespace_boundaries(pipeline.nodes)
+            modular_pipeline_ids.update(boundaries.modular_pipeline_ids)
+            for mp_id, datasets in boundaries.datasets_by_modular_pipeline.items():
+                datasets_by_modular_pipeline.setdefault(mp_id, set()).update(datasets)
+        return cls(
+            modular_pipeline_ids=modular_pipeline_ids,
+            datasets_by_modular_pipeline=datasets_by_modular_pipeline,
+        )
+
+    def owners_for_dataset(self, name: str) -> list[str] | None:
```
ravi-kumar-pilla:
```suggestion
    def modular_pipelines_for_dataset(self, name: str) -> list[str] | None:
```

ravi-kumar-pilla: or `get_modular_pipelines_for_dataset` . `owners_for_dataset` seem generic

ravi-kumar-pilla: Small naming follow up on `ModularPipelineIndex`.

`owners_for_dataset` feels a bit generic in the modular pipeline context. It also does not line up with the API field we are populating, which is `modular_pipelines`.

Could we rename to **`modular_pipelines_for_dataset`**? The call site would read:

```python
modular_pipelines=self._modular_pipeline_index.modular_pipelines_for_dataset(base_name)
```

Same suggestion for the local variable `owners` inside that method. Something like `modular_pipeline_ids` would be clearer.

While you are in there, worth sweeping the other "owner" wording in this PR so it stays consistent:

**Method and call sites**
* `owners_for_dataset` → `modular_pipelines_for_dataset` in `modular_pipelines.py`, `graph_builder.py`, and tests

**Docstrings and comments**
* "dataset ownership" → "which modular pipelines a dataset belongs to" (in `modular_pipeline_algebra.py` and module docstrings)
* "Union ownership calculated" → "Union calculated independently per registered pipeline"
* "Parameter references have no owners" → "Parameters are not assigned to any modular pipeline"
* "modular owner in this pipeline" → "modular pipeline assignment in this pipeline view"
* "unowned datasets" → "datasets not assigned to any modular pipeline"

**Tests**
* `test_dataset_owners_agree_with_the_tree_for_an_ancestor_boundary` → `test_dataset_modular_pipelines_agree_with_the_tree_for_an_ancestor_boundary`
* `test_owners_for_dataset_accepts_a_transcoded_name` → `test_modular_pipelines_for_dataset_accepts_a_transcoded_name`

About 20 lines across 4 files. Behaviour unchanged, just naming.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
```diff
@@ -0,0 +1,375 @@
+"""Build modular pipeline groups, tree, and boundary connections for Viz.
```
ravi-kumar-pilla: `modular_pipelines.py` is much better after the algebra extraction and the Index/View split, but at ~375 lines it still carries four jobs: index lookup, tree building, flowchart rendering, and boundary edges.

Would it help to turn this into a package? Something like:

```
integrations/kedro/inspection/
  snapshot_source.py          # load snapshot
  graph_builder.py            # orchestrates full GraphAPIResponse
  layers.py
  modular_pipelines/          # package, not a UI layer
    __init__.py               # exports ModularPipelineIndex, ModularPipelineView
    index.py                  # ModularPipelineIndex
    tree.py                   # tree entry, tree builder, tree response
    view.py                   # ModularPipelineView, boundary edges, cycle cleanup
    boundaries.py             # namespace boundary calculation (private, not exported)
```

Each file answers one question:

* `index.py` → which modular pipelines does this dataset belong to?
* `tree.py` → how is the folder hierarchy built?
* `view.py` → how do group nodes and boundary edges get wired into the graph?
* `boundaries.py` → what are the free inputs/outputs per namespace?

Move the current `modular_pipeline_algebra.py` content into `boundaries.py`. I prefer that name over `algebra`. It matches the Boundary I/O language in the concepts docstring and reads clearer.

`graph_builder.py` stays the single entry point for the adapter. It keeps importing the same two public names:

```python
from kedro_viz.integrations.kedro.inspection.modular_pipelines import (
    ModularPipelineIndex,
    ModularPipelineView,
)
```

Dependency order (inner modules first):

```
boundaries.py   (no internal imports)
     ↓
index.py        imports boundaries
     ↓
tree.py         imports boundaries, index
     ↓
view.py         imports tree
     ↓
__init__.py     re-exports index, view
```

When Kedro precomputes boundaries on the snapshot later, delete `boundaries.py` and read from the snapshot in `index.py` and `tree.py`. Public API stays the same.

Not blocking for merge. Good follow up once the naming sweep lands.

ravi-kumar-pilla:
Hi @jitu5 , 

I think the PR is almost ready. The split made lot of concepts easy to follow. I left few comments around structuring the whole modular pipelines code which will help us maintain in long term. There are no behavioral changes but organizing them so the debugging becomes easy in future. Thanks for resolving the suggestions patiently.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Thanks, updated. The dataset lookup is now `modular_pipelines_for_dataset`, and the related wording has been cleaned up.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines.py
jitu5: Thanks, updated. I split the modular-pipeline code into boundaries, index, tree, and view, while keeping `ModularPipelineIndex` and `ModularPipelineView` as the public entry points. I thought lets keep this changes in this PR as its really good structural improvement. https://github.com/kedro-org/kedro-viz/pull/2726/changes/30a021baf9c84152e4c3440900069d9a02bd2051

package/tests/test_inspection_adapter/test_modular_pipelines.py
```diff
@@ -0,0 +1,275 @@
+"""Hermetic tests for modular pipelines, their tree and their edges."""
```
ravi-kumar-pilla: The production code split into the package nicely, but tests only partially followed. `test_modular_pipeline_boundaries.py` covers boundaries, but this file still mixes index, tree, view and edge tests (~275 lines).

Could we mirror the package layout?

```
tests/test_inspection_adapter/modular_pipelines/
  test_boundaries.py    # move/rename existing boundaries tests
  test_index.py
  test_tree.py
  test_view.py
  conftest.py           # shared _node, _tree_builder helpers
```

Rough mapping from this file:

* **index** → dataset belongs to enclosing pipelines, parameters, unassigned datasets, transcoded names
* **tree** → tree IDs, free I/O, internal datasets, nesting, root children, tags, reserved namespace
* **view** → boundary edges, cycle removal, stable ordering, `extend_graph` integration

Same assertions, just split so a tree failure opens `test_tree.py`. Good follow up while the package split is fresh.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines/tree.py
```diff
@@ -0,0 +1,191 @@
...
+            if node.namespace is None:
+                root.children.add(
+                    (
+                        _create_task_node_id_from_node_snapshot(node),
+                        GraphNodeType.TASK.value,
+                    )
+                )
+
+
+def _build_modular_pipeline_tree_response(
```
ravi-kumar-pilla, line 175: [Nit] `_build_modular_pipeline_tree_response` as a module function makes sense to me. It is tree → API serialization and `tree.py` owns both the internal tree and its API representation. `ModularPipelineView` should orchestrate, not serialize.

Two optional tidy ups, neither blocking:

1. Move this onto `_ModularPipelineTreeBuilder` as something like `to_api_response(tree)` if you prefer it inside a class.
2. Prefix `compute_namespace_boundaries` with `_` in `boundaries.py` if nothing outside the package should call it directly.

Current layout is reasonable after the package split.

package/tests/test_inspection_adapter/test_modular_pipelines.py
ravi-kumar-pilla: **Which tests go where**

**test_index.py**

test_dataset_belongs_to_every_enclosing_modular_pipeline
test_parameters_never_belong_to_a_modular_pipeline
test_unassigned_dataset_has_no_modular_pipelines
test_modular_pipelines_for_dataset_accepts_a_transcoded_name

**test_tree.py**

test_tree_ids_include_every_ancestor_namespace
test_tree_exposes_free_inputs_and_outputs
test_internal_dataset_is_not_a_boundary
test_nested_pipeline_is_a_child_of_its_parent
test_root_holds_nodes_without_a_namespace
test_tree_without_namespaces_has_only_root
test_source_and_sink_nodes_are_modular_pipeline_children
test_modular_node_tags_union_the_whole_subtree
test_namespace_using_reserved_root_is_rejected
test_dataset_modular_pipelines_agree_with_the_tree_for_ancestor_boundary (index + tree integration, fits tree or a small test_integration.py)

**test_view.py**

test_modular_edges_connect_boundary_datasets
test_root_has_no_modular_edges
test_cyclic_modular_edge_is_removed
test_acyclic_modular_edges_are_kept
test_modular_edges_are_emitted_in_a_stable_order
test_view_extends_graph_and_returns_tree

Shared helpers (_node, _modular_pipeline_index, _tree_builder) can live in modular_pipelines/conftest.py.

package/kedro_viz/integrations/kedro/inspection/modular_pipelines/tree.py
jitu5: Thanks. I kept the tree serialization as a module function and made `_compute_namespace_boundaries` package-private since it is only used internally.

package/tests/test_inspection_adapter/test_modular_pipelines.py
jitu5: Thanks, updated. The modular-pipeline tests now mirror the production package, with separate boundary, index, tree, and view test modules and shared factories in conftest.py.

package/tests/test_inspection_adapter/modular_pipelines/test_boundaries.py
```diff
@@ -0,0 +1,29 @@
+"""Hermetic tests for modular-pipeline boundary calculations."""
```
ravi-kumar-pilla: [Nit] We do not use Hermetic anywhere in our tests I believe. Lets remove this and follow our doc strings we have in other tests. Same comment for other test files

ravi-kumar-pilla (approved):
Thank you @jitu5 , this looks great. I left a minor comment to fix. Awesome work on this !!

PR #2733, opened 2026-08-11
Group Dependabot security updates

jitu5:
## Summary

- Group Python security updates into a single Dependabot PR.
- Group npm security updates across both npm manifests.
- Preserve the existing monthly Python version-update schedule.

## Validation

- Validated the Dependabot YAML configuration.
- Passed the pre-commit checks.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,8 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
-# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
+# Dependabot configuration for kedro-viz.
+#
+# Routine Python version updates run monthly. Routine npm version updates stay
+# disabled. Security updates are grouped separately by ecosystem to reduce the
+# number of security PRs.
 version: 2
 updates:
   - package-ecosystem: "pip" # See documentation for possible values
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -15,3 +18,24 @@ updates:
...
+        patterns:
+          - "*"
+
+  - package-ecosystem: "npm"
+    directories:
+      - "/"
+      - "/tools/test-lib/react-app"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 0
```
Huongg: should we apply limit here? maybe 5 max every month?

jitu5: We can keep it but we already have lot of noise coming from Dependabot security alerts for npm version bumps. If we set this to 5 then will add more.

ravi-kumar-pilla: This will not give any npm updates unless it is a security issue. I feel we should either run a cron for may be 2 or 3 months cadence instead of monthly for both npm and pypi.

Grouping security updates is great. I would also recommend having version updates atleast on a 3 month cadence (quaterly) even if it is not a security threat. We can always close the PR if it is a patch or minor update.

Thank you

jitu5: Thanks, that makes sense. `open-pull-requests-limit: 0` currently disables regular npm version updates. I’ll change both pip and npm version-update schedules to quarterly and allow up to five open version-update PRs. The existing security-update grouping will remain unchanged.

.github/dependabot.yml
```diff
@@ -1,5 +1,8 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
+# Dependabot configuration for kedro-viz.
+#
+# Routine Python and npm version updates run quarterly.
```
Huongg: i think it will be useful to leave a comment here to briefly explain why we set quarterly with max 5 PRs, just so next time some one need to adjust they know the context

Huongg (approved):
Thank you @jitu5, it looks good to me. Just one small point about the comment, and also could you run the format check against this file to make sure the formatting is consistent?

Since this only changes the Dependabot config and nothing else, I think we can merge it and see how Dependabot behaves. It’s quite hard to fully predict the behaviour from reading the YAML config

SajidAlamQB (approved):

.github/dependabot.yml
jitu5: Added thanks.

PR #2735, opened 2026-08-12
Enhance GitHub Actions workflows with concurrency settings and branch filter

### LINKED ISSUE #2734: Optimise GitHub Actions runs for pull requests and push
## Description

When commits are pushed to a branch with an open pull request, some expensive GitHub Actions workflows run twice: once for `push` and once for `pull_request`. This duplicates the same validation, consumes runner resources, and increases CI waiting time.

## Context

PR checks should provide fast feedback without repeating equivalent work. Reducing duplicate and unnecessary runs would lower CI usage and make pull requests quicker to validate and merge.

## Possible Implementation

- Run feature-branch validation through `pull_request`, and restrict `push` triggers to permanent branches such as `main` and `demo`.
- Add workflow concurrency with `cancel-in-progress: true` so that a new commit cancels checks for an older commit on the same PR.
- Review broad path filters such as `.github/**` and run backend, frontend, and documentation checks only when relevant files change.
- Confirm that repository rules and required checks continue to work when workflows are skipped by path filters.

Example trigger:

```yaml
on:
  pull_request:
    branches: [main]
  push:
    branches: [main]

concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

## Possible Alternatives

- Keep both triggers and skip branch pushes when an open PR exists by querying the GitHub API. This is more complex and may introduce race conditions.
- Use a smaller test matrix for PRs and reserve the full matrix for `main` or scheduled runs.

## Checklist

jitu5:
Resolves https://github.com/kedro-org/kedro-viz/issues/2734 

## Description

Reduce unnecessary GitHub Actions runs and prevent outdated PR checks from consuming CI resources.

Previously, backend and JavaScript workflows could run twice for the same commit: once for `push` and once for `pull_request`.

## Development notes

- Restricted backend and JavaScript `push` workflows to `main`.
- Kept pull request checks available for PRs targeting intermediate branches.
- Added concurrency rules to cancel outdated runs for the same PR.
- Added documentation-specific path filters.
- Prevented JavaScript checks from running for Python package, documentation, Markdown, and OpenSpec-only changes.

## QA notes

- Validated the modified workflow files as YAML.
- Verified representative path-filter scenarios.
- Ran `git diff --check`.
- Confirmed that different PRs and workflows use separate concurrency groups.

## Checklist

- [ ] Opened this PR as a Draft Pull Request if it is work-in-progress
- [x] Documentation changes are not required
- [x] Release notes are not required
- [x] Workflow configuration changes were validated

### DIFF
--- a/.github/workflows/build-backend.yml
+++ b/.github/workflows/build-backend.yml
@@ -4,6 +4,8 @@ name: Build backend
 
 on:
   push:
+    branches:
+      - main
     paths:
       - 'package/**'
       - '.github/**'
@@ -12,6 +14,10 @@ on:
       - 'package/**'
       - '.github/**'
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/docs-checks.yml
+++ b/.github/workflows/docs-checks.yml
@@ -5,6 +5,15 @@ on:
     paths:
       - "docs/**"
       - '**.md'
+      - '.github/styles/**'
+      - '.github/workflows/docs-checks.yml'
+      - 'Makefile'
+      - 'mkdocs.yml'
+      - 'package/pyproject.toml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/docs-linkcheck.yml
+++ b/.github/workflows/docs-linkcheck.yml
@@ -4,10 +4,27 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'docs/**'
+      - '**.md'
+      - 'Makefile'
+      - 'mkdocs.yml'
+      - 'package/pyproject.toml'
+      - '.github/workflows/docs-linkcheck.yml'
   pull_request:
     branches:
       - main
+    paths:
+      - 'docs/**'
+      - '**.md'
+      - 'Makefile'
+      - 'mkdocs.yml'
+      - 'package/pyproject.toml'
+      - '.github/workflows/docs-linkcheck.yml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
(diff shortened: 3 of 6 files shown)


SajidAlamQB (approved):
Nice one @jitu5!

.github/workflows/docs-checks.yml
```diff
@@ -5,6 +5,15 @@ on:
     paths:
       - "docs/**"
       - '**.md'
+      - '.github/styles/**'
+      - '.github/workflows/docs-checks.yml'
+      - 'Makefile'
+      - 'mkdocs.yml'
+      - 'package/pyproject.toml'
```
ravi-kumar-pilla, line 12: why do we need to have these for doc checks ?

.github/workflows/docs-linkcheck.yml
```diff
@@ -4,10 +4,27 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'docs/**'
+      - '**.md'
+      - 'Makefile'
+      - 'mkdocs.yml'
+      - 'package/pyproject.toml'
+      - '.github/workflows/docs-linkcheck.yml'
```
ravi-kumar-pilla, line 13: same as above

.github/workflows/docs-checks.yml
jitu5: These are the files the docs jobs actually read, so a change to any of them can change the result: `make check-docs` runs `mkdocs build --strict` using the `[docs]` extra from `package/pyproject.toml`, and Vale loads its rules from `.github/styles` (per .vale.ini). The rest are the lint configs and the workflow file itself.

Without them, bumping a docs dependency or editing a Vale rule would skip docs CI entirely. Happy to drop any you think are overkill.

.github/workflows/docs-linkcheck.yml
jitu5: Same reason, smaller set — this workflow only runs make check-docs, so it needs the Makefile target, `mkdocs.yml`, and the `[docs]` deps in `package/pyproject.toml`.

.github/workflows/build-backend.yml
```diff
@@ -13,6 +15,10 @@ on:
       - '.github/**'
   workflow_dispatch:
 
+concurrency:
```
Huongg, line 18: can we move this to `.github/actions/setup-concurrency/action.yml`? and then can re-use across different workflows?

Huongg: actually i don't think github actions let you do that. Lets keep the current approach for now and maybe later we can add/create `.github/workflows/_concurrency-template.yml` but this can be done later, once we see how everything is working with this new change first

Huongg (approved):
thank you @jitu5, looks good to me. Looking forward to seeing the CI resource usage being reduced soon 😄

Issue #2660, opened 2026-05-26
Load shared inspection inputs at the context boundary and serve run status from the project context


jitu5:
### Description

**Parent issue:** #2265
**Depends on:** #2659, #2687

> **Reduced V1 scope:** This issue now records the shared inspection preparation and context-bound run-status work delivered for V1. The original node-metadata migration is tracked in #2775; context-backed export and deployment are tracked in #2776. The original issue description is preserved below.

## Current V1 boundary

- Main and pipeline graph responses come from inspection through the project context.
- Run status comes from the project's run-event file through the context, not from the snapshot.
- Node metadata remains live-backed.
- Static export and deployment remain on the legacy path.
- #2774 defers live loading until metadata or legacy export needs it when hooks are disabled. With hooks enabled, live project loading can still happen eagerly to obtain hook-modified layers.
- Lite compatibility and its remaining verification are tracked in #2661. V1 does not remove the metadata panel or promise fully session-free node detail.

The inspection context is passed explicitly. This is not a claim that all legacy process-wide state has been removed.

## Acceptance criteria for this reduced issue

- [x] Inspection inputs are prepared at the context boundary for graph construction.
- [x] File-backed extras can be prepared without a live catalog.
- [x] Full/lite graph-to-legacy-metadata ID compatibility has regression coverage.
- [x] Task ID generation is checked against the recorded run-status-hook baseline.
- [x] Run-status HTTP uses the explicit project root and retains the existing response contract.
- [x] Node metadata remains on the existing endpoint; its context migration is not required to close this reduced issue.

## Deferred work

- #2775 — snapshot-backed node metadata and same-ID live enrichment.
- #2776 — context-backed static export and deployment.
- #2723 — snapshot-backed task source handling.
- #2724 — final legacy-backend removal after consumer migrations.

## Delivered

- #2763 centralised inspection input preparation at the project-context boundary.
- #2764 consolidated file-backed stats and styles without requiring a live catalog for that preparation.
- #2773 moved `/api/run-status` to the project-root-bound `RunStatusService`. Event transformation and response formats were preserved; file lookup no longer depends on a later working-directory change.
- Inspection graphs remain compatible with legacy node-detail IDs. The full/lite compatibility test checks supported graph nodes against the legacy metadata endpoint. Separate ID tests compare task IDs with the recorded graph/run-status-hook baseline.

Closing this issue records completion of its reduced scope, not completion of the original metadata/export migration or the entire V1 release.

<details>
[...]

PR #2736, opened 2026-08-12
Populate graph nodes with parameter values  

### LINKED ISSUE #2687: Populate snapshot graph nodes with resolved parameter values

ravi-kumar-pilla:
## Description

Resolves #2687

Populate task-node `parameters` in the snapshot `GraphBuilder` using resolved values from `_InspectionSession.parameters()` (Kedro config loader, with `--params` already merged).

## Development notes

- Extended `GraphBuilder.__init__` with an optional `parameters: dict[str, Any] | None` argument.
- Added `_task_parameters()` to populate each task node's `parameters` field in the same shape as the legacy backend:
  - `"parameters"` input → full resolved mapping
  - `"params:x"` input → `{x: value}`
  - `"params:a.b.c"` input → dotted key lookup via `_resolve_param()`
  - missing keys → `None` (graph construction does not fail)
- Values are read from the existing `_InspectionSession.parameters()` helper — no new loader or manual `--params` merge.
- Update tests

## QA notes

- `pytest tests/test_inspection_adapter/`
- CI unit tests pass on Ubuntu (3.10–3.14)

## Checklist


jitu5 (approved):
The approach is right, and taking the values from the config loader is what makes this work in lite mode later.

Thanks also for the `trufflehog-ignore.txt` fix. That unblocks lint for everyone on `feat/backend_v2`, not just this PR.

### One thing to fix before merge

`_task_parameters` returns early when it sees a `parameters` input, so anything after it is dropped. I compared against the legacy backend:

```
inputs ["parameters", "params:model_options.test_size"]
  legacy  = {'model_options': {...}, 'other': 1, 'model_options.test_size': 0.2}
  this PR = {'model_options': {...}, 'other': 1}          <- dotted key missing
```

Legacy is also order-sensitive: `parameters` replaces what came before, and later `params:x` add to it. So the equivalent is:

```python
result: dict[str, Any] = {}
for ref in inputs:
    if ref == "parameters":
        result = dict(self._parameters)      # replace, like legacy
    elif ref.startswith("params:"):
        name = ref[len("params:") :]
        result[name] = _resolve_param(self._parameters, name)
return result
```

A node taking both inputs is unusual, so the demo baseline does not cover it. Worth a small test for that case.

### Coordination with #2750

1. Nothing calls `GraphBuilder` with `parameters` yet, since the provider does not exist on this branch. I will add `parameters=session.parameters()` in #2750, so no action needed from you.
2. We both touch `GraphBuilder.__init__`, so there will be a conflict. Suggest we land it as `def __init__(self, snapshot, catalog_config=None, *, parameters=None, layer_by_dataset=None)`, with both new arguments keyword-only. Whoever merges second can do it.
3. #2750 currently fills task parameters from the live node as a stopgap, marked with a `TODO(#2736)`. I will remove it once this lands, so there is only one source for that field.

### Optional, not blocking

- The ticket also mentions the override showing in that parameter node's `/api/nodes/{id}`. That looks like #2660 territory to me. Fine either way, just worth agreeing where it lands so it does not slip.
- A catalog templating case (`${runtime_params:...}`) would be good to have eventually. Related: `get_project_snapshot` does not accept runtime params, so a no-default `${runtime_params:x}` in the catalog makes the snapshot raise. Separate upstream issue, nothing for this PR.
- Might be worth reverting the `requirements.txt` trufflehog ignore once the pinned Kedro release lands, so we do not lose scanning on that file permanently.

ravi-kumar-pilla:
Hi @jitu5 ,

I made the fix. Please have a look at it and let me know if there is anything else that should be addressed. Thank you

Issue #2724, opened 2026-08-05
Remove the legacy graph-building backend


jitu5:
**Parent issue:** #2265  
**Depends on:** #2659, #2660, #2661, #2687, #2723, #2757

## Summary

Remove the legacy graph-construction path after every supported consumer uses `VizProjectContext` and `InspectionGraphService`, and parity has been proven.

Keep only the minimal live-object functionality still required for previews.

## Scope

- Delete the old pipeline traversal, graph-edge construction, registered-pipeline graph repositories and legacy graph-response builders.
- Remove the legacy free `get_pipeline_response()` and repository-backed compatibility context.
- Remove any fallback that can still serve a second graph engine.
- Remove transitional live enrichment that has been replaced by snapshot, configuration or file-backed sources.
- Keep the smallest same-ID live preview index required for preview functionality.
- Move the plain node-ID-to-layer sorter to the shared layer service.
- Remove the old `GraphNode`-based sorter and `CatalogRepository.layers_mapping` when unused.
- Verify that HTTP, export, deploy, VSCode and NotebookVisualizer all receive `VizProjectContext` explicitly.
- Verify the VSCode and NotebookVisualizer migrations completed in #2757 before deleting their legacy dependencies.
- Remove obsolete provider, adapter and compatibility terminology from production code and tests.

## Acceptance criteria

- `InspectionGraphService` is the only engine producing main and pipeline graph responses.
- Full mode, `--params`, lite mode, export, metadata and run-status tests pass.
- Full-mode previews retain their existing behaviour.
- VSCode uses a project-backed `VizProjectContext`, no longer depends on the legacy graph response builder, and passes compatibility tests.
- `NotebookVisualizer` uses an inspection-compatible in-memory `VizProjectContext` and passes its integration tests.
- HTTP, export and deployment use explicit context handoff.
- No production graph consumer reads `data_access_manager`.
- Repository-wide searches show no production references to the deleted graph-construction code.
- Parity and integration tests prove that deleting the legacy backend does not change supported responses.

PR #2716, opened 2026-07-30
Adopt `NodeSnapshot.func_name` from Snapshot

jitu5:
**Sub-task:** #2715 · **Parent:** #2265

## Summary

Uses `NodeSnapshot.func_name` to reproduce the existing Kedro-Viz task node IDs from inspection snapshots. This keeps the snapshot adapter compatible with the legacy graph and run-status IDs.

## Changes

- Pins Kedro to the commit that introduces `NodeSnapshot.func_name`.
- Reconstructs the legacy node string before hashing it.
- Handles automatic names, explicit names, namespaces and partial functions.
- Uses `func_name` when determining task display names.
- Adds ID parity tests against the captured legacy baseline.
- Updates the node-ID baseline report.

Dataset IDs remain unchanged.

## Testing

Tests cover:

- ID parity across every demo pipeline.
- Automatic and explicit node names.
- Namespaced and partial-function nodes.
- Nodes with empty inputs or outputs.
- Display-name behaviour.

## Release requirement

The temporary Kedro Git pin must be replaced with the released Kedro version containing `NodeSnapshot.func_name` before publishing Kedro-Viz.

PR #2750, opened 2026-08-14
Serve `/api/main` and `/api/pipelines/{id}` from the inspection snapshot

jitu5:
**Sub-task:** #2659  
**Parent:** #2265

## Description

Moves `/api/main` and `/api/pipelines/{id}` off the live graph engine and onto the inspection snapshot. No other endpoint implementation changes.

The routes are bound to a project-scoped `VizProjectContext` when the FastAPI app is created. `InspectionGraphService` owns the prepared snapshot, and the router receives the context explicitly—there is no process-wide provider slot or fallback to another engine. If the context cannot be built, startup fails loudly.

The context provides the extension point for #2660, while #2724 can later remove the legacy graph engine without changing these two routes again.

## Implementation

- `integrations/kedro/inspection/enrichment.py` (new): captures live-only fields as explicit, defensively copied inputs. The graph service does not retain or access the global live repositories.
- `integrations/kedro/inspection/graph_service.py` (new): reads the snapshot once at startup and serves both graph routes. `--pipeline` narrows the snapshot, and `--lite` reads it inside the existing import stubs.
- `api/rest/router.py` and `api/apps.py`: bind the graph routes to the context supplied when each FastAPI app is created. Separate apps cannot overwrite each other’s graph state.
- `server.py`: creates the context after the transitional live load and passes it into the app. The shared loader, `build` and `deploy` remain on the existing path. Context-construction failures are logged and re-raised.

Two things worth a closer look:

- **Fields supplied from the transitional live project.** The snapshot does not provide `.viz` stats or styles, or the abbreviated dataset type used by frontend icons. Startup copies those fields into explicit `EnrichmentSources`, keyed by the existing live-node IDs, and the service applies them without reading the global repositories or changing graph topology. Task parameters are not a live overlay: the merged #2736 owns parameter resolution and its tests, while this PR passes `session.parameters()` into `GraphBuilder`.
- **With `--include-hooks`, layers come from the populated catalog instead of the raw catalog config.** A hook can add, change or remove layer metadata, and only the populated catalog reflects that. Without hooks the raw config is used, as before.

## QA notes

- Full-response comparison against the captured baseline, field for field, for all six demo pipelines after a real live load. The baseline holds the existing graph responses, so this compares against real output rather than the adapter against itself.
- Route tests run over HTTP with a graph-service spy bound to the app context. They prove exact pipeline-ID delegation, 404 handling and isolation between two apps in one process. One test runs end to end through the real inspection service.
- Context-creation and server tests cover CLI-option forwarding, creation after live population, hook-modified layers, shared-loader and file-startup isolation, and failure propagation.
- Focused unit tests cover the frozen enrichment model, defensive input copying, graph-shape preservation, pipeline narrowing, lite-mode construction and graph-service behavior.
- The transitional fixture in `conftest.py` supplies a repository-backed `VizProjectContext` for older synthetic endpoint tests. It goes away with the live backend in #2724.
- Manual: run `kedro viz run` on `demo-project` and compare `/api/main` before and after, then verify `--pipeline`, `--params`, `--lite` and `--include-hooks`. `build` and `deploy` are unchanged.

## Deferred and related work

- `/api/nodes/{id}`, `/api/run-status`, static export and `--save-file` still use the live path: #2660. A saved file can therefore differ from what the endpoints serve; there is a TODO at the call site.
- Deleting the live graph engine: #2724.
- #2736 has merged into `feat/backend_v2`. It owns parameter resolution and its tests; this PR only supplies `session.parameters()` to `GraphBuilder`.
- `get_project_snapshot` takes no `runtime_params`, so a no-default `${runtime_params:x}` in the catalog or `parameters.yml` makes the snapshot raise and startup fail. This is an upstream Kedro gap and is not addressed here.
- Six modular-pipeline IDs now return 404 on `/api/nodes/{id}` where they returned `200 {}` before, because the live path registered those group nodes as a side effect of the old `/api/main`. No UI consumer requests them; this resolves with #2660.
- The lite-mode project-walk cache was extracted into the independent #2755 and is not part of this PR.
- `RELEASE.md` will be updated when the complete `backend_v2` work targets `main`.

## Verification

- Backend unit tests pass on Python 3.10 through 3.14.
- Full statement and branch coverage for the project context, graph service, explicit enrichment, route binding, graph-builder inputs and server context creation.
- `ruff check`, `ruff format --check` and both `mypy` runs are clean.
- DCO passes.
- The remaining end-to-end, documentation and secret-scan failures come from the temporary direct Kedro git dependency on `feat/backend_v2`, not from this PR.

## Checklist

- [ ] Opened this PR as a Draft Pull Request if it is work in progress (not applicable: ready for review)
- [ ] Updated the documentation (not applicable: no user-facing behaviour change)
- [ ] Added a `RELEASE.md` entry (deferred until the complete backend work targets `main`)
- [x] Added tests to cover the changes

### DIFF
--- a/package/kedro_viz/api/apps.py
+++ b/package/kedro_viz/api/apps.py
@@ -18,7 +18,9 @@ from werkzeug.utils import secure_filename
 from kedro_viz import __version__
 from kedro_viz.api.rest.responses.utils import EnhancedORJSONResponse
 from kedro_viz.integrations.kedro import telemetry as kedro_telemetry
+from kedro_viz.integrations.kedro.inspection import VizProjectContext
 
+from .rest.router import create_graph_router
 from .rest.router import router as rest_router
 
 _HTML_DIR = Path(__file__).parent.parent.absolute() / "html"
@@ -50,18 +52,24 @@ def _create_base_api_app() -> FastAPI:
 
 
 def create_api_app_from_project(
-    project_path: Path, autoreload: bool = False
+    context: VizProjectContext,
+    project_path: Path,
+    autoreload: bool = False,
 ) -> FastAPI:
     """Create an API from a real Kedro project by adding the router to the FastAPI app.
 
     Args:
-        project_path: Path to the Kedro project
+        context: Project-scoped services built from the inspection snapshot and enrichment
+            sources.
+        project_path: Path to the Kedro project.
         autoreload: Whether the app should autoreload based on content change
-            in the Kedro project
+            in the Kedro project.
+
     Returns:
-        The FastAPI app
+        The FastAPI app.
     """
     app = _create_base_api_app()
+    app.include_router(create_graph_router(context))
     app.include_router(rest_router)
 
     # Check for html directory existence.
--- a/package/kedro_viz/integrations/kedro/inspection/context.py
+++ b/package/kedro_viz/integrations/kedro/inspection/context.py
@@ -1,0 +1,60 @@
+"""Project-scoped services passed explicitly to Kedro-Viz consumers."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+from kedro_viz.integrations.kedro.inspection.enrichment import EnrichmentSources
+from kedro_viz.integrations.kedro.inspection.graph_service import (
+    InspectionGraphService,
+)
+
+
+class VizProjectContext:
+    """Services prepared for one Kedro project load."""
+
+    def __init__(self, graph: InspectionGraphService) -> None:
+        self.graph = graph
+
+    @classmethod
+    def from_project(
+        cls,
+        project_path: str | Path,
+        *,
+        env: str | None = None,
+        pipeline_name: str | None = None,
+        runtime_params: dict[str, Any] | None = None,
+        package_name: str | None = None,
+        is_lite: bool = False,
+        enrichment: EnrichmentSources | None = None,
+    ) -> VizProjectContext:
+        """Build project-scoped services from one inspection snapshot.
+
+        Args:
+            project_path: The Kedro project root.
+            env: The Kedro environment, honouring ``--env``.
+            pipeline_name: Restrict the context to one registered pipeline, honouring
+                ``--pipeline``.
+            runtime_params: Typed parameter overrides from ``--params``.
+            package_name: Project package used to identify imports in lite mode.
+            is_lite: Whether missing project dependencies should be temporarily mocked.
+            enrichment: Explicit fields supplied by the transitional live load.
+
+        Returns:
+            A project context containing the prepared inspection graph service.
+
+        Raises:
+            PipelineNotFoundError: If ``pipeline_name`` is not registered.
+        """
+        return cls(
+            graph=InspectionGraphService.from_project(
+                project_path,
+                env=env,
+                pipeline_name=pipeline_name,
+                runtime_params=runtime_params,
+                package_name=package_name,
+                is_lite=is_lite,
+                enrichment=enrichment,
+            )
+        )
--- a/package/kedro_viz/integrations/kedro/inspection/enrichment.py
+++ b/package/kedro_viz/integrations/kedro/inspection/enrichment.py
@@ -1,0 +1,77 @@
+"""Add live-only fields to graph responses without changing their structure."""
+
+from __future__ import annotations
+
+from collections.abc import Iterable, Mapping
+
+from pydantic import BaseModel, Field, field_validator
+
+from kedro_viz.api.rest.responses.pipelines import (
+    DataNodeAPIResponse,
+    GraphAPIResponse,
+    NodeExtrasAPIResponse,
+)
+from kedro_viz.models.flowchart.nodes import DataNode, GraphNode
+from kedro_viz.models.metadata import NodeExtras
+
+
+class EnrichmentSources(BaseModel, frozen=True):
+    """Fields copied from the transitional live load, keyed by existing node IDs.
+
+    The IDs come from the already-built live nodes rather than being recalculated here. This
+    keeps enrichment independent of graph construction while the live backend remains available.
+    """
+
+    node_extras_by_node_id: Mapping[str, NodeExtras] = Field(default_factory=dict)
+    dataset_type_by_node_id: Mapping[str, str | None] = Field(default_factory=dict)
+    layer_by_dataset: Mapping[str, str] | None = None
+
+    @field_validator(
+        "node_extras_by_node_id",
+        "dataset_type_by_node_id",
+        "layer_by_dataset",
+        mode="before",
+    )
+    @classmethod
+    def _copy_mapping(cls, value: Mapping | None) -> dict | None:
+        """Copy caller-owned mappings before storing them on the prepared model."""
+        return None if value is None else dict(value)
+
+    @classmethod
+    def from_live_nodes(
+        cls,
+        nodes: Iterable[GraphNode],
+        *,
+        layer_by_dataset: Mapping[str, str] | None = None,
+    ) -> EnrichmentSources:
+        """Copy the fields needed by the inspection graph from already-built live nodes."""
+        node_extras_by_node_id: dict[str, NodeExtras] = {}
+        dataset_type_by_node_id: dict[str, str | None] = {}
+        for node in nodes:
+            if node.node_extras is not None:
+                node_extras_by_node_id[node.id] = node.node_extras
+            if isinstance(node, DataNode):
+                dataset_type_by_node_id[node.id] = node.dataset_type
+        return cls(
+            node_extras_by_node_id=node_extras_by_node_id,
+            dataset_type_by_node_id=dataset_type_by_node_id,
+            layer_by_dataset=layer_by_dataset,
+        )
+
+
+def enrich_graph_response(
+    response: GraphAPIResponse, sources: EnrichmentSources
+) -> None:
+    """Mutate live-only response fields without adding, removing or renaming nodes."""
+    for node in response.nodes:
+        node_extras = sources.node_extras_by_node_id.get(node.id)
+        if node_extras is not None:
+            node.node_extras = NodeExtrasAPIResponse(
+                stats=node_extras.stats,
+                styles=node_extras.styles,
+            )
+        if (
+            isinstance(node, DataNodeAPIResponse)
+            and node.id in sources.dataset_type_by_node_id
+        ):
+            node.dataset_type = sources.dataset_type_by_node_id[node.id]
--- a/package/kedro_viz/integrations/kedro/inspection/graph_service.py
+++ b/package/kedro_viz/integrations/kedro/inspection/graph_service.py
@@ -1,0 +1,124 @@
+"""Serve graph responses from one Kedro inspection snapshot."""
+
+from __future__ import annotations
+
+import dataclasses
+from contextlib import nullcontext
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+from kedro_viz.api.rest.responses.pipelines import GraphAPIResponse
+from kedro_viz.integrations.kedro.inspection.enrichment import (
+    EnrichmentSources,
+    enrich_graph_response,
+)
+from kedro_viz.integrations.kedro.inspection.graph_builder import GraphBuilder
+from kedro_viz.integrations.kedro.inspection.snapshot_source import (
+    _InspectionSession,
+    lite_import_stubs,
+)
+
+if TYPE_CHECKING:
+    from kedro.inspection.models import ProjectSnapshot
+
+
+class PipelineNotFoundError(ValueError):
+    """Raised when a requested pipeline is not present in the inspection snapshot."""
+
+
+class InspectionGraphService:
+    """Build graph responses from one prepared inspection snapshot."""
+
+    def __init__(
+        self,
+        builder: GraphBuilder,
+        enrichment: EnrichmentSources | None = None,
+    ) -> None:
+        self._builder = builder
+        self._enrichment = enrichment if enrichment is not None else EnrichmentSources()
+
+    @classmethod
+    def from_project(
+        cls,
+        project_path: str | Path,
+        *,
+        env: str | None = None,
+        pipeline_name: str | None = None,
+        runtime_params: dict[str, Any] | None = None,
+        package_name: str | None = None,
+        is_lite: bool = False,
+        enrichment: EnrichmentSources | None = None,
+    ) -> InspectionGraphService:
+        """Read one project snapshot and prepare the graph service.
+
+        Args:
+            project_path: The Kedro project root.
+            env: The Kedro environment, honouring ``--env``.
+            pipeline_name: Restrict the view to one registered pipeline, honouring
+                ``--pipeline``.
+            runtime_params: Typed parameter overrides from ``--params``.
+            package_name: Project package used to identify imports in lite mode.
+            is_lite: Whether missing project dependencies should be temporarily mocked.
+            enrichment: Explicit fields supplied by the transitional live load.
+
+        Raises:
+            PipelineNotFoundError: If ``pipeline_name`` is not registered.
+        """
+        sources = enrichment if enrichment is not None else EnrichmentSources()
+        import_context = (
+            lite_import_stubs(project_path, package_name) if is_lite else nullcontext()
+        )
+        with import_context:
+            session = _InspectionSession(
+                project_path, env=env, runtime_params=runtime_params
+            )
+            snapshot = session.snapshot()
+            catalog_config = session.catalog_config()
+            parameters = session.parameters()
+            if pipeline_name is not None:
+                snapshot = cls._filter_to_pipeline(snapshot, pipeline_name)
+            builder = GraphBuilder(
+                snapshot,
+                catalog_config,
+                parameters=parameters,
+                layer_by_dataset=sources.layer_by_dataset,
+            )
+        return cls(builder, sources)
+
+    def get_pipeline_response(self, pipeline_id: str | None = None) -> GraphAPIResponse:
+        """Return the graph for one registered pipeline.
+
+        Raises:
+            PipelineNotFoundError: If ``pipeline_id`` is not registered.
+        """
+        selected_pipeline_id = (
+            self._builder.default_pipeline_id() if pipeline_id is None else pipeline_id
+        )
+        if not self._builder.has_pipeline(selected_pipeline_id):
+            raise PipelineNotFoundError(
+                f"Invalid pipeline ID: {selected_pipeline_id!r}"
+            )
+        response = self._builder.build(selected_pipeline_id)
+        enrich_graph_response(response, self._enrichment)
+        return response
+
+    @staticmethod
+    def _filter_to_pipeline(
+        snapshot: ProjectSnapshot, pipeline_name: str
+    ) -> ProjectSnapshot:
+        """Return a copy of the snapshot containing only the requested pipeline.
+
+        Raises:
+            PipelineNotFoundError: If ``pipeline_name`` is not registered.
+        """
+        matching = [
+            pipeline
+            for pipeline in snapshot.pipelines
+            if pipeline.name == pipeline_name
+        ]
+        if not matching:
+            available = sorted(pipeline.name for pipeline in snapshot.pipelines)
+            raise PipelineNotFoundError(
+                f"Pipeline {pipeline_name!r} not found in snapshot; available: {available}"
+            )
+        return dataclasses.replace(snapshot, pipelines=matching)
--- a/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
+++ b/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
@@ -50,9 +50,9 @@ def lite_import_stubs(
 class _InspectionSession:
     """Read a project's snapshot and config, bootstrapping and building the loader once.
 
-    Create one session per request. The project is bootstrapped and the Kedro config loader is built
-    lazily on first use and then cached, so the catalog config and parameters reuse a single loader
-    instead of rebuilding it.
+    Create one session per adapter build. The project is bootstrapped and the Kedro config loader
+    is built lazily on first use and then cached, so the catalog config and parameters reuse a
+    single loader instead of rebuilding it.
     """
 
     def __init__(
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -43,7 +43,7 @@ def populate_data(
     data_access_manager.add_pipelines(pipelines)
 
 
-def _load_and_populate_project_data(
+def load_and_populate_data(
     path: Path,
     env: Optional[str] = None,
     include_hooks: bool = False,
@@ -52,7 +52,11 @@ def _load_and_populate_project_data(
     extra_params: Optional[Dict[str, Any]] = None,
     is_lite: bool = False,
 ) -> DataAccessManager:
-    """Load one project and populate the transitional live repositories."""
+    """Load a project and return the populated legacy repositories.
+
+    VSCode and deployment call this entry point and ignore its return value. The HTTP
+    server uses the returned repositories to build its project-scoped inspection context.
+    """
 
     # Loads data from underlying Kedro Project
     catalog, pipelines, node_extras_dict = kedro_data_loader.load_data(
@@ -68,31 +72,6 @@ def _load_and_populate_project_data(
     # Creates data repositories which are used by Kedro Viz Backend APIs
     populate_data(data_access_manager, catalog, pipelines, node_extras_dict)
     return data_access_manager
-
-
-def load_and_populate_data(
-    path: Path,
-    env: Optional[str] = None,
-    include_hooks: bool = False,
-    package_name: Optional[str] = None,
-    pipeline_name: Optional[str] = None,
-    extra_params: Optional[Dict[str, Any]] = None,
-    is_lite: bool = False,
-) -> None:
-    """Load a project and populate the legacy repositories.
-
-    VSCode and deployment call this compatibility entry point directly. Project-scoped
-    inspection services are created only by their own entry points.
-    """
-    _load_and_populate_project_data(
-        path,
-        env,
-        include_hooks,
-        package_name,
-        pipeline_name,
-        extra_params,
-        is_lite,
-    )
 
 
 def _create_viz_project_context(
@@ -198,7 +177,7 @@ def run_server(
     path = Path(project_path) if project_path else Path.cwd()
 
     if load_file is None:
-        live_data = _load_and_populate_project_data(
+        live_data = load_and_populate_data(
             path, env, include_hooks, package_name, pipeline_name, extra_params, is_lite
         )
         # Copy enrichment from the transitional live repositories into the project-scoped
--- a/package/kedro_viz/server.py
+++ b/package/kedro_viz/server.py
@@ -1,6 +1,7 @@
 """`kedro_viz.server` provides utilities to launch a webserver
 for Kedro pipeline visualisation."""
 
+import logging
 from pathlib import Path
 from typing import Any, Dict, Optional
 
@@ -11,8 +12,14 @@ from kedro_viz.autoreload_file_filter import AutoreloadFileFilter
 from kedro_viz.constants import DEFAULT_HOST, DEFAULT_PORT
 from kedro_viz.data_access import DataAccessManager, data_access_manager
 from kedro_viz.integrations.kedro import data_loader as kedro_data_loader
+from kedro_viz.integrations.kedro.inspection import (
+    EnrichmentSources,
+    VizProjectContext,
+)
 from kedro_viz.launchers.utils import _check_viz_up, _wait_for, display_cli_message
 from kedro_viz.models.metadata import NodeExtras
+
+logger = logging.getLogger(__name__)
 
 DEV_PORT = 4142
 
@@ -36,7 +43,7 @@ def populate_data(
     data_access_manager.add_pipelines(pipelines)
 
 
-def load_and_populate_data(
+def _load_and_populate_project_data(
     path: Path,
     env: Optional[str] = None,
     include_hooks: bool = False,
@@ -44,8 +51,8 @@ def load_and_populate_data(
     pipeline_name: Optional[str] = None,
     extra_params: Optional[Dict[str, Any]] = None,
     is_lite: bool = False,
-):
-    """Loads underlying Kedro project data and populates Kedro Viz Repositories"""
+) -> DataAccessManager:
+    """Load one project and populate the transitional live repositories."""
 
     # Loads data from underlying Kedro Project
     catalog, pipelines, node_extras_dict = kedro_data_loader.load_data(
@@ -60,6 +67,87 @@ def load_and_populate_data(
 
     # Creates data repositories which are used by Kedro Viz Backend APIs
     populate_data(data_access_manager, catalog, pipelines, node_extras_dict)
+    return data_access_manager
+
+
+def load_and_populate_data(
+    path: Path,
+    env: Optional[str] = None,
+    include_hooks: bool = False,
+    package_name: Optional[str] = None,
+    pipeline_name: Optional[str] = None,
+    extra_params: Optional[Dict[str, Any]] = None,
+    is_lite: bool = False,
+) -> None:
+    """Load a project and populate the legacy repositories.
+
+    VSCode and deployment call this compatibility entry point directly. Project-scoped
+    inspection services are created only by their own entry points.
+    """
+    _load_and_populate_project_data(
+        path,
+        env,
+        include_hooks,
+        package_name,
+        pipeline_name,
+        extra_params,
+        is_lite,
+    )
+
+
+def _create_viz_project_context(
+    path: Path,
+    live_data: DataAccessManager,
+    *,
+    env: Optional[str] = None,
+    pipeline_name: Optional[str] = None,
+    extra_params: Optional[Dict[str, Any]] = None,
+    package_name: Optional[str] = None,
+    is_lite: bool = False,
+    include_hooks: bool = False,
+) -> VizProjectContext:
+    """Create the explicit context used by the HTTP graph routes.
+
+    Args:
+        path: The Kedro project root.
+        live_data: Repositories populated by the transitional live load.
+        env: The Kedro environment, honouring ``--env``.
+        pipeline_name: Restrict the view to one registered pipeline, honouring ``--pipeline``.
+        extra_params: Typed parameter overrides from ``--params``.
+        package_name: The Kedro project package, used to identify project imports in lite mode.
+        is_lite: Whether to mock missing project dependencies while building the snapshot.
+        include_hooks: Whether hook-modified layers should replace raw catalog layers.
+
+    Raises:
+        Exception: If the inspection context cannot be constructed.
+    """
+    try:
+        # A hook can add, change or remove layer metadata, and only the populated catalog
+        # reflects that, so with hooks the builder reads layers from there instead of the
+        # raw catalog config.
+        layer_by_dataset = (
+            dict(live_data.catalog.layers_mapping) if include_hooks else None
+        )
+        enrichment = EnrichmentSources.from_live_nodes(
+            live_data.nodes.as_list(),
+            layer_by_dataset=layer_by_dataset,
+        )
+        return VizProjectContext.from_project(
+            path,
+            env=env,
+            pipeline_name=pipeline_name,
+            runtime_params=extra_params,
+            package_name=package_name,
+            is_lite=is_lite,
+            enrichment=enrichment,
+        )
+    # Context construction is an all-or-nothing startup requirement. Log and propagate every
+    # failure rather than serving an app whose graph routes cannot work.
+    except Exception:
+        logger.exception(
+            "Could not build the Kedro inspection context, so the graph cannot be served."
+        )
+        raise
 
 
 def run_server(
@@ -110,11 +198,26 @@ def run_server(
     path = Path(project_path) if project_path else Path.cwd()
 
     if load_file is None:
-        load_and_populate_data(
+        live_data = _load_and_populate_project_data(
             path, env, include_hooks, package_name, pipeline_name, extra_params, is_lite
         )
+        # Copy enrichment from the transitional live repositories into the project-scoped
+        # context. The graph service keeps no reference to the global repositories.
+        context = _create_viz_project_context(
+            path,
+            live_data,
+            env=env,
+            pipeline_name=pipeline_name,
+            extra_params=extra_params,
+            package_name=package_name,
+            is_lite=is_lite,
+            include_hooks=include_hooks,
+        )
+
         # [TODO: As we can do this with `kedro viz build`,
         # we need to shift this feature outside of kedro viz run]
+        # TODO(#2660): make ``--save-file`` and ``kedro viz build`` use the project
+        # context so static exports match the HTTP graph responses.
         if save_file:
             from kedro_viz.api.rest.responses.save_responses import (
                 save_api_responses_to_fs,
@@ -122,7 +225,7 @@ def run_server(
 
             save_api_responses_to_fs(save_file, fsspec.filesystem("file"), True)
 
-        app = apps.create_api_app_from_project(path, autoreload)
+        app = apps.create_api_app_from_project(context, path, autoreload)
     else:
         app = apps.create_api_app_from_file(f"{path}/{load_file}/api")
 
--- a/package/tests/test_inspection_adapter/test_graph_routes.py
+++ b/package/tests/test_inspection_adapter/test_graph_routes.py
@@ -1,0 +1,131 @@
+"""Tests that graph routes use the context bound when the API app is created.
+
+Everything here goes over HTTP via ``TestClient``. A spy service records what the routes ask
+for, so these fail if ``/api/main`` or ``/api/pipelines/{id}`` bypasses the explicit context.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+from fastapi.testclient import TestClient
+
+from kedro_viz.api import apps
+from kedro_viz.api.rest.responses.pipelines import GraphAPIResponse
+from kedro_viz.integrations.kedro.inspection import (
+    InspectionGraphService,
+    VizProjectContext,
+)
+from kedro_viz.integrations.kedro.inspection.graph_service import (
+    PipelineNotFoundError,
+)
+
+DEMO_PROJECT = Path(__file__).resolve().parents[3] / "demo-project"
+
+
+class _SpyGraphService(InspectionGraphService):
+    """Record the pipeline IDs requested by routes."""
+
+    def __init__(self) -> None:
+        self.requested: list[str | None] = []
+
+    def get_pipeline_response(self, pipeline_id: str | None = None) -> GraphAPIResponse:
+        self.requested.append(pipeline_id)
+        if pipeline_id == "unknown":
+            raise PipelineNotFoundError("Invalid pipeline ID: 'unknown'")
+        return GraphAPIResponse(
+            nodes=[],
+            edges=[],
+            layers=[],
+            tags=[],
+            pipelines=[],
+            modular_pipelines={},
+            selected_pipeline=pipeline_id or "__default__",
+        )
+
+
+@pytest.fixture
+def spy_service() -> _SpyGraphService:
+    return _SpyGraphService()
+
+
+@pytest.fixture
+def client(spy_service: _SpyGraphService) -> TestClient:
+    context = VizProjectContext(graph=spy_service)
+    return TestClient(apps.create_api_app_from_project(context, Path.cwd()))
+
+
+def test_main_route_asks_the_service_for_the_default_pipeline(
+    client: TestClient, spy_service: _SpyGraphService
+) -> None:
+    """``/api/main`` delegates with no ID, leaving default selection to the service."""
+    response = client.get("/api/main")
+    assert response.status_code == 200
+    assert spy_service.requested == [None]
+
+
+def test_pipeline_route_passes_the_requested_id_through(
+    client: TestClient, spy_service: _SpyGraphService
+) -> None:
+    """The pipeline ID in the URL reaches the bound service unchanged."""
+    response = client.get("/api/pipelines/data_ingestion")
+    assert response.status_code == 200
+    assert spy_service.requested == ["data_ingestion"]
+    assert response.json()["selected_pipeline"] == "data_ingestion"
+
+
+def test_unknown_pipeline_is_translated_to_http_404(
+    client: TestClient, spy_service: _SpyGraphService
+) -> None:
+    """The HTTP layer translates the service's domain error into the established response."""
+    response = client.get("/api/pipelines/unknown")
+    assert response.status_code == 404
+    assert response.json() == {"message": "Invalid pipeline ID"}
+    assert spy_service.requested == ["unknown"]
+
+
+def test_each_app_uses_the_context_bound_when_it_was_created() -> None:
+    """Two apps in one process do not share or overwrite graph state."""
+    first_service = _SpyGraphService()
+    second_service = _SpyGraphService()
+    first_client = TestClient(
+        apps.create_api_app_from_project(
+            VizProjectContext(graph=first_service), Path.cwd()
+        )
+    )
+    second_client = TestClient(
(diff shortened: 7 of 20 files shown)


ravi-kumar-pilla (commented):
## Kedro PR Review

Thanks for the thorough work on #2659 — the provider-slot design, fail-loud startup, and baseline parity tests are all solid. This is **not an approval**; a few items need attention before merge.

### Context

This PR implements [#2659](https://github.com/kedro-org/kedro-viz/issues/2659) (serve `/api/main` and `/api/pipelines/{id}` from the inspection snapshot) under parent [#2265](https://github.com/kedro-org/kedro-viz/issues/2265). Deferred work for [#2660](https://github.com/kedro-org/kedro-viz/issues/2660), [#2724](https://github.com/kedro-org/kedro-viz/issues/2724), and [#2736](https://github.com/kedro-org/kedro-viz/issues/2736) is documented appropriately.

### Overview
- **Critical:** 3 | **Suggestions:** 5

### Critical (must fix before merge)
- [ ] **DCO** — the DCO check is `ACTION_REQUIRED`; commits need sign-off.
- [ ] `package/tests/test_server.py` — CI reports `ruff format --check` failure on this file.
- [ ] **CI e2e (Python 3.13)** — e2e fails at package install (`hatchling` direct-reference metadata error). Likely branch-wide, but it blocks green CI on this PR.

- [ ] `package/kedro_viz/integrations/kedro/lite_parser.py` — document or handle `@lru_cache` behaviour under autoreload.
- [ ] `package/tests/test_inspection_adapter/test_inspection_adapter_parity.py` — add an end-to-end `--include-hooks` case for the layer merge-over-replace behaviour.
- [ ] Track double-bootstrap (live load + snapshot read) as follow-up once #2661 lands.
- [ ] Squash commit history before merge to `main` (27 commits include pre-squash #2726 work; diff is clean but history is noisy).
- [ ] `package/kedro_viz/api/inspection_adapter_provider.py` — note lite-mode parameters limitation near the overlay code (#2736).

### What looks good
- Clean `GraphDataProvider` seam; routes delegate correctly (spy + e2e HTTP tests).
- Provider installed after live load in `run_server`; `load_and_populate_data` unchanged for deploy/build.
- Live overlay bounded to three fields; topology unchanged (dedicated test).
- Layer merge fix (hook layers over raw catalog) with realistic factory-dataset test.
- Shared cached `unresolved_modules()` avoids double lite-mode project walks.
- Captured baseline parity across all six demo pipelines compares against legacy output.

### Alignment with #2659 acceptance criteria
| Criterion | Status |
|-----------|--------|
| Graph endpoints served from inspection adapter | ✅ |
| Matches captured baseline (IDs, parameters, node_extras, dataset types) | ✅ |
| Honours `--env`, `--pipeline`, `--params` | ✅ |
| Unknown pipeline → 404 | ✅ |
| Provider built once per server load | ✅ |
| No silent fallback to live graph engine | ✅ |

### Notes
- **General review:** Strong implementation and test depth; scope matches the ticket.
- **PR checklist:** RELEASE.md deferral is acceptable on `feat/backend_v2`; DCO and format CI must be fixed.
- **API usage:** N/A (kedro-viz repo).

jitu5:
@ravi-kumar-pilla 

> DCO — the DCO check is ACTION_REQUIRED; commits need sign-off.
 
 Fixed. I squashed the branch into a single signed-off commit, which also takes care of your suggestion about the noisy 27 commits.

> package/tests/test_server.py — CI reports ruff format --check failure on this file.

`test_server.py` formatting, you were right, and thanks for catching it. It only shows up under the pinned ruff==0.15.8; I had been checking with an older ruff locally, which formatted that lambda differently. Fixed and re-verified with the pinned version.

> CI e2e (Python 3.13)

this one looks branch-wide rather than specific to this PR. Your #2736 shows the same failure across 3.10 to 3.14, and it doesn't touch requirements.txt either. Same root cause as the secret-scan one: the direct-reference kedro pin from #2716. Probably worth its own ticket, since it will keep every PR on feat/backend_v2 red until the kedro release lands.

jitu5:
> package/kedro_viz/integrations/kedro/lite_parser.py — document or handle @lru_cache behaviour under autoreload.

Documented in one line on the function. Short version: every restart path starts a new process, --autoreload goes through watchfiles.run_process and the Jupyter launcher uses a spawned multiprocessing.Process, so a reload always gets an empty cache.

> package/tests/test_inspection_adapter/test_inspection_adapter_parity.py — add an end-to-end --include-hooks case for the layer merge-over-replace behaviour.

This is the one that paid off, though not how I expected. I wrote the test, it passed, then I checked whether it actually discriminates and it didn't: it passed with the merge, with a plain replace, and even with layer_by_dataset ignored entirely.

> Track double-bootstrap (live load + snapshot read) as follow-up once https://github.com/kedro-org/kedro-viz/issues/2661 lands.

Agreed, the live load and the snapshot read both bootstrap. Leaving it for #2661 as you suggested.

> Squash commit history before merge to main (27 commits include pre-squash https://github.com/kedro-org/kedro-viz/pull/2726 work; diff is clean but history is noisy).

Done, 27 commits into 1.

> package/kedro_viz/api/inspection_adapter_provider.py — note lite-mode parameters limitation near the overlay code (https://github.com/kedro-org/kedro-viz/pull/2736).

Added a TODO(#2736) next to the overlay noting that lite mode has no live node so parameters stay empty there until your PR lands.

Huongg:
Hey @jitu5, thanks for drafting up the PR. 

This is quite a large and important change and I’m finding it difficult to do a meaningful review as it is currently grouped into two commits.

To review big changes such as this we need to have the work broken into meaningful commits that tell the story of the implementation, so that it is easier to understand what changed, why, and review each step with clearer context.

As commit history is part of our code documentation, for significant changes like this we might come back to this in a couple of months or so and `git blame` and the commit history should help us understand how and why we got here.

Happy to talk to provide more clarity on the ask if needed

package/kedro_viz/api/inspection_adapter_provider.py
```diff
@@ -0,0 +1,173 @@
...
+                # The API exposes resolved types only for plain data nodes; transcoded and
+                # parameter nodes use ``None``.
+                graph_node.dataset_type = (
+                    live_node.dataset_type if isinstance(live_node, DataNode) else None
+                )
+            if isinstance(graph_node, TaskNodeAPIResponse) and isinstance(
+                live_node, TaskNode
+            ):
+                # TODO(#2736): resolve parameters from the config loader and drop this
+                # overlay. Lite mode has no live node, so parameters are empty there.
```
SajidAlamQB: Is this describing lite mode after #2661 lands? If so could we reword to future tense, since as written
it reads like current behaviour.

jitu5:
> Hey @jitu5, thanks for drafting up the PR.
> 
> This is quite a large and important change and I’m finding it difficult to do a meaningful review as it is currently grouped into two commits.
> 
> To review big changes such as this we need to have the work broken into meaningful commits that tell the story of the implementation, so that it is easier to understand what changed, why, and review each step with clearer context.
> 
> As commit history is part of our code documentation, for significant changes like this we might come back to this in a couple of months or so and `git blame` and the commit history should help us understand how and why we got here.
> 
> Happy to talk to provide more clarity on the ask if needed


Thanks @Huongg , this is fair feedback.
Some context: this branch originally had around 27 commits because the work was developed incrementally on the feature branch, including fixups and changes carried through the stacked work. Ravi previously asked me to squash that noisy history before merge, so I cleaned it up quite aggressively.
That explains the current two-commit structure, but I agree it has gone too far in the other direction and no longer tells the implementation story clearly. The aim was to remove development noise.

I’m going to extract the lite-parser caching optimisation into a small independent PR against `feat/backend_v2`. I’ll then reorganise this PR into five curated, signed-off commits covering the GraphBuilder inputs, provider seam, inspection adapter and parity, server installation, and final route cutover. The relevant tests will remain with each step.
This should make the PR easier to review commit by commit without restoring the original development noise. Let me know if this make sense.

package/kedro_viz/server.py
```diff
@@ -62,6 +71,58 @@ def load_and_populate_data(
...
+
+    Raises:
+        Exception: If the inspection provider cannot be constructed.
+    """
+    try:
+        # A hook can add, change or remove layer metadata, and only the populated catalog
+        # reflects that, so with hooks the builder reads layers from there instead of the
+        # raw catalog config.
+        layer_by_dataset = (
+            dict(data_access_manager.catalog.layers_mapping) if include_hooks else None
```
SajidAlamQB: Edgecase but a dataset with a layer in raw config that never materialises into `catalog.keys()`. `_extract_layers` resolves patterns against snapshot dataset names, but `layers_mapping`
iterates the populated catalog, so an unmaterialised factory entry is invisible to it. Since the builder
does a full replace rather than a merge, that layer would be dropped under `--include-hooks` is that intended?

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -22,23 +22,25 @@ def lite_import_stubs(
...
         Metadata.set_has_missing_dependencies(True)
-        sys_modules_patch.update(lite_parser.create_mock_modules(modules_to_mock))
-        logger.warning(
-            "Kedro-Viz --lite: building the snapshot with %d project dependency module(s) "
-            "mocked. Install them for full functionality:\n%s",
+        sys_modules_patch.update(
+            LiteParser(package_name).create_mock_modules(modules_to_mock)
+        )
+        # The live loader already warned the user about these; this is for diagnosis only.
+        logger.debug(
```
SajidAlamQB: Could we add a `TODO(#2724)` here noting the warning has to come back when the live loader goes?

package/kedro_viz/api/inspection_adapter_provider.py
jitu5: Good catch. This describes the future state after #2661 removes the live loader. I’ll reword it in future tense.

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: Agreed, and that code has since moved. It's now in #2755, where the warning drops to debug because the live loader already warns about the same modules. I'll add the TODO(#2724) there.

package/kedro_viz/server.py
jitu5: Good catch. For a factory dataset that resolves successfully, `add_catalog()` materializes every dataset referenced by the pipelines before `layers_mapping` is read. Unused factory patterns are also absent from the snapshot graph.

A factory dataset that fails to materialize is the remaining edge case. In that situation, the current live backend also has no layer because it reads from the populated catalog without falling back to raw config.
Replacement is intentional because the populated catalog is authoritative with `--include-hooks`, including when a hook removes a layer. Merging would incorrectly restore the raw-config layer.

I’m happy to add a focused test to make this behaviour explicit.

package/kedro_viz/api/data_provider.py
```diff
@@ -0,0 +1,55 @@
...
+
+    from kedro_viz.api.rest.responses.pipelines import GraphAPIResponse
+
+
+class GraphDataProvider(Protocol):
+    """What the graph endpoints need from whatever is serving them."""
+
+    def get_pipeline_response(
+        self, pipeline_id: str | None = None
+    ) -> "GraphAPIResponse | JSONResponse": ...
```
ravi-kumar-pilla: **Suggestion:** Consider adding a one-line note on `GraphDataProvider` that implementors may return `JSONResponse` for 404s, not only `GraphAPIResponse`. FastAPI already handles this, but the protocol docstring currently only describes the happy path.

package/kedro_viz/server.py
```diff
@@ -62,6 +71,58 @@ def load_and_populate_data(
...
+        provider = InspectionAdapterProvider(
+            path,
+            env=env,
+            pipeline_name=pipeline_name,
+            runtime_params=extra_params,
+            package_name=package_name,
+            is_lite=is_lite,
+            layer_by_dataset=layer_by_dataset,
+        )
+    except Exception:
```
ravi-kumar-pilla, line 125: **Suggestion:** The broad `except Exception` is justified (log, clear slot, re-raise). A short comment that this is intentional startup failure handling would help readers who might otherwise flag it in a future review.

package/tests/test_inspection_adapter/test_inspection_adapter_parity.py
```diff
@@ -0,0 +1,96 @@
+"""Parity between the served graph and the captured legacy response.
+
+The baseline under ``baseline/`` was captured from the live backend, so these compare the whole
+served response against it, field for field, for every registered pipeline. The provider is built
+after a real live load, so the live-only fields are populated too.
```
ravi-kumar-pilla: **Suggestion:** Parity scope is live-load-only. CLI variants (`--params`, `--env`, `--lite`, `--pipeline`, `--include-hooks`) are covered in unit/installer tests but not in full-response parity. A brief note here pointing readers to where those variants are tested would prevent future "why no parity for X?" questions.

package/tests/test_inspection_adapter/test_graph_routes.py
```diff
@@ -0,0 +1,102 @@
...
+    response = client.get("/api/pipelines/data_ingestion")
+    assert response.status_code == 200
+    assert spy.requested == ["data_ingestion"]
+    assert response.json()["selected_pipeline"] == "data_ingestion"
+
+
+def test_routes_serve_the_real_adapter_response(
+    client, _restore_kedro_project_state
+) -> None:
+    """End to end: the real adapter, over HTTP, returns the demo project's graph.
```
ravi-kumar-pilla: **Suggestion:** This e2e test uses an empty live repository, so it validates routing + snapshot rendering but not overlay fields. A one-line docstring note that overlay enrichment is covered in `test_inspection_adapter_provider.py` and the parity tests would make the split explicit.

ravi-kumar-pilla (commented):
## Kedro PR Review

Thanks for the thorough work on #2659 — the provider-slot design, fail-loud startup, and baseline parity tests are all solid.

### Context

This PR implements #2659 (serve `/api/main` and `/api/pipelines/{id}` from the inspection snapshot) under parent #2265. Deferred work for #2660, #2724, and #2736 is documented appropriately.

### Overview

- **Critical:** 0 | **Suggestions:** 4

### Critical (must fix before merge)

No issues found in the code changes (e2e excluded). Previous blockers (DCO, ruff format) appear addressed; unit tests pass across Python 3.10–3.14.

- [ ] `package/kedro_viz/api/data_provider.py` — Document that `get_pipeline_response` may return `JSONResponse` for error cases.
- [ ] `package/kedro_viz/server.py` — Add a brief comment that the broad `except Exception` is intentional startup failure handling.
- [ ] `package/tests/test_inspection_adapter/test_inspection_adapter_parity.py` — Note where CLI-variant behaviour (`--params`, `--lite`, etc.) is tested, since parity is live-load-only.
- [ ] `package/tests/test_inspection_adapter/test_graph_routes.py` — Clarify in the e2e test docstring that overlay fields are covered elsewhere.

### What looks good

- Clean `GraphDataProvider` seam; routes delegate correctly (spy + e2e HTTP tests).
- Provider installed after live load in `run_server`; `load_and_populate_data` unchanged for deploy/build.
- Live overlay bounded to three fields; topology unchanged (dedicated test).
- Layer merge fix (hook layers over raw catalog) with realistic factory-dataset test.
- Captured baseline parity across all six demo pipelines compares against legacy output.
- Six curated commits tell the implementation story clearly.

### Alignment with #2659 acceptance criteria

| Criterion | Status |
|-----------|--------|
| Graph endpoints served from inspection adapter | ✅ |
| Matches captured baseline (IDs, parameters, node_extras, dataset types) | ✅ |
| Honours `--env`, `--pipeline`, `--params` | ✅ |
| Unknown pipeline → 404 | ✅ |
| Provider built once per server load | ✅ |
| No silent fallback to live graph engine | ✅ |

### Notes

- **General review:** Architecture is clean and maintainable. The provider slot is the right abstraction for #2660/#2724 follow-ups. Live-field overlay is bounded, documented, and temporary (#2736).
- **PR checklist:** RELEASE.md deferral is acceptable on `feat/backend_v2`. Tests are comprehensive.
- **API usage:** N/A (kedro-viz repo).

ravi-kumar-pilla:
Hi @jitu5 ,

This PR re-introduces the same pattern of globals we wanted to avoid earlier (`DataAccessManager`) singleton. Have a look at the below approach which avoids globals, keeps the design simple using a service layer. This will help us re-use the context across consumers like http, VSCode, Notebook and any future consumers. Let me know what you think of this approach. 

---

> **Goal:** Serve graph JSON from the Kedro inspection snapshot, remove globals, keep the design simple, and make the same service usable from HTTP, VSCode, static export, deploy, and notebook paths.

---

## Summary

`GraphBuilder` (already on `feat/backend_v2`) produces the graph JSON. We do **not** need `api/data_provider.py` or `api/inspection_adapter_provider.py`.

Instead:

1. Add **`InspectionGraphService`** under `integrations/kedro/inspection/` — orchestrates snapshot load, `GraphBuilder`, and enrichment.
2. Wrap it in **`VizProjectContext`** — one object created at startup, passed explicitly to every consumer.
3. **HTTP:** bind `context` when the FastAPI app is created (router/app factory with closure) — no module-level `_context`, no `install_viz_context()`.
4. **Save / export / VSCode / notebook / deploy:** receive `context` as an explicit function argument.
5. Phase out **`data_access_manager`** for graph construction; delete it entirely in [#2724](https://github.com/kedro-org/kedro-viz/issues/2724).

No module globals. No provider slot. No `get_pipeline_response()` reading hidden state.

---

## Problem with the current direction

PR #2750 would add a second global:

| Global | Purpose |
|---|---|
| `data_access_manager` | Legacy graph + node metadata |
| `GraphDataProvider` slot | New snapshot graph |

We are trying to remove globals ([#2724](https://github.com/kedro-org/kedro-viz/issues/2724)), not add another one. Meanwhile `GraphBuilder` already returns `GraphAPIResponse` — the PR's new `api/` files are mostly glue that belongs next to the adapter.

---

## Core idea: one `VizProjectContext`, many consumers

```text
run_server / CLI entry point
    │
    ▼
VizProjectContext.from_project(...)
    │
    ├─► create_api_app(context)          → HTTP (routes close over context)
    ├─► save_api_responses_to_fs(context) → export / --save-file
    ├─► get_kedro_project_json_data(context) → VSCode
    └─► NotebookContext.from_...(context)    → notebook (future)
    │
    ▼
context.graph.get_pipeline_response()
    │
    InspectionGraphService
    │
    ├─► GraphBuilder (topology)
    └─► enrich_graph_response (EnrichmentSources)
```

**One bootstrap, one object, explicit handoff to each consumer.**

---

## Package layout

All adapter code stays under `integrations/kedro/inspection/` (per [#2265 discussion](https://github.com/kedro-org/kedro-viz/issues/2265#issuecomment-4515097337)):

```text
package/kedro_viz/integrations/kedro/inspection/
├── snapshot_source.py      # exists — _InspectionSession, lite_import_stubs
├── graph_builder.py        # exists — pure snapshot → GraphAPIResponse
├── enrichment.py           # NEW — EnrichmentSources + enrich_graph_response()
├── graph_service.py        # NEW — InspectionGraphService
├── context.py              # NEW — VizProjectContext factory
└── __init__.py             # export public API
```

**Do not add:**

- `api/data_provider.py`
- `api/inspection_adapter_provider.py`
- `install_viz_context()` / `require_viz_context()` / module-level `_context`

---

## Layer responsibilities

### 1. `GraphBuilder` — unchanged, stays pure

- **Input:** `ProjectSnapshot`, catalog config, optional `layer_by_dataset`
- **Output:** `GraphAPIResponse` (nodes, edges, layers, tags, modular pipelines, IDs)
- **No:** HTTP, globals, live overlay, KedroSession

Already implemented on `feat/backend_v2`.

### 2. `EnrichmentSources` — what it is

`EnrichmentSources` is a **plain immutable dataclass** — a bag of inputs for fields the snapshot does not yet supply. It is not a service and not a global.

```python
@dataclass(frozen=True)
class EnrichmentSources:
    """Inputs merged onto GraphBuilder output. Does not change topology or IDs."""

    # From .viz/stats.json and .viz/styles.json
    node_extras: dict[str, NodeExtras]

    # Resolved parameter values, keyed by task node id
    parameters_by_task_id: dict[str, dict[str, Any]]

    # Transitional: abbreviated dataset type for frontend icons
    # e.g. "pandas.csv_dataset.CSVDataset" vs snapshot's "pandas.CSVDataset"
    dataset_type_by_node_id: dict[str, str] | None = None

    # When --include-hooks: hook-modified layers from populated catalog
    layer_by_dataset: dict[str, str] | None = None
```

#### Where each field comes from

| Field | Transitional (#2659) | After #2736 / #2661 | After #2724 |
|---|---|---|---|
| `node_extras` | `.viz/` JSON files | same | same |
| `parameters_by_task_id` | live `TaskNode.parameters` or config loader | config loader `[parameters]` | config loader |
| `dataset_type_by_node_id` | live `DataNode.dataset_type` | normalize snapshot types in adapter | normalize in adapter |
| `layer_by_dataset` | populated catalog if `--include-hooks` | same | same |

#### How it is used

```python
def enrich_graph_response(
    response: GraphAPIResponse,
    sources: EnrichmentSources,
) -> None:
    """Mutate response in place. Topology and IDs are unchanged."""
    for node in response.nodes:
        # attach node_extras, parameters, dataset_type from sources
        ...
```

Built once at startup via `EnrichmentSources.from_project(...)`. Deleted incrementally as [#2736](https://github.com/kedro-org/kedro-viz/pull/2736) and [#2661](https://github.com/kedro-org/kedro-viz/issues/2661) land.

### 3. `InspectionGraphService` — single graph entry point

```python
class InspectionGraphService:
    """Serve graph responses from one inspection snapshot."""

    @classmethod
    def from_project(
        cls,
        project_path: Path,
        *,
        env: str | None = None,
        pipeline_name: str | None = None,
        runtime_params: dict[str, Any] | None = None,
        package_name: str | None = None,
        is_lite: bool = False,
        enrichment: EnrichmentSources | None = None,
    ) -> InspectionGraphService:
        """
        1. lite_import_stubs if is_lite
        2. _InspectionSession → snapshot + catalog_config
        3. filter to --pipeline if set (fail at startup if unknown)
        4. GraphBuilder(snapshot, catalog_config, layer_by_dataset=enrichment.layer_by_dataset)
        """
        ...

    def get_pipeline_response(
        self, pipeline_id: str | None = None
    ) -> GraphAPIResponse | JSONResponse:
        """Default pipeline, 404 check, build(), enrich_graph_response()."""

    def pipeline_ids(self) -> list[str]:
        """Registered pipeline IDs — used by save/export enumeration."""
```

This replaces `InspectionAdapterProvider` from PR #2750, but lives in `inspection/` and takes explicit `EnrichmentSources` instead of reading `data_access_manager.nodes`.

**Do not** add `InspectionGraphService.get_pipeline_response()` as a classmethod with hidden class-level state — that recreates the singleton problem under a different name.

### 4. `VizProjectContext` — shared by all consumers

```python
@dataclass
class VizProjectContext:
    """Everything Kedro-Viz needs from one project load. Created once, passed everywhere."""

    graph: InspectionGraphService
    # Added in #2660:
    # nodes: NodeMetadataService
    # export: StaticExportService

    @classmethod
    def from_project(
        cls,
        project_path: Path,
        *,
        env: str | None = None,
        pipeline_name: str | None = None,
        runtime_params: dict[str, Any] | None = None,
        package_name: str | None = None,
        is_lite: bool = False,
        include_hooks: bool = False,
    ) -> VizProjectContext:
        enrichment = EnrichmentSources.from_project(
            project_path,
            env=env,
            runtime_params=runtime_params,
            include_hooks=include_hooks,
            is_lite=is_lite,
        )
        graph = InspectionGraphService.from_project(
            project_path,
            env=env,
            pipeline_name=pipeline_name,
            runtime_params=runtime_params,
            package_name=package_name,
            is_lite=is_lite,
            enrichment=enrichment,
        )
        return cls(graph=graph)

    # Future (#2660): in-memory notebook path
    # @classmethod
    # def from_pipeline_and_catalog(cls, pipeline, catalog) -> VizProjectContext: ...
```

---

## Making the service available to all consumers

### Recommendation: explicit `context` everywhere

| Consumer | How it gets the service |
|---|---|
| **`kedro viz run` (HTTP)** | `context = VizProjectContext.from_project(...)` → `create_api_app(context)` |
| **`kedro viz run --save-file`** | Same `context` → `save_api_responses_to_fs(path, fs, context, ...)` |
| **`kedro viz build` / deploy** | `BaseDeployer` receives or builds `context`, passes to save |
| **VSCode** | `get_kedro_project_json_data(context, pipeline_name)` |
| **Notebook visualizer** | `VizProjectContext.from_pipeline_and_catalog(...)` per cell (future) |
| **Unit / parity tests** | `InspectionGraphService.from_project(DEMO_PROJECT)` directly |

**Rule:** If a function can take `context: VizProjectContext`, it **must**. No hidden lookups.

---

## HTTP — app factory with closure (no module global)

Today `create_api_app_from_project(project_path)` builds a app with a static router that imports `data_access_manager`. Change it to accept `context`:

```python
def create_api_app(
    context: VizProjectContext,
    project_path: Path,
    autoreload: bool = False,
) -> FastAPI:
    app = _create_base_api_app()
    router = APIRouter(prefix="/api")

    @router.get("/main", response_model=GraphAPIResponse)
    async def main():
        return context.graph.get_pipeline_response()

    @router.get("/pipelines/{registered_pipeline_id}", response_model=GraphAPIResponse)
    async def get_single_pipeline_data(registered_pipeline_id: str):
        return context.graph.get_pipeline_response(registered_pipeline_id)

    # #2660: node metadata routes close over the same context
    # @router.get("/nodes/{node_id}")
    # async def get_single_node_metadata(node_id: str):
    #     return context.nodes.get_metadata(node_id)

    app.include_router(router)
    app.state.viz = context  # optional — middleware, debugging, Depends fallback
    ...
    return app
```

`run_server`:

```python
context = VizProjectContext.from_project(path, env=env, ...)
app = create_api_app(context, path, autoreload)
if save_file:
    save_api_responses_to_fs(save_file, fsspec.filesystem("file"), context, True)
uvicorn.run(app, ...)
```

Routes are short (`context.graph.get_pipeline_response()`) and **close over `context`** — no module global, no facade reading hidden state.

### Deprecate `get_pipeline_response()` without context

The old free function in `pipelines.py` reads `data_access_manager` today. **Do not** replace it with `require_viz_context()`.

Instead:

- **Delete** it from the HTTP path once routes close over `context`.
- **Keep** a named helper only when it takes context explicitly:

```python
def get_pipeline_response(
    context: VizProjectContext,
    pipeline_id: str | None = None,
):
    return context.graph.get_pipeline_response(pipeline_id)
```

Useful for save/export and VSCode without duplicating the call chain.

---

## Save / export

```python
# save_responses.py — proposed
def save_api_responses_to_fs(
    path: str,
    remote_fs: Any,
    context: VizProjectContext,
    is_all_previews_enabled: bool,
):
    write_api_response_to_fs(
        main_path, context.graph.get_pipeline_response(), remote_fs
    )
    for pipeline_id in context.graph.pipeline_ids():
        write_api_response_to_fs(
            pipeline_path,
            context.graph.get_pipeline_response(pipeline_id),
            remote_fs,
        )
    # #2660: enumerate metadata nodes from context.nodes
    ...
```

Call sites:

```python
# server.py
context = VizProjectContext.from_project(path, ...)
app = create_api_app(context, path, autoreload)
if save_file:
    save_api_responses_to_fs(save_file, fs, context, True)

# base_deployer.py
save_api_responses_to_fs(self._path, self._fs, self._context, is_all_previews_enabled)
```

---

## VSCode

```python
def get_kedro_project_json_data(
    context: VizProjectContext,
    pipeline_name: str | None = None,
):
    response = context.graph.get_pipeline_response(pipeline_name)
    return encode_to_json_dict(response)
```

VSCode bootstrap:

```python
context = VizProjectContext.from_project(project_path, ...)
return get_kedro_project_json_data(context, pipeline_name)
```

No import-time singleton. Same bootstrap as the server.

---

## Notebook visualizer

Today: `data_access_manager.reset_fields()` + `populate_data()` on every cell — a global workaround.

Proposed (when inspection supports in-memory objects, #2724 prerequisite):

```python
context = VizProjectContext.from_pipeline_and_catalog(pipeline, catalog)
return context.graph.get_pipeline_response()
```

Each cell gets a **fresh context**. No reset, no shared mutable state.

Until in-memory snapshot exists, notebook can keep the live path temporarily, but should not depend on HTTP globals or facades.

---

## Startup flows

### Full mode (transitional, #2659)

```text
kedro viz run
    │
    ▼
context = VizProjectContext.from_project(path, env, pipeline, params, hooks)
    │
    ├─► EnrichmentSources.from_project()
    │       ├─ load .viz/stats.json, styles.json
    │       ├─ [transitional] KedroSession → parameters, dataset types, hook layers
    │       └─ build immutable EnrichmentSources
    │
    ├─► _InspectionSession → get_project_snapshot()
    ├─► GraphBuilder(snapshot) → topology
    └─► InspectionGraphService(builder + enrichment)
    │
    ▼
app = create_api_app(context, path)
save_api_responses_to_fs(..., context, ...)   # if --save-file
uvicorn.run(app)
```

Live load during transition **only feeds enrichment**, not graph topology.

### Lite mode (end state, #2661)

```text
kedro viz run --lite
    │
    ▼
context = VizProjectContext.from_project(..., is_lite=True)
    │
    ├─► EnrichmentSources — config loader + files only, no KedroSession
    ├─► lite_import_stubs → snapshot
    ├─► GraphBuilder → topology
    └─► enrich (no live catalog)
    │
    ▼
app = create_api_app(context, path)
```

### End state (#2724)

Single path for full and lite. No `data_access_manager`. No `populate_data()` for graphs. Enrichment from files + config loader only.

---

## Comparison with PR #2750

| | PR #2750 | This proposal |
|---|---|---|
| Module globals | +1 (`GraphDataProvider` slot) | **0** |
| New `api/` files | 2 | 0 |
| HTTP wiring | `get_graph_data_provider()` | `create_api_app(context)` — closure |
| Save / export | Old `get_pipeline_response()` + repositories | `save_api_responses_to_fs(context, ...)` |
| VSCode | Global `get_kedro_project_json_data()` | `get_kedro_project_json_data(context)` |
| Notebook | `reset_fields()` on global | Fresh `VizProjectContext` per cell |
| Adapter location | Split across `api/` and `inspection/` | All in `integrations/kedro/inspection/` |
| Tests | Spy provider + slot cleanup | Direct `from_project()` or `TestClient(create_api_app(context))` |
| #2660 extension | Extend provider protocol | Add services to `VizProjectContext` |

---

## Phased migration

| Phase | Ticket | Work |
|---|---|---|
| **1** | #2659 (revised) | Add `InspectionGraphService`, `EnrichmentSources`, `VizProjectContext`. Change `create_api_app(context)`. Transitional live load for enrichment only. |
| **2** | #2660 | Add `NodeMetadataService` to context. Wire node routes via closure. Migrate save/export fully. |
| **3** | #2661 | Lite mode: skip KedroSession; enrichment from config loader + files. |
| **4** | #2736 | Parameters from config loader; drop live parameter overlay. |
| **5** | #2724 | Delete `DataAccessManager` graph path, repositories, `data_access_manager` singleton. |

---

## Testing strategy

### Unit tests — no HTTP, no globals

```python
service = InspectionGraphService.from_project(DEMO_PROJECT)
response = service.get_pipeline_response("__default__")
assert response.nodes
```

### Parity tests

Compare `service.get_pipeline_response(pid)` against captured legacy baseline field-for-field.

### Route tests — bind context at app creation

```python
context = VizProjectContext.from_project(DEMO_PROJECT)
app = create_api_app(context, DEMO_PROJECT)
client = TestClient(app)
assert client.get("/api/main").status_code == 200
```

No `install_viz_context`. No patching module-level state.

### Save / export tests

```python
context = VizProjectContext.from_project(DEMO_PROJECT)
save_api_responses_to_fs(tmp_path, fs, context, is_all_previews_enabled=False)
# assert files match baseline
```

---

## What to avoid

- **`GraphDataProvider` protocol** — only one implementation exists (snapshot);
- **`install_viz_context()` / `require_viz_context()` / module `_context`** — these are globals; use app factory + explicit arguments instead.
- **Second singleton mirroring `data_access_manager`** — the provider slot in PR #2750.
- **`InspectionGraphService.get_pipeline_response()` as classmethod** — hidden class state; use instance methods.
- **`get_pipeline_response()` with no arguments** — reads hidden state; require `context` or use route closure.
- **Enrichment inside `GraphBuilder`** — keep builder pure; enrichment is temporary and deletion-bound.
- **ContextVar as primary pattern** — explicit `context` arguments are clearer; ContextVar only for a legacy call site that cannot be updated immediately.

---

## Migration steps for PR #2750

If adopting this on the `#2750` / `#2659` branch:

1. **Drop** `api/data_provider.py` and `api/inspection_adapter_provider.py`.
2. **Add** `enrichment.py`, `graph_service.py`, `context.py` under `integrations/kedro/inspection/`.
3. **Keep** `GraphBuilder` layer-input changes and parity tests from #2750.
4. **Change** `create_api_app_from_project` → `create_api_app(context, project_path, ...)` with routes closing over `context`.
5. **Update** `server.py` to build `VizProjectContext`, pass it to `create_api_app` and `save_api_responses_to_fs`.
6. **Remove** or rewrite `get_pipeline_response()` to require `context`; migrate VSCode and deploy in #2660.

Same acceptance criteria as #2659.

---

## One-sentence summary

**Create `VizProjectContext` once in the entry point, pass it explicitly to save/export/VSCode/notebook and into `create_api_app(context)` for HTTP (routes close over it) — no module globals, because `GraphBuilder` already has the JSON.**

---

## Related issues

- [#2265](https://github.com/kedro-org/kedro-viz/issues/2265) — Parent: modularity of Kedro-Viz
- [#2659](https://github.com/kedro-org/kedro-viz/issues/2659) — Serve graph endpoints through inspection adapter
- [#2660](https://github.com/kedro-org/kedro-viz/issues/2660) — Node metadata and static export
- [#2661](https://github.com/kedro-org/kedro-viz/issues/2661) — Lite mode without KedroSession
- [#2724](https://github.com/kedro-org/kedro-viz/issues/2724) — Remove legacy graph backend
- [#2736](https://github.com/kedro-org/kedro-viz/pull/2736) — Parameter values from config loader
- [PR #2750](https://github.com/kedro-org/kedro-viz/pull/2750) — Current provider-slot implementation (alternative to this proposal)

jitu5:
Thanks @ravi-kumar-pilla, this is a good catch. I agree that the provider slot introduces another process-wide global, which is the pattern we are trying to remove.

I’ve reworked this locally and will update #2750 with the following structure.

### Adopting in this PR
- Remove `api/data_provider.py`, the `GraphDataProvider` protocol, and `InspectionAdapterProvider`.
- Add `InspectionGraphService`, `EnrichmentSources`, and a minimal `VizProjectContext` under `integrations/kedro/inspection/`.
- Create the context once during server startup and pass it explicitly to the app.
- Bind `/api/main` and `/api/pipelines/{id}` using a router factory that closes over the context. There is no module-level context or provider slot.
- Pass enrichment explicitly from the transitional live load. The enrichment maps use IDs already assigned to live nodes, so the ID logic is not recalculated.
- Keep resolved parameters in GraphBuilder, consistent with #2736, instead of introducing a separate task-ID enrichment map.
- Have InspectionGraphService raise UnknownPipelineError; the HTTP route translates that into the existing 404 response, so the inspection package does not depend on FastAPI.


### Keeping this PR focused
The following work remains in its existing follow-up ticket:
- #2660: node metadata, run status, --save-file, static export, and deployment.
- #2661: remove the live project load from lite mode.
- #2757: migrate NotebookVisualizer and VSCode to explicit contexts.
- #2724: remove the legacy graph backend and data_access_manager.

`load_and_populate_data()` and `get_kedro_project_json_data()` remain unchanged for now, so VSCode, NotebookVisualizer, deployment, and static export continue using the existing path until their migration tickets.

### Result
The two HTTP graph endpoints are snapshot-backed and global-free. The legacy backend remains temporarily for unmigrated consumers and live-only functionality, which preserves the planned safe stopping point without expanding #2659 into the later work.

ravi-kumar-pilla:
Thanks @jitu5 — the refactor to `VizProjectContext` + `create_api_app(context)` aligns with the architecture we discussed. A few suggestions below (not an approval):

1. **#2736 coordination** — this PR adds parameter resolution in `GraphBuilder` (`_task_parameters`, `_resolve_param`, edge-case tests). Please confirm the rebase plan with #2736 so we don't end up with duplicate or conflicting parameter logic when both land.

2. **`--save-file` mismatch** — worth calling out explicitly in the PR description: `--save-file` during `kedro viz run` still uses the legacy `get_pipeline_response()` path until #2660, so saved output can differ from what the HTTP graph endpoints serve (the TODO in `server.py` covers this; a user-facing note in the PR body would help reviewers).

3. **Optional: `InspectionGraphService.pipeline_ids()`** — `GraphBuilder` already exposes this; a thin delegate on the service would make #2660 save/export enumeration from `context.graph` slightly cleaner.

4. **Optional: `/api/main` error handling** — `/api/pipelines/{id}` catches `UnknownPipelineError` and returns 404; `/api/main` does not. Probably fine today since `main` always uses the default pipeline, but either wrap it the same way or add a brief comment explaining why `UnknownPipelineError` cannot occur on that route.

5. **Merge hygiene** — commit history still includes messages from the earlier provider-slot approach (e.g. "Add the graph data provider seam") before the removal commit. Consider squashing or curating on merge so `git blame` reflects the final design.

package/kedro_viz/api/apps.py
```diff
@@ -50,18 +52,23 @@ async def set_secure_headers(request, call_next):
...
-    project_path: Path, autoreload: bool = False
+    context: VizProjectContext,
+    project_path: Path,
+    autoreload: bool = False,
 ) -> FastAPI:
     """Create an API from a real Kedro project by adding the router to the FastAPI app.
 
     Args:
-        project_path: Path to the Kedro project
+        context: Services prepared for this project load.
```
ravi-kumar-pilla:
```suggestion
        context: Project-scoped graph service built from the inspection snapshot and enrichment sources.
```

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -0,0 +1,44 @@
...
+        project_path: str | Path,
+        *,
+        env: str | None = None,
+        pipeline_name: str | None = None,
+        runtime_params: dict[str, Any] | None = None,
+        package_name: str | None = None,
+        is_lite: bool = False,
+        enrichment: EnrichmentSources | None = None,
+    ) -> VizProjectContext:
+        """Create the services needed to visualise one project."""
```
ravi-kumar-pilla: """Builds a project-scoped context for serving API responses.

Reads the Kedro inspection snapshot once and wraps the resulting graph service.

Args:
    project_path: Root directory of the Kedro project.
    env: Kedro configuration environment (``--env``).
    pipeline_name: If set, restrict the snapshot to this registered pipeline
        (``--pipeline``). An unknown name fails at construction time.
    runtime_params: Typed parameter overrides from ``--params``.
    package_name: Project package name, used to identify imports in lite mode.
    is_lite: When ``True``, mock missing project dependencies while reading the snapshot.
    enrichment: Live-only fields to merge onto graph responses without changing topology or node IDs.

Returns:
    A frozen context holding the prepared class `InspectionGraphService`.

Raises:
    UnknownPipelineError: If ``pipeline_name`` is not a registered pipeline.
"""

package/kedro_viz/integrations/kedro/inspection/enrichment.py
```diff
@@ -0,0 +1,86 @@
...
+from kedro_viz.models.metadata import NodeExtras
+
+
+@dataclass(frozen=True)
+class EnrichmentSources:
+    """Fields copied from the transitional live load, keyed by existing node IDs.
+
+    The IDs come from the already-built live nodes rather than being recalculated here. This
+    keeps enrichment independent of graph construction while the live backend remains available.
+    """
```
ravi-kumar-pilla, line 23: Not sure if we are formatting and linting. Lets make sure the format and lint tests pass. (Only the e2e tests can be expected to fail)

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Any
+
+from kedro_viz.integrations.kedro.inspection.enrichment import EnrichmentSources
+from kedro_viz.integrations.kedro.inspection.graph_service import (
+    InspectionGraphService,
+)
+
+
+@dataclass(frozen=True)
+class VizProjectContext:
```
ravi-kumar-pilla, line 14: Hi @jitu5 ,

The design looks great and easy to follow. I would also like us to follow the below convention for backend -

**Modelling convention for the inspection backend**

For `backend_v2` we should use one rule in `integrations/kedro/inspection/`:

**Pydantic for structured data. Plain classes for services. No dataclasses.**

Pydantic fits value objects and data passed between layers. Plain Python classes fit behaviour and orchestration (`InspectionGraphService`, `GraphBuilder`, `_InspectionSession`). The new backend should not mix in dataclasses. 

**Changes requested in this PR**

1. **`EnrichmentSources`**  
   Convert from `@dataclass(frozen=True)` to a frozen Pydantic model. It is pure data and already uses Pydantic `NodeExtras`. Replace `__post_init__` and `MappingProxyType` with a validator that copies mappings on construction.

2. **`VizProjectContext`**  
   Replace the dataclass with a **plain class** whose only job is grouping services at the composition root. It holds live service instances, not serialisable data, so it should not be Pydantic or a dataclass.

3. **Internal types from #2726**  
   Convert `_NamespaceBoundaries` and `_ModularPipelineTreeEntry` to Pydantic in this PR as well, so `inspection/` does not keep a dataclass enclave.

**Leave as plain service classes**

`InspectionGraphService`, `GraphBuilder`, and the existing `*APIResponse` models are fine as they are.

package/kedro_viz/integrations/kedro/inspection/graph_service.py
```diff
@@ -0,0 +1,122 @@
...
+from kedro_viz.integrations.kedro.inspection.snapshot_source import (
+    _InspectionSession,
+    lite_import_stubs,
+)
+
+if TYPE_CHECKING:
+    from kedro.inspection.models import ProjectSnapshot
+
+
+class UnknownPipelineError(ValueError):
```
ravi-kumar-pilla: `UnknownPipelineError` reads as if the error type itself is unknown. The intent is that the requested pipeline ID is not in the inspection snapshot.

Please rename to `PipelineNotFoundError` or `UnregisteredPipelineError`

jitu5:
> Thanks @jitu5 — the refactor to `VizProjectContext` + `create_api_app(context)` aligns with the architecture we discussed. A few suggestions below (not an approval):
> 
> 1. **[Populate graph nodes with parameter values   #2736](https://github.com/kedro-org/kedro-viz/pull/2736) coordination** — this PR adds parameter resolution in `GraphBuilder` (`_task_parameters`, `_resolve_param`, edge-case tests). Please confirm the rebase plan with [Populate graph nodes with parameter values   #2736](https://github.com/kedro-org/kedro-viz/pull/2736) so we don't end up with duplicate or conflicting parameter logic when both land.
> 2. **`--save-file` mismatch** — worth calling out explicitly in the PR description: `--save-file` during `kedro viz run` still uses the legacy `get_pipeline_response()` path until [Route node metadata and static export through the inspection adapter #2660](https://github.com/kedro-org/kedro-viz/issues/2660), so saved output can differ from what the HTTP graph endpoints serve (the TODO in `server.py` covers this; a user-facing note in the PR body would help reviewers).
> 3. **Optional: `InspectionGraphService.pipeline_ids()`** — `GraphBuilder` already exposes this; a thin delegate on the service would make [Route node metadata and static export through the inspection adapter #2660](https://github.com/kedro-org/kedro-viz/issues/2660) save/export enumeration from `context.graph` slightly cleaner.
> 4. **Optional: `/api/main` error handling** — `/api/pipelines/{id}` catches `UnknownPipelineError` and returns 404; `/api/main` does not. Probably fine today since `main` always uses the default pipeline, but either wrap it the same way or add a brief comment explaining why `UnknownPipelineError` cannot occur on that route.
> 5. **Merge hygiene** — commit history still includes messages from the earlier provider-slot approach (e.g. "Add the graph data provider seam") before the removal commit. Consider squashing or curating on merge so `git blame` reflects the final design.

Thanks @ravi-kumar-pilla 

1. #2736 is merged and this PR is rebased on top of it. The duplicate `_task_parameters` and `_resolve_param` are gone, and all your parameter tests came through from the base branch, including `test_task_parameters_input_with_dotted_params`. This PR now keeps only two things in that area: `parameters=session.parameters()` in the graph service, and moving parameters behind the `*` so it and `layer_by_dataset` are both keyword-only, as we agreed.

2. Agreed, and this is already in the PR body. The "What still uses the live path" section says `--save-file` still goes through the legacy `get_pipeline_response()` until #2660, so a saved file can differ from what the HTTP endpoints serve. Happy to make it more prominent if it's easy to miss where it currently sits.

3. Agree it will be useful, but I've left it out for now since nothing in this PR calls it and I'd rather not add an unused method. It's written into #2660's scope as "Expose registered pipeline IDs from the graph service for export enumeration", so it lands with the caller that needs it.

4. Added the comment rather than the try/except, since main never passes an ID and the service picks from its own registered pipelines, so `PipelineNotFoundError` can't reach that route. Deliberately scoped the wording to that specific error rather than implying the route can never raise.

5. Done. History is curated into five commits with no merge commit and no trace of the provider slot, so `git blame` only shows the final design.

package/kedro_viz/integrations/kedro/inspection/enrichment.py
jitu5: These pass. Lint is green on 3.10 through 3.14, and unit tests pass on Linux and Windows. The only red checks are e2e, which is failing branch-wide, and Read the Docs, which is expected while the Kedro git pin is in place, that's the note you added to `requirements.txt` in #2736.

package/kedro_viz/integrations/kedro/inspection/graph_service.py
jitu5: Agreed, renamed to `PipelineNotFoundError`.

package/kedro_viz/integrations/kedro/inspection/context.py
jitu5: Agreed, and all four are converted. `EnrichmentSources` and `_NamespaceBoundaries` are frozen Pydantic models; `_ModularPipelineTreeEntry` is Pydantic but deliberately not frozen because the tree builder mutates `inputs`, `outputs` and `children`. `VizProjectContext` is a plain class. No dataclasses left in `inspection/`.

Two notes while doing it. Pydantic's frozen=True blocks reassigning a field but doesn't deep-freeze a dict, so I replaced MappingProxyType with a mode="before" validator that copies incoming mappings, and the tests now promise copied inputs and frozen fields rather than deep immutability.

Also changed the two `tree.setdefault(...)` calls to direct construction and lookup. `setdefault` was building a throwaway entry on every iteration, which was cheap for a dataclass but not for a validated model. It's  Safe to change because parents are always created before their children.

package/kedro_viz/server.py
```diff
@@ -60,6 +67,87 @@ def load_and_populate_data(
 
     # Creates data repositories which are used by Kedro Viz Backend APIs
     populate_data(data_access_manager, catalog, pipelines, node_extras_dict)
+    return data_access_manager
+
+
+def load_and_populate_data(
```
ravi-kumar-pilla: why do we need both `load_and_populate_data` and `_load_and_populate_project_data` ? Will this be removed in future tasks ?

ravi-kumar-pilla (approved):
Left some minor comments, but overall this looks good to me. Thank you @jitu5

package/kedro_viz/server.py
jitu5: Good catch, the split isn't worth it.

The only reason it exists is that `run_server` needs the populated manager so it can pass enrichment into the context explicitly instead of reading the global, and I didn't want to change the signature of the public one. But the three callers (VSCode, `kedro viz build`, `kedro viz deploy`) all call it as a plain statement and ignore the result, so adding a return value is invisible to them.

I'll merge them into one `load_and_populate_data` that returns the `DataAccessManager`.

On whether it gets removed: not for a now. We've decided to take the safe stop(options 2). This function is the legacy path VSCode and NotebookVisualizer still use, so it stays as long as that dual-path state does.

The cleanup happens in two steps, both in the later phase. #2757 moves VSCode and NotebookVisualizer onto `VizProjectContext`, and #2724 then deletes `data_access_manager` and this loader with it.

SajidAlamQB (approved):
Thank you @jitu5 this was great work thanks for taking your time to explain an address it all! 😄

Huongg (approved):
Thank you @jitu5 the new changes look good and very easy to follow as well. 🙌

PR #2753, opened 2026-08-21
Fix version checks for static and embedded Viz.

### LINKED ISSUE #2378: React component makes failed request to `/api/version` when embedded
## Description

When using the `@quantumblack/kedro-viz` React component inside a custom frontend application (i.e. not running the full Kedro backend server), the component makes a request to:

`GET /api/version`

This results in a 404 Not Found response because no Kedro backend is running to serve that endpoint.

The version request appears to be triggered unconditionally, even when the component is used in embedded mode and provided with data directly.

This produces unnecessary network noise and console errors.
Image

## Steps to Reproduce
- Create a frontend app (e.g. Vite, CRA, Next.js, etc.) or use [kedro-viz-standalone](https://github.com/kedro-org/kedro-viz-standalone/tree/main)
- Install: npm install @quantumblack/kedro-viz
- Render the Kedro Viz React component using local pipeline JSON data (without running kedro viz backend server).
- Open browser DevTools → Network tab.
- Reload the page.

## Expected Result
When Kedro-Viz is used as an embedded React component with supplied data, it should:
❌ Not attempt to call /api/version
❌ Not make any backend API calls unless explicitly configured
✅ Operate fully in client-only mode

No spurious network requests should occur.

## Actual Result
See above image

## Additional Context

From initial inspection, it appears the component is attempting to fetch version information regardless of whether it is running:
- As the standalone Kedro Viz app (with backend), or
- As an embedded React component.

The component should ideally detect embedded mode and skip version checks unless explicitly requested.

## Your Environment

* NodeJS version used (if relevant): NodeJS 22.15.0, @quantumblack/kedro-viz 11.0.2

## Checklist

jitu5:
## Description

Fixes [#2378](https://github.com/kedro-org/kedro-viz/issues/2378).

Kedro-Viz previously called `/api/version` and `/api/run-status` on load regardless of how it was running. Two contexts have no backend to answer those requests:

- **Embedded as a React component** with data supplied directly—the reported bug.
- **Static `kedro viz build` / `deploy` output** (for example, demo.kedro.org), where `/api/version` was not written as a static file and returned 404.

## Development notes

**Frontend** — skip both calls when no backend is available:

- `wrapper.jsx`: only call `/api/version` when `dataSource === 'json'`.
- `app.jsx`: only dispatch `loadInitialRunStatusData` when `data === 'json' && !runData`.

**Backend** — write and serve `api/version` consistently with the other saved API responses:

- `save_responses.py`: adds `save_api_version_response_to_fs`, wired into `save_api_responses_to_fs`.
- `version.py`: adds `get_static_version_response()`, which reports only the installed version and does not make a live PyPI request. `is_outdated` is therefore always `false` for static exports.
- `apps.py`: adds `/api/version` to the file-backed API app, so a Viz saved with `kedro viz run --save-file` can be loaded with `kedro viz run --load-file` without a version-endpoint 404.

## QA notes

- Embedded (`data={...}`, no backend): no `/api/version` or `/api/run-status` requests.
- Standalone (`kedro viz run`): both requests still run as before.
- `kedro viz build` / `deploy` output: `/api/version` returns 200 instead of 404.
- Saved-file mode: `/api/version` returns the saved version response when loaded through `kedro viz run --load-file`.
- Tests cover the frontend gates, static version response saving, and the file-backed version endpoint.

## Checklist


ravi-kumar-pilla (approved):
LGTM.. thank you @jitu5

src/components/app/app.jsx
```diff
@@ -45,8 +45,8 @@ class App extends React.Component {
       this.store.dispatch(loadInitialPipelineData());
     }
 
-    // If runData is not provided, load it from the API
-    if (!this.props.runData) {
+    // Only fetch run status when a backend is actually available
+    if (this.props.data === 'json' && !this.props.runData) {
```
Huongg: instead of doing a check here and `Wrapper` where you also do `dataSource !== 'json'`

Can we have something in the `src/utils/data-source` to do 

```
/**
 * Determine whether there's a live backend to fetch pipeline/run-status
 */
export const isBackendAvailable = (dataSource) => dataSource === 'json';
```

Huongg (commented):
hey @jitu5 thanks for the fix, it looks good to me, just 1 minor thing about duplicating ad hoc rather than reusing the existing pattern

Huongg (approved):
thank you @jitu5 LGTM 👍

Issue #2661, opened 2026-05-26
Preserve lite-mode behavior with inspection graphs and deferred live metadata


jitu5:
### Description

**Parent issue:** #2265
**Built on:** #2659, #2687, #2763, #2764, #2773, #2774

> **Reduced V1 scope:** Keep inspection-backed graphs and the existing live-backed metadata endpoint working together in lite mode. Snapshot-backed node metadata and context-backed export are deferred to #2775 and #2776.

## Summary

Complete and verify `kedro viz run --lite` with the wiring already in place:

- Main and pipeline graphs come from inspection.
- Node metadata remains on the legacy endpoint.
- Without hooks, #2774 defers live loading until metadata or legacy export needs it.
- With hooks enabled, live project loading can still happen eagerly.
- Run status reads the project’s event file through the context.

#2773 and #2774 are merged. This issue covers the remaining compatibility fixes and integrated verification, not another routing migration.

The metadata panel stays available. Deferred loading does not make node metadata session-free.

## Scope

- Keep real dependencies such as NumPy and pandas loaded across inspection and legacy loading. Remove temporary mocks without affecting unrelated modules.
- Prevent repeated project initialization from breaking the first metadata request, including settings that register an OmegaConf resolver.
- Check that temporary mocks do not leak into later imports or other requests.
- Preserve missing-dependency fallback and lazy, factory and transcoded dataset handling.
- Pass `--env`, `--pipeline` and typed `--params` consistently through inspection and legacy loading.
- Keep graph IDs and metadata consistent when pipeline definitions are evaluated again.
- Handle first, repeated and concurrent metadata requests safely. Define failure/retry behavior without exposing partially loaded data.
- Preserve stats, styles, dataset icons, layers and graph parameters.
- Preserve the `--include-hooks` behavior, project-bound run status and existing export behavior.
- Test startup through the first metadata request, not just graph loading.
- Verify process shutdown and distinguish application failures from test-harness or unrelated CI failures.

## Acceptance criteria

- [ ] Without hooks, startup and graph requests do not trigger the legacy live loader.
- [ ] The first metadata request loads the required live data; repeated or concurrent requests do not repeat a successful load.
- [ ] Failed loading returns a clear error, leaves no partially usable state and follows a tested retry policy.
- [ ] NumPy and pandas remain importable across both loading stages.
- [ ] A project registering an OmegaConf resolver in its settings works through startup and the first metadata request.
[...]

PR #2755, opened 2026-08-26
Cache the lite-mode project import scan

jitu5:
## Description

Both the live lite loader and the inspection snapshot stubs need the same unresolved-project-import list. Cache that scan by project path and package so starting Kedro-Viz in lite mode does not walk the project twice.

This is an independent optimisation extracted from #2750. It does not change the graph or endpoint behaviour.

- Add a cached `unresolved_modules()` helper.
- Reuse it from the live loader and snapshot import stubs.
- Keep the existing missing-dependency mocking and user-facing lite warning.
- Add focused coverage for flattened results and cache keys.

## Verification

- `pytest package/tests/test_integrations/test_lite_parser.py`: 25 passed
- `pytest package/tests/test_inspection_adapter/test_snapshot_source.py -k lite_import_stubs`: 2 passed
- `pytest package/tests/test_integrations/test_kedro_data_loader.py`: 23 passed
- `ruff check`: clean

### DIFF
--- a/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
+++ b/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
@@ -37,6 +37,7 @@ def lite_import_stubs(
         sys_modules_patch.update(
             LiteParser(package_name).create_mock_modules(modules_to_mock)
         )
+        # TODO(#2724): Restore this warning when the live loader is removed.
         # The live loader already warned the user about these modules.
         logger.debug(
             "Building the snapshot with %d project dependency module(s) mocked:\n%s",
--- a/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
+++ b/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
@@ -22,23 +22,24 @@ def lite_import_stubs(
     import sys
     from unittest.mock import patch
 
-    from kedro_viz.integrations.kedro.lite_parser import LiteParser
+    from kedro_viz.integrations.kedro.lite_parser import (
+        LiteParser,
+        unresolved_modules,
+    )
     from kedro_viz.models.metadata import Metadata
 
-    lite_parser = LiteParser(package_name)
-    unresolved = lite_parser.parse(Path(project_path)) or {}
-    modules_to_mock: set[str] = set()
-    for module_set in unresolved.values():
-        modules_to_mock |= module_set
+    modules_to_mock = set(unresolved_modules(str(project_path), package_name))
 
     sys_modules_patch = sys.modules.copy()
     if modules_to_mock:
         # Same banner the live --lite loader sets, so the UI flags limited functionality.
         Metadata.set_has_missing_dependencies(True)
-        sys_modules_patch.update(lite_parser.create_mock_modules(modules_to_mock))
-        logger.warning(
-            "Kedro-Viz --lite: building the snapshot with %d project dependency module(s) "
-            "mocked. Install them for full functionality:\n%s",
+        sys_modules_patch.update(
+            LiteParser(package_name).create_mock_modules(modules_to_mock)
+        )
+        # The live loader already warned the user about these modules.
+        logger.debug(
+            "Building the snapshot with %d project dependency module(s) mocked:\n%s",
             len(modules_to_mock),
             sorted(modules_to_mock),
         )
--- a/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
+++ b/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
@@ -24,11 +24,11 @@ def lite_import_stubs(
 
     from kedro_viz.integrations.kedro.lite_parser import (
         LiteParser,
-        unresolved_modules,
+        get_unresolved_modules,
     )
     from kedro_viz.models.metadata import Metadata
 
-    modules_to_mock = set(unresolved_modules(str(project_path), package_name))
+    modules_to_mock = set(get_unresolved_modules(project_path, package_name))
 
     sys_modules_patch = sys.modules.copy()
     if modules_to_mock:
@@ -37,7 +37,7 @@ def lite_import_stubs(
         sys_modules_patch.update(
             LiteParser(package_name).create_mock_modules(modules_to_mock)
         )
-        # TODO(#2724): Restore this warning when the live loader is removed.
+        # TODO(#2661): Restore this warning when --lite startup skips the live loader.
         # The live loader already warned the user about these modules.
         logger.debug(
             "Building the snapshot with %d project dependency module(s) mocked:\n%s",
--- a/package/kedro_viz/integrations/kedro/lite_parser.py
+++ b/package/kedro_viz/integrations/kedro/lite_parser.py
@@ -3,13 +3,35 @@
 import ast
 import importlib.util
 import logging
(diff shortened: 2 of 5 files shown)


package/kedro_viz/integrations/kedro/lite_parser.py
```diff
@@ -3,15 +3,37 @@
...
 from unittest.mock import MagicMock
 
 from kedro_viz.utils import Spinner, is_file_ignored, load_gitignore_patterns
 
 logger = logging.getLogger(__name__)
 
 
+@lru_cache(maxsize=None)
+def unresolved_modules(
+    project_path: str, package_name: Union[str, None] = None
```
ravi-kumar-pilla, line 18: **Suggestion:** Consider normalising `project_path` at the top of this helper (e.g. `project_path = str(Path(project_path).resolve())`) so the cache key is stable if callers pass relative vs absolute paths (or symlinked equivalents). Both call sites currently use `str(project_path)`, so this is low risk today, but cheap hardening for the optimisation goal.

package/tests/test_integrations/test_lite_parser.py
```diff
@@ -248,3 +248,29 @@ def test_directory_parse_non_package_path(self, sample_project_path, mock_spinne
         # ignore files in other packages if
         # LiteParser is instantiated with a package_name
         assert len(unresolvable_imports) == 0
+
+
+class TestUnresolvedModules:
```
ravi-kumar-pilla, line 253: **Suggestion:** `TestUnresolvedModules` correctly calls `unresolved_modules.cache_clear()`, but other tests in this module (and `test_snapshot_source.py`) now hit the global `@lru_cache` without clearing it. Pytest's unique `tmp_path` values make cross-test leakage unlikely today; an autouse fixture that calls `cache_clear()` after each test would be belt-and-braces if paths are ever reused.

ravi-kumar-pilla (commented):
## Kedro PR Review

### Overview
- **Critical:** 0 | **Suggestions:** 2

Focused optimisation — looks good overall. Sharing one cached unresolved-import scan between the live lite loader and inspection snapshot stubs is a clean extraction from #2750. Behaviour appears unchanged; unit tests cover the core caching invariant.

### Critical (must fix before merge)
No issues found.

See inline comments.

### Notes
- The snapshot warning downgrade (`logger.warning` → `logger.debug`) with `TODO(#2724)` is sensible while the live loader still warns.
- Unit tests and lint (3.12) pass on the branch. E2E failures look likely base-branch (`feat/backend_v2`) rather than this diff — worth confirming before merge.

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -22,23 +22,25 @@ def lite_import_stubs(
     import sys
     from unittest.mock import patch
 
-    from kedro_viz.integrations.kedro.lite_parser import LiteParser
+    from kedro_viz.integrations.kedro.lite_parser import (
+        LiteParser,
+        unresolved_modules,
```
ravi-kumar-pilla, line 27:
```suggestion
        get_unresolved_modules,
```

package/kedro_viz/integrations/kedro/lite_parser.py
```diff
@@ -3,15 +3,37 @@
...
+from typing import Dict, FrozenSet, List, Set, Union
 from unittest.mock import MagicMock
 
 from kedro_viz.utils import Spinner, is_file_ignored, load_gitignore_patterns
 
 logger = logging.getLogger(__name__)
 
 
+@lru_cache(maxsize=None)
+def unresolved_modules(
```
ravi-kumar-pilla, line 17:
```suggestion
def get_unresolved_modules(
```

ravi-kumar-pilla:
Hi @jitu5 ,

**Transitional cache which can be removed with #2724**

Confirming my understanding: the `@lru_cache` on `unresolved_modules` exists only because backend v2 currently runs **two** lite-mode bootstraps at startup (live loader in `load_data` + `lite_import_stubs` for the inspection snapshot). 

Once [#2724](https://github.com/kedro-org/kedro-viz/issues/2724) removes the live loader from that path, only one caller will need the unresolved-import scan and the cache becomes unnecessary.

**Suggestion:** add a `TODO(#2724)` on the helper (e.g. above `@lru_cache`) so future readers know this is a bridge, not a permanent part of `--lite`:

```python
# TODO(#2724): Remove @lru_cache once the live loader is removed from startup.
# Only the snapshot path will need the scan; caching avoids a duplicate walk until then.
@lru_cache(maxsize=None)
def get_unresolved_modules(...):
```

The flattening helper itself may still be worth keeping (or inlining) after #2724 — it's only the cache that can go.

Can you confirm on this ?

Thank you

package/kedro_viz/integrations/kedro/lite_parser.py
jitu5: Good point. I’ve normalised the project path before entering the cached implementation. Normalising it inside the @lru_cache-decorated function would happen after the cache key was created, so I used a small public wrapper around the private cached function.

package/tests/test_integrations/test_lite_parser.py
jitu5: Agreed. I added automatic cache cleanup around the lite-parser tests and extended the existing snapshot-source fixture to clear the same cache.

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: Updated.

package/kedro_viz/integrations/kedro/lite_parser.py
jitu5: Updated.

jitu5:
> Hi @jitu5 ,
> 
> **Transitional cache which can be removed with #2724**
> 
> Confirming my understanding: the `@lru_cache` on `unresolved_modules` exists only because backend v2 currently runs **two** lite-mode bootstraps at startup (live loader in `load_data` + `lite_import_stubs` for the inspection snapshot).
> 
> Once [#2724](https://github.com/kedro-org/kedro-viz/issues/2724) removes the live loader from that path, only one caller will need the unresolved-import scan and the cache becomes unnecessary.
> 
> **Suggestion:** add a `TODO(#2724)` on the helper (e.g. above `@lru_cache`) so future readers know this is a bridge, not a permanent part of `--lite`:
> 
> ```python
> # TODO(#2724): Remove @lru_cache once the live loader is removed from startup.
> # Only the snapshot path will need the scan; caching avoids a duplicate walk until then.
> @lru_cache(maxsize=None)
> def get_unresolved_modules(...):
> ```
> 
> The flattening helper itself may still be worth keeping (or inlining) after #2724 — it's only the cache that can go.
> 
> Can you confirm on this ?
> 
> Thank you

@ravi-kumar-pilla 

Yes, that matches my understanding: the cache is transitional while lite startup performs the scan through both the live loader and the inspection snapshot path.

One small issue-number adjustment: the current scope of #2661 explicitly owns short-circuiting the live loader during `--lite` startup, so that is when the duplicate scan disappears. #2724 depends on #2661 and completes the broader legacy-backend removal afterwards.
I’ve therefore added `TODO(#2661)` above the cache and updated the existing warning TODO to #2661 as well. The flattening helper remains independent of that future cache removal.

Issue #2075, opened 2024-09-04
Include a feedback component


astrojuanlu:
(Assigning to @stephkaiser to fill the rest of the details)

stephkaiser:
Notes for engineering and designs can be found here - https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=2989-19349&t=jT2I8cMoewhOHFfy-4

PR #2756, opened 2026-08-26
Add help menu with links to Kedro support channels

### LINKED ISSUE #2096: Help users find Kedro Viz support channels
## Description

> AS A Kedro-Viz and Kedro user
> I WANT TO easily find a way to get support
> SO THAT I can get help troubleshooting issues

## Possible Implementation
(We should look at how other similar products do it)

## Possible Alternatives
A 🛟 icon on the bottom left that displays some sort of modal with links to GitHub and Slack

lifering

## Checklist

deepyaman:
## Description

Resolves #2096.

Adds a `?` button at the bottom of the global toolbar which opens a menu of Kedro-Viz support channels, so users have an in-app route to help instead of having to go looking for it.

Implemented per [Steph's Figma design](https://www.figma.com/design/3kSpvIO1veLKfy9qHxuXwF/Kedro-WIP?node-id=2989-19349), using the icons @Huongg exported to `feat/kedro-viz-support-channel`.

## Development notes

New `HelpMenu` component (`src/components/help-menu/`), rendered as the last item of `pipeline-global-control-toolbar`. It's an `IconButton` (tooltip: "Help and resources") plus a flyout anchored to the right of the toolbar, with four links:

| Item | Destination |
| --- | --- |
| Report problem | `issues/new?template=bug-report.md` |
| Share idea | `issues/new?template=feature-request.md` |
| Kedro Docs | `docs.kedro.org` |
| Kedro Slack | `slack.kedro.org` |

All four URLs live in `src/config.js` alongside the existing `KEDRO_VIZ_*_URL` constants.

@Huongg @stephkaiser — one copy question: the Figma labels this "Kedro Docs", so it now points at `docs.kedro.org` to match the label (and to match its "Kedro Slack" neighbour, which is org-wide). It was pointing at the Kedro-Viz docs, which contradicted the label. If the intent was Viz-specific docs, say so and I'll relabel it "Kedro-Viz docs" instead.

**Give feedback is deliberately left out** — per [Huong's note](https://github.com/kedro-org/kedro-viz/issues/2096#issuecomment-5408557262), it should wait for the reusable feedback component tracked in #2075. When that lands, adding it here is one entry in the `helpMenuItems` array.

### A note on the icons

Two things to flag about the Figma export on `feat/kedro-viz-support-channel`:

1. **The filenames didn't match what the SVGs draw**, so I renamed them to match rather than carry the mismatch forward:
   - `feedback.jsx` (? in a circle) → `help.jsx`, the toolbar button
   - `help.jsx` (open book) → `docs.jsx`, Kedro Docs
   - `report.jsx` (! in a speech bubble) and `slack.jsx` were already correct
   - `docs.jsx` (★ in a speech bubble) is the "Give feedback" icon — not included here, since that item isn't; it stays on the branch for whoever picks up #2075.
2. **There was no lightbulb for "Share idea"**, so I added `idea.jsx` from Material Symbols `emoji_objects` at weight 300, converted to the same 24×24 grid as the others. @Huongg / @stephkaiser — worth a glance to confirm it matches the Figma one, and happy to swap in the real export if you'd rather.

### Accessibility

The button is a **disclosure**, not a menu widget: `aria-expanded` and `aria-controls` on the toggle, plain links inside. Four links out to GitHub, docs and Slack don't need composite menu keyboard behaviour, and `role="menu"` would promise arrow-key/Home/End navigation that isn't there — so the roles are omitted rather than half-implemented. <kbd>Esc</kbd> closes and returns focus to the toggle, guarded on focus having been inside the menu.

This needed two new props on `IconButton` (`ariaControls`/`ariaExpanded`), which it forwards to the `<button>` alongside the `ariaLabel`/`ariaLive` it already handled — without them, `...rest` puts stray ARIA on the wrapper `<li>`.

## QA notes

`src/components/help-menu/help-menu.test.jsx` covers open/close via the button, `aria-expanded` tracking state, `aria-controls` resolving to the panel, the href/`target`/`rel` of each link, and closing on link click, outside click and <kbd>Esc</kbd> — plus focus returning to the toggle. The button-count assertion in `global-toolbar.test.jsx` goes 6 → 7; `icon-button.test.jsx` ends up unchanged from `main`, since the two new props are covered where they actually matter, through the menu.

Verified visually in both light and dark themes at 1440×900. The menu sits above the sidebar, and the button tooltip is suppressed while the menu is open so the two don't overlap. Panel background, shadow and hover reuse the variables the existing dropdown and menu-option styles already use, so it matches the app's other popup surfaces in both themes.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/components/help-menu/help-menu.jsx
```diff
@@ -0,0 +1,116 @@
...
+      window.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [isOpen]);
+
+  return (
+    <li className="pipeline-icon--container" ref={containerRef}>
+      <IconButton
+        active={isOpen}
+        ariaLabel={`${isOpen ? 'Close' : 'Open'} help and resources`}
+        ariaLive="polite"
```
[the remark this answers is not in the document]
deepyaman: Good catch on the state not being exposed — fixed in fecef4b, along with the related point about `role="menu"`. I took the second option you offered there (drop the roles) rather than building out menu keyboard support, and that decision shapes the rest of the fix:

- **`aria-expanded` / `aria-controls`** — agreed and added. `IconButton` now takes `ariaControls`/`ariaExpanded` props so they land on the `<button>`, alongside the `ariaLabel`/`ariaLive` it already forwards. The label is now static ("Help and resources") since `aria-expanded` carries the state; `aria-live` went away with it.
- **`role="menu"` / `role="menuitem"`** — dropped. You're right that they promised arrow-key/Home/End behaviour that isn't implemented. For four links out to GitHub, docs and Slack, a disclosure with native `Tab` is the correct pattern, so I removed the roles rather than building a composite widget around them.
- **Escape unmounting the focused element** — real bug, fixed. Escape now returns focus to the toggle, guarded on focus actually having been inside the menu so it can't yank focus from elsewhere in the app. Covered by a test.

One thing I've deliberately left out: **`aria-haspopup="menu"`**. Now that this isn't a menu, that attribute would re-assert the exact semantics we just removed and put us back where we started — announcing a menu to screen reader users and setting up the same expectation of arrow-key navigation. `aria-haspopup` is for genuine popup widgets (menu, listbox, tree, grid, dialog); a disclosure revealing a group of links isn't one, so `aria-expanded` plus `aria-controls` is the complete story here.

src/components/help-menu/help-menu.jsx
```diff
@@ -0,0 +1,118 @@
...
+];
+
+const menuId = 'help-menu';
+
+/**
+ * Global toolbar button which reveals a list of Kedro-Viz support channels
+ */
+export const HelpMenu = () => {
+  const [isOpen, setIsOpen] = useState(false);
+  const containerRef = useOutsideClick(() => setIsOpen(false));
```
ravi-kumar-pilla: **Suggestion:** Wrap the `useOutsideClick` callback in `useCallback` so the hook does not remove and re-add its document listener on every render. The hook depends on `[callback, ref]`, and an inline `() => setIsOpen(false)` gets a new identity each render.

```javascript
const closeMenu = useCallback(() => setIsOpen(false), []);
const containerRef = useOutsideClick(closeMenu);
```

src/components/help-menu/help-menu.test.jsx
```diff
@@ -0,0 +1,118 @@
...
+describe('HelpMenu', () => {
+  const getButton = () =>
+    screen.getByRole('button', { name: 'Help and resources' });
+
+  const openMenu = () => fireEvent.click(getButton());
+
+  it('renders the toolbar button with the menu closed', () => {
+    render(<HelpMenu />);
+
+    expect(getButton()).toHaveAttribute('aria-expanded', 'false');
```
ravi-kumar-pilla, line 14: **Suggestion:** Add an assertion that `aria-controls` is absent when the menu is closed. Commit `9b26e86` fixed advertising a control for an unmounted panel — a test here would guard against regression.

```javascript
expect(getButton()).not.toHaveAttribute('aria-controls');
```

ravi-kumar-pilla (commented):
## Kedro PR Review

### Overview
- **Critical:** 0 | **Suggestions:** 3

No blocking issues from this review — the implementation looks solid overall, and Copilot's earlier ARIA feedback appears fully addressed. Leaving as comment-only while the draft/design questions are still open.

- [ ] `src/components/help-menu/help-menu.jsx:L54` — Stabilize `useOutsideClick` callback with `useCallback`
- [ ] `src/components/help-menu/help-menu.test.jsx:L14` — Assert `aria-controls` is absent when closed
- [ ] PR description — Update stale QA note about `icon-button.test.jsx`

### Notes
- **General review:** Focused, well-scoped PR. Disclosure pattern (`aria-expanded` / `aria-controls`, plain links, no `role="menu"`) is the right call for four external links. Unit test coverage is strong.
- **PR checklist:** `RELEASE.md` updated correctly; all CI checks pass. Still marked draft — makes sense while awaiting design sign-off on the lightbulb icon and the Kedro Docs vs Kedro-Viz docs label/URL.
- **API usage:** N/A (frontend-only).

src/components/help-menu/help-menu.jsx
```diff
@@ -0,0 +1,118 @@
...
+    const handleKeyDown = (event) => {
+      if (event.key !== 'Escape') {
+        return;
+      }
+      // Closing unmounts whichever link has focus, so hand it back to the
+      // button rather than letting it fall through to the document body
+      const hadFocus = containerRef.current.contains(document.activeElement);
+      setIsOpen(false);
+      if (hadFocus) {
+        containerRef.current.querySelector('button').focus();
```
jitu5: [nit] `querySelector('button')` returns the first button in the container, which is
the toggle only because the panel happens to contain nothing but links. If
anything button-shaped lands in the panel later, focus restoration moves to the
wrong element.

src/components/help-menu/help-menu.jsx
deepyaman: Both fixed in `8b64107f`, thanks — you're right on the identity churn. `useOutsideClick` keys its effect on `[callback, ref]`, so the inline arrow had it detaching and reattaching the document listener on every render. Memoised exactly as you wrote it.

The `aria-controls` assertion is in too, folded into the existing "renders the toolbar button with the menu closed" case rather than a new one, since that test already pins the closed state:

```js
expect(getButton()).toHaveAttribute('aria-expanded', 'false');
expect(getButton()).not.toHaveAttribute('aria-controls');
expect(screen.queryByRole('link')).not.toBeInTheDocument();
```

And the stale QA note is corrected — `icon-button.test.jsx` ends up unchanged from `main`; I dropped the case I'd added there because it only checked React forwarded two props, which the menu's own tests already cover where it matters.

src/components/help-menu/help-menu.jsx
deepyaman: Fair — that was load-bearing on an assumption I never stated. Fixed in `8b64107f` by matching the toggle as a direct child of the container, which is true by construction rather than by what the panel happens to contain:

```js
containerRef.current.querySelector(':scope > button').focus();
```

The panel is a sibling `div`, so anything button-shaped added inside it now can't shadow the toggle. Verified in Chrome as well as jsdom — `:scope > button` resolves to the toggle with the menu open, and Escape from a focused link lands focus back on it.

The clean version of this is a ref rather than a query, but `IconButton` isn't `forwardRef` and making it one to serve a single caller felt like more churn than the nit warranted. Happy to do it if you'd rather.

deepyaman:
Didn't expect the reviews already, but they're addressed, and I've also tested it myself locally, and it seems to work.

To be completely honest, in retrospect I probably shouldn't have picked this up, since I don't know much about the Viz codebase. But the code looks reasonable, it seems to work, and I've tried to ensure consistency with the existing codebase—I'd just be lying if I said I really understood the tradeoffs on everything that's produced. 😅

src/components/help-menu/help-menu.jsx
```diff
@@ -0,0 +1,120 @@
...
+    // IconButton renders this wrapper itself, but it cannot forward a ref to
+    // it, and both the outside click and the focus restoration need one
+    <li className="pipeline-icon--container" ref={containerRef}>
+      <IconButton
+        active={isOpen}
+        ariaControls={isOpen ? menuId : undefined}
+        ariaExpanded={isOpen}
+        ariaLabel="Help and resources"
+        className="pipeline-menu-button--help pipeline-menu-button--large"
+        container={React.Fragment}
```
Huongg: i dont think you need this IconButton has its own <li> wrapper

src/components/help-menu/help-menu.jsx
```diff
@@ -0,0 +1,120 @@
...
+  },
+  {
+    href: KEDRO_SLACK_URL,
+    icon: SlackIcon,
+    id: 'kedro-slack',
+    label: 'Kedro Slack',
+  },
+];
+
+const menuId = 'help-menu';
```
Huongg: maybe don't hard coded the menu Id as if we re-use this component somewhere else it will collide if its rendered twice

Huongg (approved):
thank you @deepyaman I tested the new icon component locally and all working as expected to me. Just left 2 small comments, we have a release for viz soon so if you want to squeeze this change in it would be great too 😄

jitu5 (approved):
Thanks @deepyaman

PR #2763, opened 2026-09-04
Centralize inspection inputs in `VizProjectContext`

jitu5:
Part 1 of 6 for #2660. Existing native GitHub stack #2769.

Please start review here. The other five PRs remain drafts and are available for context or early feedback.

## Description

Move inspection loading to the project context so services can share the same prepared inputs. This follows Ravi’s clarified composition-root design.

## Development notes

- Add the frozen Pydantic `InspectionInputs` model with only `snapshot`, `catalog_config` and `parameters`.
- Use module-level `load_inspection_inputs()` and `filter_inspection_inputs()` helpers. The context coordinates loading and selection; the graph service consumes prepared inputs.
- Preserve existing parameter behavior and `GraphBuilder(parameters=...)`. No annotation validation, flattened feed or typed-field expansion is introduced.
- Keep Kedro inspection imports lazy for lite mode.
- Use explicit `inspection_inputs` and `enrichment_sources` names. Export only `EnrichmentSources`, `PipelineNotFoundError` and `VizProjectContext`; tests import internal classes from their defining modules.

## QA notes

Tests cover shared inputs, filtering and errors, runtime parameters, mapping copies, lazy imports, and projects with parameter values that do not match pipeline annotations. The six existing graph baselines are unchanged.

Local verification at `9ee1bcf84`: **695 tests passed with 100% statement coverage** on both the repository-pinned Kedro revision and released Kedro 1.6.0, using Python 3.13. Ruff, formatting and both MyPy targets passed.

These are local results, not a claim that GitHub CI has passed on the rewritten head. Earlier revisions had an E2E packaging setup failure involving Hatch and direct dependency references; dependency/packaging configuration is unchanged here. Read the Docs also previously failed and has not been diagnosed as part of this correction. Check the current-head runs below for CI status.

## Scope and checklist

- [x] One corrected commit on the previous stack layer

### DIFF
--- a/package/kedro_viz/integrations/kedro/inspection/__init__.py
+++ b/package/kedro_viz/integrations/kedro/inspection/__init__.py
@@ -2,16 +2,10 @@
 
 from kedro_viz.integrations.kedro.inspection.context import VizProjectContext
 from kedro_viz.integrations.kedro.inspection.enrichment import EnrichmentSources
-from kedro_viz.integrations.kedro.inspection.graph_builder import GraphBuilder
-from kedro_viz.integrations.kedro.inspection.graph_service import (
-    InspectionGraphService,
-    PipelineNotFoundError,
-)
+from kedro_viz.integrations.kedro.inspection.errors import PipelineNotFoundError
 
 __all__ = [
     "EnrichmentSources",
-    "GraphBuilder",
-    "InspectionGraphService",
     "PipelineNotFoundError",
     "VizProjectContext",
 ]
--- a/package/kedro_viz/integrations/kedro/inspection/context.py
+++ b/package/kedro_viz/integrations/kedro/inspection/context.py
@@ -9,6 +9,10 @@ from kedro_viz.integrations.kedro.inspection.enrichment import EnrichmentSources
 from kedro_viz.integrations.kedro.inspection.graph_service import (
     InspectionGraphService,
 )
+from kedro_viz.integrations.kedro.inspection.snapshot_source import (
+    filter_inspection_inputs,
+    load_inspection_inputs,
+)
 
 
 class VizProjectContext:
@@ -47,14 +51,20 @@ class VizProjectContext:
         Raises:
             PipelineNotFoundError: If ``pipeline_name`` is not registered.
         """
+        inspection_inputs = load_inspection_inputs(
+            project_path,
+            env=env,
+            runtime_params=runtime_params,
+            package_name=package_name,
+            is_lite=is_lite,
+        )
+        if pipeline_name is not None:
+            inspection_inputs = filter_inspection_inputs(
+                inspection_inputs, pipeline_name
+            )
         return cls(
-            graph=InspectionGraphService.from_project(
-                project_path,
-                env=env,
-                pipeline_name=pipeline_name,
-                runtime_params=runtime_params,
-                package_name=package_name,
-                is_lite=is_lite,
+            graph=InspectionGraphService.from_inspection_inputs(
+                inspection_inputs,
                 enrichment=enrichment,
             )
         )
--- a/package/kedro_viz/integrations/kedro/inspection/graph_service.py
+++ b/package/kedro_viz/integrations/kedro/inspection/graph_service.py
@@ -2,28 +2,16 @@
 
 from __future__ import annotations
 
-import dataclasses
-from contextlib import nullcontext
-from pathlib import Path
-from typing import TYPE_CHECKING, Any
-
 from kedro_viz.api.rest.responses.pipelines import GraphAPIResponse
 from kedro_viz.integrations.kedro.inspection.enrichment import (
     EnrichmentSources,
     enrich_graph_response,
 )
+from kedro_viz.integrations.kedro.inspection.errors import PipelineNotFoundError
 from kedro_viz.integrations.kedro.inspection.graph_builder import GraphBuilder
 from kedro_viz.integrations.kedro.inspection.snapshot_source import (
-    _InspectionSession,
-    lite_import_stubs,
+    InspectionInputs,
 )
-
-if TYPE_CHECKING:
-    from kedro.inspection.models import ProjectSnapshot
-
-
-class PipelineNotFoundError(ValueError):
-    """Raised when a requested pipeline is not present in the inspection snapshot."""
 
 
 class InspectionGraphService:
@@ -38,52 +26,23 @@ class InspectionGraphService:
         self._enrichment = enrichment if enrichment is not None else EnrichmentSources()
 
     @classmethod
-    def from_project(
+    def from_inspection_inputs(
         cls,
-        project_path: str | Path,
+        inspection_inputs: InspectionInputs,
         *,
-        env: str | None = None,
-        pipeline_name: str | None = None,
-        runtime_params: dict[str, Any] | None = None,
-        package_name: str | None = None,
-        is_lite: bool = False,
         enrichment: EnrichmentSources | None = None,
     ) -> InspectionGraphService:
-        """Read one project snapshot and prepare the graph service.
-
-        Args:
-            project_path: The Kedro project root.
-            env: The Kedro environment, honouring ``--env``.
-            pipeline_name: Restrict the view to one registered pipeline, honouring
-                ``--pipeline``.
-            runtime_params: Typed parameter overrides from ``--params``.
-            package_name: Project package used to identify imports in lite mode.
-            is_lite: Whether missing project dependencies should be temporarily mocked.
-            enrichment: Explicit fields supplied by the transitional live load.
-
-        Raises:
-            PipelineNotFoundError: If ``pipeline_name`` is not registered.
-        """
-        sources = enrichment if enrichment is not None else EnrichmentSources()
-        import_context = (
-            lite_import_stubs(project_path, package_name) if is_lite else nullcontext()
+        """Prepare the graph service from already-loaded inspection inputs."""
+        enrichment_sources = (
+            enrichment if enrichment is not None else EnrichmentSources()
         )
-        with import_context:
-            session = _InspectionSession(
-                project_path, env=env, runtime_params=runtime_params
-            )
-            snapshot = session.snapshot()
-            catalog_config = session.catalog_config()
-            parameters = session.parameters()
-            if pipeline_name is not None:
-                snapshot = cls._filter_to_pipeline(snapshot, pipeline_name)
-            builder = GraphBuilder(
-                snapshot,
-                catalog_config,
-                parameters=parameters,
-                layer_by_dataset=sources.layer_by_dataset,
-            )
-        return cls(builder, sources)
+        builder = GraphBuilder(
+            inspection_inputs.snapshot,
+            dict(inspection_inputs.catalog_config),
+            parameters=dict(inspection_inputs.parameters),
+            layer_by_dataset=enrichment_sources.layer_by_dataset,
+        )
+        return cls(builder, enrichment_sources)
 
     def get_pipeline_response(self, pipeline_id: str | None = None) -> GraphAPIResponse:
         """Return the graph for one registered pipeline.
@@ -101,24 +60,3 @@ class InspectionGraphService:
         response = self._builder.build(selected_pipeline_id)
         enrich_graph_response(response, self._enrichment)
         return response
-
-    @staticmethod
-    def _filter_to_pipeline(
-        snapshot: ProjectSnapshot, pipeline_name: str
-    ) -> ProjectSnapshot:
-        """Return a copy of the snapshot containing only the requested pipeline.
-
-        Raises:
-            PipelineNotFoundError: If ``pipeline_name`` is not registered.
-        """
-        matching = [
-            pipeline
-            for pipeline in snapshot.pipelines
-            if pipeline.name == pipeline_name
-        ]
-        if not matching:
-            available = sorted(pipeline.name for pipeline in snapshot.pipelines)
-            raise PipelineNotFoundError(
-                f"Pipeline {pipeline_name!r} not found in snapshot; available: {available}"
-            )
-        return dataclasses.replace(snapshot, pipelines=matching)
--- a/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
+++ b/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
@@ -2,16 +2,40 @@
 
 from __future__ import annotations
 
+import dataclasses
 import logging
-from collections.abc import Iterator
-from contextlib import contextmanager
+from collections.abc import Iterator, Mapping
+from contextlib import contextmanager, nullcontext
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
+
+from pydantic import BaseModel, Field, field_validator
+
+from kedro_viz.integrations.kedro.inspection.errors import PipelineNotFoundError
 
 if TYPE_CHECKING:
     from kedro.inspection.models import ProjectSnapshot
 
 logger = logging.getLogger(__name__)
+
+
+class InspectionInputs(BaseModel, frozen=True):
+    """Snapshot and resolved config shared by a project's inspection services."""
+
+    # Kedro has already built the snapshot. Keep it opaque to Pydantic so importing
+    # this model does not load Kedro inspection modules before lite-mode stubs.
+    if TYPE_CHECKING:
+        snapshot: ProjectSnapshot
+    else:
+        snapshot: Any
+    catalog_config: Mapping[str, Any] = Field(default_factory=dict)
+    parameters: Mapping[str, Any] = Field(default_factory=dict)
+
+    @field_validator("catalog_config", "parameters", mode="before")
+    @classmethod
+    def _copy_mapping(cls, value: Mapping[str, Any]) -> dict[str, Any]:
+        """Copy caller-owned mappings before storing the prepared inputs."""
+        return dict(value)
 
 
 @contextmanager
@@ -48,11 +72,10 @@ def lite_import_stubs(
 
 
 class _InspectionSession:
-    """Read a project's snapshot and config, bootstrapping and building the loader once.
+    """Read a project's snapshot and config with one cached config loader.
 
-    Create one session per adapter build. The project is bootstrapped and the Kedro config loader
-    is built lazily on first use and then cached, so the catalog config and parameters reuse a
-    single loader instead of rebuilding it.
+    The catalog config and parameters share the loader built lazily here. Kedro's snapshot API
+    performs its own internal project bootstrap and config loading.
     """
 
     def __init__(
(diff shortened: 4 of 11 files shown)


package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -47,14 +51,18 @@ def from_project(
         Raises:
             PipelineNotFoundError: If ``pipeline_name`` is not registered.
         """
+        project_data = load_inspection_project_data(
```
ravi-kumar-pilla:
```suggestion
        project_inspection_data = load_project_inspection_data(
```

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -47,14 +51,18 @@ def from_project(
...
+            project_path,
+            env=env,
+            runtime_params=runtime_params,
+            package_name=package_name,
+            is_lite=is_lite,
+        )
+        if pipeline_name is not None:
+            project_data = filter_inspection_project_data(project_data, pipeline_name)
```
ravi-kumar-pilla:
```suggestion
            project_inspection_data = filter_project_inspection_data(project_data, pipeline_name)
```

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -47,14 +51,18 @@ def from_project(
...
         return cls(
-            graph=InspectionGraphService.from_project(
-                project_path,
-                env=env,
-                pipeline_name=pipeline_name,
-                runtime_params=runtime_params,
-                package_name=package_name,
-                is_lite=is_lite,
+            graph=InspectionGraphService.from_project_data(
```
ravi-kumar-pilla:
```suggestion
            graph=InspectionGraphService.from_project_inspection_data(
```

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -47,14 +51,18 @@ def from_project(
...
+                project_data,
```
ravi-kumar-pilla:
```suggestion
                project_inspection_data,
```

package/kedro_viz/integrations/kedro/inspection/context.py
ravi-kumar-pilla: Rename it to avoid conflict between actual project data

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
```diff
@@ -109,7 +110,7 @@ def __init__(
         snapshot: ProjectSnapshot,
         catalog_config: dict[str, Any] | None = None,
         *,
-        parameters: dict[str, Any] | None = None,
+        parameter_feed: Mapping[str, Any] | None = None,
```
ravi-kumar-pilla: any reason to change this ?

package/kedro_viz/integrations/kedro/inspection/graph_service.py
```diff
@@ -38,51 +26,20 @@ def __init__(
         self._enrichment = enrichment if enrichment is not None else EnrichmentSources()
 
     @classmethod
-    def from_project(
+    def from_project_data(
```
ravi-kumar-pilla:
```suggestion
    def from_project_inspection_data(
```

package/kedro_viz/integrations/kedro/inspection/graph_service.py
```diff
@@ -38,51 +26,20 @@ def __init__(
...
         cls,
-        project_path: str | Path,
+        project_data: InspectionProjectData,
```
ravi-kumar-pilla:
```suggestion
        project_inspection_data: InspectionProjectData,
```

package/kedro_viz/integrations/kedro/inspection/parameters.py
```diff
@@ -0,0 +1,55 @@
...
+    parameter_feed: Mapping[str, Any],
+) -> dict[str, Any]:
+    """Return values for parameter inputs using Kedro's prepared reference feed."""
+    result: dict[str, Any] = {}
+    for reference in inputs:
+        if reference == "parameters":
+            result = dict(parameter_feed.get(reference, {}))
+        elif reference.startswith("params:"):
+            result[reference.removeprefix("params:")] = parameter_feed.get(reference)
+    return result
```
ravi-kumar-pilla: why did we change all the parameter handling which we had earlier to this PR ?

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -2,18 +2,55 @@
...
 logger = logging.getLogger(__name__)
 
 
+class InspectionProjectData(BaseModel, frozen=True):
+    """Snapshot and resolved config read together for one inspection context build."""
+
+    snapshot: ProjectSnapshot
+    catalog_config: Mapping[str, Any] = Field(default_factory=dict)
+    parameters: Mapping[str, Any] = Field(default_factory=dict)
+    parameter_feed: Mapping[str, Any] = Field(default_factory=dict)
```
ravi-kumar-pilla: A bit confusing on why we have 2 parameter handling variables

package/kedro_viz/integrations/kedro/inspection/context.py
ravi-kumar-pilla: what was the reason to extract data here ? Can we have a similar architecture as before i.e., VizProjectContext will provide the required context of the project and then the individual apis will construct the data they need. If there is a need to reuse functions around data creation, it should be done via a service and not within this class. 

This class is supposed to do only 1 thing - provide context. Loading project data here pollutes the purpose.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla, line 1: I think we need to stick to the architecture here - https://github.com/kedro-org/kedro-viz/pull/2750#issuecomment-5460527238

## Core idea: one `VizProjectContext`, many consumers

```text
run_server / CLI entry point
    │
    ▼
VizProjectContext.from_project(...)
    │
    ├─► create_api_app(context)          → HTTP (routes close over context)
    ├─► save_api_responses_to_fs(context) → export / --save-file
    ├─► get_kedro_project_json_data(context) → VSCode
    └─► NotebookContext.from_...(context)    → notebook (future)
    │
    ▼
context.graph.get_pipeline_response()
    │
    InspectionGraphService
    │
    ├─► GraphBuilder (topology)
    └─► enrich_graph_response (EnrichmentSources)
```

For node metadata it should be -

```
context.graph.get_node_metadata_response(node_id) 
-> InspectionGraphService -> NodeMetadataBuilder
```

ravi-kumar-pilla: GraphBuilder should be responsible for only building the flowchart...similarly we need to apply this for other files. This way we avoid strong coupling and it is easy for us to maintain

package/kedro_viz/integrations/kedro/inspection/parameters.py
jitu5: I introduced the shared parameter lookup so graph and snapshot-backed node metadata resolve parameter references consistently, including typed fields and literal/nested dotted keys.

I included it here because both builders would consume the prepared inspection inputs. However, I can see that it makes this context-loading PR broader than necessary.

Would you prefer moving the parameter preparation changes and their tests into #2765, alongside the static metadata builder, and keeping #2763 focused on context loading? I’ve left that part unchanged pending your feedback.

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
jitu5: `parameters` contains the resolved nested values; `parameter_feed` is a lookup derived from them, keyed by pipeline references such as `params:model.test_size`. They aren’t independent parameter sources.

I agree that storing both on `InspectionProjectData` makes this confusing. I propose keeping only `parameters` there and deriving the feed once in `InspectionGraphService`, where both builders can reuse it. That would make the relationship explicit without changing the lookup behavior.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
ravi-kumar-pilla: Hi @jitu5, 

The direction of loading the project once at the boundary is right, and I want to make sure my inline comments don't read as "revert this." A few things to align on before this lands:

**1. Split out the parameter rewrite.**

The stated scope is "move project loading into VizProjectContext." But this PR also re-architects parameter handling that #2736 only just landed: ParameterValidator (pipeline-annotation validation), build_parameter_feed() with recursive typed-field expansion, parameters_for_inputs(), and the GraphBuilder(parameters=...) → parameter_feed=... change. That is a behavior change, the graph endpoint can now raise on parameter validation for a project that renders fine today and it deserves its own PR with its own parity baseline against #2736. Can we reduce this PR to InspectionProjectData{snapshot, catalog_config, parameters} with today's semantics, and keep GraphBuilder's constructor unchanged?

---

**2. VizProjectContext.from_project — keep it as a thin composition root.**

Loading once here is correct and consistent with the #2750 design (no globals, explicit passing). What I would like: from_project should just load → optionally filter → construct each service, which is basically what it does now via load_inspection_project_data(). No need to push loading back down into each service. My concern is only that the loading logic stays out of the class body — a stateless module function is fine for that.

```
@classmethod
def from_project(cls, project_path, *, env=None, pipeline_name=None, ...):
    data = load_inspection_project_data(project_path, env=env, ...)   # one read
    if pipeline_name is not None:
        data = filter_inspection_project_data(data, pipeline_name)
    return cls(
        graph=InspectionGraphService.from_project_data(data),
        nodes=NodeMetadataService.from_project_data(data),   # same object
    )
```
---

**3. One parameter representation, not two.**
parameters and parameter_feed on the same frozen model are mutually derivable and will drift. Pick parameters as canonical and derive any feed view on demand in the consumer. Related: building the feed inside a @model_validator(mode="before") hides non-trivial computation in Pydantic validation — do it explicitly in the loader instead.

---

**4. Import hygiene on the lite path.**
snapshot_source.py now imports ProjectSnapshot at module top level (was TYPE_CHECKING), and parameters.py hard-imports kedro.validation.utils at load time. The rest of this module keeps Kedro imports lazy so lite mode can mock modules first. Please keep new Kedro imports inside the functions.

---

**5. Naming.**

Can we name these after what they carry rather than stacking three nouns? 

InspectionInputs for the model and load_inspection_inputs() / filter_inspection_inputs() for the helpers — avoids colliding with Kedro's "project data" and each name reads on its own.

package/kedro_viz/integrations/kedro/inspection/context.py
jitu5: I put loading here initially so the graph and metadata services could share one project load. Agreed that the context shouldn’t own that preparation.

I’ve moved loading, filtering, and node-extras preparation into `InspectionGraphService` in the local revision. `VizProjectContext.from_project()` now delegates to the service and wires up the context. The prepared inputs are reused across requests without reloading the project. This hasn’t been pushed yet.

package/kedro_viz/integrations/kedro/inspection/graph_builder.py
jitu5: Thanks @ravi-kumar-pilla , this makes the direction clearer.

1. **Parameter rewrite:** I’ll move it into a separate PR with its own parity checks against #2736. This PR will keep the existing parameter behavior and `GraphBuilder(parameters=...)` constructor.

2. **Context responsibility:** I’d interpreted the earlier comment as moving loading into the service. I’ll adjust that: `from_project()` will coordinate one load, optional filtering, and service construction. The loading logic will stay in module-level helpers.

3. **One parameter representation:** I’ll keep only `parameters` on the inputs model and remove the stored `parameter_feed` and its model validator. Any derived lookup will be explicit where needed.

4. **Lite-path imports:** I’ll move the new Kedro imports into the functions that need them and verify the lite path.

5. **Naming:** I’ll use `InspectionInputs`, `load_inspection_inputs()`, and `filter_inspection_inputs()` consistently.



One clarification on metadata ownership: the earlier comment suggested `context.graph.get_node_metadata_response()`, while the new example shows a separate `nodes` service. Which structure would you prefer?

ravi-kumar-pilla: > One clarification on metadata ownership: the earlier comment suggested context.graph.get_node_metadata_response(), while the new example shows a separate nodes service. Which structure would you prefer?

I would prefer `NodeMetadataService` now that context holds the load once. This feels clean. wdyt ?

jitu5: Thanks @ravi-kumar-pilla , I’ve pushed the corrections across the stack.

 1. Removed the parameter rewrite. #2763 preserves existing behavior and `GraphBuilder(parameters=...)`. #2765 only extracts the existing lookup for reuse, without changing its logic.
 2. Kept the context as a thin composition root: load once, optionally filter, then construct the services using the same inputs.
 3. Kept only `parameters` on `InspectionInputs`, with no stored feed or feed-building validator.
 4. Kept Kedro inspection imports lazy and verified the lite path.
 5. Applied `InspectionInputs`, `load_inspection_inputs()` and `filter_inspection_inputs()` consistently.

 Metadata now uses the separate `context.nodes` / `NodeMetadataService`, as you suggested. The graph service stays graph-only.

 Local verification passes on pinned Kedro and Kedro 1.6, with unchanged metadata parity expectations. #2763 is ready for review; the remaining PRs stay in draft.

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -2,18 +2,42 @@
...
+from kedro_viz.integrations.kedro.inspection.errors import PipelineNotFoundError
+
 if TYPE_CHECKING:
     from kedro.inspection.models import ProjectSnapshot
 
 logger = logging.getLogger(__name__)
 
 
+class InspectionInputs(BaseModel, frozen=True):
+    """Snapshot and resolved config shared by one project's inspection services."""
```
ravi-kumar-pilla:
```suggestion
    """Snapshot and resolved config shared by a project's inspection services."""
```

package/kedro_viz/integrations/kedro/inspection/graph_service.py
```diff
@@ -38,51 +26,20 @@ def __init__(
         self._enrichment = enrichment if enrichment is not None else EnrichmentSources()
 
     @classmethod
-    def from_project(
+    def from_inspection_inputs(
         cls,
-        project_path: str | Path,
+        inputs: InspectionInputs,
```
ravi-kumar-pilla:
```suggestion
        inspection_inputs: InspectionInputs,
```

package/kedro_viz/integrations/kedro/inspection/graph_service.py
```diff
@@ -38,51 +26,20 @@ def __init__(
...
-            runtime_params: Typed parameter overrides from ``--params``.
-            package_name: Project package used to identify imports in lite mode.
-            is_lite: Whether missing project dependencies should be temporarily mocked.
-            enrichment: Explicit fields supplied by the transitional live load.
-
-        Raises:
-            PipelineNotFoundError: If ``pipeline_name`` is not registered.
-        """
+        """Prepare the graph service from already-loaded inspection inputs."""
         sources = enrichment if enrichment is not None else EnrichmentSources()
```
ravi-kumar-pilla:
```suggestion
        enrichment_sources = enrichment if enrichment is not None else EnrichmentSources()
```

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -113,3 +136,60 @@ def parameters(self) -> dict[str, Any]:
...
+            runtime_params=runtime_params,
+        )
+        return InspectionInputs(
+            snapshot=session.snapshot(),
+            catalog_config=session.catalog_config(),
+            parameters=session.parameters(),
+        )
+
+
+def filter_inspection_inputs(
```
ravi-kumar-pilla, line 171: Awesome !!

package/kedro_viz/integrations/kedro/inspection/snapshot_source.py
```diff
@@ -113,3 +136,60 @@ def parameters(self) -> dict[str, Any]:
             return self.config_loader["parameters"]
         except (KeyError, MissingConfigException):
             return {}
+
+
+def load_inspection_inputs(
```
ravi-kumar-pilla, line 141: Easy to follow and we have limited loading and filtering of inputs at the snapshot_source, instead of filtering and passing around the project. This looks great @jitu5

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -47,14 +51,18 @@ def from_project(
         Raises:
             PipelineNotFoundError: If ``pipeline_name`` is not registered.
         """
+        inputs = load_inspection_inputs(
```
ravi-kumar-pilla:
```suggestion
        inspection_inputs = load_inspection_inputs(
```

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -47,14 +51,18 @@ def from_project(
...
+            project_path,
+            env=env,
+            runtime_params=runtime_params,
+            package_name=package_name,
+            is_lite=is_lite,
+        )
+        if pipeline_name is not None:
+            inputs = filter_inspection_inputs(inputs, pipeline_name)
```
ravi-kumar-pilla:
```suggestion
            inspection_inputs = filter_inspection_inputs(inputs, pipeline_name)
```

package/kedro_viz/integrations/kedro/inspection/__init__.py
```diff
@@ -2,16 +2,20 @@
...
 )
 
 __all__ = [
     "EnrichmentSources",
     "GraphBuilder",
     "InspectionGraphService",
+    "InspectionInputs",
     "PipelineNotFoundError",
     "VizProjectContext",
 ]
```
ravi-kumar-pilla, line 11: we need to make sure to export only the required modules here (i.e., modules used across package outside of inspection). If we think there are no consumers for the modules, we should not export them.

For now keep it minimal like -

```
__all__ = [
    "EnrichmentSources",
    "PipelineNotFoundError",
    "VizProjectContext",
]
```

Drop `GraphBuilder`, `InspectionGraphService`, and `InspectionInputs` (and their imports) until a cross‑package consumer exists.

ravi-kumar-pilla:
Hi @jitu5 ,

Left minor comments, but overall the PR looks great. The data flow looks clean. Thank you 💯

package/kedro_viz/integrations/kedro/inspection/__init__.py
jitu5: Done, kept only the three exports you suggested and removed the unused package-level imports. Tests now import internal classes directly from their defining modules.

ravi-kumar-pilla (approved):
Awesome work @jitu5

PR #2764, opened 2026-09-07
Prepare project enrichment through one inspection loader

jitu5:
Part 2 of 6 for #2660. 

## Description

Prepare project enrichment through one entry point in `inspection/enrichment.py`, returning one `EnrichmentSources` object to the context. Use the released Kedro dependency range `kedro>=1.6.0`.

## Development notes

- Rename `layer_by_dataset` to `layer_by_dataset_name` consistently across preparation, graph consumers and tests, making the mapping key explicit without changing behavior.
- Replace the separate `load_node_extras()` and `EnrichmentSources.from_live_nodes()` preparation paths with `load_enrichment_sources()`.
- Gather file-backed stats/styles, copied live graph fields and supplied layer overrides into the prepared model. Keep the small JSON-reading helpers.
- Retain file extras by name for static metadata and exact live-ID overlays for graph responses. These views can differ; preserving both avoids changing live/transcoded selection or precedence.
- Reuse supplied file extras, including an explicit empty mapping, without rereading files. File-only preparation does not require a catalog or live nodes.
- Keep the legacy data-loader entry point as a delegating wrapper. The context loads enrichment only when no prepared object was supplied.
- Require `kedro>=1.6.0` exactly as agreed: no upper bound. CI uses the project requirement instead of installing Kedro main first; `requirements.txt` remains covered by secret scanning.

## QA notes

Local verification at `889219644`: **699 backend tests passed with 100% statement coverage**, using Python 3.13 and Kedro 1.6.0. Ruff, formatting and both MyPy targets passed. 

Tests cover stats/styles loading, file-only preparation, copied mappings, no rereads for supplied or empty extras, live-ID preservation, and differing file/live values. Existing graph parity expectations are unchanged.

At the final stack tip, 103 parity/enrichment/route tests passed in a clean checkout. Targeted Bandit checks and secret scanning passed again. Wheel building and strict docs passed on the preceding revision; this naming-only update changes neither packaging nor docs.

These are local results, not a claim that current-head GitHub checks are green. The previously diagnosed lite-mode E2E defect—whole-`sys.modules` restoration causing NumPy reinitialization—is inherited from V2 and deliberately unchanged. In the preceding CI runs, both full-mode scenarios passed and both lite-related scenarios errored. This revision does not claim to fix that defect; check the new runs below.

## Scope and checklist

- [x] One commit on `feat/backend_v2`, following merged #2763

### DIFF
--- a/package/kedro_viz/integrations/kedro/inspection/enrichment.py
+++ b/package/kedro_viz/integrations/kedro/inspection/enrichment.py
@@ -1,8 +1,12 @@
-"""Add live-only fields to graph responses without changing their structure."""
+"""Load file-backed node extras and enrich inspection graph responses."""
 
 from __future__ import annotations
 
+import json
+import logging
 from collections.abc import Iterable, Mapping
+from pathlib import Path
+from typing import Any
 
 from pydantic import BaseModel, Field, field_validator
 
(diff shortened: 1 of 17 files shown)


package/kedro_viz/integrations/kedro/inspection/node_extras.py
```diff
@@ -0,0 +1,111 @@
+"""Load file-backed Kedro-Viz node extras without constructing a catalog."""
```
ravi-kumar-pilla: can we do this as part of enrichments ?

jitu5: Yes, that makes sense. I’ll move the loader into the existing enrichment.py and remove node_extras.py. Loading behavior will stay unchanged, and I’ll update the imports through the stack.

package/kedro_viz/integrations/kedro/inspection/enrichment.py
```diff
@@ -75,3 +82,103 @@ def enrich_graph_response(
...
             node.dataset_type = sources.dataset_type_by_node_id[node.id]
+
+
+def _read_and_validate_json(
+    file_path: Path,
+    file_type: str,
+    project_path: Path,
+    fallback_message: str,
+) -> dict[str, Any]:
+    """Read a JSON object, returning an empty mapping when it cannot be used."""
```
ravi-kumar-pilla, line 75:
```suggestion
    """Read a JSON file, returning its object as a dict or ``{}`` on any failure."""
```

package/kedro_viz/integrations/kedro/inspection/enrichment.py
```diff
@@ -11,22 +15,28 @@
...
+    node_extras_by_name: Mapping[str, NodeExtras] = Field(default_factory=dict)
     node_extras_by_node_id: Mapping[str, NodeExtras] = Field(default_factory=dict)
     dataset_type_by_node_id: Mapping[str, str | None] = Field(default_factory=dict)
     layer_by_dataset: Mapping[str, str] | None = None
 
     @field_validator(
+        "node_extras_by_name",
         "node_extras_by_node_id",
         "dataset_type_by_node_id",
         "layer_by_dataset",
```
ravi-kumar-pilla: [Nit] everything is clear here except for layer_by_dataset - should if rename it to layer_by_dataset_name ? If you feel this is good, leave it.

jitu5: Updated. `layer_by_dataset_name` this make more sense.

SajidAlamQB (approved):
Thank you @jitu5!

ravi-kumar-pilla (approved):

PR #2765, opened 2026-09-07
Build static node metadata from inspection state

jitu5:
Part 3 of 6 for #2660. 

#2764 is merged. This PR now targets `feat/backend_v2` and contains one commit.

## Description

Build static node-detail responses from the shared inspection inputs, independently of live Kedro objects.

## Development notes

- Add `test_static_node_metadata_parity.py`: compare all 57 supported demo nodes with independently built legacy responses, without any demo data files. Compare every static field, including `None`, and check node IDs and all four response types. Resolve configured dataset aliases and relative paths for semantic comparison; live-only source/preview fields and exact enriched response strings remain covered in later layers.
- Keep the context as a thin composition root: load once, optionally filter, then construct independent services from the same `InspectionInputs`.
- Rename `InspectionGraphService` to `GraphService`. Give it only `EnrichmentSources.graph_extras`; pass name-keyed file extras to `NodeMetadataService` through its `enrichment` argument.
- Keep one enrichment-loading function. `GraphExtras` holds copied graph overlays and layer overrides; file extras retain their names for metadata. Neither container retains live nodes.
- Put shared dataset indexing and parameter preparation in `snapshot_source.py`. Graph indexing remains selected-pipeline scoped; metadata indexing spans the loaded snapshot.
- Preserve first-encounter name/free-input status, last-producer assignment, ordered transcoded inputs, and the distinct graph rule that any transcoded reference marks a graph dataset as transcoded.
- Use `build_parameters_from_inputs()` for both task and parameter-node metadata, with `_resolve_parameters()` private to the source module. Existing nested lookup semantics and `GraphBuilder(parameters=...)` are unchanged; no validation or typed-field rewrite is introduced.
- Keep `NodeMetadataBuilder` responsible for responses, returning dataset metadata to its caller for insertion. Reuse canonical IDs and return fresh response copies.
- Leave live enrichment and HTTP routing to following layers. The proposed services/builders/data_sources folder reorganisation is deferred.

## QA notes

Verified this PR independently on September 16 at `a048c4e98`, using Python 3.13.1 and Kedro 1.6.0 in an isolated checkout with built frontend assets supplied for backend tests:

- Full backend suite: **745 passed, 100% statement coverage**.
- Both MyPy targets, Ruff lint, formatting and diff checks passed.
- The new static parity module runs from #2765 onward. It compares real-project legacy expectations without using inspection builders to calculate them, keeps explicit static/live field partitions, and restores legacy mutable state.
- Dataset type aliases and relative paths are explicitly resolved only for the static semantic comparison. Tests verify different dataset types and paths remain distinguishable. No fields are silently dropped to excuse mismatches.

Existing complete-response parity in #2767 and HTTP parity in #2768 remain unchanged. Independent archived-source capture at `a048c4e98` reproduced all 57 metadata responses, 16 preview hashes and recorded dependency versions exactly. Only the baseline source-commit reference changed; response expectations were not changed. No production code changed for this feedback.

These are local results. Current-head GitHub checks still need to complete. Packaging, E2E and strict docs were not rerun locally for this test-only addition.

## Scope and checklist

- [x] One corrected commit on the previous stack layer


package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -72,5 +79,9 @@ def from_project(
             graph=InspectionGraphService.from_inspection_inputs(
                 inspection_inputs,
                 enrichment=enrichment_sources,
```
ravi-kumar-pilla, line 81: if `InspectionGraphService` does not need `node_extras_by_name` then lets have it like -

```python
graph=InspectionGraphService.from_inspection_inputs(
                inspection_inputs,
                enrichment=enrichment_sources.graph_extras,
```
and for `NodeMetadataService`, have it like 

```python
nodes=NodeMetadataService.from_inspection_inputs(
                inspection_inputs,
                enrichment=enrichment_sources.node_extras_by_name,
            ),
```

**Nit:** To be consistent with NodeMetadataService naming, if possible, lets rename `InspectionGraphService` to `GraphService`

package/kedro_viz/integrations/kedro/inspection/parameters.py
```diff
@@ -0,0 +1,33 @@
...
+    for ref in inputs:
+        if ref == "parameters":
+            result = dict(parameters)
+        elif ref.startswith("params:"):
+            name = ref[len("params:") :]
+            result[name] = resolve_parameter(parameters, name)
+    return result
+
+
+def resolve_parameter(parameters: dict[str, Any], dotted: str) -> Any:
```
ravi-kumar-pilla, line 25:
```suggestion
def _resolve_parameters(parameters: dict[str, Any], dotted: str) -> Any:
```

package/kedro_viz/integrations/kedro/inspection/parameters.py
```diff
@@ -0,0 +1,33 @@
+"""Resolve parameter references using the existing inspection graph semantics."""
+
+from typing import Any
+
+
+def parameters_for_inputs(
```
ravi-kumar-pilla, line 6:
```suggestion
def resolve_node_parameters(
```

ravi-kumar-pilla: or - `build_parameters_from_inputs`

package/kedro_viz/integrations/kedro/inspection/parameters.py
```diff
@@ -0,0 +1,33 @@
+"""Resolve parameter references using the existing inspection graph semantics."""
```
ravi-kumar-pilla, line 1: I feel https://github.com/kedro-org/kedro-viz/blob/feat/backend_v2/package/kedro_viz/integrations/kedro/inspection/snapshot_source.py

file will be a great place to have all inspection related inputs we use in our builders via service. i.e., 1 datasource -> 1 context -> individual services -> individual builders

If we have all our data loading, filtering and any data manipulations needed within snapshot_source.py, it would be a good thing to follow. wdyt ?

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -12,6 +12,9 @@
 from kedro_viz.integrations.kedro.inspection.graph_service import (
     InspectionGraphService,
 )
+from kedro_viz.integrations.kedro.inspection.node_metadata_service import (
```
ravi-kumar-pilla, line 15: at some point, lets have subfolders for services, builders and data_sources like we had for modular pipelines.

services (folder) -> graph_service, node_metadata_service
builders (folder) -> graph_builder, node_metadata_builder
data_sources (folder) -> snapshot_source, enrichment

ravi-kumar-pilla:
Hi @jitu5 ,

Great work 💯 ... I left some comments inline and I would like to get some clarity on the ones mentioned below that might be worth fixing - 

## node_metadata_builder.py and graph_builder.py both index datasets

I noticed `_index_datasets` and `_DatasetEntry` in `node_metadata_builder.py` are doing pretty much the same walk over node inputs and outputs that `GraphBuilder._register_dataset` already does in `graph_builder.py`, stripping transcoding, tracking the first name seen, figuring out free inputs. They can't just be merged since the metadata builder also needs the full list of transcoded variants and which one is the producer, but now we have two places that need to agree on the same legacy behavior. Is there a way to pull just the indexing pass into something shared, maybe living in `snapshot_source.py` like I mentioned in my other comment? Would rather not have two implementations quietly drift apart later.

## Worth a comment on the DatasetEntry docstring

`_index_datasets` matches the live repo correctly, last producer wins for `output`, but first encounter wins for `name` and `is_free_input`. Both are right and both are tested, but they are opposite rules living right next to each other. Can we add a line in the `_DatasetEntry` docstring spelling that out plainly, something like "producer assignment goes with the last one seen, name and free input status go with the first one seen"? Would save whoever reads this next from having to trace it back through the tests to figure out why it works that way.

## _build_parameters_metadata reinvents what parameters.py already does

`_build_parameters_metadata` strips the `params:` prefix and calls `resolve_parameter` by hand instead of just calling `parameters_for_inputs` (or whatever we end up renaming it to) with a single item list, like `parameters_for_inputs([reference], self._parameters)`. Small thing, but that module exists specifically so we are not repeating this logic. Can we swap it over so there is one place that owns the params: parsing?

## Passing a dict around instead of returning one

`_register_dataset_metadata` takes `metadata_by_node_id` as an argument and mutates it inside `_build_metadata_index`. Given the file already leans on building little entry objects like `_DatasetEntry`, would it read cleaner if `_register_dataset_metadata` just returned the entry and the caller put it in the dict? Not a big deal, just flagging since it would make `_build_metadata_index` easier to follow top to bottom, index datasets, build task entries, then dataset entries.


Thank you

package/kedro_viz/integrations/kedro/inspection/context.py
```diff
@@ -21,8 +24,9 @@
 class VizProjectContext:
     """Services prepared for one Kedro project load."""
 
-    def __init__(self, graph: InspectionGraphService) -> None:
+    def __init__(self, graph: GraphService, nodes: NodeMetadataService) -> None:
```
ravi-kumar-pilla, line 27:
```suggestion
    def __init__(self, graph: GraphService, node_metadata: NodeMetadataService) -> None:
```

package/kedro_viz/integrations/kedro/inspection/node_metadata_builder.py
```diff
@@ -0,0 +1,149 @@
...
+        self._snapshot = snapshot
+        self._parameters = dict(parameters)
+        self._node_extras_by_name = dict(node_extras_by_name or {})
+        self._metadata_by_node_id = self._build_metadata_index()
+
+    def build(self, node_id: str) -> NodeMetadataAPIResponse:
+        """Return fresh metadata for ``node_id``.
+
+        Raises:
+            NodeNotFoundError: If the ID is unknown or represents an unsupported node kind.
```
ravi-kumar-pilla, line 52: is this NodeNotFoundError or NodeMetadataNotFoundError ?

jitu5: The intention is to use `NodeNotFoundError` for an unknown node ID. In #2768, known modular nodes without metadata are handled separately with `NodeMetadataNotAvailableError`, preserving their existing empty response. That distinction is introduced later in the stack; #2765 currently uses `NodeNotFoundError` for any ID missing from the metadata index.

package/kedro_viz/integrations/kedro/inspection/context.py
jitu5: Updated. GraphService now receives only `graph_extras`, while `NodeMetadataService` receives the name-keyed node extras through enrichment. Also renamed `InspectionGraphService` to `GraphService` throughout the stack.

package/kedro_viz/integrations/kedro/inspection/parameters.py
jitu5: Agreed. Shared dataset indexing and parameter preparation now live in `snapshot_source.py`, alongside snapshot loading and filtering. The context prepares the inputs for each service, and the builders remain responsible for constructing responses. Enrichment loading stays in `enrichment.py`.

package/kedro_viz/integrations/kedro/inspection/context.py
jitu5: Agreed. I’ve left the folder layout unchanged in this stack to keep the review focused. We can group services, builders and data sources separately as a follow-up.

jitu5:
> Hi @jitu5 ,
> 
> Great work 💯 ... I left some comments inline and I would like to get some clarity on the ones mentioned below that might be worth fixing -
> 
> ## node_metadata_builder.py and graph_builder.py both index datasets
> I noticed `_index_datasets` and `_DatasetEntry` in `node_metadata_builder.py` are doing pretty much the same walk over node inputs and outputs that `GraphBuilder._register_dataset` already does in `graph_builder.py`, stripping transcoding, tracking the first name seen, figuring out free inputs. They can't just be merged since the metadata builder also needs the full list of transcoded variants and which one is the producer, but now we have two places that need to agree on the same legacy behavior. Is there a way to pull just the indexing pass into something shared, maybe living in `snapshot_source.py` like I mentioned in my other comment? Would rather not have two implementations quietly drift apart later.
> 
> ## Worth a comment on the DatasetEntry docstring
> `_index_datasets` matches the live repo correctly, last producer wins for `output`, but first encounter wins for `name` and `is_free_input`. Both are right and both are tested, but they are opposite rules living right next to each other. Can we add a line in the `_DatasetEntry` docstring spelling that out plainly, something like "producer assignment goes with the last one seen, name and free input status go with the first one seen"? Would save whoever reads this next from having to trace it back through the tests to figure out why it works that way.
> 
> ## _build_parameters_metadata reinvents what parameters.py already does
> `_build_parameters_metadata` strips the `params:` prefix and calls `resolve_parameter` by hand instead of just calling `parameters_for_inputs` (or whatever we end up renaming it to) with a single item list, like `parameters_for_inputs([reference], self._parameters)`. Small thing, but that module exists specifically so we are not repeating this logic. Can we swap it over so there is one place that owns the params: parsing?
> 
> ## Passing a dict around instead of returning one
> `_register_dataset_metadata` takes `metadata_by_node_id` as an argument and mutates it inside `_build_metadata_index`. Given the file already leans on building little entry objects like `_DatasetEntry`, would it read cleaner if `_register_dataset_metadata` just returned the entry and the caller put it in the dict? Not a big deal, just flagging since it would make `_build_metadata_index` easier to follow top to bottom, index datasets, build task entries, then dataset entries.
> 
> Thank you

Addressed all four:
1. Both builders now use `build_dataset_index` from `snapshot_source.py`, while preserving their existing pipeline scopes and transcoding behaviour.
2. `DatasetEntry` documents the ordering explicitly: first encounter determines the name and free-input status; last producer wins.
3. Parameter metadata now calls `build_parameters_from_inputs` instead of repeating the parsing.
4. `_build_dataset_metadata` returns the response, and the caller owns insertion into the index.

The parity checks passed without changing the expected metadata responses or preview hashes.

package/kedro_viz/integrations/kedro/inspection/context.py
jitu5: Renamed `nodes` to `node_metadata` in the context constructor, attribute and callers throughout the stack.

package/tests/test_inspection_adapter/test_node_metadata_builder.py
```diff
@@ -0,0 +1,737 @@
+"""Tests for static node metadata prepared from inspection inputs."""
```
ravi-kumar-pilla, line 1: **Missing automated live-parity test for NodeMetadataService**

`test_inspection_adapter_parity.py` has `live_enriched_graph_service`, which builds a `GraphService` from `load_inspection_inputs()` and compares it against the live DataAccessManager-built graph. There is no equivalent for `NodeMetadataService` — the new builder's tests (test_node_metadata_builder.py) all use hand-constructed NodeSnapshot/PipelineSnapshot fixtures, not the real demo project.

Given this PR is where node metadata reaches feature-parity (analogous to where the graph parity test was introduced), it'd be worth adding a `live_metadata_service` fixture alongside `live_enriched_graph_service` that diffs `NodeMetadataBuilder` output against live per-node metadata for the demo project, so this parity is enforced in CI rather than only verified manually once.

ravi-kumar-pilla (approved):
Added a minor comment but overall looks good. Thank you @jitu5

package/tests/test_inspection_adapter/test_node_metadata_builder.py
jitu5: updated.

Issue #2723, opened 2026-08-05
Serve task source metadata from the Kedro inspection snapshot


jitu5:
**Parent issue:** #2265
**Depends on:** #2660, https://github.com/kedro-org/kedro/issues/5668 the stable upstream Kedro node-source API

## Summary

Use Kedro inspection's node-source metadata to serve task `code` and `filepath` without looking up a live Kedro node. Task IDs remain the existing Kedro-Viz IDs restored by #2715/#2716.

## Scope

- Read source location information from the stable Kedro inspection API.
- Map it to task IDs with the shared node-ID helper.
- Resolve project-relative paths against the project root, validate paths according to the final upstream contract, use Python-aware source-file decoding, and validate the 1-based inclusive line range before slicing.
- Support the same static source path in full and lite mode.
- Treat missing source as an expected optional case.
- Keep task and dataset previews on the live-object path.

## Acceptance criteria

- Source content matches the current full-mode response when Kedro reports a source location.
- Project-relative POSIX paths from Kedro are converted to the existing Viz filepath response shape, or an intentional response-format change is documented and tested.
- Source lookup does not change graph IDs or introduce an ID-translation map.
- Lite mode serves available source without a `KedroSession` or `DataCatalog`.
- `source=None`, unreadable files, and unsupported functions do not cause a server error.

PR #2766, opened 2026-09-07
Extract stateless live metadata helpers

### LINKED ISSUE #2660: Load shared inspection inputs at the context boundary and serve run status from the project context

jitu5:
Part 4 of 6 for #2660.

Review order: [#2763](https://github.com/kedro-org/kedro-viz/pull/2763) → [#2764](https://github.com/kedro-org/kedro-viz/pull/2764) → [#2765](https://github.com/kedro-org/kedro-viz/pull/2765) → **#2766** → [#2767](https://github.com/kedro-org/kedro-viz/pull/2767) → [#2768](https://github.com/kedro-org/kedro-viz/pull/2768).

#2765 is merged. This PR targets `feat/backend_v2` and contains one commit.

## Description

Extract the metadata calculations that require live Kedro objects into shared helpers. Legacy metadata models use them here; #2767 will reuse them to enrich snapshot-backed metadata without duplicating the logic.

This PR also removes request-specific objects from metadata model classes. Validators now read the node being validated from Pydantic's per-instance validation data.

## Development notes

- Add `models/flowchart/live_node_metadata.py` for task source, dataset paths and types, and task/dataset previews. Keep these helpers alongside the live models because both legacy consumers and inspection enrichment need them.
- Keep response models and static metadata ownership unchanged. This PR does not wire live enrichment into the context or change HTTP routes.
- Pass dataset preview policy explicitly to the helpers. Retain the existing class-level preview switch for legacy callers; it no longer needs to be the policy source for new consumers.
- Preserve plain-dataset release once before metadata reads. Transcoded metadata does not release its datasets.
- Preserve task preview behaviour independently of the dataset preview switch, including supported payloads and error handling.
- Remove the file-wide private-member lint exemption from `node_metadata.py`; the necessary dataset `_describe()` calls are narrowly annotated in the helper module.

## QA notes

Verified independently on September 16 at `2403162b9`, using Python 3.13.1 and Kedro 1.6.0. The rebased head `768011d7c` was checked on September 17 and has an identical file tree; these are the earlier local results, not a new test run:

- **759 backend tests passed with 100% statement coverage**, in an isolated checkout with built frontend assets supplied for the backend tests.
- Both MyPy targets, Ruff lint, formatting and diff checks passed.
- Tests cover preview arguments and opt-outs, preview error handling, task preview independence, release timing, run commands, and removal of request objects from model classes.
- The real-project static metadata parity tests introduced in #2765 also pass at this layer.

Complete enriched-response parity remains in #2767, and HTTP parity remains in #2768. Those later checks complement this PR's tests; they are not part of this diff.

Local verification does not establish that current-head GitHub checks are complete. Packaging, E2E and strict documentation builds were not rerun locally for this revision.

## Scope and checklist

- [x] Focused helper and legacy-model tests
- [x] Full backend suite and local quality checks for this layer
- [x] One commit on top of the merged #2765 work
- [x] Stack release note included in #2768

This PR prepares the shared live-metadata logic for #2767. Context-backed export/deployment and parameter-behaviour changes remain separate work; this stack does not close #2660.


package/kedro_viz/models/flowchart/node_metadata.py
```diff
@@ -3,23 +3,27 @@
...
+    get_data_node_preview,
+    get_data_node_preview_type,
+    get_data_node_type,
+    get_task_code,
+    get_task_filepath,
+    get_task_preview,
+    get_transcoded_data_node_filepath,
+    get_transcoded_data_node_original_type,
+    get_transcoded_data_node_types,
+)
```
ravi-kumar-pilla, line 24: we have already built node extras and other metadata required in previous PRs. Why all this is required ?

package/kedro_viz/models/flowchart/node_metadata.py
```diff
@@ -77,119 +81,46 @@ class TaskNodeMetadata(GraphNodeMetadata):
...
-        if inspect.ismethod(func):
-            func = func.__func__
-
-        if inspect.isfunction(func):
-            code = inspect.getsource(_extract_wrapped_func(func))
-            return code
-
-        return None
+    def set_code(cls, _, info: ValidationInfo):
+        return get_task_code(cast(TaskNode, info.data["task_node"]))
```
ravi-kumar-pilla, line 89: I thought we will eventually get rid of TaskNode and other classes here. Was that changed ? If we are doing cast like this, is this only for v1 or even after v2 we want this to remain as-is ?

package/kedro_viz/models/flowchart/node_metadata.py
jitu5: The earlier PRs handle file-backed extras and static metadata. This PR extracts the existing previews, task source, and live dataset type/path calculations so the legacy validators and #2767 can share them without duplicating the logic. It doesn’t rebuild node extras. 

The live type/path handling preserves existing full-mode responses; task source will move to the snapshot in #2723, using the source locations available in Kedro 1.6.

package/kedro_viz/models/flowchart/node_metadata.py
jitu5: Yes, removing the old node models is still the plan. They are a temporary dependency for this migration, not the intended final design. The cast only helps type checking.

The new helpers and #2767 still depend on these models, so this stack does not remove that dependency. #2768 moves the HTTP routes to the context; #2724 handles the remaining legacy cleanup, keeping only the live objects needed for previews. I should make that temporary dependency and its removal clearer in this PR.

jitu5:
Closing as metadata endpoint work shifted to V2 of Kedro-Viz modularity work.

package/kedro_viz/models/flowchart/node_metadata.py
ravi-kumar-pilla: Having live kedro obj for previews make sense. We should handle metadata at a later stage and see if we can get previews from kedro inspection api, so we can get rid of live kedro_obj. Lets park this for now and shift this to v2 as discussed

PR #2770, opened 2026-09-10
Fix toolbar `tooltip` not dismissing on mouse leave

### LINKED ISSUE #2282: Tooltip takes long time before disappearing even when not hovered
## Description

The tooltip stays even when not hovered

image

## Context

https://github.com/kedro-org/kedro-viz/pull/2251
https://github.com/kedro-org/kedro-viz/pull/2251#issuecomment-2685712064

## Steps to Reproduce

1. Open Kedro-Viz
2. Hover on any icon on the toolbar

## Expected Result

Tooltip should disappear after the user removes focus from the label/icon

## Actual Result

Sometimes the tooltip stays on the screen forever

## Your Environment

## Checklist

ravi-kumar-pilla:
## Description

Resolves https://github.com/kedro-org/kedro-viz/issues/2282

## Development notes

**Root cause**

`IconButton` used a plain local variable to hold the tooltip's show-timeout id:

```jsx
let inTimeout; // recreated on every render
```

`showTooltip` (on `mouseEnter`) schedules a `setTimeout(..., 333)` that flips the tooltip visible, and `hideTooltip` (on `mouseLeave`) is supposed to `clearTimeout(inTimeout)`. Because `inTimeout` is re-created on every render, if the component re-renders during that 333 ms window (very common in Kedro-Viz - hover/layout/store updates re-render the connected toolbar), the new render's `hideTooltip` calls `clearTimeout(undefined)` and never cancels the pending timer. That orphaned timer then fires **after** the pointer has already left, and since no further `mouseleave` event is coming, the tooltip stays on screen indefinitely.

**Fix** (`src/components/ui/icon-button/icon-button.jsx`)

- Store the timeout id in a `useRef` so the handle is stable across re-renders and `hideTooltip` always cancels the right timer.
- Add an `isHovered` ref; the delayed callback bails out if the pointer left while it was pending.
- Add a `useEffect` cleanup that clears any pending timer on unmount (covers the "button unmounted / covered by a panel while hovered" variants too).

## QA notes

**Behaviour change:** toolbar tooltips now always disappear when the pointer leaves the icon. A tooltip can no longer appear (or remain) while the cursor is elsewhere.

**Automated tests:** added a regression test in `src/components/ui/icon-button/icon-button.test.jsx` - `does not show tooltip if the pointer leaves before the show delay elapses, even across a re-render` - which reproduces the exact race (mouseEnter -> re-render -> mouseLeave -> run timers) and asserts the tooltip stays hidden.

**Manual steps to reproduce the original bug:**

The bug needs a re-render of the `IconButton` to land inside the 333 ms show-delay while the cursor never leaves the button. Browser zoom fires a `resize` -> Redux `updateChartSize` -> the toolbar re-renders, and it's keyboard-only so the mouse stays parked.

1. To make the timing window comfortable, temporarily widen the delay in `src/components/ui/icon-button/icon-button.jsx` (line ~41): change `333` to `4000`.
2. `make run` (or your usual dev start) and open the app in the browser.
3. Move the mouse onto a left-hand toolbar icon - e.g. **Hide menu**, the theme toggle, or **Expand all pipelines**. Leave it there, don't move it.
4. Within the 4 s window, press **Cmd + `-`** (zoom out) once or twice. The cursor stays on the button.
5. Move the mouse well away - into the flowchart canvas - before the 4 s elapses.
6. Wait for the delay to elapse.

- **On `main`:** the tooltip label appears with the cursor nowhere near it and stays on screen indefinitely (until you hover + unhover that same button again).

Screenshot 2026-09-10 at 12 44 01 PM

---

- **With this fix:** the tooltip never appears.

image

## Checklist

### DIFF
--- a/src/components/ui/icon-button/icon-button.jsx
+++ b/src/components/ui/icon-button/icon-button.jsx
@@ -49,7 +49,6 @@ const IconButton = ({
       if (!isHovered.current) {
         return;
       }
-      window.localStorage.setItem('kedro-viz-tooltip-show', true);
       setIsTooltipVisible(true);
     }, 333);
   };
--- a/src/components/ui/icon-button/icon-button.jsx
+++ b/src/components/ui/icon-button/icon-button.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import './icon-button.scss';
@@ -25,8 +25,15 @@ const IconButton = ({
   ...rest
 }) => {
   const Icon = icon;
-  let inTimeout;
+  const inTimeout = useRef(null);
+  const isHovered = useRef(false);
   const [isTooltipVisible, setIsTooltipVisible] = useState(false);
+
+  // Clear any pending show-timeout when the button unmounts, so the tooltip
+  // can't be shown after the pointer has already left.
+  useEffect(() => {
+    return () => clearTimeout(inTimeout.current);
+  }, []);
 
   const labelPosition = labelPositionTypes.includes(
     labelTextPosition.toLowerCase()
@@ -35,14 +42,21 @@ const IconButton = ({
     : 'right';
 
   const showTooltip = () => {
-    inTimeout = setTimeout(() => {
+    isHovered.current = true;
+    clearTimeout(inTimeout.current);
+    inTimeout.current = setTimeout(() => {
+      // Bail out if the pointer left while the timeout was pending.
+      if (!isHovered.current) {
+        return;
+      }
       window.localStorage.setItem('kedro-viz-tooltip-show', true);
       setIsTooltipVisible(true);
     }, 333);
   };
 
   const hideTooltip = () => {
-    clearTimeout(inTimeout);
+    isHovered.current = false;
+    clearTimeout(inTimeout.current);
     setIsTooltipVisible(false);
   };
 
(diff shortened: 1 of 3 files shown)


src/components/ui/icon-button/icon-button.jsx
```diff
@@ -25,24 +25,38 @@ const IconButton = ({
...
   const showTooltip = () => {
-    inTimeout = setTimeout(() => {
+    isHovered.current = true;
+    clearTimeout(inTimeout.current);
+    inTimeout.current = setTimeout(() => {
+      // Bail out if the pointer left while the timeout was pending.
+      if (!isHovered.current) {
+        return;
+      }
       window.localStorage.setItem('kedro-viz-tooltip-show', true);
```
jitu5: Looks like dead code,  nothing in `src/` ever reads `kedro-viz-tooltip-show` key. Please check before you delete.

jitu5 (approved):
Added a small comment, but overall looks good. Thanks.

src/components/ui/icon-button/icon-button.jsx
ravi-kumar-pilla: Looks like deadcode from 2022 (https://github.com/kedro-org/kedro-viz/commit/ed52e17e , https://github.com/kedro-org/kedro-viz/commit/9b7b58ff ) . Let me clean it. Thank you

Huongg (approved):
LGTM thank you @ravi-kumar-pilla

PR #2768, opened 2026-09-07
Serve node metadata and run status from VizProjectContext

jitu5:
Part 6 of 6 for #2660. Existing native GitHub stack #2769.

Review order: [1: #2763](https://github.com/kedro-org/kedro-viz/pull/2763) → [2: #2764](https://github.com/kedro-org/kedro-viz/pull/2764) → [3: #2765](https://github.com/kedro-org/kedro-viz/pull/2765) → [4: #2766](https://github.com/kedro-org/kedro-viz/pull/2766) → [5: #2767](https://github.com/kedro-org/kedro-viz/pull/2767) → [6: #2768](https://github.com/kedro-org/kedro-viz/pull/2768).

Depends on #2767. Please review this diff against its immediate parent; it contains one commit.

## Description

Serve project metadata and run status through the context, completing the HTTP portion of this stack.

## Development notes

- Restacked onto the revised #2765 interfaces and shared input preparation; this layer keeps its original responsibility.
- Bind graph routes to `context.graph`, metadata routes to `context.node_metadata`, and run status to `context.run_status`.
- Add `RunStatusService` bound to the explicit project root; keep the legacy zero-argument run-status reader for export.
- Distinguish known modular nodes (200 with `{}`) from unknown IDs (404), including before any graph request and after pipeline filtering.
- Preserve None exclusion, logging, error responses, OpenAPI documentation and cross-app isolation. Remove duplicate legacy routes.
- Add the release note for all six layers.

## QA notes

Verified this PR independently on September 16 at `718b92b12`, using Python 3.13.1 and Kedro 1.6.0 in an isolated checkout with built frontend assets supplied for backend tests:

- Full backend suite: **849 passed, 100% statement coverage**.
- Both MyPy targets, Ruff lint, formatting and diff checks passed.
- The new static parity module runs from #2765 onward. It compares real-project legacy expectations without using inspection builders to calculate them, keeps explicit static/live field partitions, and restores legacy mutable state.
- Dataset type aliases and relative paths are explicitly resolved only for the static semantic comparison. Tests verify different dataset types and paths remain distinguishable. No fields are silently dropped to excuse mismatches.

Existing complete-response parity in #2767 and HTTP parity in #2768 remain unchanged. Independent archived-source capture at `a048c4e98` reproduced all 57 metadata responses, 16 preview hashes and recorded dependency versions exactly. Only the baseline source-commit reference changed; response expectations were not changed. No production code changed for this feedback.

[...]

PR #2773, opened 2026-09-18
Serve run status from the project context

jitu5:
## Description

Move `/api/run-status` into `VizProjectContext` so it reads the event file belonging to the project being served, rather than relying on the process's current working directory.

This keeps run status tied to the correct project if the working directory changes or multiple project apps are created in the same process. The response format and event processing stay the same.

## Changes

- Add `RunStatusService`, which resolves the project path once and reads its event file on each request, so new runs are reflected without restarting Viz.
- Use a context-bound project router for graph and run-status endpoints.
- Extract event-file reading into a shared function, keeping the existing wrapper for legacy callers such as static export.
- Preserve missing-file handling, logging and the existing HTTP error response.
- Move the event-transformation functions from `run_events.py` to `inspection/run_status_builder.py` unchanged, so run status follows the source → service → builder shape.
- `run_events.py` keeps the response models and the legacy export wrapper.

## How this fits into V1

This carries forward only the run-status part of #2768 after the V1 scope reduction and cleanup in #2772. It targets `feat/backend_v2` and contributes to #2660; it does not complete that issue.

Graph endpoints already use the project context. This PR brings run status alongside them without migrating `/api/nodes/{id}`. Node metadata remains on the legacy path. Lite-mode work and context-backed export/deployment are separate, and the existing metadata PR stack is unchanged.

## Verification

- Backend suite: **718 tests passed, 100% statement coverage**.
- Frontend pre-push suite: **950 tests passed**.
- Both MyPy checks, Ruff lint, formatting and diff checks passed.

Tests cover project isolation, working-directory changes, updated event files between requests, missing and invalid event files, route registration, and the existing HTTP error response. Existing graph and legacy metadata compatibility tests also pass.

These are local results; packaging, E2E and documentation checks were not rerun locally for this change.


package/kedro_viz/api/rest/responses/run_events.py
ravi-kumar-pilla, line 1: Do you think moving this to run_status_builder.py make sense ? The inspection pattern would be consistent - snapshot_source/enrichment -> service -> builders ?

package/kedro_viz/integrations/kedro/inspection/run_status_service.py
```diff
@@ -0,0 +1,23 @@
...
+class RunStatusService:
+    """Read run status for one resolved Kedro project root."""
+
+    def __init__(self, project_path: str | Path) -> None:
+        project_root = Path(project_path).expanduser().resolve()
+        self._event_file_path = project_root / PIPELINE_EVENT_FULL_PATH
+
+    def get_run_status_response(self) -> RunStatusAPIResponse:
+        """Return transformed events from this service's project file."""
+        return read_run_status_response(self._event_file_path)
```
ravi-kumar-pilla, line 29: If this is placed in builder as suggested in above comment, it will be consistent design

package/kedro_viz/api/rest/responses/run_events.py
jitu5: Yes, that makes sense. I’ll separate the event-to-response logic into `run_status_builder.py` and have the service handle the project-bound file loading. The response models will stay in the response module, and the legacy export wrapper will keep its existing behavior. I’ll move the relevant tests with the builder and preserve the current HTTP and service coverage.

package/kedro_viz/integrations/kedro/inspection/run_status_builder.py
```diff
@@ -0,0 +1,342 @@
+"""Build run-status responses from recorded pipeline events."""
```
ravi-kumar-pilla, line 1: Just to confirm @jitu5 , this is just moving code from run_events.py or is there any change on how this builder is done ?

jitu5: Yes, this is just moving the existing response-building logic from `run_events.py` into `run_status_builder.py`.

ravi-kumar-pilla (approved):

PR #2774, opened 2026-09-22
Live load deferred for metadata

ravi-kumar-pilla:
## Checklist


jitu5:
Thanks @ravi-kumar-pilla 

### 1. concurrent metadata requests can return another node’s data

I reproduced overlapping parameter metadata construction:

```text
Expected: [{'a': 1}, {'b': 2}]
Actual:   [{'b': 2}, {'b': 2}]
```

The loader’s lock protects initial loading—not metadata construction afterward.

### 2. lite mode still fails on the first node click

Using the large benchmark project in a fresh process:

```text
GET /api/main          → 200
GET /api/nodes/{id}    → 500
```
It fails with:

```text
ImportError: cannot load module more than once per process
```

The existing lite import-cache restoration removes imported NumPy modules; the later live load attempts to import them again.

**This is an existing problem moved from startup to the first click—not solved by lazy loading.**

### 3. removing live dataset-type enrichment changes graph responses

I found **17 dataset-type differences** against the existing demo baseline. For example:

```text
Before: plotly.json_dataset.JSONDataset
After:  plotly.JSONDataset
```

All six graph-parity cases fail. This also affects frontend behavior

ravi-kumar-pilla:
Hi @jitu5,

Thanks for testing this out.

> ### 1. concurrent metadata requests can return another node’s data
> The loader’s lock protects initial loading—not metadata construction afterward.

Fixed this. The issue was present even before this change but was un-noticed as we had `async` route. Now that we changed it, this surfaced. Please test it again.

> ### 2. lite mode still fails on the first node click

Will fix the lite mode in the following PR

> ### 3. removing live dataset-type enrichment changes graph responses
> All six graph-parity cases fail. This also affects frontend behavior

Fixed it. Please test it again. 

Thank you

jitu5 (approved):

